Issue #14, opened 2024-11-14
Rollout Scheduling


HCookie:
Our current rollout implementation is very focused on sequential epoch increments, it would be good to generalise this to provide schedulers to control rollout.

Work was done in `aifs-mono` to enable this. [here](https://github.com/ecmwf-lab/aifs-mono/tree/feature/rollout-strategies) 
I think this can be generalised and provide more general applicability.

## Features
Below is a list of features and requirements as I see them
- Epoch Step rollout
- Static Rollout - constant (already supported) 
- Upon hitting a threshold begin another strategy
- Random selection of rollout between bounds
- Dynamic selection of increments 
   - e.g. At epoch 5, increment by 2, at epoch 13, increment by 3

## Improvements
Setup config at begin of training with rollout increment be

## Questions
- Will the rollout only change between epochs? Could within an epoch it change?

What other features may be needed?

PR #46, opened 2024-12-20
feat(training,rollout)!: Rollout Schedulers

HCookie:
Closes https://github.com/ecmwf/anemoi-core/issues/14

# Rollout Schedulers

Expand the ways to describe rollout, and provide an interface to schedule updates
## New default rollout config
```yaml
rollout:
  _target_: anemoi.training.schedulers.rollout.stepped.EpochStepped
  minimum: 1
  maximum: 12
  # increase rollout every n epochs
  every_n_epochs: 1
  # Control the incrementing of the rollout window
  increment:
    step:
      0: 0
      200000: 1 # After 200k steps, increment by 1 every 1 epoch
```
Can step by epoch, step, and control the increment based on either the step or epoch.

Additonally, formally add random steppers.
## Todo

- [x] Integrate with the data loader
- [x] Ensure that the randomness is seeded appropriately 
- [x] Randomness broadcast
- [x] Ensure restartability
- [x] Ability to change config

# Tested
Tested with restart, and with change in config after restart.

----
📚 Documentation preview 📚: https://anemoi-training--46.org.readthedocs.build/en/46/

----
📚 Documentation preview 📚: https://anemoi-graphs--46.org.readthedocs.build/en/46/

----
📚 Documentation preview 📚: https://anemoi-models--46.org.readthedocs.build/en/46/


training/src/anemoi/training/train/train.py
```diff
@@ -405,6 +419,7 @@ def train(self) -> None:
             use_distributed_sampler=False,
             profiler=self.profiler,
             enable_progress_bar=self.config.diagnostics.enable_progress_bar,
+            reload_dataloaders_every_n_epochs=self._need_to_reload_dataloaders,
```
anaprietonem: Is there a reason why the type of `reload_dataloaders_every_n_epochs` has been changes from `int` to `bool`? Wondering because looking at PTL docs the type of flag is `int` (https://lightning.ai/docs/pytorch/stable/_modules/lightning/pytorch/trainer/trainer.html#Trainer.__init__) and that's used by the `[data_connector](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/trainer/connectors/data_connector.py#L51)`

training/src/anemoi/training/diagnostics/callbacks/rollout.py
```diff
@@ -0,0 +1,70 @@
...
+        trainer : pl.Trainer
+            Pytorch Lightning trainer
+        pl_module : pl.LightningModule
+            Model
+        checkpoint : dict
+            Checkpoint dictionary
+        """
+        self._update_rollout(trainer, pl_module, epoch=checkpoint["epoch"], step=checkpoint["global_step"])
+
+    def on_validation_epoch_end(self, trainer: pl.Trainer, pl_module: pl.LightningModule, *_) -> None:
```
anaprietonem: If someone sets the limit_batches for validation to 0, to skip validation this hook wouldn't be triggered?

training/src/anemoi/training/train/forecaster.py
```diff
@@ -146,18 +150,15 @@ def __init__(
         self.warmup_t = getattr(config.training.lr, "warmup_t", 1000)
         self.lr_iterations = config.training.lr.iterations
         self.lr_min = config.training.lr.min
-        self.rollout = config.training.rollout.start
-        self.rollout_epoch_increment = config.training.rollout.epoch_increment
-        self.rollout_max = config.training.rollout.max
+
+        self.rollout: RolloutScheduler = instantiate(config.training.rollout)
```
anaprietonem, line 178: Nice, looks much cleaner!

training/src/anemoi/training/train/forecaster.py
```diff
@@ -451,7 +451,7 @@ def rollout_step(
         )
         assert batch.shape[1] >= rollout + self.multi_step, msg
 
-        for rollout_step in range(rollout or self.rollout):
+        for rollout_step in range(rollout or int(self.rollout)):
```
anaprietonem: Do we still need to pass rollout as a value to this function is you are already making it a self variable of the `GraphForecaster` ?

training/src/anemoi/training/train/forecaster.py
```diff
@@ -605,6 +605,20 @@ def calculate_val_metrics(
...
+        # Sync the rollout at the start of training
+        self.rollout.sync(step=self.global_step, epoch=self.current_epoch)
+
+    def on_load_checkpoint(self, checkpoint: dict) -> None:
+        # Sync the rollout on the load of a checkpoint
+        self.rollout.sync(step=checkpoint["global_step"], epoch=checkpoint["epoch"])
+
+    def on_train_epoch_start(self) -> None:
+        # Sync the rollout at the start of each epoch
+        # Cannot use stepping due to inconsistent behaviour with Pytorch Lightning
```
anaprietonem: Just to understand it better, what do you mean here with the inconsistent behaviour with PTL?

training/src/anemoi/training/train/forecaster.py
```diff
@@ -605,6 +605,20 @@ def calculate_val_metrics(
 
         return metrics
 
+    def on_train_start(self) -> None:
+        # Sync the rollout at the start of training
+        self.rollout.sync(step=self.global_step, epoch=self.current_epoch)
+
+    def on_load_checkpoint(self, checkpoint: dict) -> None:
```
anaprietonem, line 640: In the docs for the `Modelhooks` I can't find a hook for `on_load_checkpoint` https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.core.hooks.ModelHooks.html#lightning.pytorch.core.hooks.ModelHooks so maybe I am missing something but how is this working?

anaprietonem (commented):
Started to go through the PR and left some comments! I still need to understand better some of the functionality so hope te questions makes sense. Thanks for this Harrison!

FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=46) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=46) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=46) it.</sub>

training/src/anemoi/training/train/train.py
HCookie: Due to python duck typing I suppose a `True` will evaluate to `1` when used as an int. So it works, but if it is clearer, I can change it to an int.

training/src/anemoi/training/diagnostics/callbacks/rollout.py
HCookie: Good point, I'll need to take a look.

training/src/anemoi/training/train/forecaster.py
HCookie: Passing it as an arg allows for override, which I think is used in some of the callbacks.

training/src/anemoi/training/train/forecaster.py
HCookie: If I use the step function, it gets triggered in sanity checking, and other places where I don't want it,

training/src/anemoi/training/train/forecaster.py
HCookie: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.core.hooks.CheckpointHooks.html
It seems to be here?

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -85,12 +85,19 @@ scale_validation_metrics:
...
   # increase rollout every n epochs
-  epoch_increment: 0
-  # maximum rollout to use
-  max: 1
+  every_n_epochs: 1
+  # Control the incrementing of the rollout window
+  increment:
+    step:
+      0: 0
+      200000: 1 # After 200k steps, increment by 1 every 1 epoch
```
anaprietonem: I am probably just being slow but how does this interact with the `limit batches`?  What would be the difference between doing the above, and the 'old configuration' with a limit batches of  200000?

training/tests/schedulers/__init__.py
```diff
@@ -0,0 +1,8 @@
+# (C) Copyright 2024 Anemoi contributors.
```
anaprietonem, line 1: Minor thing, but just a remainder that all of these headers would need to be updated to 2025 before merging

training/src/anemoi/training/train/train.py
anaprietonem: I see now, thanks for the clarification! all good then

training/src/anemoi/training/train/train.py
```diff
@@ -377,6 +377,20 @@ def strategy(self) -> DDPGroupStrategy:
             static_graph=not self.config.training.accum_grad_batches > 1,
         )
 
+    @cached_property
+    def _need_to_reload_dataloaders(self) -> bool:
```
anaprietonem: Have you seen any difference in terms of runtime from using this?

training/src/anemoi/training/train/forecaster.py
anaprietonem: Nice yep you're right, I did not know about that! so all good

training/src/anemoi/training/train/train.py
HCookie: Unfortunately, yes, I need to still quantify it, but it does slow down the transition between epochs.

training/src/anemoi/training/config/training/default.yaml
HCookie: The `limit_batches` ends the training, this will continue on, and then begin updating the rollout.

training/tests/schedulers/__init__.py
HCookie: Even the code I wrote in 2024? I honestly have no idea

training/src/anemoi/training/schedulers/rollout/randomise.py
```diff
@@ -0,0 +1,371 @@
...
+        int
+            Either broadcasted value or update value from broadcast
+        """
+        self._dist_rollout = torch.tensor([value])
+        try:
+            if dist.get_world_size() > 1:
+                dist.broadcast(self._dist_rollout, src=0)
+        except ValueError:
+            pass
+        return self._dist_rollout[0]
```
HCookie: @cathalobrien Can you verify this does what I think it does?

cathalobrien: Looks good! Each process picks a value and stores it in `_dist_rollout`. Then, if you have more then 1 process, the global rank zero process overwrites everyone elses value with its copy. 

You have the `try...except ValueError` in case torch.dist is not initalised I guess? You could try replace with

```python
if dist.is_initialized() and dist.get_world_size() > 1:
   dist.broadcast(self._dist_rollout, src=0)
```
Maybe the world_size check is not needed, I would hope a broadcast with a single member would just complete straight away.

One question I have here, should the rollout value be constant across and within all models? I don't know enough about rollout. But here you dont pass a model_comm_group to broadcast so that's what you'd get.

HCookie: Good point about the constant value across all models, 
What would I need for the model_comm_group to be used here?

training/src/anemoi/training/schedulers/rollout/randomise.py
```diff
@@ -0,0 +1,371 @@
...
+        -------
+        int
+            Either broadcasted value or update value from broadcast
+        """
+        self._dist_rollout = torch.tensor([value])
+        try:
+            if dist.get_world_size() > 1:
+                dist.broadcast(self._dist_rollout, src=0)
+        except ValueError:
+            pass
```
HCookie, line 72:
```suggestion
        if dist.is_initialized() and dist.get_world_size() > 1:
                dist.broadcast(self._dist_rollout, src=0)
```

training/src/anemoi/training/schedulers/rollout/randomise.py
cathalobrien: Your call to broadcast would change slightly, and you'd have to pass model_comm_group into the function somehow
```python
- dist.broadcast(self._dist_rollout, src=0)
+ dist.broadcast(self._dist_rollout, group=model_comm_group, group_src=0)
```
Then each process within a `model_comm_group` would get the same rollout value. Each `model_comm_group` could have a different value

sahahner:
This PR is great. Thank you for that! I have invented a few scheduling protocols and most of them work. There are a few edge cases that we might want to support
- The scheduler allows negative steps, which is useful in the case when the user wants to decrease the number of rollout steps. Nevertheless, the actual rollout should never be negative, we would be going in the wrong direction with our forecast. This should be easy to fix by comparing to the minimum rollout provided how we also limit by the maximum rollout provided.
- At the moment, when the counter reaches the maximum rollout step it is still increased internally. I think this should not be the case in order to allow decreasing this number again after some number of epochs/steps.

training/docs/user-guide/training.rst
```diff
@@ -242,19 +242,52 @@ performed on 12 forecast steps (equivalent to 72 hours) incrementally.
...
+
+**Default Rollout Scheduler**
+
+Shown below is the default rollout scheduler in ``config.training``.
+
+.. code:: yaml
+
+   # Dataloader rollout counter can only be updated at the end of each epoch
+   # So updates during an epoch will only be reflected at the end of said epoch.
+   rollout:
```
sahahner, line 277: Should the specifications for training.rollout be indented?

training/src/anemoi/training/train/forecaster.py
```diff
@@ -485,15 +488,15 @@ def _step(
...
+            rollout=int(self.rollout),
             training_mode=True,
             validation_mode=validation_mode,
         ):
             loss += loss_next
             metrics.update(metrics_next)
             y_preds.extend(y_preds_next)
 
-        loss *= 1.0 / self.rollout
+        loss *= 1.0 / int(self.rollout)
```
sahahner: I see `int(self.rollout)` twice in the `_step` function. Does it make sense to ensure that self.rollout is always an int?

HCookie: So that is can be used as for the divide it needs to be cast to an int.
If I add rdiv to the class it will work without the cast.

training/docs/user-guide/training.rst
HCookie: Probably....

training/src/anemoi/training/schedulers/schedulers.py
```diff
@@ -0,0 +1,282 @@
...
+# nor does it submit to any jurisdiction.
+
+from __future__ import annotations
+
+import enum
+from abc import ABC
+from typing import Literal
+from typing import Union
+
+from typing_extensions import Self
```
sahahner, line 17: This is not working for all supported python versions.

training/src/anemoi/training/train/forecaster.py
```diff
@@ -605,6 +608,26 @@ def calculate_val_metrics(
...
+        if "rollout" in checkpoint:
+            self.rollout = instantiate(self.config.training.rollout).get_state_from(checkpoint["rollout"])
+
+    def on_save_checkpoint(self, checkpoint: dict) -> None:
+        checkpoint["rollout"] = self.rollout
+
+    def on_train_batch_end(self, *_) -> None:
+        self.rollout.step()
+
+    def on_train_epoch_end(self, *_) -> None:
```
sahahner, line 653: Can we add the new rollout step to these logger outputs?

HCookie: As in log the rollout when stepped?

sahahner: I mean adding the rollout to these logging statements:
`LOGGER.debug("Stepping Rollout on train epoch end")`

sahahner: I don't think logging all changes to rollout step in the info terminal output is necessary as the rollout is tracked as a model metric.

HCookie: Resolved in cc060b0db307537ba83ed50fb0d4cab4a0252d79

sahahner: Should the print come after the update of the rollout step? LOGGER.info needs some string formatting.
```suggestion
    def on_train_epoch_end(self, *_) -> None:
        if self.trainer.limit_val_batches == 0:
            self.rollout.step_epoch()
            LOGGER.info("Stepping Rollout on train epoch end, rollout: %i", int(self.rollout))

    def on_validation_epoch_end(self, *_) -> None:
        if not self.trainer.sanity_checking:
            self.rollout.step_epoch()
            LOGGER.info("Stepping Rollout on validation epoch end, rollout: %i", int(self.rollout))
```

HCookie: Fixed

training/src/anemoi/training/train/train.py
ssmmnn11: this will clash with the data loader seeding -> we get a base seed at the beginning of the training run. can you check that the runs are still shuffled differently for every epoch? otherwise we need to increment the random seed according to e.g. epoch

training/src/anemoi/training/schedulers/rollout/__init__.py
```diff
@@ -0,0 +1,240 @@
...
+        """Get rollout value as int."""
+        return int(self.rollout)
+
+    def __index__(self) -> int:
+        """Get rollout value as index."""
+        return int(self.rollout)
+
+    def rollout_at(
+        self,
+        step: int | None = None,
```
ssmmnn11, line 74: Does this work now with restaring?

So when my job get canceld at rollout  = 3, and I am halfway through the dataloader, does the module now when I restart that it only needs to do 1/2 of the dataloader, and skip the rest before incrementing the rollout again?

ssmmnn11 (changes requested):
nice, please see comments

training/src/anemoi/training/schedulers/rollout/__init__.py
HCookie: As the rollout is calculated as a product of step and epoch, restarting works just fine from the rollout's perspective. The step and epoch are stored in the last checkpoint, and only incremented when appropriate. Regarding the dataloader part that is out of scope of this one, and will require work on that class.

HCookie:
@ssmmnn11 
> can you check that the runs are still shuffled differently for every epoch?

They are not, the indices are the same for each epoch. 
However, I have been thinking further on the need to reload, I think it may be possible to do it without it, but I'll need to figure out how to broadcast a value to all the dataloader workers, as this is why they do not update. Rebuilding them obviously fixed that but causes other issues.

ssmmnn11:
persistent_workers=False means we have an initial seed and then just draw randon numbers from the generator, hence there should not be any repetition. 

What we could do: before we did not have the epoch information in the dataset. This should now be the case and you could simply add the epoch to the random seed:

https://github.com/ecmwf/anemoi-core/blob/83d72e17b5c8017a1d9a47f75e0848e0bbb080a3/training/src/anemoi/training/data/dataset.py#L224

something like: base_seed = get_base_seed() + self.epoch * 100

though we should probably rename base_seed here to dataset_seed otherwise it might be confusing -> we call refer to the base seed elsewhere in the code.

training/src/anemoi/training/train/train.py
HCookie: Fixed it in upcoming commit hopefully

PR #47, opened 2024-12-20
feat(models): normalization layers

jakob-schloer:
### Describe your changes

This PR makes it possible to switch the implementation of Linear and LayerNorm kernels in the config.

At the moment we use torch.NN implementation for many layers in Anemoi model e.g. torch.nn.layerNorm, torch.NN.linear. This has the advantage of being available out of the box with torch and portable to many systems (CPU, AMD and Nvidia GPUs). However, other layer implementations might be more efficient for certain hardware, or we might want to use a custom layer.

This PR adds the following block to config/model/.yaml:
```
  layer_kernels:
    LayerNorm:
      #_target_: "transformer_engine.pytorch.LayerNorm"
      _target_: "liger_kernel.transformers.rms_norm.LigerRMSNorm"
      #_target_: "torch.nn.LayerNorm" #the default PyTorch implementation
      _partial_: True
      #Any arguments to your chosen function go here e.g.
      #bias: False
    Linear:
      #_target_: "transformer_engine.pytorch.Linear"
      _target_: "torch.nn.Linear"
      _partial_: True
```

You can pass any parameters to your new kernels in the config file, after "partial : True". Hydra tries to load the desired kernel in "models/encoder_processor_decoder.py". If the desired library isn't available, torch currently will fall back to torch.nn..

The calls to torch.nn are then replaced with

```
- self.layer_norm = nn.LayerNorm(normalized_shape=num_channels)
+ self.layer_norm = layer_kernels['LayerNorm'](normalized_shape=num_channels)
```

In the future, this syntax could be extended to replace other layers if required. 

The `ConditionalLayerNorm` (originally implemented by @ssmmnn11) is for instance a custom normalization layer required by the ensemble-based model. 

### Type of change

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

### Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated the documentation and docstrings to reflect the changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have ensured that the code is still pip-installable after the changes and runs
- [x] I have not introduced new dependencies in the inference partion of the model
- [x] I have ran this on single GPU
- [x] I have ran this on multi-GPU or multi-node
- [ ] I have ran this to work on LUMI (or made sure the changes work independently.)
- [ ] I have ran the Benchmark Profiler against the old version of the code


models/src/anemoi/models/layers/block.py
```diff
@@ -330,37 +347,41 @@ def __init__(
...
+        self.lin_key = linear(in_channels, num_heads * self.out_channels_conv)
+        self.lin_query = linear(in_channels, num_heads * self.out_channels_conv)
+        self.lin_value = linear(in_channels, num_heads * self.out_channels_conv)
+        self.lin_self = linear(in_channels, num_heads * self.out_channels_conv, bias=bias)
+        self.lin_edge = linear(edge_dim, num_heads * self.out_channels_conv)  # , bias=False)
 
         self.conv = GraphTransformerConv(out_channels=self.out_channels_conv)
 
-        self.projection = nn.Linear(out_channels, out_channels)
+        # Why does the GraphTransformer not have a layer_norm_mlp like the Transformer?
```
ssmmnn11: it is part of the mlp iirc

models/src/anemoi/models/layers/block.py
```diff
@@ -474,7 +500,7 @@ def __init__(
             **kwargs,
         )
 
-        self.layer_norm2 = nn.LayerNorm(in_channels)
+        self.layer_norm_attention_2 = layer_kernels["LayerNorm"](normalized_shape=in_channels)
```
ssmmnn11: I don't like the naming layer_norm_attention and layer_norm_attention_2

maybe layer_norm_attention_src and layer_norm_attention_dst ?

models/src/anemoi/models/layers/block.py
```diff
@@ -489,9 +515,10 @@ def forward(
         x_skip = x
 
         x = (
-            self.layer_norm1(x[0]),
-            self.layer_norm2(x[1]),
-        )  # Why does this use layer_norm2? And only is a mapper thing?
+            self.layer_norm_attention(x[0]),
+            self.layer_norm_attention_2(x[1]),
+        )  # Why does this use layer_norm_attention_2? And only is a mapper thing?
```
ssmmnn11: we can remove this comment. we have two layernorms here, because we  are doing cross attention and normalise the src and dst nodes.

models/src/anemoi/models/layers/chunk.py
```diff
@@ -37,6 +38,7 @@ def __init__(
         num_layers: int,
         *args,
         activation: str = "GELU",
+        layer_norm: Optional[dict] = None,
```
ssmmnn11: layer_kernels?

models/src/anemoi/models/layers/normalization.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def __init__(
+        self,
+        normalized_shape: Union[int, list, Size],
+        condition_shape: int = 16,
+        w_one_bias_zero_init: bool = True,
+        autocast: bool = True,
+    ):
+        super().__init__()
+        self.norm = nn.LayerNorm(normalized_shape, elementwise_affine=False)  # no learnable parameters
```
ssmmnn11, line 50: I guess this one is a bit more complex, because here we would like to be able to configure the layer kernel inside of the layernorm ... can you think about a way how to do this? maybe overkill for this first implementation.

training/src/anemoi/training/config/model/gnn.yaml
```diff
@@ -4,6 +4,17 @@ num_channels: 512
 model:
   _target_: anemoi.models.models.encoder_processor_decoder.AnemoiModelEncProcDec
 
+layer_kernels:
+    LayerNorm:
+      _target_: torch.nn.LayerNorm #the default PyTorch implementation
```
ssmmnn11: the gnn should run with the autocast layernorm, otherwise the memory requirements will be large.

ssmmnn11 (changes requested):
great work!

models/src/anemoi/models/layers/normalization.py
jakob-schloer: I thought we might define custom normalization layers in the `normalization.py` which we would then specify in the config, e.g. 
```
layer_kernels:
    LayerNorm:
      _target_: anemoi.models.layers.normalization.AutocastLayerNorm
      _partial_: True
```

How would you want to use the AutocastLayerNorm?

FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=47) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=47) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: jakob-schloer<br/>:x: cathalobrien<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=47) it.</sub>

models/src/anemoi/models/layers/normalization.py
ssmmnn11: What I mean is that we might want to use conditional layernorms, but with e.g. a rms norm from liger instead of the pytorch implementation. 

A first workaround would be of course to have multiple conditional layernorms, but with different layernorm implementations. This can also be a future iteration and does not have to be implemented here.

jakob-schloer: I was discussing this with @cathalobrien but we couldn't think of an easy way to do that yet. I would probably do that at a later stage.

training/src/anemoi/training/config/model/gnn.yaml
```diff
@@ -4,6 +4,16 @@ num_channels: 512
 model:
   _target_: anemoi.models.models.encoder_processor_decoder.AnemoiModelEncProcDec
 
+layer_kernels:
+    LayerNorm:
+      _target_: anemoi.models.layers.normalization.AutocastLayerNorm #the default PyTorch implementation
```
ssmmnn11: can you remove #the default ...

ssmmnn11 (changes requested):
One last thing :-)

training/src/anemoi/training/config/model/gnn.yaml
jakob-schloer: Done

ssmmnn11 (approved):
looks good!

PR #50, opened 2024-12-20
chore(training): Add default config files for 2 and 3 level hierarchical processors

JPXKQX:
This PR includes a default config file for hierarchical graphs.


training/src/anemoi/training/config/model/gnn.yaml
```diff
@@ -2,7 +2,7 @@ activation: GELU
 num_channels: 512
 
 model:
-  _target_: anemoi.models.models.encoder_processor_decoder.AnemoiModelEncProcDec
+  _target_: anemoi.models.models.AnemoiModelEncProcDec
```
mchantry, line 5: Why is this changing here?

JPXKQX: Now the different model classes are included in the `__all__` attribute and I think it may be more friendly for the user to access each model class directly from `anemoi.models.models`. But it may have some negative implications I am not aware of, so feel free to raise any concerns

floriankrb (approved):

PR #52, opened 2024-12-27
fix!: Rework Loss Scalings to provide better modularity

sahahner:
Solve the problem explained in issue #7  by refactoring the variable scalings into a general variable scaling and a pressure level scaling.
@mc4117 , @pinnstorm and me came up with a new structure. This PR implements this.

**Changes by this PR**
- collect all available loss scalings in **anemoi/training/losses/scaling**
  - the new modular structure makes it easier to implement new scalings
- move **configuration** of all available scalings into config/training/scalers
  - we provide a list of possible scalers
  - list the scalings in a dictionary you want to apply as scalers in the training_loss configuration
    -> in comparison to before, no changes to the code are necessary to apply additional scalings, but can simply be added to the training-config files
  - the new default config files do **not change the default** scalers to be applied
- **new features** of this PR include
  - define new scalers that are applied to a group of model/pressure level variables
  - the grouping of variables into surface and pressure level variables is no longer defined by the parsing of strings but is defined in config/training/scalers
  - introduce a util function to retrieve this grouping at other places in the code
  - **tendency scaler**: 
    - introduce the ability to scale the losses by the statistical tendencies in the dataset. At infinite precision, this is equivalent to training towards a tendency loss
  - A new node attribute `MaskedPlanarAreaWeights` has been added to compare results between stretched and lam experiments. This new node attribute should be added to your graph configuration files and used in the validation metrics.
- rename loss scalars to scalers, as this is the correct naming for the feature

While this PR does introduce a breaking change to the training-config files, it comes with neat new features, that should be of use to many of us.
We did our best to document the changes to the training-config files as best as possible.

Tasklist
- [x] allow several variable level scaling (i.e. pressure level and model level)
- [x] implement/update tests
- [x] decide: do we want to allow scaling by variable_ref and variable_name, i.e. scale q_50 by q and q_50?
- [x] get variable level and name from dataset metadata if available
- [x] Change of name: loss scalars to scalers.
- [x] move node weights into new scaling submodule

**Breaking Changes**
- change configuration of loss functions
- loss function scalers need to be initialized in config files (both scalers on variable level and on node level)
  - scalers are initialized in in the new config file level training.scalers
- rename loss scalars to scalers (also in the training config file)

----
📚 Documentation preview 📚: https://anemoi-training--52.org.readthedocs.build/en/52/

----
📚 Documentation preview 📚: https://anemoi-graphs--52.org.readthedocs.build/en/52/

----
📚 Documentation preview 📚: https://anemoi-models--52.org.readthedocs.build/en/52/


b8raoult:
Please consider using the knowledge about variables that come from the dataset metadata. See https://github.com/ecmwf/anemoi-transform/blob/7cbf5f3d4baa37453022a5a97e17cc71a5b8ceeb/src/anemoi/transform/variables/__init__.py#L47

sahahner:
> Please consider using the knowledge about variables that come from the dataset metadata. See https://github.com/ecmwf/anemoi-transform/blob/7cbf5f3d4baa37453022a5a97e17cc71a5b8ceeb/src/anemoi/transform/variables/__init__.py#L47

We have given this some thought, and after wanting to use the information from the dataset in the beginning, I have opted for allowing the definition of our own groups here to use different scaling for self-defined groups.
Also, I was also told that it is possible to build datasets without information about the variable types and therefore not to rely on that metadata.
If you have strong opinions on this I am happy to discuss it again.

training/src/anemoi/training/train/forecaster.py
```diff
@@ -98,7 +98,18 @@ def __init__(
 
         self.logger_enabled = config.diagnostics.log.wandb.enabled or config.diagnostics.log.mlflow.enabled
 
-        variable_scaling = self.get_variable_scaling(config, data_indices)
+        variable_scaling = GeneralVariableLossScaler(
+            config.training.variable_loss_scaling,
+            data_indices,
+        ).get_variable_scaling()
+
+        # Instantiate the pressure level scaling class with the training configuration
```
mc4117: I don't know if this is possible but I wonder if here we could instantiate from a list of scalars rather than specific ones? I think this is how it is done for the validation metrics

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+        for variable_name, idx in self.data_indices.internal_model.output.name_to_index.items():
+            _, variable_ref, _ = self.get_variable_group(variable_name)
+            # Apply variable scaling
+            variable_loss_scaling[idx] = self.scaling_config.get(
+                variable_ref,
+                1.0,
+            )
+            # TODO(all): do we want to allow scaling by variable ref and variable name?,
```
mc4117: I'm not sure what the difference between variable_ref and variable_name is here?

sahahner: `variable_ref` refers to this variable in the config weight list. 

So for q_10 the values would be:
- `variable_name=q_10`
- `variable_ref=q` (without the variable level)

I will add a few comments.

training/src/anemoi/training/train/forecaster.py
sahahner: Yes, this would be useful, as we want to allow more scalar methods for model levels/tendency/etc.
I will have a look.

training/src/anemoi/training/train/forecaster.py
```diff
@@ -98,7 +98,18 @@ def __init__(
...
+        pressurelevelscaler = instantiate(
+            config.training.pressure_level_scaler,
```
mc4117: I think this config location is wrong? It should be training.variable_loss_scaling.pressure_level_scalar

sahahner: Indeed, in the config file, the pressure_level_scalar should be defined directly in training as before.

FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=52) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=52) before we can accept your contribution.<br/>**0** out of **2** committers have signed the CLA.<br/><br/>:x: sahahner<br/>:x: mc4117<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=52) it.</sub>

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,221 @@
...
+            if isinstance(variables, str):
+                variables = [variables]
+            for variable in variables:
+                self.group_variables[variable] = group
+        self.default_group = self.scaling_config.variable_groups.default
+
+    @abstractmethod
+    def get_variable_scaling(self) -> np.ndarray: ...
+
+    def get_variable_group(self, variable_name: str) -> tuple[str, str, int]:
```
mc4117: I think maybe we should put this function in an utils. I only say this because I want to use it in another application outside of the loss scalings

JPXKQX:
Hi, I would like to know what you think about making all scalers explicit in the config file. Something similar to the `additional_scalers:` field, but including not only the scalers per variable, but also the `node_loss_weight`,... The positive aspect I see is that there would be more homogeneity in the scalers defined in the `metrics/loss` fields.

training/src/anemoi/training/train/scaling.py
sahahner: Yes, it is very similar to the function from the plotting function. Will make sure to have only one function that can be used in both cases.

mc4117:
> Hi, I would like to know what you think about making all scalers explicit in the config file. Something similar to the `additional_scalers:` field, but including not only the scalers per variable, but also the `node_loss_weight`,... The positive aspect I see is that there would be more homogeneity in the scalers defined in the `metrics/loss` fields.

Seems like a good idea! Would you like to add this in this PR?

JPXKQX:
> > Hi, I would like to know what you think about making all scalers explicit in the config file. Something similar to the `additional_scalers:` field, but including not only the scalers per variable, but also the `node_loss_weight`,... The positive aspect I see is that there would be more homogeneity in the scalers defined in the `metrics/loss` fields.
> 
> Seems like a good idea! Would you like to add this in this PR?

I’m not sure what the best approach is. On the one hand, adding more work to this PR would increase its complexity, which might make it more logical to address this refactor in a future PR. On the other hand, this PR already introduces some changes to the configs, and the future PRs would also involve changes to the configs. From this, it might be better to have 1 PR and communicate all the changes to users at once. What do you think?

pinnstorm:
> > > Hi, I would like to know what you think about making all scalers explicit in the config file. Something similar to the `additional_scalers:` field, but including not only the scalers per variable, but also the `node_loss_weight`,... The positive aspect I see is that there would be more homogeneity in the scalers defined in the `metrics/loss` fields.
> > 
> > 
> > Seems like a good idea! Would you like to add this in this PR?
> 
> I’m not sure what the best approach is. On the one hand, adding more work to this PR would increase its complexity, which might make it more logical to address this refactor in a future PR. On the other hand, this PR already introduces some changes to the configs, and the future PRs would also involve changes to the configs. From this, it might be better to have 1 PR and communicate all the changes to users at once. What do you think?

I'm happy for it to be included in this PR! Not sure if @sahahner or @mc4117 have other views?

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -109,33 +110,49 @@ lr:
...
+    y_intercept: 0.2
+    slope: 0.001
+    scale_dim: -1 # dimension on which scaling applied
+    name: "variable_pressure_level"
+  # tendency scalers
+  # scale the prognostic losses by the stdev of the variable tendencies (e.g. the 6-hourly differences of the data)
+  # useful if including slow vs fast evolving variables in the training (e.g. Land/Ocean vs Atmosphere)
+  # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
+  # stdev tendency scaler
+  # - _target_: anemoi.training.data.scaling.StdevTendencyScaler
```
mc4117: These should be train.scaling I think

pinnstorm: Good catch! I have just pushed an update now 🙏

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,285 @@
...
+        -------
+        str
+            Group of the variable given in the training-config file.
+        str
+            Variable reference which corresponds to the variable name without the variable level
+        str
+            Variable level, i.e. pressure level or model level
+
+        """
+        split = variable_name.split("_")
```
floriankrb: Parsing the variable name ? There must be another way to get the pressure/model levels. If not, we need to implement it.

floriankrb: as an alternative, you could enumerate the whole list (or use wildcard) in the config:
```
variable_groups:
    default: sfc
    pl: ['q_*', 't_*', , ...]
```

But the best thing is to get the information from the dataset metadata.

sahahner: I see that getting information from the metadata is convenient. You made the point earlier that the information might not always be available. I implemented the retrieval of information from the metadata now, falling back to the parsing of the variable name in case it is not available. (also answering comment from @b8raoult from earlier)

We need to keep in might, that the variable metadata in our datasets is quite mars specific and accessed via 
`metadata["dataset"]["variables_metadata"][variable_name]["mars"]["param"]`
Is that correct or is the information somewhere else? 

**Attention**: accessing information from the metadata might break backwards compatibility for some cases. I am not sure of all the scenarios and whether this might turn into an issue.

training/src/anemoi/training/train/scaling.py
sahahner: resolved by af10173b93117bcfbcc9e65a9c3d31ba5e22f5d8

training/src/anemoi/training/train/forecaster.py
```diff
@@ -95,10 +96,38 @@ def __init__(
...
+                        scaling_config=config.training.variable_loss_scaling,
+                        data_indices=data_indices,
+                        statistics=statistics,
+                        statistics_tendencies=statistics_tendencies,
+                    )
+                    if scalar_config["name"] == "tendency"
+                    else instantiate(
+                        scalar_config,
+                        scaling_config=config.training.variable_loss_scaling,
+                        data_indices=data_indices,
```
HCookie: Given that the tendency scalar is a subclass of the same generic Scalar base class this overly specific `if` should be removed. My suggestion would be to add to the `statistics, & tendencies` to the base class and leave the usage up to the implementation.

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,316 @@
...
+
+import logging
+import warnings
+from abc import ABC
+from abc import abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from omegaconf import DictConfig
+    from anemoi.models.data_indices.collection import IndexCollection
```
HCookie: This will raise an issue once the CI is back, this block should be after any other imports.

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,316 @@
...
+class BaseVariableLossScaler(ABC):
+    """Configurable method converting variable to loss scaling."""
+
+    def __init__(
+        self,
+        scaling_config: DictConfig,
+        data_indices: IndexCollection,
+        metadata_variables: dict | None = None,
+    ) -> None:
+        """Initialise Scaler.
```
HCookie: Could it make sense to further 'genericsize' this class? `BaseLossScalar`?

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -109,33 +110,49 @@ lr:
 # Variable loss scaling
 # 'variable' must be included in `scalars` in the losses for this to be applied.
 variable_loss_scaling:
+  variable_groups:
+    default: sfc
+    pl: [q, t, u, v, w, z]
```
HCookie: A comment on the use of this would be useful here

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,316 @@
...
+        self.group_variables = {}
+        for group, variables in self.variable_groups.items():
+            if isinstance(variables, str):
+                variables = [variables]
+            for variable in variables:
+                self.group_variables[variable] = group
+        self.default_group = self.scaling_config.variable_groups.default
+        self.metadata_variables = metadata_variables
+
+        self.ExtractVariableGroupAndLevel = ExtractVariableGroupAndLevel(
```
HCookie: The variable should not be camel case, and instead snake_case

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,316 @@
...
+            self.group_variables,
+            self.metadata_variables,
+            self.default_group,
+        )
+
+    @abstractmethod
+    def get_variable_scaling(self) -> np.ndarray: ...
```
HCookie: A quick documentation here on what this method is expected to return would be good

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,316 @@
...
+            Variable level, i.e. pressure level or model level
+
+        """
+        return self.ExtractVariableGroupAndLevel.get_group_and_level(
+            variable_name,
+        )
+
+
+class GeneralVariableLossScaler(BaseVariableLossScaler):
+    """General scaling of variables to loss scaling."""
```
HCookie: This documentation is hard to read, maybe I'm just missing the point.

HCookie (changes requested):

training/src/anemoi/training/train/scaling.py
sahahner: This class focuses on the loss scaling at the variable dimension, i.e. we pass the variable loss scaling configuration from the training config file. Only then the variable groupings are relevant. 
I like the idea of introducing another parent class for all future scalars. Possibly then @JPXKQX can use it for the node-weights?

training/src/anemoi/training/train/scaling.py
sahahner: That is true. I have written something more significant.

training/src/anemoi/training/train/scaling.py
sahahner: In commit [d0046fa](https://github.com/ecmwf/anemoi-core/pull/52/commits/d0046fa8ba330deaa2480c3eca48a689c844f3e3) I introduce another parent class. What do you think about that?

training/src/anemoi/training/config/training/default.yaml
sahahner: Added.

training/src/anemoi/training/train/scaling.py
HCookie: A great start, once we add more classes we can revise the scope of functionality.

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,335 @@
...
+
+        for variable_name, idx in self.data_indices.internal_model.output.name_to_index.items():
+            _, variable_ref, _ = self.get_variable_group(variable_name)
+            # Apply variable scaling by base variable name (variable_ref: variable name without variable level)
+            variable_loss_scaling[idx] = self.scaling_config.get(
+                variable_ref,
+                1.0,
+            )
+            # TODO(all): do we want to allow scaling by variable_ref and variable_name?
+            # i.e. scale q_50 by value for q_50 AND q
```
sahahner: Are there any thoughts on this? At the moment, the code allows the scaling of, i.e. `q_50`, by a factor for `q` (`variable_ref`) and another one for `q_50` (`variable_name`). While we do not allow this functionality now (on develop), it might be wanted in the future, so I do not see need to forbid it any longer.

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -109,33 +110,52 @@ lr:
...
+  tp: 0.025
+  cp: 0.0025
+additional_scalars:
+  # pressure level scalar
+  - _target_: anemoi.training.train.scaling.ReluVariableLevelScaler
+    group: pl
+    y_intercept: 0.2
+    slope: 0.001
+    scale_dim: -1 # dimension on which scaling applied
+    name: "variable_pressure_level"
```
HCookie: I would suggest that instead of providing name as an attribute, `additional_scalars` should be `dict[str, dict]`,
so,
```yaml
additional_scalars:
   variable_pressure_level:
       _target_: ------
       ....
```
This then will be used to name the scalar for the loss function to choose between.

training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -0,0 +1,76 @@
...
+        self,
+        group_variables: dict,
+        metadata_variables: dict | None = None,
+        default_group: str = "sfc",
+    ) -> None:
+        self.group_variables = group_variables
+        self.metadata_variables = metadata_variables
+        self.default_group = default_group
+
+    def get_group_and_level(self, variable_name: str) -> tuple[str, str, int]:
```
sahahner, line 47: Apply this new utils function to the `get_val_metric_ranges` in forecaster as well. Thanks, @HCookie for the pointer.

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,339 @@
+# (C) Copyright 2024 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
```
HCookie: Given our expectation that more scalars will be created this may be better placed into a submodule

```text
scaling/
    __init__.py
    variable.py
    tendency.py
    ....
```

training/src/anemoi/training/config/training/default.yaml
JPXKQX: This would be consistent with the way we create the nodes in anemoi-graphs.

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -109,33 +110,53 @@ lr:
 # Variable loss scaling
 # 'variable' must be included in `scalars` in the losses for this to be applied.
 variable_loss_scaling:
+  # Variable groups definition for scaling by variable level.
+  # The variable level scaling methods are defined under additional_scalars
+  # The default group is appended as prefix to the metric of all variables not asigned to a group.
```
jakob-schloer: I would add here, that the group "default" is required.

jakob-schloer:
> > > Hi, I would like to know what you think about making all scalers explicit in the config file. Something similar to the `additional_scalers:` field, but including not only the scalers per variable, but also the `node_loss_weight`,... The positive aspect I see is that there would be more homogeneity in the scalers defined in the `metrics/loss` fields.
> > 
> > 
> > Seems like a good idea! Would you like to add this in this PR?
> 
> I’m not sure what the best approach is. On the one hand, adding more work to this PR would increase its complexity, which might make it more logical to address this refactor in a future PR. On the other hand, this PR already introduces some changes to the configs, and the future PRs would also involve changes to the configs. From this, it might be better to have 1 PR and communicate all the changes to users at once. What do you think?

I fully agree with @JPXKQX. I personally think that all config keywords related to the loss and its scaling are a bit scattered in the training. Why is variable scaling under training while nodeweights is under losses? In my opinion everything should be under losses. 
I would suggest bringing the restructuring into this PR. If this PR goes in, it will break old configs, and the restructuring will again break old configs:

Ideally, the the config could look something like that:
```	
loss_scaling:
	default: 1
	groups:
		default: sfc
		pl: []
	scalers:
		- _target_: anemoi.training.losses.scaling.ConstVariableScaler
			default: 1
			variables:
				  q: 0.6 #1
				  t: 6   #1
				  u: 0.8 #0.5
				  v: 0.5 #0.33
				  w: 0.001
				  z: 12  #1
				  sp: 10
				  10u: 0.1
				  10v: 0.1
				  2d: 0.5
				  tp: 0.025
				  cp: 0.0025
		- _target_: anemoi.training.losses.scaling.ReluVariableLevelScaler
		    group: pl
		    y_intercept: 0.2
		    slope: 0.001
		    scale_dim: -1 # dimension on which scaling applied
		    name: "variable_pressure_level"
		- target_: anemoi.training.losses.nodeweights.GraphNodeAttribute
		  target_nodes: ${graph.data}
		  node_attribute: area_weight

**Review by jakob-schloer — DISMISSED** [review]
Great effort! The flexible scaling of the loss is really nice. 

However, in its current version all config keywords related to the loss and its scaling as well as the code is a bit scattered.  Why is variable scaling under training while nodeweights is under losses? In my opinion everything should be under losses. See my comment above.

File: training/src/anemoi/training/train/forecaster.py [hunk 19]
```diff
@@ -95,12 +100,38 @@ def __init__(
...
+        variable_scaling = GeneralVariableLossScaler(
+            config.training.variable_loss_scaling,
+            data_indices,
+            metadata["dataset"].get("variables_metadata"),
+        ).get_scaling()
+
+        # Instantiate the pressure level scaling class with the training configuration
+        config_container = OmegaConf.to_container(config.training.additional_scalars, resolve=False)
+        if isinstance(config_container, list):
+            scalar = [
```
jakob-schloer: The variable naming is not very intuitive. I know that you did not introduce `self.scalar`, but maybe this is a good chance to rename it to `self.loss_scaling` or something like that.

training/src/anemoi/training/train/forecaster.py
```diff
@@ -120,6 +151,9 @@ def __init__(
             "loss_weights_mask": ((-2, -1), torch.ones((1, 1))),
             "limited_area_mask": (2, limited_area_mask),
         }
+        # add addtional user-defined scalars
+        [self.scalars.update({scale.name: (scale.scale_dim, scale.get_scaling())}) for scale in scalar]
```
jakob-schloer: I'm not sure if this is the right place, but it would be great to print or log (to MLflow) the combined scaling per variable and pressure level. What do you think?

mc4117: My only worry with this, is that the more params you add to mlflow the longer it takes to load the experiment. We could add it to the log in debug mode, though?

training/src/anemoi/training/train/forecaster.py
mc4117: I'm happy to change this as long as you're ok with this @HCookie ?

training/src/anemoi/training/config/training/default.yaml
sahahner: This makes sense. I can have a look at it.

training/src/anemoi/training/train/forecaster.py
HCookie: Go for it, potentially could make sense to go for an even more generic name.

training/src/anemoi/training/train/scaling.py
```diff
@@ -0,0 +1,335 @@
...
+        for variable_name, idx in self.data_indices.internal_model.output.name_to_index.items():
+            _, variable_ref, _ = self.get_variable_group(variable_name)
+            # Apply variable scaling by base variable name (variable_ref: variable name without variable level)
+            variable_loss_scaling[idx] = self.scaling_config.get(
+                variable_ref,
+                1.0,
+            )
+            # TODO(all): do we want to allow scaling by variable_ref and variable_name?
+            # i.e. scale q_50 by value for q_50 AND q
+            if variable_ref != variable_name:
```
sahahner: == instead of != and assert if trying to scale pressure level variable by its actual name

training/src/anemoi/training/train/forecaster.py
sahahner: I think this PR is already quite big. I see why this can be benifitial and would opt for working on this in another PR as printing this list requires no changes to the config file. Or do you @HCookie see a quick way to print all the scalers for dimension -1? Since the NaN mask covers dimensions (-2,-1) I didn't see a quick way to print the final weights apllied only on the variable dimension.

training/src/anemoi/training/train/scaling.py
sahahner: != was actually correct. Now scaling `q_50` by both `q` and `q_50` is raising an error.

training/src/anemoi/training/train/forecaster.py
HCookie: Subsetting over -1, should provide the right scalar for printing across the variable dimension.

training/src/anemoi/training/scaling/__init__.py
HCookie, line 1: Should scaling be a top level or under the losses module?
Open to discussions

training/src/anemoi/training/config/training/scalers/scalers.yaml
```diff
@@ -0,0 +1,58 @@
...
+  var_tendency:
+    _target_: anemoi.training.scaling.variable_tendency.VarTendencyScaler
+    scale_dim: -1 # dimension on which scaling applied
+
+  node_weights:
+    _target_: anemoi.training.losses.nodeweights.GraphNodeAttribute
+    target_nodes: ${graph.data}
+    node_attribute: area_weight
+    scale_dim: 2 # dimension on which scaling applied
+
```
HCookie, line 57: The class being intialised here doesn't contain a `scale_dim` key.
Should this be abstracted into a general scaler by node attribute?

training/src/anemoi/training/train/forecaster.py
```diff
@@ -111,25 +134,24 @@ def __init__(
 
         # Kwargs to pass to the loss function
```
HCookie: This should be added to the new structure.

training/src/anemoi/training/config/training/scalers/scalers.yaml
```diff
@@ -0,0 +1,58 @@
...
+    scale_dim: -1 # dimension on which scaling applied
+  var_tendency:
+    _target_: anemoi.training.scaling.variable_tendency.VarTendencyScaler
+    scale_dim: -1 # dimension on which scaling applied
+
+  node_weights:
+    _target_: anemoi.training.losses.nodeweights.GraphNodeAttribute
+    target_nodes: ${graph.data}
+    node_attribute: area_weight
+    scale_dim: 2 # dimension on which scaling applied
```
HCookie: This class doesn't have a scale_dim attribute.
It may also be useful to add a general scale by node attribute scaler.

training/src/anemoi/training/scaling/variable_level.py
```diff
@@ -0,0 +1,159 @@
...
+            y_intercept=1.0,
+            slope=0.0,
+            scale_dim=scale_dim,
+        )
+
+    @staticmethod
+    def get_level_scaling(variable_level: float) -> np.ndarray:
+        del variable_level  # unused
+        # no scaling, always return 1.0
+        return 1.0
```
HCookie: If this is supposed to scale by 1, at all time, why not remove the options for a user, and hardcode in the intercept and slope?

sahahner: I decided on that when I introduced the scaling, because I don't want the parameters to be wrong in the config file that is uploaded to mlflow as this might lead to confusion.

training/src/anemoi/training/scaling/__init__.py
sahahner: Yes, good comment. I though about that as well. In the end, I opted for top level but now that you say it, it actually does make sense to move in under the losses module as the scaling is for the losses.

training/src/anemoi/training/config/training/scalers/scalers.yaml
sahahner: Not yet. Refactor is still ongoing.

training/src/anemoi/training/scaling/variable_level.py
HCookie: If the options are removed, they can never be given in the config, and thus can't be wrong?

sahahner: As far as I know the config file is pushed to mlflow before the call of this function.

sahahner: Ah, you recommend to remove them from the function definition?

HCookie: Yep, they are not needed for this scalar

training/src/anemoi/training/train/forecaster.py
sahahner: Final variable scalings are now printed by the logger in debug mode.

training/CHANGELOG.md
HCookie, line 1: The changelog no longer needs to be manually updated.

HCookie:
Are we sure of the rename of scalar to scaler?

https://en.wikipedia.org/wiki/Scalar
Vs
https://en.wikipedia.org/wiki/Scaler?wprov=sfti1

pinnstorm:
> Are we sure of the rename of scalar to scaler?
> 
> https://en.wikipedia.org/wiki/Scalar Vs https://en.wikipedia.org/wiki/Scaler

Its a good question 😅. I think we came down on "scaler" as something that "scales" (adjusts, regulates) the losses. Rather than a "scalar" quantity. But I think either can work so am happy either way! Will be nice to try and get this merged in soon 👀

training/src/anemoi/training/train/forecaster/forecaster.py
```diff
@@ -581,62 +381,36 @@ def calculate_val_metrics(
...
-                        )
-                        raise ValueError(exception_msg)
-
-                    metrics[f"{metric_name}/{mkey}/{rollout_step + 1}"] = metric(
-                        y_pred_postprocessed,
-                        y_postprocessed,
-                        scalar_indices=[..., indices],
+                metric_step_name = f"{metric_name}/{mkey}/{rollout_step + 1}"
+                if len(metric.scaler.subset_by_dim(TensorDim.VARIABLE)):
+                    exception_msg = (
```
jakob-schloer, line 405: FYI: This line fails for me when specifying wrong `scalers` in the `validation_metrics`. It would not raise the exception as intended.

JPXKQX: I think this should be fixed now. `subset_by_dim()` expects an integer, so the argument should be `TensorDim.VARIABLE.value`.

training/CHANGELOG.md
```diff
@@ -95,6 +95,9 @@ Keep it human-readable, your future self will thank you!
 
 ## [Unreleased](https://github.com/ecmwf/anemoi-training/compare/0.3.2...HEAD)
 
+### Add
+
+- Order parameters in loss-plot by pressure/model level if possible [#55](https://github.com/ecmwf/anemoi-core/pull/55)
 ### Fixed
```
HCookie: Changelog no longer needs updating

training/docs/modules/losses.rst
```diff
@@ -68,26 +68,95 @@ deterministic:
...
+i.e. model or pressure levels for upper air variables. The variable
+level scalers are applied to groups that are defined under
+`scalers.variable_groups`.
+
+For a pressure level scaler applied to all pressure level variables the
+configuration would look like this:
+
+.. code:: yaml
+
+   pressure_level:
```
HCookie, line 129: Might be useful showing where in the config this is set

training/docs/modules/losses.rst
```diff
@@ -68,26 +68,95 @@ deterministic:
...
+
+Variable Groups
+===============
+
+Define a default group and a list of groups to be used in the variable
+level scalers.
+
+.. code:: yaml
+
+   # Variable groups to be used in the variable level scalers
```
HCookie, line 147: Same point of where

training/docs/modules/losses.rst
```diff
@@ -129,17 +198,16 @@ By default, only `all` is kept in the normalised space and scaled.
 ***********************
 
 Additionally, you can define your own loss function by subclassing
-``BaseWeightedLoss`` and implementing the ``forward`` method, or by
-subclassing ``FunctionalWeightedLoss`` and implementing the
-``calculate_difference`` function. The latter abstracts the scaling, and
-node weighting, and allows you to just specify the difference
-calculation.
+``BaseLoss`` and implementing the ``forward`` method, or by subclassing
+``BaseLoss`` and implementing the ``calculate_difference`` function. The
```
HCookie:
```suggestion
``FunctionalLoss`` and implementing the ``calculate_difference`` function. The
```

training/docs/modules/losses.rst
```diff
@@ -129,17 +198,16 @@ By default, only `all` is kept in the normalised space and scaled.
...
+latter abstracts the scaling, and node weighting, and allows you to just
+specify the difference calculation.
 
 .. code:: python
 
-   from anemoi.training.losses.weightedloss import FunctionalWeightedLoss
+   from anemoi.training.losses.weightedloss import BaseLoss
 
-   class MyLossFunction(FunctionalWeightedLoss):
+   class MyLossFunction(BaseLoss):
```
HCookie:
```suggestion
   class MyLossFunction(FunctionalLoss):
```

HCookie (commented):
Amazing work all, I really like how this turned out. This will unlock some exciting avenues

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -250,33 +251,36 @@ training:
...
+      slope: 0.001
+    node_weights:
+      _target_: anemoi.training.losses.scalers.GraphNodeAttributeScaler
+      nodes_name: ${graph.data}
+      nodes_attribute_name: area_weight
+      norm: unit-sum
   metrics: []
+  variable_groups:
+    default: sfc
+    pl: [q, t, u, v, w, z]
```
MeraX: Dear @JPXKQX et al., thank you very much for updating the AICON test case.

I have a question for clarification: Is the `ReluVariableLevelScaler` actually used for any variable in this version of the configuration? I'm just wondering as the variables in this list are not part of the dataset used in this test.

mc4117: Hi, I think you're right and at the moment if q, t, u, v, w and z are not in your dataset, then this pressure level scaler is simply not being applied. Would it be correct to change this pl group just to P?

MeraX: Yes, I think we could even use all atmospheric variables, i.e.  `P, T, U, V`. And would it be possible to rename the group to `ml`, as the ICON data is based on model levels?

HCookie (approved):

PR #55, opened 2024-12-30
Training reorder parameter names for plot

sahahner:
In the newer versions of anemoi-datasets from MARS the variables are not ordered by pressure level anymore but by alphabet, meaning that the order in the loss-plot is also by alphabet (100 is then "smaller" than 20). However, the analysis of the loss weights is more intuitive when ordered by pressure level (or model level).
loss_rstep_rstep00_rank0_epoch085

This PR sorts the variables (for this specific plotting function) in one group by the variable level (if available) interpreting them as digits not strings. 

Screenshot 2025-01-29 at 17 40 56


training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -824,6 +828,24 @@ def automatically_determine_group(name: str) -> str:
...
+        position (variable level), then by the original string.
+        """
+        data = self.parameter_names
+
+        def custom_sort_key(index: int) -> tuple:
+            s = data[index]  # Access the element by index
+            parts = s.split("_")
+            alpha_part = s if len(parts) == 1 else s[: -len(parts[-1]) - 1]
+            numeric_part = int(parts[-1]) if len(parts) > 1 and parts[-1].isdigit() else float("inf")
+            return (alpha_part, numeric_part, s)
```
mc4117: Why do you need to return s here too?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -824,6 +828,24 @@ def automatically_determine_group(name: str) -> str:
             legend_patches,
         )
 
+    def argsort_name_variablelevel(self) -> list[int]:
```
mc4117: Would it be worth putting this in to a callback or loss utils? I can see this function being useful for other callbacks in the future

training/src/anemoi/training/diagnostics/callbacks/plot.py
sahahner: If `alpha_part` and `numeric part` are the same for two different variables, the entire variable name `s` is used for sorting.

training/src/anemoi/training/diagnostics/callbacks/plot.py
sahahner: That's right. Will do that.

sahahner: Should be done after PR https://github.com/ecmwf/anemoi-core/pull/52, where this function is implemented. While for grouping in plots the variable name parsing is still relevant, the variable level should indeed be taken from metadata if possible.

mc4117 (approved):
Looks good to me!

PR #66, opened 2025-01-08
Models: GraphtransformerProcessor chunking

japols:
## Describe your changes

This PR adds chunking for the GraphTransformerProcessorBlock to reduce memory usage in inference. The functionality is equivalent to the GraphTransformerMapperBlock chunking and uses the same env variable `ANEMOI_INFERENCE_NUM_CHUNKS` to control chunking behaviour. 

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated the documentation and docstrings to reflect the changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have ensured that the code is still pip-installable after the changes and runs
- [x] I have not introduced new dependencies in the inference partion of the model
- [x] I have ran this on single GPU
- [x] I have ran this on multi-GPU or multi-node
- [ ] I have ran this to work on LUMI (or made sure the changes work independently.)
- [ ] I have ran the Benchmark Profiler against the old version of the code

## Tag possible reviewers
@ssmmnn11 @gabrieloks

### DIFF
--- a/models/src/anemoi/models/layers/block.py
+++ b/models/src/anemoi/models/layers/block.py
@@ -625,11 +625,37 @@ class GraphTransformerProcessorBlock(GraphTransformerBaseBlock):
             ), "Only batch size of 1 is supported when model is sharded across GPUs"
 
         query, key, value, edges = self.shard_qkve_heads(query, key, value, edges, shapes, batch_size, model_comm_group)
-        out = self.conv(query=query, key=key, value=value, edge_attr=edges, edge_index=edge_index, size=size)
+
+        num_chunks = self.num_chunks if self.training else NUM_CHUNKS_INFERENCE
+
+        if num_chunks > 1:
+            # split 1-hop edges into chunks, compute self.conv chunk-wise and aggregate
+            edge_attr_list, edge_index_list = sort_edges_1hop_chunks(
+                num_nodes=size, edge_attr=edges, edge_index=edge_index, num_chunks=num_chunks
+            )
+            out = torch.zeros((x[1].shape[0], self.num_heads, self.out_channels_conv), device=x[1].device)
+            for i in range(num_chunks):
+                out += self.conv(
+                    query=query,
+                    key=key,
+                    value=value,
+                    edge_attr=edge_attr_list[i],
+                    edge_index=edge_index_list[i],
+                    size=size,
+                )
+        else:
+            out = self.conv(query=query, key=key, value=value, edge_attr=edges, edge_index=edge_index, size=size)
+
         out = self.shard_output_seq(out, shapes, batch_size, model_comm_group)
-        out = self.projection(out + x_r)
+
+        # compute out = self.projection(out + x_r) in chunks:
+        out = torch.cat([self.projection(chunk) for chunk in torch.tensor_split(out + x_r, num_chunks, dim=0)], dim=0)
 
         out = out + x_skip
         nodes_new = self.node_dst_mlp(out) + out
+        # compute nodes_new = self.node_dst_mlp(out) + out in chunks:
+        nodes_new = torch.cat(
+            [self.node_dst_mlp(chunk) + chunk for chunk in out.tensor_split(num_chunks, dim=0)], dim=0
+        )
 
         return nodes_new, edge_attr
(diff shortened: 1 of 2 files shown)


ssmmnn11:
Hi Jan, thank you for adding this, very nice. Looking at the code it seems that some parts do the same but look slightly different. I was wondering if this would also be a good opportunity to reduce code duplication between GraphTransformerProcessorBlock and GraphTransformerMapperBlock. Maybe encapsulated in a common routine? I think differences are

dim of x_skip
providing size (I think that goes in the block though, ... need to check)
and the part of updating the source nodes. 

diff1

japols:
> Hi Jan, thank you for adding this, very nice. Looking at the code it seems that some parts do the same but look slightly different. I was wondering if this would also be a good opportunity to reduce code duplication between GraphTransformerProcessorBlock and GraphTransformerMapperBlock. Maybe encapsulated in a common routine? I think differences are
> 
> dim of x_skip providing size (I think that goes in the block though, ... need to check) and the part of updating the source nodes.

I moved the common "attention" part to the GraphTransformerBaseBlock.

HCookie:
What's the connection between this PR and the shard everything one?

japols:
> What's the connection between this PR and the shard everything one?

There is none, I can make it ready for merge

japols:
Waiting for [PR#182](https://github.com/ecmwf/anemoi-core/pull/182) to be merged.

models/src/anemoi/models/layers/block.py
```diff
@@ -440,6 +462,39 @@ def shard_qkve_heads(
...
+    ) -> Tensor:
+        conv_size = size if isinstance(size, tuple) else None
+
+        if num_chunks > 1:
+            # split 1-hop edges into chunks, compute self.conv chunk-wise and aggregate
+            edge_attr_list, edge_index_list = sort_edges_1hop_chunks(
+                num_nodes=size, edge_attr=edges, edge_index=edge_index, num_chunks=num_chunks
+            )
+            # shape: (num_nodes, num_heads, out_channels_conv)
+            out = torch.zeros((*query.shape[:2], self.out_channels_conv), device=query.device)
```
ssmmnn11: out = torch.zeros((*query.shape[:-1], self.out_channels_conv), device=query.device)

models/src/anemoi/models/layers/block.py
```diff
@@ -440,6 +462,39 @@ def shard_qkve_heads(
...
+        self,
+        query: Tensor,
+        key: Tensor,
+        value: Tensor,
+        edges: Tensor,
+        edge_index: Adj,
+        size: Union[int, tuple[int, int]],
+        num_chunks: int,
+    ) -> Tensor:
+        conv_size = size if isinstance(size, tuple) else None
```
ssmmnn11: the size of part is now a bit confusing to me. the routine requires size as either int or tuple of int, but we check for None as well. why not keep it optional (type Size, from torch_geometric.typing import Size) like before across all layers and just hand it down to self.conv? or what has changed now? Is it for the 1hop chunking? Could we make sure that we pass around the correct size in the processor instead, and add an assert if it is missing?

japols: With chunking, size is not optional anymore because it is required by sort_edges_1hop_chunks as either an int or tuple[int, int] for the bipartite case.
The only case in which size=None would work is the non-bipartite (i.e. GTProc) case with num_chunks=1. I would rather make it non-optional instead of the assert != None for that special case.
Changed it to `conv_size = (size, size) if isinstance(size, int) else size` which is more explicit than using None for the non-bipartite case.

ssmmnn11: Sounds good. But can we make sure that size gets the same type hint everywhere?

ssmmnn11 (dismissed):
nice!

mchantry (changes requested):
@japols super work, could these new flags feature in the documentation please.
Am I right that we would expect this to work on AMD?

japols:
> @japols super work, could these new flags feature in the documentation please. Am I right that we would expect this to work on AMD?

added docs and yes, this works regardless of hardware

mchantry (approved):
Great contribution, thanks.

PR #67, opened 2025-01-08
Feature/transformer sequence sharding

japols:
This PR adds a new sharding strategy shard_sequence for the transformer processor.

The current implementation (shard_heads) alternates between sharding across the sequence to sharding across heads for the sliding window attention mechanism. This requires two all-to-all communication steps per layer.

The shard_sequence strategy simplifies this process by keeping a sequence shard on each GPU and computing the sliding window attention locally. This requires a halo communication to exchange overlapping window segments (halos) between neighboring sequence shards.

Instead of 2 all-to-all communication steps per layer, the halo exchange only requires a single point-to-point communication between neighbouring GPUs, reducing communication time and improving scalability of model sharding across multiple GPUs.

The following benchmarking results show that using a 2 neighbor all-to-all (orange) is the best communication strategy to implement the halo exchange which consistently outperforms the old head-sharding strategy (blue):

sharding_strategies 1

This is an isolated fwd+bwds pass of 16 transformer layers with o96 input shapes, 1024 channels.

For a full training run on n320, o96 hidden we get the following increases in throughput (aligning with the benchmark results):

<!DOCTYPE html>
GPUs/Model | sharding strategy | avg time/batch (s)
-- | -- | --
2 | shard_heads | 1.38495
2 | shard_sequence | 1.29771
4 | shard_heads | 0.72034
4 | shard_sequence | 0.69254

[mlflow](https://mlflow.ecmwf.int/#/metric?runs=%5B%22ff99c1c794be4c69849ca6ad7e98e21e%22,%222fb2e79ac56c4fcea0d33d05569098c8%22,%2248e3ec3a3e854702adfbd29622fac8e9%22,%22d1b8c835c9cc4fc9b40e014bc10f7333%22%5D&metric=%22train_wmse_step%22&experiments=%5B%2245%22%5D&plot_metric_keys=%5B%22train_wmse_step%22%5D&plot_layout=%7B%22autosize%22:true,%22xaxis%22:%7B%7D,%22yaxis%22:%7B%7D%7D&x_axis=relative&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=%5B%5D&last_linear_y_axis_range=%5B%

@mishooax @ssmmnn11


models/src/anemoi/models/distributed/transformer.py
```diff
@@ -130,6 +199,36 @@ def shard_sequence(input_: Tensor, shapes: list, mgroup: ProcessGroup) -> Tensor
     return _SplitSequenceParallelSection.apply(input_, shapes, mgroup)
 
 
+def halo_exchange(x: Tensor, halo_size: int, mgroup: ProcessGroup) -> Tensor:
```
ssmmnn11: I was wondering: we now have 
halo_exchange
_halo_exchange
_HaloExchange

would it make sense to come up with more unique / more descriptive names for these? I think this might be a bit confusing. I admit that the names for the other routines (shard_heads etc.) are not great either.

models/src/anemoi/models/layers/processor.py
```diff
@@ -97,6 +97,7 @@ def __init__(
         num_heads: int = 16,
         mlp_hidden_ratio: int = 4,
         dropout_p: float = 0.1,
+        shard_strategy: str = "shard_heads",
```
ssmmnn11: Add to doc string below?

models/src/anemoi/models/layers/attention.py
```diff
@@ -55,38 +58,75 @@ def __init__(
...
         if model_comm_group:
             assert (
                 model_comm_group.size() == 1 or batch_size == 1
             ), "Only batch size of 1 is supported when model is sharded accross GPUs"
 
-        query, key, value = (
-            einops.rearrange(
-                t,
-                "(batch grid) (heads vars) -> batch heads grid vars",
+        if self.shard_strategy == "shard_sequence":
```
ssmmnn11: this is now very long. can we introduce e.g. something like

`if if self.shard_strategy == "shard_sequence":
     x = self.shard_sequence(x)

query, key, value = self.lin_qkv(x).chunk(3, -1)

query, key, value = (
                einops.rearrange(
                    t,
                    "(batch grid) (heads vars) -> batch heads grid vars",
                    batch=batch_size,
                    heads=self.num_heads,
                )
                for t in (query, key, value)
            )

if if self.shard_strategy == "shard_heads"
      query = shard_heads(query, shapes=shapes, mgroup=model_comm_group)
       key = shard_heads(key, shapes=shapes, mgroup=model_comm_group)
       value = shard_heads(value, shapes=shapes, mgroup=model_comm_group)
.
.
.
.
`

models/src/anemoi/models/layers/attention.py
```diff
@@ -104,7 +144,11 @@ def forward(
                 dropout_p=dropout_p,
             )  # expects (batch heads grid variable) format
 
-        out = shard_sequence(out, shapes=shapes, mgroup=model_comm_group)
+        if self.shard_strategy == "shard_sequence":
+            out = out[:, :, halo_size_left : out.shape[-2] - halo_size_right, :]  # remove halos
```
ssmmnn11: I would prefer if this would happen in a function that lives at the same place as halo_exchange, e.g. call halo_expand first and then halo_contract (not best names).

ssmmnn11: maybe just remove_halos

ssmmnn11 (changes requested):
Very nice contribution :-)

models/src/anemoi/models/layers/attention.py
mishooax: agreed, the if & else blocks should be refactored as separate (member) functions

models/src/anemoi/models/layers/attention.py
```diff
@@ -55,38 +58,75 @@ def __init__(
...
             ), "Only batch size of 1 is supported when model is sharded accross GPUs"
 
-        query, key, value = (
-            einops.rearrange(
-                t,
-                "(batch grid) (heads vars) -> batch heads grid vars",
+        if self.shard_strategy == "shard_sequence":
+            assert (
+                shapes[-1][0] // 2 >= self.window_size[0]
+            ), "Sharded sequence length must be at least twice the window size"
```
mishooax: we could have the assert print the sharded sequence length and window size so the user sees the values that raised the error?

models/src/anemoi/models/layers/processor.py
mishooax: is this value configurable? (how can one override the default?)

mishooax:
excellent work! 👏

models/src/anemoi/models/distributed/transformer.py
japols: changed it to:
halo_exchange
_HaloExchangeParallelSection
_halo_comm

models/src/anemoi/models/layers/processor.py
japols: > Add to doc string below?

done

> is this value configurable? (how can one override the default?)

ah yes, it is configurable from config.model.processor.shard_strategy, forgot to add that to anemoi-core
also set the default strategy to `shard_sequence`

models/src/anemoi/models/layers/attention.py
japols: > I would prefer if this would happen in a function that lives at the same place as halo_exchange, e.g. call halo_expand first and then halo_contract (not best names).

good idea, I added add_halos and remove_halos

models/src/anemoi/models/layers/attention.py
japols: good point, done

models/src/anemoi/models/layers/attention.py
japols: > agreed, the if & else blocks should be refactored as separate (member) functions

I moved it to member functions get_qkv_shard_[heads/sequence]

japols: changed it to `query, key, value =  self.get_qkv_shard_sequence` and `get_qkv_shard_sequence` respectively

models/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Keep it human-readable, your future self will thank you!
 - Reduced memory usage when using chunking in the mapper [#84](https://github.com/ecmwf/anemoi-models/pull/84)
 - Added `supporting_arrays` argument, which contains arrays to store in checkpoints. [#97](https://github.com/ecmwf/anemoi-models/pull/97)
 - Add remappers, e.g. link functions to apply during training to facilitate learning of variables with a difficult distribution [#88](https://github.com/ecmwf/anemoi-models/pull/88)
+- Add sequence sharding strategy for TransformerProcessor [#67](https://github.com/ecmwf/anemoi-core/pull/67)
```
HCookie, line 21: FYI, Changelog should **not** be manually updated

HCookie:
@japols Is this a PR you are still looking to merge?

PR #70, opened 2025-01-09
Fix/combined loss

### LINKED ISSUE #68: CombinedLoss not working/not tested
### What happened?

The node_weights tensor is never passed as an argument to individual losses. The node_loss_weights resolves with hydra in a GraphNodeAttribute which does not translate anywhere into a tensor other than through the get_node_weights method from the GraphForecaster. 
It means that the node_weights tensor is provided as a kwarg to the CombinedLoss init, which does not need it, but when hydra instantiates the individual losses, there is no tensor but only a GraphNodeAttribute object.
Reminder: the CombinedLoss is not tested.

Write that in the training config (the documentation string is wrong, node_weights is not provided anywhere if you follow it). 
```

training_loss:
  losses:
    - _target_: anemoi.training.losses.mse.WeightedMSELoss
      node_weights: ${training.node_loss_weights_resolved}
    - _target_: anemoi.training.losses.mae.WeightedMAELoss
      node_weights: ${training.node_loss_weights_resolved}
  scalars:
    - variable
  loss_weights:
    - 1.0
    - 0.5
  ignore_nans: True

node_loss_weights:
  _target_: anemoi.training.losses.nodeweights.GraphNodeAttribute
  target_nodes: ${graph.data}
  node_attribute: area_weight
  
node_loss_weights_resolved: ${training.node_loss_weights}
```

### Version

latest develop

Linux balfrin-ln002 5.14.21-150400.24.81_12.0.87-cray_shasta_c

### Relevant log output

_No response_

### Accompanying data

MeteoSwiss

### Organisation

MeteoSwiss

OpheliaMiralles:
Try to fix #68 
Add tests

### DIFF
--- a/training/src/anemoi/training/losses/combined.py
+++ b/training/src/anemoi/training/losses/combined.py
@@ -81,10 +81,16 @@ class CombinedLoss(torch.nn.Module):
         assert len(losses) == len(loss_weights), "Number of losses and weights must match"
         assert len(losses) > 0, "At least one loss must be provided"
 
-        self.losses = [
-            GraphForecaster.get_loss_function(loss, **kwargs) if isinstance(loss, dict) else loss(**kwargs)
-            for loss in losses
-        ]
+        self.losses = []
+        for loss in losses:
+            if isinstance(loss, dict):
+                self.losses.append(GraphForecaster.get_loss_function(loss, **kwargs))
+            elif isinstance(loss, type):
+                self.losses.append(loss(**kwargs))
+            elif hasattr(loss, "__class__"):
+                self.losses.append(loss)
+            else:
+                raise TypeError(f"Loss {loss} is not a valid loss function")
         self.loss_weights = loss_weights
 
     def forward(
--- a/training/src/anemoi/training/losses/combined.py
+++ b/training/src/anemoi/training/losses/combined.py
@@ -27,7 +27,6 @@ class CombinedLoss(torch.nn.Module):
         self,
         losses: Sequence[torch.nn.Module],
         loss_weights: tuple[int, ...],
-        **kwargs,
     ):
         """Combined loss function.
 
@@ -44,8 +43,6 @@ class CombinedLoss(torch.nn.Module):
             Loss objects.
         loss_weights : tuple[int, ...]
             Weights of each loss function in the combined loss.
-        kwargs: Any
-            Additional arguments to pass to the loss functions
 
         Examples
         --------
--- a/training/src/anemoi/training/losses/combined.py
+++ b/training/src/anemoi/training/losses/combined.py
@@ -10,12 +10,14 @@
 from __future__ import annotations
 
 import functools
+from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
 
 import torch
 
-from anemoi.training.train.forecaster import GraphForecaster
+if TYPE_CHECKING:
+    from collections.abc import Sequence
 
 
 class CombinedLoss(torch.nn.Module):
@@ -23,8 +25,7 @@ class CombinedLoss(torch.nn.Module):
 
     def __init__(
         self,
-        *extra_losses: dict[str, Any] | Callable,
-        losses: tuple[dict[str, Any] | Callable] | None = None,
+        losses: Sequence[torch.nn.Module],
         loss_weights: tuple[int, ...],
         **kwargs,
     ):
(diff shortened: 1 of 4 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=70) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=70) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: anaprietonem<br/>:x: OpheliaMiralles<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=70) it.</sub>

training/src/anemoi/training/train/forecaster.py
```diff
@@ -121,13 +121,24 @@ def __init__(
...
+                loss_instantiated = self.get_loss_function(loss, scalars=self.scalars, **{"node_weights": loss_node_weights, "ignore_nans": ignore_nans})
+                losses.append(loss_instantiated)
+                assert isinstance(loss_instantiated, BaseWeightedLoss)
+            self.loss = instantiate({"_target_": config.training.training_loss._target_}, losses=losses, loss_weights = config.training.training_loss.loss_weights, **loss_kwargs)
+        else:
+            self.loss = self.get_loss_function(config.training.training_loss, scalars=self.scalars, **loss_kwargs)
+            assert isinstance(self.loss, BaseWeightedLoss) and not isinstance(
+                self.loss,
+                torch.nn.ModuleList,
+            ), f"Loss function must be a `BaseWeightedLoss`, not a {type(self.loss).__name__!r}"
```
HCookie: I think that this is over specific for this use case, and instantiate's objects unneccessarily

training/src/anemoi/training/losses/combined.py
```diff
@@ -81,10 +81,16 @@ def __init__(
...
-            for loss in losses
-        ]
+        self.losses = []
+        for loss in losses:
+            if isinstance(loss, dict):
+                self.losses.append(GraphForecaster.get_loss_function(loss, **kwargs))
+            elif isinstance(loss, type):
+                self.losses.append(loss(**kwargs))
+            elif hasattr(loss, "__class__"):
+                self.losses.append(loss)
```
HCookie: Why are we checking for ` __class__`? If checking for an object why not `isinstance(loss, object)`?

training/src/anemoi/training/train/forecaster.py
OpheliaMiralles: Instantiating node_weights was necessary to call the combined loss but if you find a way around it, please let me know... I have another version where all of this is implemented in the get_loss_function from the forecaster. It is cleaner so I'll try to commit it soon.

training/src/anemoi/training/losses/combined.py
OpheliaMiralles: Because it could originally only take a class (of type "type", not instantiated) as losses arguments. Indeed, loss(**kwargs) called later in the function expects init arguments from the individual loss object and not forward arguments. As I said, I'll try to commit recent changes later.

training/src/anemoi/training/train/forecaster.py
HCookie: Hi, yeah, as I wrote the loss functions code originally, I was able to find a way around, and only update the CombinedLoss class.

HCookie: If you'd like, we can work together on https://github.com/ecmwf/anemoi-core/tree/fix/combined_loss_hcookie to make sure your use case is addressed.

OpheliaMiralles: I don't know, for me the CombinedLoss is not a BaseWeightedLoss, so I don't really see the point in trying to make it fit this base class. The weights don't mean the same thing here, and the individual losses should probably all have separate node weights. I'll update this PR today. Let me know what you think, but really we should not be afraid of separating use cases when they don't match, don't you think?

HCookie: While the `CombinedLoss` may not be a clear use case of `BaseWeightedLoss`, it is still an anemoi loss function, and so inheritance based structures make sense. I am very wary of any solution that **requires** hard coding of any sort. Anemoi is designed to be a generic framework so following proper OOP principles is a must, otherwise any of these main classes end up with massive branching behaviours which is both hard to read and hard to use. (This is already the case in the GraphForecaster)

> the individual losses should probably all have separate node weights

Excluding your use case of different losses for different params, in what case will this be true? Having a weighting between losses and then different relative weightings within the losses will massively increase complexity, and *in my opinion* be very hard to interpret.

HCookie: There have been some changes implemented in #52 that I think may be interesting for your use cases? Shall we move this discussion to slack and organise a call?

OpheliaMiralles: In this usecase, the node weights are defined as several masks defining grids for different data sources (a radar mask, a station pointwise mask and a satellite mask). I believe it might be a common usecase in the scope of data assimilation, but of course it is part of a broader discussion. OK to move the discussion to slack. We can have a call next week, let's schedule on slack too.

HCookie:
Superseeded by #103

Issue #71, opened 2025-01-09
Copy Imputer


icedoom888:
### Is your feature request related to a problem? Please describe.

Within the Alpine domain, converting from Model Levels to Pressure Levels causes the raise of many NaN values on the lower Pressure Levels. This is due to steep orography. 
The missing values change dynamically in time.

### Describe the solution you'd like

Add an imputer that copies missing information from another field.
When a value is missing at a certain Pressure Level it would be useful to copy the value of the Pressure Level above. 
This technique ensures constant imputing in areas with missing Pressure Levels and will not cause a big delta in values with respect to vertical changes. 

### Additional context

Here attached is an exemple config for such imputer:

```

mean_imputer:
  default: "none"

  "mean":
   - 10u
   - 10v
   - 2d
   - 2t
   - msl
   - sp

   - q_100
   - q_150
   - q_200
   - q_250
   - q_300
   - q_400
   - q_50
   - q_500

   - t_100
   - t_150
   - t_200
   - t_250
   - t_300
   - t_400
   - t_50
   - t_500
  
   - u_100
   - u_150
   - u_200
   - u_250
   - u_300
   - u_400
   - u_50
   - u_500
  
   - v_100
   - v_150
   - v_200
   - v_250
   - v_300
   - v_400
   - v_50
   - v_500
  
   - w_100
   - w_150
   - w_200
   - w_250
   - w_300
   - w_400
   - w_50
   - w_500
  
   - z_100
   - z_150
   - z_200
   - z_250
   - z_300
   - z_400
   - z_50
   - z_500
  
copy_imputer:
  default: "none"
  
  "q_500":
    - q_600
  "q_600":
    - q_700
  "q_700":
    - q_850
  "q_850":
    - q_925

  "t_500":
    - t_600
  "t_600":
    - t_700
  "t_700":
    - t_850
  "t_850":
    - t_925


  "u_500":
    - u_600
  "u_600":
    - u_700
  "u_700":
    - u_850
  "u_850":
    - u_925

  "v_500":
    - v_600
  "v_600":
    - v_700
  "v_700":
    - v_850
  "v_850":
    - v_925
  

  "w_500":
    - w_600
  "w_600":
    - w_700
  "w_700":
    - w_850
  "w_850":
    - w_925


  "z_500":
    - z_600
  "z_600":
    - z_700
  "z_700":
    - z_850
  "z_850":
    - z_925

processors:
  mean_imputer:
[...]

PR #72, opened 2025-01-09
feat(models): Copy Imputer

icedoom888:
Add an imputer that copies missing information from another field.

When a value is missing at a certain Pressure Level it would be useful to copy the value of the Pressure Level above.
This technique ensures constant imputing in areas with missing Pressure Levels and will not cause a big delta in values with respect to vertical changes.

Solves #71


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=72) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=72) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=72) it.</sub>

HCookie:
Can you please provide some more details on the description about where this imputer may be useful and some example code/config of it in action?

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -231,6 +231,111 @@ def __init__(
...
 
 
+class CopyImputer(BaseImputer):
+    """Imputes missing values copying them from another variable.
+    ```
+    default: "none"
+    x:
+        - y
+        - q
+    ```
```
HCookie: Some comments on what these characters represent will help here

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -303,3 +408,50 @@ def __init__(
             "You are using a dynamic Imputer: NaN values will not be present in the model predictions. \
                       The model will be trained to predict imputed values. This might deteriorate performances."
         )
+
+
+class DynamicCopyImputer(CopyImputer):
```
HCookie: A little more detail on the exact differences, and uses would be helpful to other users.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -303,3 +408,50 @@ def __init__(
...
+        for idx_src, (idx_dst, value) in zip(self.index_training_input, zip(index, self.replacement)):
+            if idx_dst is not None:
+                assert not torch.isnan(
+                    x[..., self.data_indices.data.input.name_to_index[value]][nan_locations[..., idx_src]]
+                ).any(), f"NaNs found in {value}."
+                x[..., idx_dst][nan_locations[..., idx_src]] = x[
+                    ..., self.data_indices.data.input.name_to_index[value]
+                ][nan_locations[..., idx_src]]
+
+        return x
```
HCookie: Given the similarities between this and the parent class implementation, would it make sense to factor it into another function. Or a generic parent class with two children?

models/src/anemoi/models/preprocessing/imputer.py
icedoom888: Added info to DynamicMixin, as the imputing behaviour is identical to the copy imputer.

models/src/anemoi/models/preprocessing/imputer.py
icedoom888: The transform function has a completely different implementation.. i tried to combine CopyImputer and DynamicMixin, let me know if this works better for you

HCookie (approved):
Thanks for addressing my comments,
As this is only a new class, I am assume you have tested the functionality, and it behaves as expected.

icedoom888:
Yes this is tested and working!

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -414,7 +413,7 @@ def __init__(
         data_indices: Optional[IndexCollection] = None,
         statistics: Optional[dict] = None,
     ) -> None:
-        super().__init__(config, data_indices, statistics)
+        ConstantImputer.__init__(config, data_indices, statistics)
```
HCookie: This appears to be missing a `self` argument.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -424,17 +423,30 @@ def __init__(
...
-                    x[..., self.data_indices.data.input.name_to_index[value]][nan_locations[..., idx_src]]
-                ).any(), f"NaNs found in {value}."
-                x[..., idx_dst][nan_locations[..., idx_src]] = x[
-                    ..., self.data_indices.data.input.name_to_index[value]
-                ][nan_locations[..., idx_src]]
 
+        if x.shape[-1] == self.num_training_input_vars:
+            indices = self.data_indices.data.input.name_to_index
+        elif x.shape[-1] == self.num_inference_input_vars:
+            indices = self.data_indices.model.input.name_to_index
```
HCookie, line 443: If the shape doesn't match either, this will fail abruptely with an undefined indices.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -424,17 +423,30 @@ def __init__(
...
-            if idx_dst is not None:
-                assert not torch.isnan(
-                    x[..., self.data_indices.data.input.name_to_index[value]][nan_locations[..., idx_src]]
-                ).any(), f"NaNs found in {value}."
-                x[..., idx_dst][nan_locations[..., idx_src]] = x[
-                    ..., self.data_indices.data.input.name_to_index[value]
-                ][nan_locations[..., idx_src]]
 
+        if x.shape[-1] == self.num_training_input_vars:
+            indices = self.data_indices.data.input.name_to_index
```
HCookie, line 441: Is there an edge case where the shape == the number of training vars but the indices between data and model are different?

models/src/anemoi/models/preprocessing/imputer.py
icedoom888: Nice catch! Sorry I missed this :)

models/src/anemoi/models/preprocessing/imputer.py
icedoom888: Added the same ValueError as line 148 in the base imputer.

models/src/anemoi/models/preprocessing/imputer.py
icedoom888: I would day yes, this could happen if you provide new data to a pretrained model and you don't follow the same dataloader.. But this issue can happen in any imputer, same behaviour is present in the base imputer too (check the transform function).

HCookie: Okay, seems fine to me, maybe another PR in the future to resolve that.

HCookie (approved):

PR #75, opened 2025-01-10
ci: Reinstantiate CI files

JesperDramsch:
This PR adds the following workflows:

- Public Label PR
- Conventional Commit PR titles
- PR labeller for convenience
- PR Pre-commit and testing

These depend on merge of:
- https://github.com/ecmwf-actions/reusable-workflows/pull/144


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=75) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=75) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=75) it.</sub>

.github/workflows/python-pull-request.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    - name: Install basic dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest pytest-md pytest-emoji
+
+    - name: Detect changed packages
+      id: changed-packages
+      run: |
+        CHANGED_FILES=$(git diff --name-only ${{ github.event.before }} ${{ github.sha }})
```
MeraX: I'm trying to understand this line. As I understand it, it checks if there are changes between the commit pushed to `develop` (`${{ github.sha }}` and the previous HEAD of `develop` `${{ github.event.before }}`. Subsequently, you install any of the three packages from the repository, if there were and changes and use PyPI otherwise.

Does this assume, that the previous HEAD of `develop` is always equal to the version on PyPI?

MeraX:
Hi Jesper, I went through the yaml files. Most of them look uncontroversial to me. However, the `reusable workflows` aren't happy with their integration: `reusable workflows should be referenced at the top-level `jobs.*.uses' key, not within steps` https://github.com/ecmwf/anemoi-core/actions/runs/12710997502

MeraX:
And another question: The Github Summary above reports "All checks have passed", even though the `python-pull-request.yml` is red. Do you consider making `python-pull-request.yml` mandatory once you PR is merged?
image

MeraX:
Another thing that came to my mind: You are installing the three `-core` packages in individual steps. This might lead to a different resolution of package versions than installing them all at once: `pip install -e graphs/[tests] -e training/[tests] -e models/[tests]`. In particular, the current version would first install `anemoi-graphs` and `anemoi-models` from PyPI, as both are dependencies of `./training`. In subsequent steps, both packages are overwritten with local versions. However, dependencies might still be installed in versions required in the PyPI packages and not as required by the local packages.

.github/workflows/python-pull-request.yml
```diff
@@ -0,0 +1,104 @@
...
+        python-version: ${{ matrix.python-version }}
+        install-dependencies: false
+        custom-pytest: pytest models/tests
+
+    - name: Run integration tests
+      if: success()
+      uses: ecmwf-actions/reusable-workflows/.github/workflows/qa-pytest-pyproject.yml@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+        install-dependencies: false
```
MeraX: I suspect, that you have to specify `        custom-pytest: pytest tests` here as well, as the [default](https://github.com/ecmwf-actions/reusable-workflows/blob/main/.github/workflows/qa-pytest-pyproject.yml#L39) is just `pytest` which would recursively run all tests of the three packages again.

.github/workflows/python-pull-request.yml
JesperDramsch: I tried not to make implicit assumptions here. I thought it makes most sense to try and see if there are any changes compared to the previous version on `develop` and not necessarily the released version.

JesperDramsch:
> Hi Jesper, I went through the yaml files. Most of them look uncontroversial to me. However, the `reusable workflows` aren't happy with their integration: `reusable workflows should be referenced at the top-level `jobs.*.uses' key, not within steps` https://github.com/ecmwf/anemoi-core/actions/runs/12710997502

Oh that's a problem. Wasn't aware of that limitation.

.github/workflows/python-pull-request.yml
JesperDramsch: Good catch, I wasn't aware pytest does recursion.

JesperDramsch:
This now runs the tests of changed code against:

1. Currently released code on PyPI (local-changes)

This mimics the way all other code in Anemoi currently runs, specifically against the released versions of other packages.

2. All other develops

This ensures that the code will stay up to date with coming changes on develop.

## Current limitations

This can mean that if there are changes on develop that are not backwards compatible, but also aren't released yet, then some of the tests will fail. 

I believe it's important to test both against released and current code on develop. so I added  the label `skip-ci-release-tests` which can be applied to skip tests against pyPI releases, when we know there was a breaking change.

HCookie (approved):

.github/workflows/python-pull-request.yml
```diff
@@ -0,0 +1,127 @@
...
+    uses: ecmwf-actions/reusable-workflows/.github/workflows/qa-precommit-run.yml@v2
+    with:
+      skip-hooks: "no-commit-to-branch"
+
+  checks:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: ["ubuntu-latest", "macos-latest"]
+        python-version: ["3.9", "3.10", "3.11"]
```
theissenhelen: Aren't we testing 3.12 as well?

HCookie:
Could it be useful to add an inactive issue closer?
https://docs.github.com/en/actions/use-cases-and-examples/project-management/closing-inactive-issues

.github/workflows/python-pull-request.yml
JesperDramsch: Ok!

JesperDramsch:
> Could it be useful to add an inactive issue closer? https://docs.github.com/en/actions/use-cases-and-examples/project-management/closing-inactive-issues

I think we'd best add these after.

HCookie:
What's the intersection of changes between this PR and #84? 
Did this form the basis of that one?

PR #76, opened 2025-01-13
test: test aicon in CI with pytest

MeraX:
This bases on #75 and adds a short integration test to test the training of aicon.

Should be merged when the issues in #75 have been resolved.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=76) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=76) before we can accept your contribution.<br/>**1** out of **2** committers have signed the CLA.<br/><br/>:white_check_mark: JesperDramsch<br/>:x: MeraX<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=76) it.</sub>

MeraX:
@JesperDramsch or @HCookie could you have a look at the `install-mode`? It somehow evaluated to "true", which is an invalid value. The error appears on this PR https://github.com/ecmwf/anemoi-core/actions/runs/13007212878 as well as on "main" https://github.com/ecmwf/anemoi-core/actions/runs/13008047920

When this issue is solved, I think this RP is ready for review.

mchantry:
Hi MeraX, I haven't had a chance to test this myself yet. But I think you need
``` 
install-mode: ${{ fromJSON(github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'skip-ci-release-tests')) ? '["all"]' : '["local-changes", "all"]') }}
```

VeraChristina:
Hi Marek,
Sure, my pleasure! Thanks a lot for working through the pydantic changes. And I agree, let’s get it in!

Since this is the first integration test use case — congrats on that! — we’re keen to work out best practices with your help to set up a good template for how to write these tests (as far as that’s possible given different use cases and requirements). So I have a couple of small questions for you. Would you have time for a call sometime this week? 

It’s mostly questions about specific requirements of your set-up (like the grid), some thoughts on how to keep the test set-up as light for you to maintain as possible, and a few imminent changes in our integration test action and set-up (sorry! small downside of being so fast to be the first). I’ll do a review with more specific comments tomorrow morning so that you have an idea of what I’d like to understand better

MeraX:
Hi Vera, I would be available. Perhaps it would be best to coordinate via Slack (or Mail)

I hope further revisions won't take too long and that no breaking changes will be merged in the meantime. (We've had two major changes in the past that set this activity back by more than a month each time...)

tests/aicon/test_cicd_aicon_04_icon-dream_medium.py
```diff
@@ -0,0 +1,78 @@
...
+        config = compose(config_name="test_cicd_aicon_04_icon-dream_medium")
+
+    if shorten:
+        date = datetime.datetime.fromisoformat(config.dataloader.training.start)
+        date = date + datetime.timedelta(days=3)
+        config.dataloader.training.end = date.isoformat()
+        date = date + datetime.timedelta(hours=6)
+        config.dataloader.validation.start = date.isoformat()
+        date = date + datetime.timedelta(days=2)
+        config.dataloader.validation.end = date.isoformat()
```
VeraChristina: I think it might be simpler, and more explicit, to shorten the training loop by setting the config entries "dataloader.batch_size.training", "dataloader.limit_batches.training" (and corresponding for validation) -- either directly in the config yaml, or here, or as overrides in the compose.
Do we use the not shortened option?

.github/workflows/python-pull-request.yml
```diff
@@ -3,9 +3,6 @@ name: Test PR
 on:
   pull_request:
     types: [opened, synchronize, reopened]
-  push:
-    branches:
-      - develop
```
VeraChristina: Can you put this back in and change develop to main? If there has been a conversation about taking this out that I am not aware of, that's of course fine.

tests/aicon/test_cicd_aicon_04_icon-dream_medium.py
```diff
@@ -0,0 +1,78 @@
...
+
+    grid_filename = config.graph.nodes.icon_mesh.node_builder.grid_filename
+    with tempfile.NamedTemporaryFile(suffix=".nc") as grid_fp:
+        if grid_filename.startswith(("http://", "https://")):
+            import urllib.request
+
+            urllib.request.urlretrieve(grid_filename, grid_fp.name)  # noqa: S310
+            config.graph.nodes.icon_mesh.node_builder.grid_filename = grid_fp.name
```
VeraChristina: Do we need this (due to specifics of your set-up), or can we set this in the config yaml directly and let anemoi handle the grid file loading?
@b8raoult do you have more context on this than me?

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -0,0 +1,387 @@
...
+    mlflow:
+      enabled: true
+      offline: True
+      log_model: false
+      experiment_name: 'cicd_test'
+      project_name: AICON
+      system: false
+      terminal: false
+      run_name: "R3B5 6hr loss_scaling=1"
+      tracking_uri: "none" # @Helen: must be string without pydantic
```
VeraChristina: Do we want to do offline logging if it's not included in the test? Do we reach an epoch where logging is triggered? If we do, we need to think about cleaning up temporary log outputs

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -0,0 +1,387 @@
...
+    frequency: ${data.frequency}
+    drop: []
+  test:
+    dataset: {}
+    start: 2022
+    end: null
+    frequency: ${data.frequency}
+  num_workers:
+    training: 8
+    validation: 8
```
VeraChristina: I would recommend setting these to 2. This has made our tests more stable when run on different platforms.

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -0,0 +1,387 @@
...
+# Anemoi settings for execution in a CI/CD pipeline.
+#
+# This script is not part of a productive ML workflow, but is
+# used for CI/CD!
+#
+# 11/2024, DWD
+# ---------------------------------------------------------------
+
+# set to true to switch off config validation
+no_validation: True
```
VeraChristina: Do you want to run without config validation? Asking because I see that you have edited the IconNodeSchema

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -0,0 +1,387 @@
...
+    every_n_epochs:
+      save_frequency: 1
+      num_models_saved: -1
+    every_n_train_steps:
+      save_frequency: null
+      num_models_saved: 0
+  log:
+    wandb:
+      enabled: false
+      entity: "{}" # @Helen. can be `???` with pydantic, must be `{}` otherwise
```
VeraChristina: I am looking into fixing this, such that '???" entries are allowed when wandb is not enabled. In the meantime, this seems to be a good temporary fix.
@theissenhelen Additional thoughts about this?

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -0,0 +1,387 @@
...
+# set to true to switch off config validation
+no_validation: True
+
+defaults:
+  - model: gnn
+  - training: default
+  - diagnostics: evaluation
+  #- override hydra/job_logging: none
+  #- override hydra/hydra_logging: none
+  - _self_
```
VeraChristina: I am wondering about how to set this up in a way that is easy for you to maintain. It depends a bit on your aim with this test: Do you want to know when there have been changes that break the current config set-up (so that (a) you are aware, and (b) you can early-on have a look at how to fix them)? Or would you rather keep the config as flexible as possible, e.g. by setting these defaults and potentially more defaults, and only have to change things when your set-up diverges from the current default config set-up?

For context here is our current approach to handling configs in integration tests https://github.com/ecmwf/anemoi-core/blob/9a37cf4bf32c24b4426405c7aba8d019b6b386c2/training/tests/integration/conftest.py, i.e. using templates from the repo and then setting test- and dataset-specific overrides. I am aware that your case is somewhat different. We want to ensure that templates and code stay compatible, whereas you want to know whether your specific set-up still runs (so you might want to stay as close as possible to the config you actually use). Just thinking that it's worth to think about this in the context of maintenance.

tests/aicon/test_cicd_aicon_04_icon-dream_medium.py
```diff
@@ -0,0 +1,78 @@
...
+        trainer.train()
+        final_sum = torch.tensor(list(map(torch.sum, trainer.model.parameters()))).sum()
+    return trainer, initial_sum, final_sum
+
+
+@pytest.fixture
+def get_trainer() -> tuple:
+    return trainer()
+
+
```
VeraChristina: May need to add a @longtests decorator to not run this on each commit but instead nightly. I'll get back to you about this

tests/aicon/test_cicd_aicon_04_icon-dream_medium.py
VeraChristina: As discussed in the call, this is the current approach to handle the grid set-up

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
VeraChristina: As discussed earlier this week, I think it's worth making the validation work to improve config management but happy to keep it as is for now. Up to you

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
VeraChristina: As above, up to you

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
VeraChristina: I just ran the test and it adds an output dir with subfolders for logs etc. Irrespective of logging, I think we should remove those at the end of the test

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
MeraX: I fully agree. I think we should first merge this PR and then use it to fix the config validation. I've opened an issue for this https://github.com/ecmwf/anemoi-core/issues/198

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
MeraX: Thank you for the suggestion. This makes the test 30% faster, at least locally.

tests/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
MeraX: The output comes from `hardware.paths.output` and  `hardware.paths.graph`. I'll set the to temporary paths.

.github/workflows/python-pull-request.yml
MeraX: As we discussed, I'll revert this change completely and would leave a modification up to another more genuine PR.

tests/aicon/test_cicd_aicon_04_icon-dream_medium.py
MeraX: I removed all of the `shorten` logic.

tests/aicon/test_cicd_aicon_04_icon-dream_medium.py
MeraX: Hi Vera, the test is passing now. I'm wondering if `longtests` is actually required as the "integration" tests are anyway just run nightly.

We were able to enable config validation, too.

VeraChristina: That's great! Yes, the longtest flag is still required. (We want to avoid people who are used to just run `pytest tests/` running long tests unintentionally.)

VeraChristina: As to config validation, if I see correctly, the no_validation flag is still set to true

VeraChristina: Oh, sorry! I think I was not very clear on this point. The integration tests have moved to training/tests/integration. Would you move your aicon folder there? (That's why we need the longtests decorator)

MeraX: I'm sorry, too. I didn't push the modifications I was referring to. Now I also moved the test to the new location and added the tag.

MeraX:
successful integration test action: https://github.com/ecmwf/anemoi-core/actions/runs/13930810492/job/38987207691

VeraChristina (approved):
Thanks for addressing the feedback! This looks good to me

PR #77, opened 2025-01-14
feature(models): Add model comm group to predict_step 

cathalobrien:
Small change to allow predict_step to take an optional model_comm_group, like model.forward does.

This is needed for parallel-inference, which calls predict_step

### DIFF
--- a/models/src/anemoi/models/interface/__init__.py
+++ b/models/src/anemoi/models/interface/__init__.py
@@ -96,7 +96,9 @@ class AnemoiModelInterface(torch.nn.Module):
         # Use the forward method of the model directly
         self.forward = self.model.forward
 
-    def predict_step(self, batch: torch.Tensor, model_comm_group: Optional[ProcessGroup] = None) -> torch.Tensor:
+    def predict_step(
+        self, batch: torch.Tensor, model_comm_group: Optional[ProcessGroup] = None, **kwargs
+    ) -> torch.Tensor:
         """Prediction step for the model.
 
         Parameters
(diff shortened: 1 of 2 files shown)


japols (approved):
Looks good to me

HCookie (approved):

models/src/anemoi/models/interface/__init__.py
```diff
@@ -94,7 +96,7 @@ def _build_model(self) -> None:
         # Use the forward method of the model directly
         self.forward = self.model.forward
 
-    def predict_step(self, batch: torch.Tensor) -> torch.Tensor:
+    def predict_step(self, batch: torch.Tensor, model_comm_group: Optional[ProcessGroup] = None) -> torch.Tensor:
```
gmertes: Can we also add a `**kwargs` here for future proofing? Then we don't need any logic in inference to maintain backwards compatibility when new optionals are added.

cathalobrien: It would be my pleasure. And then pass that to `self()`?

gmertes: I would not pass it to self, so that that remains explicit. Passing kwargs blindly to sub-functions is something we should try to avoid in general.

cathalobrien: @gmertes like [so](https://github.com/ecmwf/anemoi-core/pull/77/commits/fdd92e6944a278aeb3c3a3efbb9c7dfb7d27d96d)?

gmertes (approved):

ssmmnn11 (approved):
looks good to me :-)

PR #78, opened 2025-01-16
feat(models): use num_layers of the processor in hierarchical graphs

JPXKQX:
This PR splits the hidden-most processor instantiation from the in-level processors. 
- The `level_process_num_layers` ([link](https://github.com/ecmwf/anemoi-core/blob/develop/training/src/anemoi/training/config/hierarchical.yaml#L23C3-L23C30)) specifies the number of layers for the intermediate level processors.
- The `num_layers` argument sets the value for the hidden-most level processor.

This PR extends the current functionality. The previous behaviour can be obtained by setting `num_layers` and `level_process_num_layers` to the same value.


HCookie (approved):
LGTM,
Just change the PR title to read a bit nicer before merge.

PR #83, opened 2025-01-22
fix: cpu memory savings of sharded dataloader

japols:
This PR brings back the cpu memory savings from sharded dataloading [(#76)](https://github.com/ecmwf/anemoi-training/pull/76). At 9km, the max cpu memory usage goes down from ~70% to ~30%.

The sharded reading clashed with the recent changes to [Support masking of unconnected nodes (LAM)](https://github.com/ecmwf/anemoi-training/pull/171). Since anemoi-datasets currently doesn't support slicing and indexing in one operation like

``
x = self.data[start : end : self.timeincrement, :, :, grid_shard_indices]
``

the new version checks for the type of `grid_shard_indices` and does it in 2 steps if necessary. 

@Rilwan-Adewoyin @JPXKQX @b8raoult @floriankrb

### DIFF
--- a/training/src/anemoi/training/data/dataset.py
+++ b/training/src/anemoi/training/data/dataset.py
@@ -279,8 +279,15 @@ class NativeGridDataset(IterableDataset):
             end = i + (self.rollout + 1) * self.timeincrement
 
             grid_shard_indices = self.grid_indices.get_shard_indices(self.reader_group_rank)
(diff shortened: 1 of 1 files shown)


JPXKQX (approved):
LAM works.

HCookie (approved):
Go for the merge.

PR #93, opened 2025-01-24
fix: default behaviour for kernel_layers when not set in config.

jakob-schloer:
When `kernel_layers` is not set in the model config, it should return the pytoch default kernel. 

Changes: 
```
- OmegaConf.select(model_config, "model.layer_kernels") -> returns None if layer_kernels is not set
+ model_config.model.get("layer_kernels", {}) -> returns {} if layer_kernels is not set 
```


anaprietonem (approved):
Approving as this PR fixes the backward compatibility issue introduced in the config with the definition of the kernel layers

PR #97, opened 2025-01-28
docs: cancel RTD builds on PRs without change

JesperDramsch:
This limits the builds on RTD to avoid always building all documentations for PRs.


anaprietonem (approved):
Thanks for this fix Jesper!

PR #98, opened 2025-01-28
fix:  bug in variables ordering in NormalizedReluBounding

lzampier:
A bug affects `anemoi.models.layers.bounding.NormalizedReluBounding` if the order of the variable douring the call in config does not match the order of the variables in the torch array. This a problem emerged only when multiple variables were bounded at the same time with this procedure.

This is fixed by making sure that `self.norm_min_val` follows the same order as `self.data_index`.

### DIFF
--- a/models/src/anemoi/models/layers/bounding.py
+++ b/models/src/anemoi/models/layers/bounding.py
@@ -13,7 +13,6 @@ from abc import ABC
 from abc import abstractmethod
 from typing import Optional
 
-import numpy as np
 import torch
 from torch import nn
 
--- a/models/src/anemoi/models/layers/bounding.py
+++ b/models/src/anemoi/models/layers/bounding.py
@@ -14,6 +14,7 @@ from abc import abstractmethod
 from typing import Optional
 
 import torch
+import numpy as np
 from torch import nn
 
 from anemoi.models.data_indices.tensor import InputTensorIndex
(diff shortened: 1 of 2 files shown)


models/src/anemoi/models/layers/bounding.py
```diff
@@ -153,6 +155,9 @@ def __init__(
                 std = self.statistics["stdev"][stat_index]
                 self.norm_min_val[ii] = min_val[ii] / std
 
+        # Reorder normalized min values based on data_index
+        self.norm_min_val = norm_min_val[torch.argsort(torch.tensor(self.data_index))]
```
sahahner: `self.data_index` should lie on the CPU, so I would use numpy operations here.

lzampier: Good point! Now using `np.array` and `np.argsort`.

lzampier:
The effect of the bug does only show in the training metrics. I can write a test that fails if the order of variables is wrong after the initialization.

lzampier:
> Would it make sense to add a test that would catch this bug next time?

Yes, more tests would be good. I would postpone this to a different pull request, given that we will refactor the interface of this function to make it simpler over the next days. The tests will also inevitably change for the refactor.

sahahner (approved):
Looks good.

PR #102, opened 2025-01-31
fix(training,config)!: refactor node_weights

JPXKQX:
- [x] Move node_weights to loss scalers
- [x] support unit-norm in node_weights
- [x] Fix Limited Area Mask 
- [x] Fix Limited area mse (mse contributions)
- [x] Fix output mask over scalers
- [x] Fix post processors over scalers

Test:
- Global
- LAM (output_mask) 
- Stretched (limited_area_mask metrics)
- combined loss

to discuss:
- MSELoss vs MSE
- scalers.builders, other options?
- filenames (base.py,...)
- validation metrics, list -> dict

----
📚 Documentation preview 📚: https://anemoi-training--102.org.readthedocs.build/en/102/

----
📚 Documentation preview 📚: https://anemoi-graphs--102.org.readthedocs.build/en/102/

----
📚 Documentation preview 📚: https://anemoi-models--102.org.readthedocs.build/en/102/


training/src/anemoi/training/losses/logcosh.py
```diff
@@ -94,4 +89,8 @@ def forward(
         """
         out = LogCosh.apply(pred - target)
         out = self.scale(out, scaler_indices, without_scalers=without_scalers)
-        return self.scale_by_node_weights(out, squash)
+
+        if squash:
+            out = self.avg_function(out, dim=-1)
+            
+        return self.sum_function(out, dim=(0, 1, 2))
```
HCookie: Considering this is identical to all losses, should this be provided as a function in the BaseLoss?

training/src/anemoi/training/losses/weightedloss.py
```diff
@@ -179,15 +138,19 @@ def forward(
...
+
 
     @property
     def name(self) -> str:
         """Used for logging identification purposes."""
         return self.__class__.__name__.lower()
 
 
-class FunctionalWeightedLoss(BaseWeightedLoss):
+class FunctionalWeightedLoss(BaseLoss):
```
HCookie: Given the Weighed part has been removed from the Base class, this should be `FunctionalLoss`. Same for the others

training/src/anemoi/training/losses/logcosh.py
JPXKQX: I may inherit from FunctionalLoss and just overwrite the calculate_difference. Does this make sense to you?

training/src/anemoi/training/config/training/scalers/global.yaml
```diff
@@ -4,13 +4,15 @@ variable_groups:
 
 # Several scalers can be added here. In order to be applied their names must be included in the loss.
 # scaler name must be included in `scalers` in the losses for this to be applied.
+# All scalers needs a `scale_dims` argument representing the dimension/s on which it is applied
```
sahahner: This should be updated as some scale_dims are now fixed.

training/src/anemoi/training/losses/base.py
```diff
@@ -179,35 +137,29 @@ def forward(
...
         out = self.scale(out, scaler_indices, without_scalers=without_scalers)
 
-        return self.scale_by_node_weights(out, squash)
+        if squash:
+            out = self.avg_function(out, dim=-1)
 
-    @property
-    def name(self) -> str:
-        """Used for logging identification purposes."""
-        return self.__class__.__name__.lower()
```
HCookie, line 193: Why has this been removed?

training/src/anemoi/training/losses/scaling/__init__.py
```diff
@@ -17,28 +17,83 @@
...
+
+        if self.norm.lower() == "unit-mean":
+            return values / np.mean(values)
+
+        error_msg = f"{self.norm} must be one of: None, unit-sum, l1, unit-mean."
+        raise ValueError(error_msg)
+
+
+class BaseDelayedScaler(BaseScaler, ABC):
+    pass
```
HCookie: This seems odd, why is there a subclass with the a parent class readded, but with no extra's included?

training/src/anemoi/training/losses/scaling/node_attributes.py
```diff
@@ -0,0 +1,66 @@
...
+
+if TYPE_CHECKING:
+    import numpy as np
+    from torch_geometric.data import HeteroData
+
+    from anemoi.models.data_indices.collection import IndexCollection
+
+
+class GraphNodeAttributeScaler(BaseScaler, ABC):
+    """Base class for all loss masks that are more than one-dimensional."""
```
HCookie: That docstring implies this class is a generic, whereas it appeats to be for graphnodes

training/src/anemoi/training/losses/scaling/__init__.py
HCookie: At least add some docstrings

training/src/anemoi/training/losses/scaling/scaling.py
```diff
@@ -0,0 +1,92 @@
...
+            delayed_scaler_builders[name] = scaler_builder
+            scalers[name] = (scaler_builder.scale_dims, np.ones(tuple([1] * len(scaler_builder.scale_dims))))
+            continue
+
+        scaler_values = scaler_builder.get_scaling()
+
+        # If a scaler needs to apply the output mask (LAM) after its creation,
+        # it must include the apply_output_mask attribue.
+        if scaler_builder.is_spatial_dim_scaled and getattr(scaler_builder, "apply_output_mask", False):
+            scaler_values = output_mask.apply(scaler_values, dim=0, fill_value=0.0)
```
HCookie: Instead of having an attribute check which seems a badly defined interface, how about you provide a generic `apply_output_mask` in the `Scalar` class which does nothing by default.

training/src/anemoi/training/losses/scaling/scaling.py
```diff
@@ -0,0 +1,92 @@
...
+    torch.Tensor
+        Final variable scaling.
+    """
+    # Subsetting over -1 to get the right scalar for printing across the variable dimension
+    final_variable_scaling = 1.0
+    for scale_dims, scaling in scalers.values():
+        if -1 in scale_dims or 3 in scale_dims:
+            final_variable_scaling = final_variable_scaling * scaling.squeeze()
+
+    return final_variable_scaling
```
HCookie: If a full `ScaleTensor` is created, you could use it's ability of subsetting instead of this.

training/src/anemoi/training/losses/scaling/scaling.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def print_final_variable_scaling(
+    scalers: dict[str, tuple[tuple[int] | torch.Tensor]],
+    data_indices: IndexCollection,
+) -> None:
+    final_variable_scaling = get_final_variable_scaling(scalers)
```
HCookie: See above comment, but having a seperate function is probably not needed

HCookie (changes requested):

training/src/anemoi/training/schemas/training.py
```diff
@@ -87,58 +87,101 @@ class LossScalingSchema(BaseModel):
...
+    mse = "anemoi.training.losses.MSELoss"
+    mae = "anemoi.training.losses.MAELoss"
+    logcosh = "anemoi.training.losses.LogCoshLoss"
 
 
 class BaseLossSchema(BaseModel):
     target_: ImplementedLossesUsingBaseLossSchema = Field(..., alias="_target_")
     "Loss function object from anemoi.training.losses."
-    scalars: list[PossibleScalars] = Field(example=["variable"])
+    scalers: list[str] = Field(example=["variable"])
```
HCookie: It may be useful to try and validate the contents of this list with the scalar dictionary.

training/src/anemoi/training/config/training/lam.yaml
```diff
@@ -0,0 +1,119 @@
...
+  mse:
+    _target_: anemoi.training.losses.MSELoss
+    # Scalers to include in loss calculation
+    # Cannot scale over the variable dimension due to possible remappings.
+    # Available scalers include:
+    # - 'loss_weights_mask': Giving imputed NaNs a zero weight in the loss function
+    # Use the `scale_validation_metrics` section to variable scale.
+    scalers: []
+    # other kwargs
+    ignore_nans: True
```
HCookie, line 78: This part of the config will need to be updated, mainly the docs, but that can be in the main branch

training/src/anemoi/training/losses/base.py
HCookie: @JPXKQX This needs to be addressed.

training/src/anemoi/training/losses/base.py
```diff
@@ -179,35 +137,29 @@ def forward(
...
+        return self.sum_function(out, dim=(0, 1, 2))
```
HCookie: Considering this behaviour is shared between the losses, it would be good to provide it as a function.

training/src/anemoi/training/losses/combined.py
HCookie, line 1: I think this file is not up to date with the current state in main.

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -0,0 +1,98 @@
...
+        assert self.scale_dims is not None, f"Class {self.__class__.__name__} must define 'scale_dims'"
+        if isinstance(self.scale_dims, int):
+            self.scale_dims = (self.scale_dims,)
+        assert all(
+            -4 <= d <= 3 for d in self.scale_dims
+        ), f"Invalid dimension for scaling in 'scale_dims': {self.scale_dims}"
+
+    @property
+    def is_spatial_dim_scaled(self) -> bool:
+        return self.scale_dims is not None and (-2 in self.scale_dims or 2 in self.scale_dims)
```
HCookie: An enum for the dimensions of the tensor would be good to define and use here.

HCookie: Plus, providing all `is_dim_name_scaled` would be useful.

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @abstractmethod
+    def get_scaling_values(self, **kwargs) -> np.ndarray:
+        """Abstract method to get loss scaling."""
+        ...
+
+    def normalise(self, values: np.ndarray) -> np.ndarray:
+        if self.norm is None:
```
HCookie, line 63: Can a docstring be added.

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -0,0 +1,98 @@
...
+    def get_scaling(self) -> SCALER_DTYPE:
+        scaler_values = self.get_scaling_values()
+        scaler_values = self.normalise(scaler_values)
+        return self.scale_dims, scaler_values
+
+
+class BaseDelayedScaler(BaseScaler):
+    """Base class for delayed Scalers.
+
+    The delayed scalers are only initialise when creating all the scalers, but its value is
```
HCookie:
```suggestion
    The delayed scalers are only initialised when creating all the scalers, but its value is
```

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -0,0 +1,98 @@
...
+    computed during the first iteration of the training loop. This delayed scalers are suitable
+    for scalers requiring information from the `model.pre_processors`.
+    """
+
+    @abstractmethod
+    def get_delayed_scaling_values(self, **kwargs) -> np.ndarray: ...
```
HCookie, line 99: What is the expected arguments this function is called with?

HCookie (changes requested):
Looks good, just a couple things to address before merging to the other branch.

training/src/anemoi/training/losses/scalers/base_scaler.py
JPXKQX: model.pre_processors, model or something similar

training/src/anemoi/training/losses/base.py
JPXKQX: function `reduce` has been created. `FunctionalLoss` has also been removed.

training/src/anemoi/training/config/training/lam.yaml
JPXKQX: I am not sure what to expect here. We have the same in all configs (stretched, global and lam)

training/src/anemoi/training/losses/base.py
JPXKQX: The name is no longer used as a class attribute. The user can specify the same class with different combinations of scalers, so we move to a dict-based definition in the config

training/src/anemoi/training/train/forecaster.py
```diff
@@ -623,7 +427,7 @@ def calculate_val_metrics(
     def training_step(self, batch: torch.Tensor, batch_idx: int) -> torch.Tensor:
         train_loss, _, _ = self._step(batch, batch_idx)
         self.log(
-            f"train_{getattr(self.loss, 'name', self.loss.__class__.__name__.lower())}",
+            "train_loss",
```
HCookie, line 427: I'm not sure I like the opaque-ness of this in the logs for the training

training/src/anemoi/training/losses/base.py
HCookie: I get that for the validation metrics, but the training loss can still benefit from a name.

training/src/anemoi/training/losses/base.py
HCookie: I disagree with the removal of `FunctionalLoss`, a user may want to subclass from the `BaseLoss`, but fully control the `forward` pass, so the `calculate_difference` becomes an obtuse function.

training/src/anemoi/training/config/training/lam.yaml
HCookie: Lets revisit this in the loss_scaling branch

training/src/anemoi/training/losses/scalers/base_scaler.py
HCookie: Then lets set in this function, so subclasses can know what to expect.

training/src/anemoi/training/losses/combined.py
HCookie, line 1: Can you update this with the new get_loss_function location?

training/src/anemoi/training/losses/base.py
JPXKQX: Thinking about this, I have put my thoughts here, let me know what you think. 
- Note that the name used in loss train_{name} is also used for val_{name}.
- I think we both agree that this name should be the same for both metrics. 
- I also agree that train_mse is better (more informative) than train_loss. 

But since the validation metrics can include other mse metrics, and the user can specify any key in the validation_metrics, I think it is more convenient to use train_loss & val_loss.

training/src/anemoi/training/train/forecaster.py
JPXKQX: Discussing this in another comment.

training/src/anemoi/training/losses/scaling/scaling.py
JPXKQX: Here the `scaling` is a np.ndarray. We could move after instantiating the loss function

training/src/anemoi/training/losses/scalers/base_scaler.py
JPXKQX: Done

training/src/anemoi/training/losses/base.py
HCookie: How about nest the training loss with another layer, a dictionary of one element with the key being the name, loss as default but a user can name it if they wish?

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -0,0 +1,111 @@
...
+
+LOGGER = logging.getLogger(__name__)
+SCALER_DTYPE = tuple[tuple[int], np.ndarray]
+
+
+class ScalerDim(Enum):
+    BATCH_SIZE = 0
+    ENSEMBLE_DIM = 1
+    GRID = 2
+    VARIABLE = 3
```
HCookie: This is more generic, a TensorDim, should it be elevated to the top level?

JPXKQX: Move to anemoi.training.utils.enums.TensorDim

PR #103, opened 2025-01-31
fix: Combined Loss

HCookie:
# Fix Combined Loss
The combined loss was never working (whoops)

This fixes it such that CombinedLoss looks like any other loss, with no interface changes to the forecaster.
Will allow for multiple loss functions to be used together
- Weighted on a single prediction
- Attending to only parts of a prediction

```yaml
        training_loss:
            losses:
                - _target_: anemoi.training.losses.mse.WeightedMSELoss
                  scalars: ['ocean']
                - _target_: anemoi.training.losses.mae.WeightedMAELoss
                  scalars: ['atmosphere']
            scalars: ['*']
            loss_weights: [1.0, 1.0]
```

### DIFF
--- a/training/docs/modules/losses.rst
+++ b/training/docs/modules/losses.rst
@@ -140,10 +140,10 @@ losses above.
 .. code:: yaml
 
    training_loss:
-      __target__: anemoi.training.losses.combined.CombinedLoss
+      _target_: anemoi.training.losses.combined.CombinedLoss
       losses:
-         - __target__: anemoi.training.losses.mse.WeightedMSELoss
-         - __target__: anemoi.training.losses.mae.WeightedMAELoss
+         - _target_: anemoi.training.losses.mse.WeightedMSELoss
+         - _target_: anemoi.training.losses.mae.WeightedMAELoss
       scalars: ['variable']
       loss_weights: [1.0,0.5]
 
--- a/training/src/anemoi/training/losses/combined.py
+++ b/training/src/anemoi/training/losses/combined.py
@@ -75,8 +75,20 @@ class CombinedLoss(BaseWeightedLoss):
             losses:
                 - _target_: anemoi.training.losses.mse.WeightedMSELoss
                 - _target_: anemoi.training.losses.mae.WeightedMAELoss
-            scalars: ['variable']
-            loss_weights: [1.0,0.5]
+            scalars: ['*']
+            loss_weights: [1.0, 0.6]
+        ```
+
+        ```
+        training_loss:
+            _target_: anemoi.training.losses.combined.CombinedLoss
+            losses:
+                - _target_: anemoi.training.losses.mse.WeightedMSELoss
+                  scalars: ['ocean']
+                - _target_: anemoi.training.losses.mae.WeightedMAELoss
+                  scalars: ['atmosphere']
+            scalars: ['*']
+            loss_weights: [1.0, 1.0]
         ```
         """
         self.losses: list[BaseWeightedLoss] = []
--- a/training/src/anemoi/training/losses/combined.py
+++ b/training/src/anemoi/training/losses/combined.py
@@ -52,7 +52,7 @@ class CombinedLoss(torch.nn.Module):
         Examples
         --------
         >>> CombinedLoss(
-                {"__target__": "anemoi.training.losses.mse.WeightedMSELoss"},
+                {"_target_": "anemoi.training.losses.mse.WeightedMSELoss"},
                 loss_weights=(1.0,),
                 node_weights=node_weights
             )
@@ -66,10 +66,10 @@ class CombinedLoss(torch.nn.Module):
 
         ```
         training_loss:
-            __target__: anemoi.training.losses.combined.CombinedLoss
+            _target_: anemoi.training.losses.combined.CombinedLoss
             losses:
-                - __target__: anemoi.training.losses.mse.WeightedMSELoss
-                - __target__: anemoi.training.losses.mae.WeightedMAELoss
+                - _target_: anemoi.training.losses.mse.WeightedMSELoss
+                - _target_: anemoi.training.losses.mae.WeightedMAELoss
             scalars: ['variable']
             loss_weights: [1.0,0.5]
         ```
--- a/training/src/anemoi/training/losses/combined.py
+++ b/training/src/anemoi/training/losses/combined.py
@@ -10,15 +10,20 @@
 from __future__ import annotations
 
 import functools
+from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
 
-import torch
+from omegaconf import DictConfig
 
-from anemoi.training.train.forecaster import GraphForecaster
+from anemoi.training.losses.utils import ScaleTensor
+from anemoi.training.losses.weightedloss import BaseWeightedLoss
+
+if TYPE_CHECKING:
+    import torch
 
 
-class CombinedLoss(torch.nn.Module):
+class CombinedLoss(BaseWeightedLoss):
     """Combined Loss function."""
 
     def __init__(
@@ -74,17 +79,25 @@ class CombinedLoss(torch.nn.Module):
             loss_weights: [1.0,0.5]
         ```
         """
-        super().__init__()
+        self.losses: list[BaseWeightedLoss] = []
+        super().__init__(node_weights=None)
 
         losses = (*(losses or []), *extra_losses)
 
         assert len(losses) == len(loss_weights), "Number of losses and weights must match"
         assert len(losses) > 0, "At least one loss must be provided"
 
-        self.losses = [
-            GraphForecaster.get_loss_function(loss, **kwargs) if isinstance(loss, dict) else loss(**kwargs)
-            for loss in losses
-        ]
+        from anemoi.training.train.forecaster import GraphForecaster
+
+        for i, loss in enumerate(losses):
+            self.losses.append(
+                (
+                    GraphForecaster.get_loss_function(loss, **kwargs)
+                    if isinstance(loss, (DictConfig, dict))
+                    else loss(**kwargs)
+                ),
+            )
+            self.add_module(self.losses[-1].name + str(i), self.losses[-1])
         self.loss_weights = loss_weights
 
     def forward(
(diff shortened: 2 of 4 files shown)


training/src/anemoi/training/losses/combined.py
```diff
@@ -66,25 +71,45 @@ def __init__(
...
         losses = (*(losses or []), *extra_losses)
 
         assert len(losses) == len(loss_weights), "Number of losses and weights must match"
         assert len(losses) > 0, "At least one loss must be provided"
 
-        self.losses = [
-            GraphForecaster.get_loss_function(loss, **kwargs) if isinstance(loss, dict) else loss(**kwargs)
-            for loss in losses
-        ]
+        from anemoi.training.train.forecaster import GraphForecaster
```
chebertpinard: Any reason why this line has been move in the class?

training/src/anemoi/training/train/forecaster.py
```diff
@@ -224,10 +224,11 @@ def get_loss_function(
         scalars_to_include = loss_config.pop("scalars", [])
 
         # Instantiate the loss function with the loss_init_config
+        kwargs["_recursive_"] = kwargs.get("_recursive_", False)
```
chebertpinard, line 254: Why do we need the recursive key?

training/src/anemoi/training/losses/combined.py
```diff
@@ -66,25 +71,45 @@ def __init__(
...
+        ```
+        training_loss:
+            _target_: anemoi.training.losses.combined.CombinedLoss
+            losses:
+                - _target_: anemoi.training.losses.mse.WeightedMSELoss
+                  scalars: ['ocean']
+                - _target_: anemoi.training.losses.mae.WeightedMAELoss
+                  scalars: ['atmosphere']
+            scalars: ['*']
+            loss_weights: [1.0, 1.0]
```
chebertpinard, line 121: Have you test this example with different scalars for each loss function?

training/src/anemoi/training/losses/combined.py
HCookie: No reason, I shall revert the change.

training/src/anemoi/training/train/forecaster.py
HCookie: The recursive key has to be set to False for the CombinedLoss so that the underlying losses are not instantiated before they need to be.

training/src/anemoi/training/losses/combined.py
HCookie: Not yet, I shall

HCookie: Added in latest commit

HCookie: Added in [5a7664d](https://github.com/ecmwf/anemoi-core/pull/103/commits/5a7664d9660bf91a778312444e37fc9c078c0c7e)

training/src/anemoi/training/train/forecaster.py
```diff
@@ -186,8 +186,8 @@ def get_loss_function(
         ----------
         config : DictConfig
             Loss function configuration, should include `scalars` if scalars are to be added to the loss function.
-        scalars : Union[dict[str, tuple[Union[int, tuple[int, ...], torch.Tensor]]], None], optional
-            Scalars which can be added to the loss function. Defaults to None., by default None
+        scalars : dict[str, tuple[Union[int, tuple[int, ...], torch.Tensor]]] | None,
+            Scalars which can be added to the loss function. Defaults to None.,
```
chebertpinard, line 215: Why is the scalars value not optional anymore? Should "Defaults to None.," be remove from the description?

training/docs/modules/losses.rst
```diff
@@ -140,16 +140,33 @@ losses above.
...
-before combining them.
+.. code:: yaml
+
+   training_loss:
+      _target_: anemoi.training.losses.combined.CombinedLoss
+      losses:
+            - _target_: anemoi.training.losses.mse.WeightedMSELoss
+            scalars: ['ocean']
+            - _target_: anemoi.training.losses.mae.WeightedMAELoss
+            scalars: ['atmosphere']
```
chebertpinard: Maybe just a last comment. With the pydantic schema, I don't think this example would work because of this line https://github.com/ecmwf/anemoi-core/blob/5628bbae297bbcad13b35503ef7ddbde4da6db10/training/src/anemoi/training/schemas/training.py#L108
https://github.com/ecmwf/anemoi-core/blob/5628bbae297bbcad13b35503ef7ddbde4da6db10/training/src/anemoi/training/schemas/training.py#L118-L123
But I don't know if the example isn't precise enough or if it is the pydantic verification that is too strict.

Plus I think scalars is missing from the CombinedLossSchema
https://github.com/ecmwf/anemoi-core/blob/5628bbae297bbcad13b35503ef7ddbde4da6db10/training/src/anemoi/training/schemas/training.py#L137-L140

Maybe just wait the pydantic merge before merging

chebertpinard (approved):

training/docs/modules/losses.rst
HCookie: There will have to be work to improve the validation of the scalars, mainly because of the loss scaling refactor currently underway. 
This example is in preperation of that PR.

training/src/anemoi/training/schemas/training.py
```diff
@@ -137,18 +140,44 @@ class WeightedMSELossLimitedAreaSchema(BaseLossSchema):
...
+
+        use_attribute_docstrings = True
+        use_enum_values = True
+        validate_assignment = True
+        validate_default = True
+        extra = "allow"  # Allow extra kwargs to be passed to underlying losses
+
+    @field_validator("losses", mode="before")
+    @classmethod
+    def add_empty_scalars(cls, losses: Any) -> Any:
```
theissenhelen, line 152: We need to be a bit careful here, because we are adding a default which does not exist for the no-validation option. For the first draft we have decided to not use defaults, so that the user configs are complete for both cases with validation and without.

training/src/anemoi/training/train/forecaster.py
```diff
@@ -232,6 +232,8 @@ def get_loss_function(
         ValueError
             If scalar is not found in valid scalars
         """
+        scalars = scalars or {}
+
         if isinstance(config, ListConfig):
```
theissenhelen, line 237: Just a thought: but couldn't we make this simpler by assuming that we are always using a "CombinedLoss". In the standard use case, we simply have a list with only one element?

training/src/anemoi/training/schemas/training.py
HCookie: Ahh gotcha. But I'm not sure this applies in this context.
This is not quite a default but an override so that the validation works. From the combinedloss perspective not having scalars is fine, but using the schemas prevents that.

HCookie: In this case, the no validation is right, and the validation must be corrected.

training/src/anemoi/training/train/forecaster.py
HCookie: Possibly, but it would be good to abstract that so a user doesn't have to specify CombinedLoss but with one entry.

training/src/anemoi/training/schemas/training.py
```diff
@@ -137,18 +140,44 @@ class WeightedMSELossLimitedAreaSchema(BaseLossSchema):
...
+    "Weightings of losses, if not set, all losses are weighted equally."
+
+    class Config:
+        """Pydantic configuration."""
+
+        use_attribute_docstrings = True
+        use_enum_values = True
+        validate_assignment = True
+        validate_default = True
+        extra = "allow"  # Allow extra kwargs to be passed to underlying losses
```
theissenhelen: With extra="allow typos in the configs will not be detected, maybe we should add a comment for that as well.

training/src/anemoi/training/schemas/training.py
theissenhelen: Right, it is not optional from the schema perspective.

training/docs/modules/losses.rst
chebertpinard: The scalars issue is still there. "*", "ocean" or "atmosphere" would not be accepted by the pydantic schema at the moment.

```
anemoi.training.schemas.utils.ValidationError: {"Value * not in ['variable', 'loss_weights_mask']"}
```

training/docs/modules/losses.rst
```diff
@@ -140,16 +140,36 @@ losses above.
...
-All kwargs passed to ``CombinedLoss`` are passed to each of the loss
-functions, and the loss weights are used to scale the individual losses
-before combining them.
+.. code:: yaml
+
+   training_loss:
+      _target_: anemoi.training.losses.combined.CombinedLoss
+      losses:
+            - _target_: anemoi.training.losses.mse.WeightedMSELoss
+            scalars: ['ocean']
```
chebertpinard: Small detail but I got a yaml error with this indentation. I had to align `scalars` with `_target_`

```yaml
training_loss:
  losses:
     - _target_: anemoi.training.losses.mse.WeightedMSELoss
       scalars: ['ocean']
```

training/src/anemoi/training/schemas/training.py
```diff
@@ -137,18 +140,45 @@ class WeightedMSELossLimitedAreaSchema(BaseLossSchema):
...
+    def check_length_of_weights_and_losses(self) -> CombinedLossSchema:
+        """Check that the number of losses and weights match, or if not set, skip."""
+        losses, loss_weights = self.losses, self.loss_weights
+        if loss_weights is None:
+            return
+
         if len(losses) != len(loss_weights):
             error_msg = "Number of losses and weights must match"
             raise ValueError(error_msg)
-        return values
```
chebertpinard, line 151: This validation function has to return something (CombinedLossSchema) otherwise you get a NoneType error.
```
ValueError: Input cfg is not an OmegaConf config object (NoneType)
```

I don't know if it is a good way to do it, but returning `self` seems to be working.

training/src/anemoi/training/schemas/training.py
```diff
@@ -137,18 +140,45 @@ class WeightedMSELossLimitedAreaSchema(BaseLossSchema):
...
+    @classmethod
+    def add_empty_scalars(cls, losses: Any) -> Any:
+        """Add empty scalars to loss functions, as scalars can be set at top level."""
+        from omegaconf.omegaconf import open_dict
+
+        for loss in losses:
+            if "scalars" not in loss:
+                with open_dict(loss):
+                    loss["scalars"] = []
+        return losses
```
chebertpinard, line 160: Would it works to set `[]` as a default value in the BaseLossSchema?

training/docs/modules/losses.rst
HCookie: This behaviour relies on the loss scaling PR

training/src/anemoi/training/schemas/training.py
HCookie: It would work, but conflict with the expected behaviour of the BaseLossSchema. Losses without scalars doesn't make too much sense.

chebertpinard (dismissed):

training/src/anemoi/training/schemas/training.py
HCookie: Removed this in the latest commit.

training/docs/modules/losses.rst
```diff
@@ -140,16 +140,36 @@ losses above.
...
+
+   training_loss:
+      _target_: anemoi.training.losses.combined.CombinedLoss
+      losses:
+            - _target_: anemoi.training.losses.mse.WeightedMSELoss
+              scalars: ['variable']
+            - _target_: anemoi.training.losses.mae.WeightedMAELoss
+              scalars: ['loss_weights_mask']
+      loss_weights: [1.0, 1.0]
+      scalars: ['*']
```
chebertpinard, line 172: scalars: ['*'] would not work with the current pydantic schema. If it is just for an example it's not too bad, but if you would like the option to be available, it might be better to add it to the PossibleScalars
```py
PossibleScalars = Annotated[str, AfterValidator(partial(allowed_values, values=["variable", "loss_weights_mask","*"]))]
```

HCookie: Good spot, forgot that

chebertpinard (approved):

PR #143, opened 2025-02-18
fix:  dataset schema too defined too strictly

### LINKED ISSUE #141: fix: dataset schema too defined too strictly
### What happened?

The DatasetSchema in the DataloaderSchema is too strictly defined too allow all options specified by anemoi-datasets. See https://anemoi.readthedocs.io/projects/datasets/en/latest/using/opening.html.

Any additional operations specified in the data loader config (e.g. joining two datasets would reproduce the bug).

training:
    #dataset: ${dataloader.dataset}
    start: "2022-12-01T00:00:00"
    end: "2022-12-03T18:00:00"
    join:
    - dataset1
    - dataset2
    frequency: ${data.frequency}
 

### Version

on branch main

Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

ECMWF

theissenhelen:
## Description

The DatasetSchema in the DatalaoderSchema also allows dictionaries now to match the broad configuration options in anemoi-datasets. This PR also sets the paths and files in the HardwareSchema that are not referenced in the code to optional fields.

## Type of Change

## Issue Number

Closes #141 

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

----
📚 Documentation preview 📚: https://anemoi-training--143.org.readthedocs.build/en/143/

----
📚 Documentation preview 📚: https://anemoi-graphs--143.org.readthedocs.build/en/143/

----
📚 Documentation preview 📚: https://anemoi-models--143.org.readthedocs.build/en/143/

### DIFF
--- a/training/src/anemoi/training/schemas/dataloader.py
+++ b/training/src/anemoi/training/schemas/dataloader.py
@@ -14,6 +14,7 @@ import datetime  # noqa: TC003
 from pathlib import Path  # noqa: TC003
 from typing import Any
 from typing import Literal
+from typing import Optional
 from typing import Union
 
 from omegaconf import DictConfig  # noqa: TC002
@@ -62,7 +63,7 @@ class Frequency(RootModel):
 class DatasetSchema(PydanticBaseModel):
     """Dataset configuration schema."""
 
-    dataset: Union[str, dict, Path, list[dict]] | None = None
+    dataset: Optional[Union[str, dict, Path, list[dict]]] = None
     "Dataset, see anemoi-datasets"
     start: Union[str, int, None] = Field(default=None)
     "Starting datetime for sample of the dataset."
--- a/training/src/anemoi/training/schemas/dataloader.py
+++ b/training/src/anemoi/training/schemas/dataloader.py
@@ -16,7 +16,9 @@ from typing import Any
 from typing import Literal
 from typing import Union
 
+from omegaconf import DictConfig  # noqa: TC002
 from pydantic import BaseModel as PydanticBaseModel
+from pydantic import ConfigDict
 from pydantic import Field
 from pydantic import PositiveInt
 from pydantic import RootModel
@@ -57,10 +59,10 @@ class Frequency(RootModel):
         return int(self.as_timedelta.total_seconds())
 
 
-class DatasetSchema(BaseModel):
+class DatasetSchema(PydanticBaseModel):
     """Dataset configuration schema."""
 
-    dataset: Union[str, dict, Path, list[dict]]
+    dataset: Union[str, dict, Path, list[dict]] | None = None
     "Dataset, see anemoi-datasets"
     start: Union[str, int, None] = Field(default=None)
     "Starting datetime for sample of the dataset."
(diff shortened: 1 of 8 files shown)


training/src/anemoi/training/schemas/dataloader.py
```diff
@@ -57,10 +59,10 @@ def as_seconds(self) -> int:
         return int(self.as_timedelta.total_seconds())
 
 
-class DatasetSchema(BaseModel):
+class DatasetSchema(PydanticBaseModel):
     """Dataset configuration schema."""
 
-    dataset: Union[str, dict, Path, list[dict]]
+    dataset: Union[str, dict, Path, list[dict]] | None = None
```
anaprietonem: I think you might need to update this type hint to avoid using | so unit tests can pass with 3.9, something like  `Union[Union[str, dict, Path, list[dict]],None]` should work

training/src/anemoi/training/schemas/hardware.py
```diff
@@ -53,19 +53,19 @@ class Logs(PydanticBaseModel):
...
+    data: Path | None = None
     "Path to the data directory."
-    graph: Path
+    graph: Path | None = None
     "Path to the graph directory."
-    output: Path
+    output: Path | None = None
     "Path to the output directory."
-    logs: Logs
+    logs: Logs | None = None
```
anaprietonem: can those be none? just wondering that if those are none and your experiment tracker it's activated then training would fail

training/src/anemoi/training/schemas/dataloader.py
```diff
@@ -104,6 +106,9 @@ class MaskedGridIndicesSchema(BaseModel):
 
 
 class DataLoaderSchema(PydanticBaseModel):
+
+    model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
```
anaprietonem, line 111: This will allow to have any entries then on the dataloaderschema right? could we just not add a new entry for the join rather than leaving it so open?

theissenhelen: I considered the same. However, I think that would block the fix as we would have to go through all possible variations of the config for anemoi-dataset first (see https://anemoi.readthedocs.io/projects/datasets/en/latest/using/opening.html or here https://anemoi.readthedocs.io/projects/datasets/en/latest/building/operations.html). I would suggest putting that as a separate PR.

training/src/anemoi/training/schemas/hardware.py
theissenhelen: Yes, they can be none. I have added a check to the respective model validator in the BaseSchema

training/src/anemoi/training/schemas/dataloader.py
anaprietonem: Okey so your proposal is to lax the schema for dataloader definition for now and then a separate PR to make strict again but with all entries of the schema according to anemoi-datasets?

anaprietonem (dismissed):

training/src/anemoi/training/schemas/dataloader.py
theissenhelen: Exactly, these schemas should probably live in Anemoi-datasets

anaprietonem (approved):

PR #148, opened 2025-02-20
feat: Log FTT2 loss + Fourier Correlation loss 

OpheliaMiralles:
## Type of Change

-   [ ] Documentation update

## Issue Number

new loss function

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

----
📚 Documentation preview 📚: https://anemoi-training--148.org.readthedocs.build/en/148/

----
📚 Documentation preview 📚: https://anemoi-graphs--148.org.readthedocs.build/en/148/

----
📚 Documentation preview 📚: https://anemoi-models--148.org.readthedocs.build/en/148/


mchantry:
Exciting contribution, thanks.

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,64 @@
...
+    power_spectra_fake = torch.abs(torch.fft.rfft2(fake_output)) ** 2
+    ratio = (power_spectra_real + epsilon) / (power_spectra_fake + epsilon)
+
+    def log10(x: torch.Tensor) -> torch.Tensor:
+        return torch.log(x) / torch.log(torch.tensor(10.0, device=x.device, dtype=x.dtype))
+
+    return (10 * log10(ratio)) ** 2
+
+
+class LogSpectralDistance(FunctionalWeightedLoss):
```
mchantry: I wonder if a different name might be better, more precise that this uses 2DFFTs for the spectra.
It's possible we may want to add a Spherical harmonic spectral distance later, and both could be interpreted as a spectral distance.

OpheliaMiralles: Which name would you suggest? I am also wondering whether this current implementation is correct (that is why I left it in draft mode) because anemoi-datasets uses flattened spatial coordinates. What I had in mind is computing the 2D fourier transform on time/batch sequences of 2D fields. I have another implementation, however, it takes as input of the loss a field shape and reshape the tensor before computing the loss, e.g. it would not work with unstructured grids. So I don't know whether to drop it or update the PR with the 2D version and if the x/y dim are not provided, it would raise an error anyway.

mchantry:
Sorry for the slow re-review Ophelia.
I think this will work for rectilinear grids. Do you have a thought on how we could ensure that it is only used in this context, and not misused?
Would you be able to provide an update to the docs please.

OpheliaMiralles:
> Sorry for the slow re-review Ophelia. I think this will work for rectilinear grids. Do you have a thought on how we could ensure that it is only used in this context, and not misused? Would you be able to provide an update to the docs please.

No worries, I was waiting for a signal to start working on this again. I might have some time next week or the following.

OpheliaMiralles:
> Sorry for the slow re-review Ophelia. I think this will work for rectilinear grids. Do you have a thought on how we could ensure that it is only used in this context, and not misused? Would you be able to provide an update to the docs please.

Updated the doc and added a warning

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,155 @@
...
+) -> torch.Tensor:
+    return power_spectra_real * power_spectra_fake.conj() + power_spectra_fake * power_spectra_real.conj()
+
+
+def get_spectra(
+    fake_output: torch.Tensor,
+    real_output: torch.Tensor,
+    dims: tuple[int, int],
+) -> tuple[torch.Tensor, torch.Tensor]:
+    assert dims[0] * dims[1] == real_output.shape[2]
```
HCookie: Can you add a message to this assert, so if it is triggered, a useful message is shown

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,155 @@
...
+        return torch.log(x) / torch.log(torch.tensor(10.0, device=x.device, dtype=x.dtype))
+
+    log_10 = (10 * log10(ratio)) ** 2
+    return torch.sqrt(torch.mean(log_10)) / 10
+
+
+class LogFFT2Distance(FunctionalLoss):
+    r"""The log spectral distance is used to compute the difference between spectra of two fields.
+
+    It is also called log spectral distorsion.
```
HCookie: `Distortion`?

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,155 @@
...
+    dims: tuple[int, int],
+) -> tuple[torch.Tensor, torch.Tensor]:
+    assert dims[0] * dims[1] == real_output.shape[2]
+    dims_total = (*real_output.shape[:2], *dims, real_output.shape[-1])
+    power_spectra_real = torch.fft.rfft2(real_output.reshape(dims_total), dim=(-2, -3))
+    power_spectra_fake = torch.fft.rfft2(fake_output.reshape(dims_total), dim=(-2, -3))
+    return power_spectra_real, power_spectra_fake
+
+
+def log_rfft2_distance(fake_output: torch.Tensor, real_output: torch.Tensor, dims: tuple[int, int]) -> torch.Tensor:
```
HCookie: Why is the prediction being called `fake_output`?

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,155 @@
...
+    power_spectra_real, power_spectra_fake = get_spectra(fake_output, real_output, dims)
+    epsilon = torch.finfo(real_output.dtype).eps  # Small epsilon to avoid division by zero
```
HCookie: Please add a docstring to this function

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,155 @@
...
+        y_dim: int,
+        ignore_nans: bool = False,
+    ) -> None:
+        super().__init__(ignore_nans)
+        LOGGER.warning(
+            "LogFFT2Distance can only be used with data on 2D grids.",
+        )
+        self.x_dim = x_dim
+        self.y_dim = y_dim
+        LOGGER.warning("Please use x_dim and y_dim such that field_shape=(x_dim, y_dim).")
```
HCookie: Rather then raise a warning here, can you add a check and exception in the `calculate` function?

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,155 @@
...
+        self,
+        pred: torch.Tensor,
+        target: torch.Tensor,
+        squash: bool = True,
+        *,
+        scalar_indices: tuple[int, ...] | None = None,
+        without_scalars: list[str] | list[int] | None = None,
+    ) -> torch.Tensor:
+        result = super().forward(pred, target, squash, scalar_indices=scalar_indices, without_scalars=without_scalars)
+        return torch.sqrt(torch.mean(result))
```
HCookie, line 104: In the case of `squash==True`, I believe the dimensions will have already been reduced and averaged. Does the `sqrt(` belong in the `calculate_difference`?

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,155 @@
...
+        y_dim: int,
+        ignore_nans: bool = False,
+    ) -> None:
+        super().__init__(ignore_nans)
+        LOGGER.warning(
+            "Fourier Correlation loss can only be used with data on 2D grids.",
+        )
+        self.x_dim = x_dim
+        self.y_dim = y_dim
+        LOGGER.warning("Please use x_dim and y_dim such that field_shape=(x_dim, y_dim).")
```
HCookie: Same note as above

training/src/anemoi/training/losses/spatial.py
OpheliaMiralles: no indeed

training/docs/modules/losses.rst
```diff
@@ -67,6 +67,25 @@ deterministic:
       _target_: anemoi.training.losses.kcrps.KernelCRPSLoss
       # loss function kwargs here
 
+************************
+ Spatial Loss Functions
+************************
```
mchantry, line 72: Could you please specify that this only works for 2D regular grids please?

mchantry: Really nice to see documentation

OpheliaMiralles: Done!

mchantry: Sorry to be a pedant, but I think the word regular is really important, all the grids are 2D (horizontal) but being able to write it as m x n is the key component, which I think should be described as "regular".
Maybe you could even say 
** to be used only with regular 2D fields (i.e. fields that can be written as `[n_lat, n_lon]`)**

What do you think?

OpheliaMiralles: Let's hope it is good now!

training/src/anemoi/training/losses/loss.py
```diff
@@ -68,6 +68,10 @@ def get_loss_function(
     """
     loss_config = OmegaConf.to_container(config, resolve=True)
     scalers_to_include = loss_config.pop("scalers", [])
+    # related to spatial losses
+    if config.get("x_dim", None) is not None and config.get("y_dim", None) is not None:
+        kwargs["x_dim"] = config.x_dim
+        kwargs["y_dim"] = config.y_dim
```
HCookie: Why can't the `x` and `y` dims be provided in the config of the loss function rather than then in the top level config?

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,160 @@
...
+def get_spectra(
+    predicted_output: torch.Tensor,
+    real_output: torch.Tensor,
+    dims: tuple[int, int],
+) -> tuple[torch.Tensor, torch.Tensor]:
+    assert dims[0] * dims[1] == real_output.shape[2], (
+        "The product of dims must match the spatial dims of the output."
+        "Please use x_dim and y_dim such that field_shape=(x_dim, y_dim)."
+    )
+    dims_total = (*real_output.shape[:2], *dims, real_output.shape[-1])
```
HCookie: Please use the `TensorDim` Enum when referencing axis points in the tensors

training/src/anemoi/training/losses/loss.py
```diff
@@ -69,7 +69,7 @@ def get_loss_function(
     loss_config = OmegaConf.to_container(config, resolve=True)
     scalers_to_include = loss_config.pop("scalers", [])
     # related to spatial losses
-    if config.get("x_dim", None) is not None and config.get("y_dim", None) is not None:
+    if loss_config.get("x_dim", None) is not None and loss_config.get("y_dim", None) is not None:
         kwargs["x_dim"] = config.x_dim
         kwargs["y_dim"] = config.y_dim
```
HCookie: If they are in the loss config, why pull them out into the kwargs, they are used to instantiate?

OpheliaMiralles: then I guess we don't need it at all indeed, could you please remove those lines? not working this afternoon but would love to see this merged

HCookie: I'm unable to test this, but I have removed it, so once it's tested we can merge

training/src/anemoi/training/losses/spatial.py
```diff
@@ -0,0 +1,163 @@
...
+import torch
+import torch.fft
+
+from anemoi.training.losses.base import FunctionalLoss
+from anemoi.training.utils.enums import TensorDim
+
+LOGGER = logging.getLogger(__name__)
+
+
+def amplitude(spectrum: torch.Tensor) -> torch.Tensor:
```
floriankrb, line 24: As I understand, all functions in here are private ? And only the classes inheriting from `FunctionalLoss` are public?
If so, they could be renamed `amplitude` -> `_amplitude` for instance.

PR #151, opened 2025-02-21
test: initial integration tests

theissenhelen:
This is a first set of integration tests in training. The idea is to use the basic config template and only modify dataset-specific and test-specific config entries; and to then run a short training cycle. This tests both if config templates are still consistent with the code, and if a basic training cycle runs through.

Todo:
- [x] confirm that pytests run on atos
- [x] confirm that pytests run via ci action
- [x] add documentation on integration tests -- added first draft, need to review
- [x] add documentation/template for settting up member state use case tests -- added first draft, need to review
- [x] merge [the reusable workflow](https://github.com/ecmwf/reusable-workflows/pull/5) and point to workflow in main
- [x] remove the push triggers in the action before merging
- [x] ask in anemoi dev if there are pending unit tests edits (where moving the unit tests in this PR would cause downstream work for people) before merging 

Note that this currently runs on cpu on a github runner. The action is the first component we are planning to swap out (see below).

Next steps (to follow soon, but not included in this PR):
- move this to hpc runners, i.e. update action and the corresponding reusable workflow
- link nightly action to release-please PR if possible
- add unit tests that cover validation of config templates+modifications
- add tests for the other config templates (so that we test all main use cases)
- other tests tbd (e.g. can we overfit on small dataset)

----
📚 Documentation preview 📚: https://anemoi-training--151.org.readthedocs.build/en/151/

----
📚 Documentation preview 📚: https://anemoi-graphs--151.org.readthedocs.build/en/151/

----
📚 Documentation preview 📚: https://anemoi-models--151.org.readthedocs.build/en/151/


training/docs/dev/testing.rst
```diff
@@ -120,14 +123,88 @@ To run tests in a specific file:
...
 .. code:: bash
 
    pytest -m slow
 
+For integration tests, ensure that you have GPU available, then from the
+top-level directory of anemoi-core run:
+
+.. code:: bash
+
+   pytest training/tests/integration --slowtest
```
gareth-j: Should this be `--longtests`?

training/tests/integration/test_training_cycle.yaml
```diff
@@ -0,0 +1,42 @@
...
+      tracking_uri: null # override '???'
+
+hardware:
+  accelerator: auto
+  num_gpus_per_node: 1
+  num_nodes: 1
+  num_gpus_per_model: 1
+
+  paths:
+    output: ${oc.env:PWD}/tmp_output/
```
gareth-j: Could you use something like `PYTEST_DEBUG_TEMPROOT` here? That would ensure the directory is cleaned up.

training/tests/integration/test_training_cycle.py
```diff
@@ -0,0 +1,48 @@
...
+
+longtests = pytest.mark.skipif("not config.getoption('longtests')", reason="need --longtests option to run")
+
+LOGGER = logging.getLogger(__name__)
+
+
+@longtests
+def test_training_cycle_architecture_configs(architecture_config: DictConfig) -> None:
+    AnemoiTrainer(architecture_config).train()
+    shutil.rmtree(architecture_config.hardware.paths.output)
```
gareth-j, line 29: This isn't run if I kill the tests, could the pytest tempdir handling be used instead?

training/docs/dev/testing.rst
VeraChristina: Yes, absolutely. Good catch!

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  integration_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ecmwf/reusable-workflows/nightly-pytest@add-integration-test-anemoi
+        with:
+          repository: ${{github.repository}}
+          ref: ${{github.ref}}
+          python-version: '3.10'
+          install-package: "-e ./training[all,tests] -e ./graphs[all,tests] -e ./models[all,tests]"
```
anaprietonem, line 16: this install training graphs and models from the branch the worflow is triggered or from main?

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -258,7 +258,7 @@ class MlflowSchema(BaseModel):
     log_model: Union[bool, Literal["all"]]
     "Log checkpoints created by ModelCheckpoint as MLFlow artifacts. \
             If True, checkpoints are logged at the end of training. If 'all', checkpoints are logged during training."
-    tracking_uri: Union[str, None]
+    tracking_uri: Union[str, None] = Field(default=None)
```
anaprietonem: Are those changes required for the integration test work, or just other fixes that could go into a separate PR?

training/docs/dev/testing.rst
```diff
@@ -120,14 +123,88 @@ To run tests in a specific file:
...
+=================================
+
+For an example, see `training/tests/integration/test_training_cycle.py`.
+The test uses a configuration based on the template
+`training/src/anemoi/training/config/basic.py`, i.e. the basic global
+model. It applies test-specific modifications to reduce batch_size etc.
+as detailed in `training/tests/integration/test_training_cycle.yaml`. It
+furthermore applies dataset-specific modifications as detailed in
+`training/tests/integration/test_basic.yaml` to provide the location of
+our testing dataset compatible with the global model.
```
anaprietonem, line 180: Really clear and well explained!

training/docs/dev/testing.rst
```diff
@@ -120,14 +123,88 @@ To run tests in a specific file:
...
+
+2. Test-specific Modifications: Apply only the necessary
+dataset-specific and testing-specific modifications to the template. Use
+a config modification yaml, or hydra overrides for parametrization of a
+small number of config values.
+
+3. Reducing Compute Load: Where possible, reduce the number of batches,
+epochs, and batch sizes.
+
+4. Subfolder Organization: Place test configuration modification files
```
anaprietonem: I like this suggestion. Though I wonder if the current tests are related to global model use case if it would make sense to then have `training/tests/integration/global` and in the future we can expect to have `training/tests/integration/lam/`,`training/tests/integration/stretch`?

training/tests/integration/test_basic.yaml
```diff
@@ -0,0 +1,14 @@
+# Modifications for the basic template "config.yaml"
+hardware:
+  files:
+    dataset: aifs-ea-an-oper-0001-mars-o48-1979-19-6h-v6-testset.zarr
+    graph: test_graph.pt
+  paths:
+    data: https://object-store.os-api.cci1.ecmwf.int/ml-tests/test-data/samples/anemoi-integration-tests
+
+
```
anaprietonem: I understand this file is related to dataset specific, and hence it should be used by all the integration tests uses cases right? I find a bit counter intuitive the fact that the dataloader section is repeated across the test_training_cycle and test_basic. Would it make sense to just have it one place? or just even just have one single file?

training/tests/integration/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+from hydra import compose
+from hydra import initialize
+from omegaconf import OmegaConf
+
+
+@pytest.fixture(
+    params=[
+        ["model=gnn"],
+        ["model=graphtransformer"],
```
anaprietonem, line 22: is there any reason why the transformer is not included in this list?

training/tests/integration/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+    ],
```
anaprietonem, line 23: Related to this, what's this conftest for? is this used to permutate the options tested in the `test_training_cycle.py` file? Might be worth mentioning this in the docs?

training/tests/integration/test_basic.yaml
```diff
@@ -0,0 +1,14 @@
+# Modifications for the basic template "config.yaml"
+hardware:
+  files:
+    dataset: aifs-ea-an-oper-0001-mars-o48-1979-19-6h-v6-testset.zarr
+    graph: test_graph.pt
+  paths:
+    data: https://object-store.os-api.cci1.ecmwf.int/ml-tests/test-data/samples/anemoi-integration-tests
```
anaprietonem: will we be downloading the graph and the dataset everytime we run those tests ?

training/tests/integration/test_basic.yaml
```diff
@@ -0,0 +1,14 @@
+# Modifications for the basic template "config.yaml"
+hardware:
+  files:
+    dataset: aifs-ea-an-oper-0001-mars-o48-1979-19-6h-v6-testset.zarr
+    graph: test_graph.pt
```
anaprietonem: Regarding use a predefined graph,  if we introduce changes in anemoi-graphs at config level those might be not captured in this file and hence it can be prone to the test failing. Could be better to avoid passing the graph so then we just trigger the build on thy fly graph to also check that as part of the global model integration test?

@JPXKQX would be good to get your thoughts here

anaprietonem (commented):
Thanks @VeraChristina  and @theissenhelen for the work here! I left some comments regarding the folder structure and the use of a preloaded graph, let me know know if something is not clear!

.github/workflows/integration-tests.yml
VeraChristina: For nightly tests, from main. For workflow dispatch, from the branch that it is triggered (that's the information passed on in "ref")

VeraChristina: Though I haven't tested the workflow dispatch because I think the workflow needs to be in main for it to appear

training/src/anemoi/training/schemas/diagnostics.py
VeraChristina: This was me trying to fix this without overwriting it in the test config. Have taken this out.

I do think we should not require the corresponding config entries to be set when logging is disabled, will create an issue for this.

training/tests/integration/test_basic.yaml
VeraChristina: This is the file specific to the use case, i.e. with the path to the dataset used for the global model. (It's a bit confusing because the template for the global model is called "basic.yaml" not "global.yaml"). So this is the file that changes per use case, whereas test_training_cycle.yaml is used for all use cases

VeraChristina: The reason that dataloader appears in both is that we have some test-specific modifications for the dataloader across all tests, and some use-case-specific modifications. Maybe I should call it use-case-specific modifications instead of dataset-specific modifications? Would that be more clear?

training/tests/integration/conftest.py
VeraChristina: Yes, it's not debugged yet, we hit a cuda memory error, and a different error on cpu. So we decided to add it in the next steps

training/tests/integration/test_basic.yaml
VeraChristina: Just the dataset, and yes, we download the parts that we need into memory -- we don't want this to be platform dependant (such that everyone can run the tests) so we can't use datasets on atos

training/tests/integration/test_basic.yaml
VeraChristina: Good spot! I'll take that line out as it's confusing. We don't actually use a predefined graph but create the graph on the fly -- for exactly your reason

training/tests/integration/test_basic.yaml
VeraChristina: The reason we separated into two files is that we want to have the same test-specific modifications for all use cases, and it's easier to maintain if we can change it in one file and don't need to change it in five files. -- This is not very visible in the current state since there is just one use case. But the idea is to add a test_lam.yml and a test_stretched.yml and to use the test_training_cycle.yml for all test_training_cycle tests. Does this make sense?

VeraChristina: Based on this question I have changed the wording around modifications. We want to have one set of _testing-specific_ modifications that we apply to all training cycle tests -- in one file, such that we only need to change these in one place. And then for each use case, a file with _use-case-specific_ modifications (e.g. the dataset will differ for global vs lam vs stretched). I hope this makes it a bit more clear?
It is not super clear from the current structure (and because the global config file is called "basic.yml" not "global.yml"), but the idea would be to add only a "test_lam.yml" and a "test_stretched.yml" for those and to re-use the "test_training_cycle.yml"

training/docs/dev/testing.rst
VeraChristina: Fair point. I would keep them together for the time being as they share some config files and some conftest functionality (and they are the basic use cases, not super specific set-ups), but that's a weak preference. @theissenhelen ?

training/tests/integration/test_training_cycle.py
VeraChristina: Very good point! I'll look into this later today

training/docs/dev/testing.rst
VeraChristina: Have discussed with Helen. We decided to move all configs to a config folder for better overview but to not separate out the basic use cases (global/stretched/lam) since they share a lot of functionality. Shall we take this up again when we add the stretched and lam use cases?

training/tests/integration/test_training_cycle.py
VeraChristina: Yes, handling this via pytest's tmp_path now. Thanks for the idea!

training/tests/integration/test_training_cycle.yaml
VeraChristina: We now overwrite this in the test, just need this for hydra, will add a comment

training/docs/dev/testing.rst
anaprietonem: That's all good, i was just getting confused because of the docs 
`Place test configuration modification files in a relevant subfolder within `training/tests/integration/` for clarity
and ease of maintenance.`

.github/workflows/integration-tests.yml
anaprietonem: Thanks Vera, and just to confirm so the workflow has been tested locally so far? and a related question that I appreciate might be broader than this PR. But have you discussed the need of having specific repos to test all of this before merging? This can be something to pick up on an anemoi meeting

training/docs/dev/testing.rst
VeraChristina: Oh, got it, I missed that! I'll remove it there and only keep it for the member state use cases (which are slightly different)

training/tests/integration/conftest.py
anaprietonem: Out of interest, for running on GPU how are we planning to handle the installation of flash attention since it's not included in the pyproject.toml? And that's okey I'd just appreciate if that can be captured in the list of next steps of a github ticket/issue we can make everyone aware

.github/workflows/integration-tests.yml
VeraChristina: Well, we've tested on this PR (that's why the push trigger is still in the file, which should be removed before merging) and this works, you can see the integration tests passing in the checks at the bottom of this PR. But this does not test the scheduled version (which differs slightly from the dispatch version). How to test this more thoroughly might be a question to discuss with the ci group. In this case, the worst case is that the first nightly run fails and then we need to fix it

training/tests/integration/conftest.py
VeraChristina: To run on GPU, we need to set up different runners and, at least in case of hpc runners, a new workflow. The workflow could handle the installation of flash attention separately (after installing the rest). Our current configurations don't use flash attention, but when we add a config with flash attention, we need to think about how to handle this

VeraChristina: Will include in the relevant ticket

.github/workflows/integration-tests.yml
anaprietonem: Great all clear now thanks Vera!

training/tests/integration/test_basic.yaml
anaprietonem: Yes I understand this better now. Minor suggestion but since test_basic.yaml refers to the global_single model (ie, deterministic) I would use a more explicit naming for such case. If you plan to address when adding test_lam and test_stretched all good, but since we are also about to to add probabilistic models in Anemoi, I think it would be easier to understand

VeraChristina: Oh, I tried to mirror the names of the config folder, where the config file for the global model is "basic.yml" (which I also found confusing) -- The plan is to have a test for each config template (i.e. basic, lam, stretched, hierarchical), and I thought the easiest way is to use the names of the corresponding config files and prepend with test_

VeraChristina: Hm, not sure if this will be more or less clear when we add the other cases

anaprietonem: I see, and it's a fair point! so let's revisit that when adding the other uses cases. Thanks Vera for addressing my comments I am happy to approve the PR!

anaprietonem (dismissed):

anaprietonem (approved):

PR #153, opened 2025-02-25
feat(models,training)!: temporal interpolation

Magnus-SI:
## Description

Adds temporal interpolation functionality to anemoi with the following changes:
- Added option to specify explicit input and target times. For a forecaster this would be [0,1] and [2] for a typical multistep 2 and rollout 1 case. For an interpolator this could be [0,6] and [1,2,3,4,5].
- Changed valid dates to instead consider explicit indices (if rollout and multistep is specified in the config, these are translated). This also now allows for gaps in missing dates to be "jumped" if requested inidces can cross the gap. E.g. with a missing date at index 8, and requested dates at 0,3,6. Indices 3,4 and 6,7 are still usable, which would not have been represented correctly before.
- Added functionality to avoid crossing model runs when training on non-analysis data.
- Added "interpolator.py" in anemoi-training which loops through all target times for a given set of input times and combines the loss. Note that this increases the effective batch size by the number of target times, and that the learning rate should be increased accordingly (Not yet done by default).
- Added option to add forcing parameters as additional input to the target time.
- Added "interpolator.py" in anemoi-models that can use the additional target forcings. It also includes an option to specify whether to perform latent and grid skip connections, with an optional index for the latter. As a more general version of encoder_processor_decoder.py, merging the two could be considered.

## Type of Change

-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Motivation

A forecaster with longer time steps might yield better predictions going days out than a 1 hour forecaster, as it has to make fewer auto-regressive steps. To produce the hourly predictions still, we can use the information available from a longer time-stepping forecaster, e.g. hours 12 and 18 as input to predict hours 13-17. These predictions can be made individually, assisted by some information about the target time as input.

Preliminary results indicate that this approach works well when combining a 6 hour forecaster and hourly interpolator during inference. Additionally, when hourly analysis is not available, there does not seem to be a major loss in training the interpolator as an emulator on nwp model runs, as long as it is combined with a forecasting model trained on analysis during inference.

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

Although the interpolator needs functionality to explicitly state time indices, I also introduced this concept to the forecaster and dataset/datamodule.py. Partially because creating duplicates of both these classes would not be ideal, but also because it could be useful for the forecaster by enabling new potential setups. E.g. using hours 0,5,6 to make a forecast at hour 7, or training a combined forecaster/interpolator.
As of now, however, anemoi datasets does not support lists, arrays or tuples of indices, and slices must be used. Since these do not support irregular intervals, the explicit indices are somewhat unnecessary for the forecaster until support is added.


models/src/anemoi/models/models/interpolator.py
```diff
@@ -0,0 +1,140 @@
...
+        config : DotDict
+            Job configuration
+        data_indices : dict
+            Data indices
+        graph_data : HeteroData
+            Graph definition
+        """
+        self.num_target_forcings = len(model_config.training.target_forcing.data) + 1
+        self.input_times = len(model_config.training.explicit_times.input)
+        super().__init__(
```
HCookie: The super call should be done first.

models/src/anemoi/models/models/interpolator.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+
+        self.latent_skip = latent_skip
+        self.grid_skip = grid_skip
+
+    def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
+        self.num_input_channels = len(data_indices.internal_model.input)
+        self.num_output_channels = len(data_indices.internal_model.output)
+        self._internal_input_idx = data_indices.internal_model.input.prognostic
+        self._internal_output_idx = data_indices.internal_model.output.prognostic
```
HCookie: As these are the same definitions as in the super class, call `super()._calculate_shapes_and_indices()` and override `input_dim`

training/src/anemoi/training/data/datamodule.py
```diff
@@ -86,6 +80,48 @@ def supporting_arrays(self) -> dict:
     def data_indices(self) -> IndexCollection:
         return IndexCollection(convert_to_omegaconf(self.config), self.ds_train.name_to_index)
 
+    def relative_date_indices(self, val_rollout: int = 1) -> list:
```
HCookie, line 83: This is not neccesarily the validation rollout, more an override for the rollout value.

training/src/anemoi/training/data/dataset.py
```diff
@@ -39,9 +39,7 @@ def __init__(
         self,
         data_reader: Callable,
         grid_indices: type[BaseGridIndices],
-        rollout: int = 1,
-        multistep: int = 1,
-        timeincrement: int = 1,
+        relative_date_indices: list,
         shuffle: bool = True,
```
HCookie, line 43: It's worth pointing out this change is in direct conflict with the rollout scheduler, so we will need to think about how to address that.

training/src/anemoi/training/train/interpolator.py
```diff
@@ -0,0 +1,135 @@
...
+            graph_data=graph_data,
+            statistics=statistics,
+            data_indices=data_indices,
+            metadata=metadata,
+            supporting_arrays=supporting_arrays,
+        )
+        self.target_forcing_indices = itemgetter(*config.training.target_forcing.data)(
+            data_indices.data.input.name_to_index,
+        )
+        if type(self.target_forcing_indices) is int:
```
HCookie: Why is this not an isinstance? That is the standard approach of the rest of the code.

training/src/anemoi/training/utils/usable_indices.py
```diff
@@ -14,46 +14,41 @@
...
-    next_invalid_dates = rollout * timeincrement
+    usable_indices = np.arange(series_length - max(relative_indices))
 
-    usable_indices = np.arange(series_length)  # set of all indices
-
-    if missing_indices is None:
-        missing_indices = set()
-
-    missing_indices |= {-1, series_length}  # to filter initial and final indices
+    # Avoid crossing model runs by selecting only relative indices with the same model run id
```
HCookie, line 43: By model runs do you mean multiple datasets each from a different source model?

HCookie (changes requested):
Left some comments, if you could provide an example config, or add one to the provided configs that would be helpful in my testing. 
Additonally, this PR will need some work to fit with the schemas.

Magnus-SI:
> Left some comments, if you could provide an example config, or add one to the provided configs that would be helpful in my testing. Additonally, this PR will need some work to fit with the schemas.

Thanks, I have added an example config under training now, and will look at your comments and do some general clean up of the code tomorrow.

training/src/anemoi/training/utils/usable_indices.py
mchantry: The training is on forecast trajectories, so you want to ensure you are on the same trajectory for the time sequence.

models/src/anemoi/models/models/interpolator.py
Magnus-SI: I did it this way because the encoder_processor_decoder super calls on _calculate_shapes_and_indices(data_indices), with only data_indices as an argument.
But the interpolator requires the extra information for the number of target forcings, hence I have to define these variables as objects in self before calling super if I want to inherit the whole init function.
However, I'd be happy to change it if there's an alternate solution.

training/src/anemoi/training/train/train.py
```diff
@@ -165,12 +165,10 @@ def model(self) -> pl.LightningModule:
             "supporting_arrays": self.supporting_arrays,
         }
 
-        train_module = importlib.import_module(
-            getattr(self.config.training, "train_module", "anemoi.training.train.forecaster"),
-        )
-        train_func = getattr(train_module, getattr(self.config.training, "train_function", "GraphForecaster"))
-        model = train_func(**kwargs)
-        # NOTE: This should be replaced with instantiate
+        module_name, class_name = self.config.training.task.rsplit(".", 1)
```
mchantry: Can this be replaced with a `get_class` 
see https://github.com/ecmwf/anemoi-core/blob/d3bf2ae37bb38d5e9b85626a295eaa0b7b8b6f3a/training/src/anemoi/training/train/train.py#L185

Magnus-SI: Thanks, that's cleaner, I've replaced it now.

training/src/anemoi/training/train/interpolator.py
Magnus-SI: My mistake, replaced it now.

anaprietonem:
Just to flag, that this PR will need some docs update. The APIs will be updated automatically with sphinx, but it would needed some info to explain the new model interface and some of the major changes. Thank you!

mchantry:
Due to clashes with ongoing doc-a-thon, I think we will put the documentation in later.

mchantry:
@Magnus-SI are you ready for to switch this away from draft?

mchantry:
@Magnus-SI please can you update the default training config to use
```
training:
  task: anemoi.training.train.forecaster.GraphForecaster
  ```
  so that the code will work without pydantic validation checking.

File: training/src/anemoi/training/schemas/training.py [hunk 8]
```diff
@@ -80,6 +80,25 @@ class LR(BaseModel):
...
+    target: list[NonNegativeInt] = Field(default=[2])
+
+
+class TargetForcing(BaseModel):
+    """Forcing parameters for target output times.
+
+    Extra forcing parameters to use as input to distinguish between different target times.
+    """
+
+    data: list[str] = Field(default="insolation")
```
HCookie: Can you provide some insight into the use cases of this? and how the default values are meaningful for the forecasting task?

training/src/anemoi/training/train/interpolator.py
```diff
@@ -0,0 +1,135 @@
...
+            device=self.device,
+            dtype=batch.dtype,
+        )
+        for interp_step in self.interp_times:
+            # get the forcing information for the target interpolation time:
+            target_forcing[..., : len(tfi)] = batch[:, self.imap[interp_step], :, :, tfi]
+            target_forcing[..., -1] = (interp_step - self.boundary_times[1]) / (
+                self.boundary_times[1] - self.boundary_times[0]
+            )
+            # TODO(Magnus-SI): make fraction time one of a config given set of arbitrary custom forcing functions.
```
mchantry: @Magnus-SI is this a to-do before merging?

training/src/anemoi/training/train/interpolator.py
```diff
@@ -0,0 +1,135 @@
...
+        tfi = self.target_forcing_indices
+        target_forcing = torch.empty(
+            batch.shape[0],
+            batch.shape[2],
+            batch.shape[3],
+            len(tfi) + 1,
+            device=self.device,
+            dtype=batch.dtype,
+        )
+        for interp_step in self.interp_times:
```
mchantry, line 114: @Magnus-SI have you tested if doing all steps each time is more efficient than just random sampling a interp_step for each element in a batch?

training/src/anemoi/training/schemas/training.py
Magnus-SI: For the interpolator, as the same model is used to predict different target times separately, it needs additional information to distinguish between them.
Although I've added a fractional time between the boundary input times as a default input across the whole grid, I initially thought this might not be enough information. Hence, I added the option for additional forcings containing temporal information for the target time, e.g. knowing where the sun is at for the time you are predicting. However, early testing seemed to indicate that this didn't have much impact on the skill, and that the fractional time might be enough.

Still, I would expect the temporal forcings to help generalize the concept of continuous time within in the model, instead of it just memorizing the distinct fractions (e.g. 0.167, 0.333, etc.). So if you want to apply it to a finer temporal resolution than you have in the training data, a model trained with more target temporal forcings might perform better. I have not tested this yet, but I still feel it is worth leaving the option in.

For the basic forecasting task, I would not expect these temporal forcings to contribute much, as it always makes predictions at the same exact time. Still, if someone wants to make a forecaster that can forecast at multiple different times, they can just use the interpolator task for that objective.

training/src/anemoi/training/train/interpolator.py
Magnus-SI: I have not tested this, but I have given it some thought.
If you want to go through all combinations of input and target times within the dataset, predicting on all targets once an input is loaded in should be slightly more effective, only considering data loading, which is why I chose to do it this way.
However, this might not matter much, and a big downside is that your effective batch size becomes larger, leaving less room for data parallel across multiple processes if you want to limit your total batch size, resulting in a longer wait in real time to train your model (but similar cost).
A potential upside to averaging over all target times for each set of inputs is that the target times become more consistent in time. Since all target times have to match simultaneously, they have to follow the exact weather flow in the training sample for minimal loss. This is just speculation for now, but something I want to test at some point.

A random choice of the interp_step can be implemented pretty easily in place of the for loop. But it would be suboptimal if the dataloader is a bottleneck, since all target times are loaded unnecessarily. Having that change as well would probably take a bit longer to implement, and might lead to more changes in forecaster related code.

With that in mind, I or someone else could easily try the suboptimal implementation and see how it impacts the model, and if useful, the optimal implementation can be added in a later PR. Does that sound reasonable?

training/src/anemoi/training/train/interpolator.py
Magnus-SI: I've been on the fence on whether this is necessary, or if the fractional time can just be used as a default. But I will try to quickly implement it now.

training/src/anemoi/training/train/interpolator.py
mchantry: Should have clarified, not requesting changes, just interested in the science. We can look into this later.

mchantry:
@Magnus-SI note that Harrison pushed some changes to split the schemas for forecasting and interpolation. Can you check you are happy with this proposal?

training/src/anemoi/training/train/interpolator.py
Magnus-SI: For the purpose of this PR I now made fractional time an optional forcing. If needed, other custom forcings can be implemented later.

Magnus-SI:
> @Magnus-SI note that Harrison pushed some changes to split the schemas for forecasting and interpolation. Can you check you are happy with this proposal?

I was not familiar with pydantic before this, so I am not 100% sure on what the changes do. Having tested a bit, it seems that by adding either explicit_times or target_forcing as arguments to the config, it uses the InterpolationSchema. And since it expects both, they both have to be included for it to run.
This means that if one wants to use the forecaster task with specified explicit indices, an unused config option of target_forcing need to be included. Using a forecaster with explicit indices could be useful for e.g. making an hourly forecaster that uses hours 0,5,6, to forecast hour 7. But it might be fine this way, as it is still possible by just specifying an empty list of target forcings.

training/src/anemoi/training/data/dataset.py
```diff
@@ -276,18 +269,22 @@ def __iter__(self) -> torch.Tensor:
         )
 
         for i in shuffled_chunk_indices:
-            start = i - (self.multi_step - 1) * self.timeincrement
-            end = i + (self.rollout + 1) * self.timeincrement
+            start = i + self.relative_date_indices[0]
+            end = i + self.relative_date_indices[-1] + 1
+            timeincrement = self.relative_date_indices[1] - self.relative_date_indices[0]
+            # NOTE: this is temporary until anemoi datasets allows indexing with arrays or lists
+            # data[start...] will be replaced with data[self.relative_date_indices + i]
```
Magnus-SI, line 277: > @Magnus-SI are you ready for to switch this away from draft?

I am not sure what explicitly follows from switching this away from draft, but as far as I see, there are currently two remaining issues.

One relates to the referenced code, wherein the explicit indices cannot be used exactly as intended until support is added in anemoi datasets. This limits the use of explicit indices to extracting data at regular intervals, e.g.
- input: [0,6], target: [1,2,3,4,5]
- input: [0,6], target: [0,3,6]
- input [0,1], target: [2]
are possible, but
- input: [0,5,6], target: [7]
- input: [0,6], target: [1,2,3,4,5,12]
are not possible.

This does not affect the base interpolator use case, so I personally don't think it's needed for this PR. It can be easily added at a later time with a small commit once functionality for explicit time indices is added to anemoi datasets. What do you think @mchantry ?


The second issue is that the explicit indices need to be merged with the rollout scheduler functionality that @HCookie added. I am not sure in what order you want this to happen, but I think I might have a decent way of doing it. It should be possible to include the rollout scheduler as an argument in dataset.py as in that branch, have the explicit indices list the indices to be extracted at full rollout, then use the rollout value updated by the rollout scheduler to exclude max_rollout - current_rollout indices from the end of self.relative_date_indices

training/src/anemoi/training/schemas/training.py
```diff
@@ -269,3 +293,22 @@ class TrainingSchema(BaseModel):
...
+    "Training objective."
+
+
+class ForecasterSchema(BaseTrainingSchema):
+    task: str = Field(default="anemoi.training.train.forecaster.GraphForecaster")
+    "Training objective."
+
+
+class InterpolationSchema(BaseTrainingSchema):
+    task: str = Field(default="anemoi.training.train.interpolator.GraphInterpolator")
```
anaprietonem: Just a comment in regards of using 'default' in Pydantic that this have some implications when using the `no_validation` flag. Discussing with @theissenhelen, this can lead to a different behaviour. So our proposal would be to rather use just an example in pydantic and provide it explicitly in our configs. If for this PR you don't want to introduce backward compatibility at config level,  I believe Helen had implemented a way to deal with this in this PR in case it helps! (Helen pls free to add to this)

https://github.com/ecmwf/anemoi-core/blob/feature/trmapper/training/src/anemoi/training/schemas/models/processor.py#L80

Magnus-SI: I already added the forecaster as a default task in the training config, so I can just change default to example. I've also added some other arguments that I can change to example. I'm not that familiar with pydantic, but I assume that default is set as an argument even when not specified in the config. What does setting an example do exactly?

Magnus-SI: This should be changed in the most recent commit now.

training/src/anemoi/training/data/dataset.py
HCookie: > rollout schedulers

It may also be possible to update the `relative_date_indices` in the dataset, like was done over in the other PR with the rollout value.

HCookie:
> Training a forecaster with explicit indices only could be useful

I see, this was an edge case I missed, maybe it makes sense to expand the prediction specification parts, such that it can be either a rollout scheduler, or a list of explicit indices?

training/src/anemoi/training/data/dataset.py
Magnus-SI: Yes, though if I understand correctly, that would require adding the relative_date_indices object to the schedulers themselves, or a similar object, which might involve a lot more additional code.

Magnus-SI: If you re-revert the rollout scheduler PR, I'd be open to try to merge it here and make something work.

mchantry: I think these issues should be addressed later.
The rollout scheduler and future work on datasets can look to address this.

training/src/anemoi/training/schemas/training.py
anaprietonem: So using `defaults` with Pydantic means, that when `no_validation:False`  then if that entry is not passed in the config, Pydantic will pick the default. The problem here is that when `no_validation:True` then Pydantic is deactivated and hence the default doesn't work, so if there is not set in the code to hardcore that value the user needs to pass that in the config and you can then have differences in configs when running anemoi-training if you activate or activate the validation flag. 

To avoid this, we either would like:
1. Raise the PR as a breaking change PR so that from now on - users will need to pass the task in their configs. (as the PR it's right now - this is the case)
2. You can use a `check_validation_extras` as Helen implemented [here](https://github.com/ecmwf/anemoi-core/blob/feature/trmapper/training/src/anemoi/training/schemas/models/processor.py#L80) that helps to bypass the need to have a config breaking change, and avoid differences in execution with validation on/off.

Example is just a friendlier way to tell the user/developer what that Field value can look like. 

I think since the kcrps PR is also extending the graphforecaster interfaces, we could keep option 1 and simply inform that this PR has a breaking change because users will need to define a task on their configs

Magnus-SI:
> > Training a forecaster with explicit indices only could be useful
> 
> I see, this was an edge case I missed, maybe it makes sense to expand the prediction specification parts, such that it can be either a rollout scheduler, or a list of explicit indices?

I think having the rollout scheduler functionality even when specifying explicit indices might be useful in e.g. the 0,5,6->7 case. For an hourly forecaster in particular, doing a significant amount of training at rollout should be important, and rollout being dynamically adjustable during training would make that a lot easier.

How about having explicit times still be an option for the forecaster, but defaulting to None? See the most recent commit for a suggested change.

training/src/anemoi/training/schemas/training.py
Magnus-SI: Okay, I think we should go with option 1 then. I did add the task to the default training config, but I guess it would still be breaking for users who use a custom config under the training folder. I'll go ahead and add the breaking change tag to this PR.

mchantry:
@Magnus-SI please can you update the branch, there will need to be some small changes to the new aicon integration test, we can help with these if you would like.

Magnus-SI:
> @Magnus-SI please can you update the branch, there will need to be some small changes to the new aicon integration test, we can help with these if you would like.

I assume config changes? If so, adding task: `anemoi.training.train.forecaster.GraphForecaster` and `explicit_times: null` to training should work. Can I run the integration tests to verify locally on my computer, or do I need GPU access?

Magnus-SI:
> > @Magnus-SI please can you update the branch, there will need to be some small changes to the new aicon integration test, we can help with these if you would like.
> 
> I assume config changes? If so, adding `task: anemoi.training.train.forecaster.GraphForecaster` and `explicit_times: null` to training should work. I can try to run the tests myself first.

I seem to be having trouble running these tests locally and on lumi. So I would appreciate some help verifying that these added config options work.

mchantry:
> > > @Magnus-SI please can you update the branch, there will need to be some small changes to the new aicon integration test, we can help with these if you would like.
> > 
> > 
> > I assume config changes? If so, adding `task: anemoi.training.train.forecaster.GraphForecaster` and `explicit_times: null` to training should work. I can try to run the tests myself first.
> 
> I seem to be having trouble running these tests locally (our local server with interactive GPU access is down for maintenance this week). So I would appreciate some help verifying that these added config options work. I've pushed the suggested fix.

I can look now.

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -124,6 +122,9 @@ def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
         self.num_output_channels = len(data_indices.internal_model.output)
         self._internal_input_idx = data_indices.internal_model.input.prognostic
         self._internal_output_idx = data_indices.internal_model.output.prognostic
+        self.input_dim = (
+            self.multi_step * self.num_input_channels + self.node_attributes.attr_ndims[self._graph_name_data]
```
HCookie, line 126: You are still referencing multi_step here, how does that work if a user set the explicit times? Would they need to align the config entries?

mchantry:
> > > > @Magnus-SI please can you update the branch, there will need to be some small changes to the new aicon integration test, we can help with these if you would like.
> > > 
> > > 
> > > I assume config changes? If so, adding `task: anemoi.training.train.forecaster.GraphForecaster` and `explicit_times: null` to training should work. I can try to run the tests myself first.
> > 
> > 
> > I seem to be having trouble running these tests locally (our local server with interactive GPU access is down for maintenance this week). So I would appreciate some help verifying that these added config options work. I've pushed the suggested fix.
> 
> I can look now.

@Magnus-SI the aicon test now passes.

mchantry:
> > > > @Magnus-SI please can you update the branch, there will need to be some small changes to the new aicon integration test, we can help with these if you would like.
> > > 
> > > 
> > > I assume config changes? If so, adding `task: anemoi.training.train.forecaster.GraphForecaster` and `explicit_times: null` to training should work. I can try to run the tests myself first.
> > 
> > 
> > I seem to be having trouble running these tests locally (our local server with interactive GPU access is down for maintenance this week). So I would appreciate some help verifying that these added config options work. I've pushed the suggested fix.
> 
> I can look now.

@Magnus-SI the integration tests all pass.

models/src/anemoi/models/models/encoder_processor_decoder.py
Magnus-SI: Good point, that is not ideal. It would be possible to use `len(explicit_times.input) if explicit_times is not None else multistep`. Or would that be too much clutter? It would need to be added to the `__init__` of both encoder_processor_decoder.py and hierarchical.py.

models/src/anemoi/models/models/interpolator.py
```diff
@@ -0,0 +1,138 @@
...
+        self.num_target_forcings = (
+            len(model_config.training.target_forcing.data) + model_config.training.target_forcing.time_fraction
+        )
+        self.input_times = len(model_config.training.explicit_times.input)
+        super().__init__(
+            model_config=model_config, data_indices=data_indices, statistics=statistics, graph_data=graph_data
+        )
+
+        self.latent_skip = model_config.model.latent_skip
+        self.grid_skip = model_config.model.grid_skip
```
Magnus-SI, line 58: Should these options be brought into the standard encoder_processor_decoder.py as well? They determine whether you want skip connections on the grid, and/or in latent space.

models/src/anemoi/models/models/encoder_processor_decoder.py
HCookie: Given the work needed, and with the knowledge of the rollout schedulers, I would propose removing explicit times from the forecaster, and looking at refactoring in a future PR.

models/src/anemoi/models/models/interpolator.py
HCookie: To simplify things, maybe in a future PR?

models/src/anemoi/models/models/encoder_processor_decoder.py
Magnus-SI: I will change it back then. Explicit indices would still be usable for forecasting anyway through the interpolator by setting no target forcings and a future target.

anaprietonem (approved):
I'm approving this PR, as the reviewers point have been addressed and code LGTM

HCookie (approved):

PR #163, opened 2025-02-28
feat!: generalise activation function

chebertpinard:
## Description

We want to make the activation function more flexible to allow adding those that are not in pytorch. In this situation the GLU derivated (SwiGLU, ReGLU, GEGLU) would be interesting to test. 

There is a pytorch implementation of the GLU function, but it splits the input in 2 parts and make the output half the size of the input. We then have issue with input size of the next layer. The suggested solution for that is to make the linear transformation on the complete initial matrix, but other options are possible and could replace that one.

```py
W = nn.Linear(dim, dim)
V = nn.Linear(dim, dim)

GLU = Sigmoid(W(input)) * V(input)
```

Equation: GLU(x, W, V, b, c) = σ(xW + b) ⊗ (xV + c)

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

There is no new dependencies

### Documentation

## Additional Notes

- Need to add tests and adjust the existing ones.
- Need to adjust type of the activation passed to the layers objects (not a string anymore).
- The \_partial\_ key should not be in the config file since it is not an option (should always be True). Maybe adding it to the DictConfig at the start of AnemoiTrainer. (something like `with open_dict(config): config.model.activation._partial_ = True`)
- Need to test on longer training to compare behaviour of the different activation function.

----
📚 Documentation preview 📚: https://anemoi-training--163.org.readthedocs.build/en/163/

----
📚 Documentation preview 📚: https://anemoi-graphs--163.org.readthedocs.build/en/163/

----
📚 Documentation preview 📚: https://anemoi-models--163.org.readthedocs.build/en/163/


JesperDramsch:
I've been asked by @anaprietonem to take a look at this and put the finishing touches on it.

JesperDramsch:
I started out with some profiling of different implementations. The profiling was done with a warmup of 5 epochs and 100 iterations on an input tensor of size `[128, 512, 2048]` (with a code snippet from @japols thank you!):

## GLU
```python
class GLU(nn.Module):
    def __init__(self, dim, variation=nn.Sigmoid()):
        super().__init__()
        self.W = nn.Linear(dim, dim)
        self.V = nn.Linear(dim, dim)
        self.variation = variation

    def forward(self, x):
        return self.variation(self.W(x)) * self.V(x)
```
Time: 182.58 ms ± 0.72 ms (min: 182.43 ms, max: 189.74 ms)
Memory: 4204.10 MB ± 42.00 MB (min: 3788.29 MB, max: 4208.30 MB)

## GLU2
Has the benefit that we don't have to supply the `dim` in the config. Makes it much easier to use.
```python
class GLU2(nn.Module):
    def __init__(self, variation=None):
        super().__init__()
        self.dim = None
        self.variation = variation if variation is not None else nn.Sigmoid()

    def forward(self, x):
        if self.dim is None:
            self.dim = x.shape[-1]
            self.W = nn.Linear(self.dim, self.dim).to(x.device)
            self.V = nn.Linear(self.dim, self.dim).to(x.device)
        return self.variation(self.W(x)) * self.V(x)
```
Time: 182.53 ms ± 0.03 ms (min: 182.42 ms, max: 182.56 ms)
Memory: 4268.13 MB ± 42.00 MB (min: 3852.32 MB, max: 4272.33 MB)

## FastGLU
This was suggested in a medium article as being "faster".
```python
class FastGLU(nn.Module):
    def __init__(self, dim, variation=nn.Sigmoid()):
        super().__init__()
        self.dim = dim
        self.projection = nn.Linear(dim, dim * 2)
        self.variation = variation

    def forward(self, x):
        out = self.projection(x)
        return self.variation(out[..., :self.dim]) * out[..., self.dim:]
```
Time: 186.45 ms ± 0.44 ms (min: 185.44 ms, max: 187.44 ms)
Memory: 4810.90 MB ± 54.40 MB (min: 4272.33 MB, max: 4816.34 MB)

## FastGLU2
And initialised on the fly.
```python
class FastGLU2(nn.Module):
    def __init__(self, variation=None):
        super().__init__()
        self.dim = None
        self.variation = variation if variation is not None else nn.Sigmoid()

    def forward(self, x):
        if self.dim is None:
            self.dim = x.shape[-1]
            self.projection = nn.Linear(self.dim, self.dim * 2).to(x.device)
        out = self.projection(x)
        return self.variation(out[..., :self.dim]) * out[..., self.dim:]
```
Time: 186.55 ms ± 0.46 ms (min: 185.26 ms, max: 187.71 ms)
Memory: 4874.94 MB ± 54.40 MB (min: 4336.36 MB, max: 4880.38 MB)

# Comparison
GLU is          0.0% slower than GLU2
GLU uses        1.5% less memory than GLU2
GLU is          2.1% faster than FastGLU
GLU uses        12.6% less memory than FastGLU
GLU is          2.1% faster than FastGLU2
GLU uses        13.8% less memory than FastGLU2
GLU2 is         2.1% faster than FastGLU
GLU2 uses       11.3% less memory than FastGLU
GLU2 is         2.2% faster than FastGLU2
GLU2 uses       12.4% less memory than FastGLU2
FastGLU is      0.1% faster than FastGLU2
FastGLU uses    1.3% less memory than FastGLU2

It is evident that the idea of "FastGLU" is both slower and more memory-intensive due to the projection parameters.

GLU2 seems to be slightly less efficient regarding memory, but might still be preferrable due to the improved usability.

JesperDramsch:
This re-work does a couple things:

## Changes proposed

1. Implement the GLU2 version to avoid having to declare dimensionality for the activation functions.
2. Move activation to layer_kernels as it's a similar usecase to switch out layers.
3. Removed activation as independent from configs, schema, function signature etc.
4. Adjust layer_kernel #47 to instantiate partial factories in the function itself (to avoid duplicate instantiates and remove the "non-optional option" from the config
5. Implemented tests for activation functions with hypothesis
6. Adjusted tests for layer_kernels
7. Adjusted processor, mapper, chunk, blocks and their tests.
8. Tidied up the function signatures in each
9. Cleaned up the docstring for each

## What needs to be done during review

In addition to checking for style and correctness of the implementation. I ran the code as `debug`, but I don't know if it slowed down in any significant way on a full run or with different model components. The memory footprint is comparable.

.gitignore
HCookie, line 1: If this is to be changed, I think it’ll need to be done in the sync files.

JesperDramsch: Happy to revert, that one accidentally slipped by

models/src/anemoi/models/layers/mapper.py
```diff
@@ -242,14 +251,13 @@ def __init__(
...
             edge_dim=self.edge_dim,
-            activation=activation,
             num_chunks=num_chunks,
             layer_kernels=layer_kernels,
         )
 
         self.offload_layers(cpu_offload)
 
-        self.emb_nodes_dst = Linear(self.in_channels_dst, self.hidden_dim)
+        self.emb_nodes_dst = Linear(self.in_channels_dst, self.hidden_dim, bias=False)
```
ssmmnn11: why set bias to False here?

models/src/anemoi/models/layers/activations.py
```diff
@@ -0,0 +1,87 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from typing import Optional
+
+import torch
+from torch import nn
+
+
+class GLU(nn.Module):
```
anaprietonem, line 47: So if someone would like to use torch version of GLU, we could still do torch.nn.GLU but for SwiGLU/ReGLU we load them via this custom implementation right?

training/src/anemoi/training/schemas/models/common_components.py
```diff
@@ -15,35 +15,53 @@
...
+    Softsign = "torch.nn.Softsign"
+
+    # Hard activation functions
+    Hardshrink = "torch.nn.Hardshrink"
+    Hardsigmoid = "torch.nn.Hardsigmoid"
+    Hardtanh = "torch.nn.Hardtanh"
+    Hardswish = "torch.nn.Hardswish"
+
+
+class ActivationFunctionSchema(BaseModel):
```
anaprietonem: Where is this `ActivationFunctionSchema` being used?

anaprietonem (commented):
I started looking at this, great work Jesper. Related to the regression we discussed, if someone would train the model with say GEGLU, and then freeze the components for encoder and decoder, would they then retrain the processor with a new activation function or that's not supported?

anaprietonem:
` I ran the code as debug, but I don't know if it slowed down in any significant way on a full run or with different model components.` 
Did you get to log any of this in mlflow?

models/src/anemoi/models/layers/activations.py
JesperDramsch: Yes

training/src/anemoi/training/schemas/models/common_components.py
JesperDramsch: Not sure, I just removed them from the TransformerModelComopnent. I'll look into it.

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -171,10 +171,8 @@ model:
     _target_: anemoi.models.models.encoder_processor_decoder.AnemoiModelEncProcDec
 
   num_channels: 5 # reduced for CI/CD test. original value 512
-  activation: GELU
```
mchantry, line 175: Where is the activation now specified for this test?

JesperDramsch: In the layer kernels

JesperDramsch:
Summary:

- Modified `layer_kernels` to be instantiated in the loading function to reduce `_partial_` usage and avoid user error
- Integrated activation function instantiation directly in the layer_kernels
- Implemented additional activation functions:
  - GLU (Gated Linear Unit) with customizable variations
  - SwiGLU (Swish GLU), ReGLU (ReLU GLU), GEGLU (GELU GLU)
  - Sine activation with parameterizable frequency and phase
- Added support to vary `layer_kernels` by model component
- Eliminated random default values that could introduce unpredictable behaviours
- Generalised testing methodology with hypothesis for better coverage
- Comprehensive test suite with various activation function properties
- Implemented layer_kernels for Transformer Mappers

## Implementation Details
- GLU variations now include support for both bounded and unbounded activation functions
- Unbounded activations (ReLU, GeLU, Swish) require a projection layer that expands input to 2x dimension to prevent numerical explosion
- Bounded activations (Sigmoid, Tanh, ReLU6) work well without the projection layer
- Performance testing shows similar values on MLFlow and evaluation metrics

## Regression Note
The previous regression to layer_kernels being global has now been solved and implemented to be per component.

## Review Considerations
- I've run the code in `debug` mode with a comparable memory footprint
- Further testing may be needed to verify there's no significant performance impact on full runs or with different model components, although I have applied

.gitignore
```diff
@@ -135,3 +135,6 @@ _version.py
 *.to_upload
 tempCodeRunnerFile.python
 Untitled-*.py
+.envrc.local
```
anaprietonem: Are those to be committed? Look like local changes

models/src/anemoi/models/layers/utils.py
```diff
@@ -31,20 +32,30 @@ def forward(self, *args, **kwargs):
...
-        kernel_config : DotDict
-            Kernel configuration, e.g. {"Linear": {"_target_": "torch.nn.Linear", "_partial_": True}}
+    This function tries to load the layer kernels from the config. If the layer kernel is not supplied, it will fall back to the torch.nn implementation.
 
-    Returns:
-        DotDict: hydra partial instantiation of the layer kernels
+    Parameters
+    ----------
+    kernel_config : DotDict
+        Kernel configuration, e.g. {"Linear": {"_target_": "torch.nn.Linear", "_partial_": True}}
```
anaprietonem: Does the `kernel_config` still need the '_partial_' or should we update this line?

models/src/anemoi/models/layers/chunk.py
```diff
@@ -189,18 +181,17 @@ def __init__(
...
         else:
             self.emb_edges = None
 
         self.build_blocks(
             GraphConvProcessorBlock,
-            num_channels,
-            num_channels,
+            in_channels=num_channels,
+            out_channels=num_channels,
+            num_chunks=1,
```
anaprietonem, line 192: where is this `num_chunks=1` coming from? (asking also since the default before seemed `num_chunks: int = 2`)

models/tests/layers/block/test_block_graphconv.py
```diff
@@ -63,7 +67,7 @@ class TestGraphConvMapperBlock:
         in_channels=st.integers(min_value=1, max_value=100),
         out_channels=st.integers(min_value=1, max_value=100),
         mlp_extra_layers=st.integers(min_value=1, max_value=5),
-        activation=st.sampled_from(["ReLU", "GELU", "Tanh"]),
+        activation=st.sampled_from(["torch.nn.ReLU", "torch.nn.GELU", "anemoi.models.layers.activations.GLU"]),
         update_src_nodes=st.booleans(),
```
anaprietonem, line 89: Since `SwiGLU` is now stable should we add it to the test to ensure it doesn't break in the future?

training/pyproject.toml
```diff
@@ -129,7 +129,7 @@ lint.select = [
   "C4",     # Comprehensions
   "C90",    # Mccabe complexity
   "COM",    # Commas
-  "CPY",    # Copyright
+  # "CPY",    # Copyright
```
anaprietonem: Is this a local change or was this giving you any errors?

training/src/anemoi/training/config/model/graphtransformer.yaml
```diff
@@ -60,13 +16,13 @@ processor:
...
-
+  # LayerKernels are optional. If not specified, the default are pytorch.nn.LayerNorm and torch.nn.Linear.
+  # TODO: Add to documentation: The query and key norms need to be autocast layer norms if flash attention is used.
 
 encoder:
   _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
   trainable_size: ${model.trainable_parameters.data2hidden}
   sub_graph_edge_attributes: ${model.attributes.edges}
-  activation: ${model.activation}
   num_chunks: 1
```
anaprietonem, line 38: Why in this case we don't have the layer kernels included? do we rely on the defaults?

training/src/anemoi/training/config/model/graphtransformer_ens.yaml
```diff
@@ -68,12 +23,26 @@ processor:
...
+      _target_: anemoi.models.layers.normalization.AutocastLayerNorm
+      bias: False
+    KeyNorm:
+      _target_: anemoi.models.layers.normalization.AutocastLayerNorm
+      bias: False
 encoder:
   _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
   trainable_size: ${model.trainable_parameters.data2hidden}
   sub_graph_edge_attributes: ${model.attributes.edges}
-  activation: ${model.activation}
```
anaprietonem, line 76: Similar to my previous question, wondering if we should state the layer kernel explicitly in encoder and decoder as done for the `gnn.yaml`

( and same question for the graphtransformer_ens and transformer files)

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
anaprietonem: If DWD has the layer kernel explicitly defined as GELU for their tests, wouldn't it be better to state clearly the layer kernel section with that activation as part of this config? (we can ping Marek if useful)

models/src/anemoi/models/layers/activations.py
```diff
@@ -39,3 +42,137 @@ def leaky_hardtanh(
...
+    def __init__(self, *, variation: Optional[nn.Module] = None, bias: bool = True, projection: bool = True):
+        super().__init__()
+        self.bias, self.projection = bias, projection
+        self.dim = self.W = self.V = None
+        self.variation = variation if variation is not None else nn.Sigmoid()
+
+    def _post_init(self, x: torch.Tensor) -> None:
+        """Initialize the weights and biases."""
+        self.dim = x.shape[-1]
+        if self.projection:
```
anaprietonem: Well done catching this and thanks for testing!

anaprietonem (commented):
Left some comments, great work!

models/src/anemoi/models/layers/processor.py
```diff
@@ -88,74 +91,74 @@ class TransformerProcessor(BaseProcessor):
...
-        mlp_hidden_ratio: int = 4,
+        num_channels: int,
+        num_chunks: int,
+        num_heads: int,
+        mlp_hidden_ratio: int,
         qk_norm=False,
         dropout_p: float = 0.0,
         attention_implementation: str = "flash_attention",
-        softcap: float = 0.0,
+        softcap: float = 0,
```
ssmmnn11, line 103: should be softcap: float = 0., to initialise as float not int?

models/src/anemoi/models/layers/processor.py
```diff
@@ -88,74 +91,74 @@ class TransformerProcessor(BaseProcessor):
...
-            ratio of mlp hidden dimension to embedding dimension, default 4
+            Ratio of mlp hidden dimension to embedding dimension
         qk_norm: bool, optional
             Normalize query and key, by default False
-        activation : str, optional
-            Activation function, by default "GELU"
         dropout_p: float, optional
-            Dropout probability used for multi-head self attention, default 0.0
-        attention_implementation: str, optional
+            Dropout probability used for multi-head self attention, default 0.1
```
ssmmnn11, line 127: Dropout does not work currently with our sharding setup.  we have re-designed the seeding of the different GPUs, we could set it to 0. as a default (if we don't want to remove it).

models/src/anemoi/models/layers/activations.py
```diff
@@ -39,3 +42,137 @@ def leaky_hardtanh(
...
+            self.projection = nn.Linear(self.dim, self.dim * 2, bias=False, device=x.device)
+        self.V = nn.Linear(self.dim, self.dim, bias=self.bias, device=x.device)
+        self.W = nn.Linear(self.dim, self.dim, bias=self.bias, device=x.device)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self.dim is None:
+            self._post_init(x)
```
ssmmnn11: Would it be possible to supply a dim / channel attribute to these in the config file instead? The issue that could pop up is that we don't have the granularity to specify every single  activation in every layer. But I would prefer the situation where this does not always works and the model fails compared to having these being added in the first forward - it makes me uneasy.

JesperDramsch: This was discussed up here https://github.com/ecmwf/anemoi-core/pull/163#issuecomment-2718563958

Is there any substance behind feeling uneasy? 
I would prefer avoiding another irrelevant config entry that can be easily derived in the code.

models/src/anemoi/models/layers/processor.py
JesperDramsch: I just changed the docs to reflect what is actually happening. Should I change the behaviour here instead?

models/src/anemoi/models/layers/processor.py
JesperDramsch: Not necessary: https://peps.python.org/pep-0484/#the-numeric-tower

models/src/anemoi/models/layers/activations.py
JesperDramsch: Thanks!

training/src/anemoi/training/config/model/graphtransformer.yaml
JesperDramsch: I was trying to reduce config bloat and rather add to the documentation instead. But I can change it if you insist.

training/src/anemoi/training/config/model/graphtransformer_ens.yaml
JesperDramsch: Will use this https://github.com/ecmwf/anemoi-core/pull/163#discussion_r2099488296 as discssuion thread.

models/src/anemoi/models/layers/chunk.py
JesperDramsch: The default before is defined here: https://github.com/ecmwf/anemoi-core/blob/07bd6183dc835a1132faa3af6ff9d0b5cd75477c/models/src/anemoi/models/layers/block.py#L282 which is `1`.

As it was never passed it used the default kwarg defined in the thing. This may also be confusing as num_chunks in the mapper is different that num_chunks in the blocks if I recall correctly.

models/src/anemoi/models/layers/activations.py
ssmmnn11: when we set up the optimiser and wrap the model (ddp) the model is not complete.

models/src/anemoi/models/layers/processor.py
ssmmnn11: excellent!

mishooax:
> I started out with some profiling of different implementations. The profiling was done with a warmup of 5 epochs and 100 iterations on an input tensor of size `[128, 512, 2048]` (with a code snippet from @japols thank you!):

How do these runtimes compare with those of the current activation "default" (GELU, IIRC)? Do the new activations play nice with `torch.compile()`, and do you see any performance benefit from jit compilation?

models/src/anemoi/models/layers/utils.py
JesperDramsch: Good catch, thanks!

training/src/anemoi/training/config/model/graphtransformer.yaml
anaprietonem: My concern is that if the defaults changes then picking that here would not be so easy. So for now I'd say we keep all of that in the config. We could look in the future of having a explicit default of the layers in the config that can be reused across the model components.

models/src/anemoi/models/layers/activations.py
JesperDramsch: I have rewritten the entire function to just take a `dim` instead, as was originally written by @theissenhelen and @chebertpinard. cc: @anaprietonem

models/tests/layers/block/test_block_graphconv.py
JesperDramsch: Should I include it here as well? It is tested as part of the GLUs in the other tests, so by extension _should_ work here. Just didn't want to blow up the functions.

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
JesperDramsch: I have added the layer_kernels GELU definition here. cc: @MeraX

training/src/anemoi/training/config/model/graphtransformer.yaml
MeraX: Ana, is your concern related to the general discussion about setting defaults at different location?

With this PR, the defaults will be described in the docs: https://github.com/ecmwf/anemoi-core/pull/163/files#diff-7218f9c7028ace974aef33a3b6bdf869f7cffb0b76e7fefb5cf183a21f323e4eR232-R235 This should stay in sync with the code and any changes to the default must be clearly communicated as breaking change.

What I am wondering if the defaults would be logged to MLFlow or if they were just logged if they were explicitly defined in the yaml files. I believe having all settings, including defaults, logged in MLFlow is quite useful. If "bloating" the config would be the only way to transparently log the defaults, I would prefer defining them in the yaml. Perhaps one could limit the bloat by using yaml anchors and aliases.

training/pyproject.toml
JesperDramsch: I don't remember honestly.

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
MeraX: @JesperDramsch, thank you very much for taking care!

models/docs/usage/create_model.rst
```diff
@@ -214,6 +203,44 @@ In this example, ``model_interface.model`` is the following:
...
+Currently, you can switch out the following layers (with a given key):
+
+-  Activation function: ``Activation`` with default ``GELU``
+-  Linear layers: ``Linear`` with default ``torch.nn.Linear``
+-  Layer Normalisation: ``LayerNorm`` with default
+   ``torch.nn.LayerNorm``
+-  Query Normalisation: ``QueryNorm`` with default
+   ``anemoi.models.layers.normalization.AutocastLayerNorm``
+-  Key Normalisation: ``KeyNorm`` with default
+   ``anemoi.models.layers.normalization.AutocastLayerNorm``
```
MeraX, line 261: Is there a list of valid `_target_`s? Are they all defined in a single or multiple files that could be referred to here? Or could you specify the API that a `_target_` has to follow?

models/tests/layers/processor/test_graphtransformer_processor.py
```diff
@@ -34,123 +49,49 @@ def fake_graph(self) -> tuple[HeteroData, int]:
...
-            cpu_offload,
-            sub_graph,
-            edge_attributes,
-            src_grid_size,
-            dst_grid_size,
-            trainable_size,
-            qk_norm,
-        )
+    def graphtransformer_init(self):
+        return GraphTransformerProcessorConfig()
```
MeraX, line 59: Hm, I'm wondering if `GraphTransformerProcessorConfig` couldn't just become a fixture itself.

models/tests/layers/processor/test_graphtransformer_processor.py
```diff
@@ -34,123 +49,49 @@ def fake_graph(self) -> tuple[HeteroData, int]:
...
+            num_heads=graphtransformer_init.num_heads,
+            mlp_hidden_ratio=graphtransformer_init.mlp_hidden_ratio,
+            trainable_size=graphtransformer_init.trainable_size,
+            sub_graph=fake_graph[("nodes", "to", "nodes")],
+            sub_graph_edge_attributes=["edge_attr1", "edge_attr2"],
+            src_grid_size=graphtransformer_init.src_grid_size,
+            dst_grid_size=graphtransformer_init.dst_grid_size,
+            qk_norm=graphtransformer_init.qk_norm,
+            cpu_offload=graphtransformer_init.cpu_offload,
+            layer_kernels=graphtransformer_init.layer_kernels,
```
MeraX: It's perhaps a matter of taste, but did you consider simplifying this code even further with `**asdict(graphtransformer_init)`?

training/src/anemoi/training/config/model/graphtransformer.yaml
JesperDramsch: I disagree with this decision, but I will make the change.

models/src/anemoi/models/layers/processor.py
JesperDramsch: @ssmmnn11 please open a separate issue to fix this, as I can't test the impact of changing this.

JesperDramsch (commented):
I have tried to address all comments to the best of my ability.

MeraX (commented):
It's always exciting to see PRs that aim to generalize, unify, and clean up code. I skimmed through the PR and left some comments. I hope they help making the PR even better.

models/src/anemoi/models/layers/utils.py
```diff
@@ -31,20 +32,30 @@ def forward(self, *args, **kwargs):
...
+        This is useful for testing purposes.
+        Defaults to True.
+
+    Returns
+    -------
+    DotDict
+        Container with layer factories.
     """
     # If self.layer_kernels entry is missing from the config, use torch.nn kernels
     default_kernels = {
```
MeraX, line 55: Are the defaults in the code still necessary since 01ea856fc80174552dc5522e5cff69f325eb00fe also defines them in the default configs?

JesperDramsch: I'd much prefer if the defaults stayed in the code than the configs, but I just respond to the reviews. (Especially since this also defines the Queries, which aren't defined for all models)

models/tests/layers/processor/test_graphtransformer_processor.py
JesperDramsch: I was planning to set this up as something that can be parametrised eventually.

models/docs/usage/create_model.rst
JesperDramsch: Not really, basically you could replace the Query normalisation (or any other kernel) with any type of nn.Module and it should still work. Whether it makes sense is a different question. 

(The `_target_` technically is just the `hydra` instantiation pattern, but I'm believe you were rather asking for the API/interface rather than the mechanism.)

models/tests/layers/processor/test_graphtransformer_processor.py
JesperDramsch: Love this. I was looking for that but didn't find it.

models/docs/usage/create_model.rst
MeraX: I might have placed my comment inaccurately. Actually, I wanted to ask this for all the different layer types that I would be able to configure in the future.

How will a user know about the all the options available in Anemoi-models? 

Perhaps one could refer the advanced user to the respective classes in  `models/src/anemoi/models/layers/activations.py`, `models/src/anemoi/models/layers/attention.py`, etc.

models/src/anemoi/models/layers/utils.py
MeraX: I wouldn't object to either approach. However, I do believe it is crucial to have all settings, including active defaults, logged by MLFlow. If this can be easily achieved by defining defaults within the code, I would be equally satisfied. Incorporating all defaults into the code would result in significantly shorter default configuration files.

I have commented on this matter as it initially appeared to me that the current state involves duplicating defaults and introducing unnecessary default-handling logic within the code, which could be avoided by fully relying on the defaults specified in the YAML defaults file. Nevertheless, I must admit that I am not sufficiently familiar with the software engineering aspects of this part to definitively determine whether this is actual duplication or actually necessary.

training/src/anemoi/training/config/model/transformer_ens.yaml
```diff
@@ -71,34 +33,47 @@ processor:
...
   use_alibi_slopes: False # Transformer only
   cpu_offload: ${model.cpu_offload}
+  layer_kernels:
+    LayerNorm:
+      _target_: anemoi.models.layers.normalization.ConditionalLayerNorm
+      normalized_shape: ${model.num_channels}
+      condition_shape: ${model.noise_injector.noise_channels_dim}
+      w_one_bias_zero_init: True
+      autocast: false
+      #Any arguments to your chosen function go here
```
anaprietonem, line 49: are we missing here an entry for `KeyNorm` and `QueryNorm`?

training/src/anemoi/training/config/model/transformer.yaml
```diff
@@ -61,42 +26,41 @@ processor:
   softcap: 0.0 # Transformer only
   use_alibi_slopes: False # Transformer only
   cpu_offload: ${model.cpu_offload}
-  qk_norm: False
```
anaprietonem, line 64: was this deleted intentionally or by accident?

training/src/anemoi/training/config/model/transformer.yaml
```diff
@@ -61,42 +26,41 @@ processor:
...
   use_rotary_embeddings: False
-
+  layer_kernels: ${model.layer_kernels}
```
anaprietonem, line 36: I believe we are missing kernel for Keynorm and QueryNorm

training/src/anemoi/training/config/model/graphtransformer.yaml
```diff
@@ -1,57 +1,20 @@
...
-      _target_: torch.nn.Linear
-      _partial_: True
-      #Any arguments to your chosen function go here e.g.
+  LayerNorm:
+    # The GNN requires the autocast layer norm, otherwise its memory usage is too high.
+    _target_: anemoi.models.layers.normalization.AutocastLayerNorm
+  Linear:
+    _target_: torch.nn.Linear
+  Activation:
+    _target_: torch.nn.GELU
```
anaprietonem, line 21: Similar comment to transformer, missing query and key info

training/src/anemoi/training/config/model/graphtransformer_ens.yaml
```diff
@@ -7,59 +6,24 @@ cpu_offload: False
...
 noise_injector:
   _target_: anemoi.models.layers.ensemble.NoiseConditioning
   noise_std: 1
   noise_channels_dim: 4
   noise_mlp_hidden_dim: 32
   inject_noise: True
 
+layer_kernels:
+  LayerNorm:
+    # The GNN requires the autocast layer norm, otherwise its memory usage is too high.
```
anaprietonem: This would need update since I believe the layer kernel is not the same for the GNN and for the graph transformer

training/src/anemoi/training/config/model/gnn_ens.yaml
```diff
@@ -32,33 +11,53 @@ noise_injector:
   noise_mlp_hidden_dim: 32
   inject_noise: True
 
+
+layer_kernels:
+  LayerNorm:
+    # The GNN requires the autocast layer norm, otherwise its memory usage is too high.
+    _target_: anemoi.models.layers.normalization.AutocastLayerNorm
+  Linear:
+    _target_: torch.nn.Linear
```
anaprietonem, line 21: Do we need to define the Linear for the encoder and the processor?

training/src/anemoi/training/config/model/transformer.yaml
anaprietonem: I realised this was a duplicated key, so ignore the above comment

models/docs/usage/create_model.rst
JesperDramsch: Done! Thanks :)

models/src/anemoi/models/layers/utils.py
anaprietonem: @MeraX  the use of the defaults in the code here came from the initial design of the layer_kernels as a non-breaking change. For now, we have decided to keep the defaults in the configs as revisiting the use of defaults is probably outside of the scope of the PR and should look across all configs in training

PR #56, opened 2024-12-31
feat/migrate-graph-creation-to-torch

JPXKQX:
This PR improves the performance of creating a graph:

- Update the **KNNEdges** & **CutOffEdges** to rely on `torch-cluster`. This method also supports moving the tensors to the GPU.
- **EdgeAttributeBuilder**'s are now implemented as message passing graphs. This method supports moving the tensors to the GPU.

The performance improvement for generating a graph n320 (~500k nodes) -> o96 (~40k nodes) -> n320 is:

version | Device | CutOffEdges(cutoff_factor=0.6) | CutOffEdges attribute | KNNEdges(num_nearest_neighbors=3) | KNNEdges attributes
-- | -- | -- | -- | -- | --
v3.0.0 | CPU | 2.25 s | 0.55 s | 11.83 s | 1.21 s
torch-support | CPU | 0.45 s | 0.31 s | 0.30 s | 0.54 s
torch-support. | 1 GPU | 0.19 s | 0.14 s | 0.15 s | 0.00 s

The main difference is that the Haversine distance is not supported in the torch-cluster. To solve this problem, we transform the 2D coordinates into 3D coordinates (sphere) before calculating the Euclidean distance. The edge direction calculation has also been refactored to use torch.tensor's instead of np.array's & scipy (using Rodrigues' rotation formula).

PR #175, opened 2025-03-07
Feature/attribute based reweight

### LINKED ISSUE #173: Multiple Node Weight Scalers

Apply multiple scalers to node weights instead of only one at a time.

_No response_

_No response_

### Additional context

_No response_

### Organisation

MeteoSwiss

### LINKED ISSUE #174: New Node Attribute: distance from point

A new attribute computing the distance of each node from a given point (eg. center of LAM area).

_No response_

_No response_

### Additional context

_No response_

### Organisation

MeteoSwiss

icedoom888:
## Description

Introduces the possibility of scaling node weights with a sequence of operations rather than only one.
Introduces a new node attribute: distance from a given point coordinates.
 
## Type of Change

-   [x] Documentation update

## Issue Number

Fixes #174 
Fixes #173 

## Code Compatibility

-   [x] I have run the Benchmark Profiler against the old version of the code

### Dependencies

### Documentation


JPXKQX:
Hi Alberto,

Thanks for the PR! Last week, we merged [a PR](https://github.com/ecmwf/anemoi-core/pull/56) which will probably cause some conflicts with this one. When rebasing onto the latest main, feel free to reach out if you run into any issues.

A couple of points regarding this PR:

- The functionality for multiplying 2 or more node_weights is already being included in the [loss scaling PR](https://github.com/ecmwf/anemoi-core/pull/52). Since it will be handled there, it doesn’t make sense to implement it here. In this PR, I would expect to focus only on adding a new node attribute, `DistanceFromArea`.
- From a "theoretical" perspective, do we want to calculate the distance to a point (the centre of limited area)? If I understand correctly, all points in your region should have the same weight. Wouldn't using the KNNAreaMaskBuilder be a more suitable approach?

Let me know your thoughts!

graphs/src/anemoi/graphs/nodes/attributes.py
```diff
@@ -312,3 +317,104 @@ class BooleanOrMask(BooleanOperation):
...
+        # Compute mean of Cartesian coordinates
+        mean_xyz = xyz.mean(axis=0)
+
+        # Normalize to project back onto the sphere
+        mean_xyz /= np.linalg.norm(mean_xyz)
+
+        # Convert back to lat-lon
+        return cartesian_to_latlon_rad(mean_xyz[np.newaxis, :])[0]
+
+    def haversine_distance(self, src_point, dst_point):
```
JPXKQX: Check out the utils.py. We have a `haversine_distance` function there.

icedoom888:
> * The functionality for multiplying 2 or more node_weights is already being included in the [loss scaling PR](https://github.com/ecmwf/anemoi-core/pull/52). Since it will be handled there, it doesn’t make sense to implement it here. In this PR, I would expect to focus only on adding a new node attribute, `DistanceFromArea`.

Hey Mario, 
Thanks for looking into this!
Will modify the PR to focus on  `DistanceFromArea` only then!

icedoom888:
> * From a "theoretical" perspective, do we want to calculate the distance to a point (the centre of limited area)? If I understand correctly, all points in your region should have the same weight. Wouldn't using the KNNAreaMaskBuilder be a more suitable approach?

Regarding this:
The idea was to use this kind of loss over the global nodes only. That way you could penalise less errors in regions that are further away from your LAM and focusing on solving issues closer to boundary regions. All nodes of the LAM area should have constant weighting.

graphs/src/anemoi/graphs/nodes/attributes.py
icedoom888: Will change this ;)

JPXKQX:
> Regarding this: The idea was to use this kind of loss over the global nodes only. That way you could penalise less errors in regions that are further away from your LAM and focusing on solving issues closer to boundary regions. All nodes of the LAM area should have constant weighting.

The `node_weights` are multiplied by the loss values at all points, so if we want to compute the loss only over the global values, setting the local values to 1 makes sense. As for a function to apply, I would consider f such that f(0)=1 and is decreasing with dist -> inf. A few options could be:
- 1 / (1 + dist) ^ r
- e ^ (- dist ^ 2)

icedoom888:
> > Regarding this: The idea was to use this kind of loss over the global nodes only. That way you could penalise less errors in regions that are further away from your LAM and focusing on solving issues closer to boundary regions. All nodes of the LAM area should have constant weighting.
> 
> The `node_weights` are multiplied by the loss values at all points, so if we want to compute the loss only over the global values, setting the local values to 1 makes sense. As for a function to apply, I would consider f such that f(0)=1 and is decreasing with dist -> inf. A few options could be:
> 
> * 1 / (1 + dist) ^ r
> * e ^ (- dist ^ 2)

Hey Mario, 
In here I set the distance of all points inside the LAM to 0 and only loop through the outside points to assign the actual distance. 
Effectively later on when scaling by distance, all LAM points will not be affected. Or are you saying it should be 1?

```
def get_raw_values(self, nodes: NodeStorage, **kwargs) -> np.ndarray:

        distances = np.zeros(nodes.x.shape[0])

        if self.from_lam:
            assert isinstance(nodes["_dataset"], dict), "The 'dataset' attribute must be a dictionary."
            assert "cutout" in nodes["_dataset"], "The 'dataset' attribute must contain a 'cutout' key."
            num_lam, _ = open_dataset(nodes["_dataset"]).grids

            coords = nodes.x[num_lam:, :]
            dst_point = self.central_point_on_sphere(nodes.x[:num_lam])

        else:
            coords = nodes.x
            dst_point = self.dst_point

        # Compute distances for all points from the central point
        for i, p in enumerate(coords):
            distances[num_lam + i] = self.haversine_distance(dst_point, p)

        return distances
```

Also how should i modify this PR to not conflict with your new multi node weight loss scaling?

JPXKQX:
Hi Alberto, we have merged the loss refactor PR. Feel free to update the PR when you have time. I would expect the changes to only affect `graph/nodes/attributes`, `graph/schemas` and `graph/docs`. Let me know if you want to discuss anything.

PR #181, opened 2025-03-10
compile transformer gnn

ssmmnn11:
Draft pull request to compile GraphTransformerConv

### DIFF
--- a/models/src/anemoi/models/compile/decorators.py
+++ b/models/src/anemoi/models/compile/decorators.py
@@ -1,0 +1,20 @@
+# (C) Copyright 2024 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+def torch_compile():
+    def decorator(func):
+        import torch
+        import torch_geometric
+        req = torch.__version__ >= "2.6" and torch_geometric.__version__ >= "2.6"
+        if req:
+            compile_func = torch.compile(dynamic=True)
+            return compile_func(func)
+        return func
+    return decorator
--- a/models/src/anemoi/models/layers/conv.py
+++ b/models/src/anemoi/models/layers/conv.py
@@ -22,6 +22,7 @@ from torch_geometric.utils import scatter
 from torch_geometric.utils import softmax
 
 from anemoi.models.layers.mlp import MLP
+from anemoi.models.compile.decorators import torch_compile
 from anemoi.utils.config import DotDict
 
 
@@ -81,7 +82,7 @@ class GraphConv(MessagePassing):
 
         return out, edges_new
 
-
+@torch_compile()
 class GraphTransformerConv(MessagePassing):
     """Message passing part of graph transformer operator.
 


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=181) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=181) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=181) it.</sub>

models/src/anemoi/models/compile/decorators.py
HCookie, line 1: Can you please add a simple docstring to the function detailing the minimum requirements for compilation to occur?
Logging the time taken to compile may also be useful information.

HCookie: Additionally, will extra arguments to the decorator be expected?
If not, the `torch_compile` and `decorator` function could be merged, and it used as 
```python
@torch_compile
def func():
	...
```

HCookie (changes requested):

models/src/anemoi/models/compile/decorators.py
ssmmnn11: Good question. At some point we want this to be configurable because it is possible to test a wide range of options on torch compile, see torch._inductor.list_options() -> also output the kernel it generates etc. 

compilation will happen the first time the module / function is called.

models/src/anemoi/models/compile/decorators.py
```diff
@@ -0,0 +1,22 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+def torch_compile():
+    def decorator(func):
+        import torch
+        import torch_geometric
+
+        req = torch.__version__ >= "2.6" and torch_geometric.__version__ >= "2.6"
```
mchantry: Agreed that we should make the minimal version of torch be `2.6` and remove this.

HCookie:
When running all the tests with torch>2.6 and geometric>2.6 no issues are raised.
If that is the case, it may be possible to update the min dependencies, and remove this decorator, and directly use `torch.compile`.

HCookie (approved):

models/src/anemoi/models/layers/conv.py
```diff
@@ -82,6 +83,7 @@ def aggregate(self, edges_new: Tensor, edge_index: Adj, dim_size: Optional[int]
         return out, edges_new
 
 
+@torch_compile()
```
japols: Do we want to make these compile options configurable?
Benchmarking results on an A100 suggest dynamic=False, mode='max-autotune' performs best for fp16:

|      module                                   | time_ms/iter | peak_memory_MB  |
|-----------------------------------------------|--------------|-----------------|
| GraphTransformerConv                          |      59.96   |       7565.30   |
| CompiledGraphTransformerConv(dynamic=True)    |      63.61   |       3641.00   |
| CompiledGraphTransformerConv(dynamic=False)   |       9.39   |       3641.00   |

supporting dynamic=False would require compiling the Encoder/Decoder separately. Happy to add that to this PR.

icedoom888 (dismissed):
lgtm

einrone:
I heard great news about this new feature. Just builded a new 2.6.0 torch (for rocm 6.2.4)  and torch geometric container on LUMI. Will perform some tests for stretched-grid and benchmark.

ssmmnn11:
> I heard great news about this new feature. Just builded a new 2.6.0 torch (for rocm 6.2.4) and torch geometric container on LUMI. Will perform some tests for stretched-grid and benchmark.

Hi, it depends a bit on the configuration you are running. Can you try float16 and then set dynamic=False in the decorator? We still need to make it more configurable.

einrone:
> > I heard great news about this new feature. Just builded a new 2.6.0 torch (for rocm 6.2.4) and torch geometric container on LUMI. Will perform some tests for stretched-grid and benchmark.
> 
> Hi, it depends a bit on the configuration you are running. Can you try float16 and then set dynamic=False in the decorator? We still need to make it more configurable.

Yes, I will try to do that. Do you mean pure float16 or mixed-16 (float32 and float16)?

ssmmnn11:
Mixed :-)

einrone:
However, would it not be better to not pin minimum dependency to torch 2.6.0 and torch 2.6.0. I am just thinking of users how run on systems that does not support latest version of these packages. Then instead of throwing an error, it just skips the @torch_compile decorator (as it is now). Just a suggestion from my site. Probably not an issue though.

fprill:
Just a beginner's question: Is there a reason why you did not implement this feature something like
```python
if hasattr(torch, 'compile'):
    torch_compile = torch.compile(fullgraph=True, mode='max-autotune',  dynamic=False)
else:
    torch_compile = lambda func: func  # Identity fallbac
```

HCookie:
@fprill, returning the function directly is akin to returning an identity function and removes unneccessary layer a lambda will cause.

HCookie:
@einrone In our testing we saw no reason to avoid increasing the minimum dependancy, if this is an issue for you, we would be happy to revert that, as with the current implementation, it would just skip the compile.

models/src/anemoi/models/compile/decorators.py
```diff
@@ -0,0 +1,37 @@
...
+        if req:
+            compile_func = torch.compile(dynamic=dynamic, **kwargs)
+            return compile_func(func)
+
+        LOG.debug("Requirements not met for torch.compile, returning original function.")
```
gabrieloks: Can we make this more explicit? Maybe we should use a flag in the config files. Something like:

compile: True

And if compile== True but torch and torch_geometric are not on the correct version, the code should crash.

ssmmnn11:
Please note that this is here primarily to have something to point people to. Please test different settings and give feedback how it performs for you. We need to gather some experience what combinations - precision, hardware etc. - work well and then we can make decisions how to bring it in :-).

einrone:
My apologies for late reply. I have now created a container with torch 2.6.0 based on LUMI's torch singularity images using rocm 6.2.4. However when trying to run a normal job on standard-g, the job hangs and gets killed due to NCCL watchdog timeout. Tried different env setups for NCCL/RCCL but no luck. Currently in the loop with LUMI about how to fix this issue. The work is a bit slow due to large queues.

When it comes to minimum dependency @HCookie . I think having it not pinned to torch 2.6.0 and above would prevent this issue since the user (who uses LUMI)  is not locked into use this functionality (torch.compile), and can still train models with an older torch version.

mchantry:
Thanks @einrone we are not planning to pin the torch version here, but without this version of torch, some of the compile options won't be supported.

einrone:
Hi, managed to create a container on LUMI with pytorch 2.7.0 and flash-attn. However, I used the old code for GT compile. Observed approx 30% reduction in memory. 

see slack for submit and container script:
[](https://app.slack.com/client/T06JL8GP4P5/C079023RETD)

Name of container: /pfs/lustrep4/scratch/project_465000527/anemoi/containers/pytorch-2.7.0-rocm-6.2.4-py-3.12.9-v2.0.sif

ssmmnn11:
> I made some changes to this PR, but happy to spin them off to another branch if people think it's beyond the scope.
> 
> @ssmmnn11 noticed that compiling the ConditonalLayerNorm is 3x faster than the uncompiled ConditionalLayerNorm (at 1024c) layernorm-perf
> 
> So my changes also compiles that component. @mishooax had the idea of a 'ConditionalCompile' wrapper (open for better name suggestions :D). You can give options to this wrapper which are passed on to torch.compile
> 
> ```python
> # models/layers/normalization.py
> 
> from anemoi.models.compile.conditional import ConditionalCompile
> 
> 
> class ConditionalLayerNorm(nn.Module):
>     """Conditional Layer Normalization.
> 
>     x_norm = a(u) * (x - mean) / sqrt(var + eps) + b(u)
> 
>     """
> 
>     @ConditionalCompile(dynamic=True)
>     def forward(self, x: Tensor, cond: Tensor) -> Tensor:
>          ...
> ```
> 
> And then I have added a `compile` field to the models config. This is a list of modules to be compiled. Currently you need both the hydra entry, and the @ConditionalCompile wrapper for any compilation to occur.
> 
> ```yaml
> # training/src/anemoi/training/config/model/transformer_ens.yaml
> 
> # A list of elements to be compiled. these will be given the self.compile=True variable, which will then be
> # acted on by ConditionalCompile
> compile:
>   - anemoi.models.layers.normalization.ConditionalLayerNorm
>   - anemoi.models.layers.conv.GraphTransformerConv
> ```
> 
> Thoughts? @HCookie @ssmmnn11

could we also supply options for each element? maybe make them a dict or something? so some could be dynamic=True, others not? :-)

ssmmnn11:
> > I made some changes to this PR, but happy to spin them off to another branch if people think it's beyond the scope.
> > @ssmmnn11 noticed that compiling the ConditonalLayerNorm is 3x faster than the uncompiled ConditionalLayerNorm (at 1024c) layernorm-perf
> > So my changes also compiles that component. @mishooax had the idea of a 'ConditionalCompile' wrapper (open for better name suggestions :D). You can give options to this wrapper which are passed on to torch.compile
> > ```python
> > # models/layers/normalization.py
> > 
> > from anemoi.models.compile.conditional import ConditionalCompile
> > 
> > 
> > class ConditionalLayerNorm(nn.Module):
> >     """Conditional Layer Normalization.
> > 
> >     x_norm = a(u) * (x - mean) / sqrt(var + eps) + b(u)
> > 
> >     """
> > 
> >     @ConditionalCompile(dynamic=True)
> >     def forward(self, x: Tensor, cond: Tensor) -> Tensor:
> >          ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And then I have added a `compile` field to the models config. This is a list of modules to be compiled. Currently you need both the hydra entry, and the @ConditionalCompile wrapper for any compilation to occur.
> > ```yaml
> > # training/src/anemoi/training/config/model/transformer_ens.yaml
> > 
> > # A list of elements to be compiled. these will be given the self.compile=True variable, which will then be
> > # acted on by ConditionalCompile
> > compile:
> >   - anemoi.models.layers.normalization.ConditionalLayerNorm
> >   - anemoi.models.layers.conv.GraphTransformerConv
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thoughts? @HCookie @ssmmnn11
> 
> could we also supply options for each element? maybe make them a dict or something? so some could be dynamic=True, others not? :-)

please leave everything in this branch, I think that is nice.

training/src/anemoi/training/train/train.py
```diff
@@ -180,6 +180,48 @@ def truncation_data(self) -> dict:
 
         return truncation_data
 
+    def set_compile_flags(self, compile_config):
+        # I guess min triton version is enforced by torch 2.6
+        # TODO allow setting different requirements for each class
+        import torch_geometric
+
+        req = torch.__version__ >= "2.6" and torch_geometric.__version__ >= "2.6"
```
mchantry: Remove this and go with toml to control version?

cathalobrien: yeah that's better

cathalobrien:
> > @ssmmnn11 Now the compilation is controlled entirely from the config, you can set options too
> > ```yaml
> > #config/model/graphtransformer_ens.yaml
> > 
> > # Torch compile configuration
> > # A list of modules present in the model, which will be compiled
> > # You can optionally pass options to torch.compile with the 'options' key
> > compile:
> >   - module: anemoi.models.layers.conv.GraphTransformerConv
> >     options:
> >       dynamic: false
> >       #mode: max-autotune # an example of setting the compilation mode
> >   - module: anemoi.models.layers.normalization.ConditionalLayerNorm
> >     options:
> >       dynamic: false
> > ```
> 
> Can you explain `dynamic` in a comment and when the user might want to change it please.

@mchantry I have added the following explanation

```yaml
# Torch compile configuration
# A list of modules present in the model, which will be compiled
# You can optionally pass options to torch.compile with the 'options' key
#
# Below is an explanation of some common parameters to torch.compile
# For a full list of possible parameters, consult the documenation for torch compile
#       https://docs.pytorch.org/docs/stable/generated/torch.compile.html
#
# dynamic (bool): When True, it will try to avoid recompilation by generating
#       as general a kernel as possible. But the performance of the general
#       kernel might be worse. When False, it will generate a specific
#       kernel for each input shape (until the configurable recompile
#       limit has been hit), leading to possibly better performance but
#       more recompilations
# mode (str): Different compilation modes, allowing you to trade off
#       compilation time versus potential performance. See the
#       torch.compile documentation for list of possible modes
# fullgraph (bool): When True, torch.compile will error when it hits a
#       section of code it can't compile. When False, it will fallback to
#       non-compiled ("eager") execution for those lines.
# options (dict): a dict of further options which can be passed to torch.compile
```

For a visual example of dynamic in action, here I'm running with dynamic=False. The first 8 input shapes have kernels generated specifically for them. But then I hit the recompile limit, so it effectively falls back to dynamic=True, leading to worse performance
layernorm-perf (1)

models/src/anemoi/models/schemas/models.py
```diff
@@ -223,6 +225,8 @@ class BaseModelSchema(PydanticBaseModel):
         discriminator="target_",
     )
     "GNN decoder schema."
+    compile: Optional[list[dict[str, Any]]]
```
HCookie:
```suggestion
    compile: Optional[list[dict[str, Any]]] = Field(None)
```

cathalobrien: done, tysm

cathalobrien:
correctness integration tests fails bc of CPU OOM, 
memory seems to accumulate on CPU across pytests. I saw the same on the benchmark branch.
My hacky fix there was to take a a whole nodes memory. that's not an option here because I understand the correctness integration tests run on a login node. 
So would be good to track down the cause here

mchantry:
Do we expect this to work on AMD hardware? Do we need testers?

mchantry:
Marked as ATS approval for brief discussion on significant torch version bump.

training/pyproject.toml
```diff
@@ -56,8 +56,8 @@ dependencies = [
   "pyshtools>=4.13",
   "pytorch-lightning>=2.1",
   "timm>=0.9.2",
-  "torch>=2.2",
-  "torch-geometric>=2.3.1",
+  "torch>=2.6",
+  "torch-geometric>=2.6",
```
anaprietonem: I believe newer versions of torch-geometric (from 2.5 onwards) introduce a class called inspector that can be intrusive and break checkpoints so this might be worth flagging or think if I can have a migration checkpoint script t o address it

floriankrb:
Is there anything to do in inference to run checkpoints of models that have been compiled? 
(i.e. would inference need to compile again or use the exact same GPU/libraries as training?)

training/pyproject.toml
cathalobrien: good to know. I made reverted the pyproject and made torch-geometric a soft dependancy for compile only.

models/src/anemoi/models/compile/decorators.py
cathalobrien: I have made it so that, if you try to compile something and you dont meet the requirements it warns you and keeps on running.

training/pyproject.toml
anaprietonem: When talking about this at ATS today, @bdvllrs pointed out that the anemoi-models migration https://anemoi.readthedocs.io/projects/models/en/latest/cli/migration.html could handle this case of 'new breaking class in torch-geometric' - so Cathal we could look at adding a migration script here as part of this PR

training/src/anemoi/training/config/model/graphtransformer.yaml
```diff
@@ -68,6 +68,29 @@ trainable_parameters:
   hidden2data: 8
   hidden2hidden: 8 # GNN and GraphTransformer Processor only
 
+# Torch compile configuration
+# A list of modules present in the model, which will be compiled
+# You can optionally pass options to torch.compile with the 'options' key
+#
+# Below is an explanation of some common parameters to torch.compile
```
cathalobrien, line 75: should we move this explanation from each config to the anemoi training documentation, to reduce duplication?

training/src/anemoi/training/utils/compile.py
```diff
@@ -0,0 +1,92 @@
...
+
+def mark_for_compilation(model: BaseGraphModule, compile_config: DictConfig | None) -> BaseGraphModule:
+    """Compiles parts of 'model' according to 'config.model.compile'."""
+    if compile_config is None:
+        return model
+
+    if not _meets_library_versions_for_compile():
+        return model
+
+    LOGGER.info("The following modules will be compiled: %s", str(compile_config))
```
cathalobrien: right now i dont check that all entries in compile_config are present in the model, which could be confusing

e.g. if you ask for condLayerNorm compiled but you dont use that layer in your model, it will say it will mark compile condLayerNorm and then nothing will happen.
I could warn for unused entries at the end of this function, at least its more explicit

training/src/anemoi/training/config/model/transformer.yaml
```diff
@@ -77,6 +77,33 @@ attributes:
...
+# mode (str): Different compilation modes, allowing you to trade off
+#       compilation time versus potential performance. See the
+#       torch.compile documentation for list of possible modes
+# fullgraph (bool): When True, torch.compile will error when it hits a
+#       section of code it can't compile. When False, it will fallback to
+#       non-compiled ("eager") execution for those lines.
+# options (dict): a dict of further options which can be passed to torch.compile
+compile:
+  - module: anemoi.models.layers.conv.GraphTransformerConv
+    options:
```
cathalobrien: should standardise these options across all configs (i removed dynamic: false from the GT configs bc LAM or sharded struggled with it)

training/pyproject.toml
icedoom888: The change to torch 2.6 will probably break compatibility for Meteoswiss on our Balfrin cluster.

Balfrin is running with Driver Version: 525.105.17 & CUDA Version: 12.0
 PyTorch 2.6 + CUDA 12.6 requires Driver Version ≥ 535.xx

cathalobrien: Hey Alberto,
I removed the strict torch 2.6 requirement in the pyproject. now its an optional dependancy to enable compilation. if you dont have torch 2.6, compilation is skipped and you run as norml.

If the optional torch 2.6 requirement is too high for Meteoswiss and you want to try torch compile, feel free to edit the function `anemoi.training.utils.compile._meets_library_versions_for_compile`, to lower the requirements

training/src/anemoi/training/train/train.py
```diff
@@ -522,6 +523,9 @@ def train(self) -> None:
             check_val_every_n_epoch=getattr(self.config.diagnostics, "check_val_every_n_epoch", 1),
         )
 
+        if hasattr(self.config.model, "compile"):
```
anaprietonem: can this be moved to a function to it's a bit cleaner?

training/src/anemoi/training/train/train.py
```diff
@@ -530,8 +534,8 @@ def train(self) -> None:
             ckpt_path=None if (self.load_weights_only) else self.last_checkpoint,
         )
 
-        if self.config.diagnostics.print_memory_summary:
-            LOGGER.info("memory summary: %s", torch.cuda.memory_summary())
+        if self.config.diagnostics.print_memory_summary and rank_zero_only.rank == 0:
```
anaprietonem: why do we need to add this for the memory summary is that because of the compiler? Otherwise I'd suggest to undo these changes

ssmmnn11: I agree, we want the memory summary for all ranks.

models/src/anemoi/models/utils/compile.py
```diff
@@ -0,0 +1,102 @@
...
+        LOGGER.warning(msg)
+        has_triton = False
+
+    version_req = torch.__version__ >= "2.6" and torch_geometric.__version__ >= "2.6"
+
+    if not version_req:
+        msg = "Minimum library versions for compilation not met. "
+        msg += f"torch: v{torch.__version__}<2.6 or torch_geometric: v{torch_geometric.__version__}<2.6. "
+        msg += "Please upgrade these libraries to enable compilation."
+        LOGGER.warning(msg)
```
anaprietonem, line 56: shouldn't this raise an error? so we stop the execution or what would happen if the user tries to use compile with wrong versions?

models/tests/utils/test_compile.py
```diff
@@ -0,0 +1,72 @@
+# (C) Copyright 2024 Anemoi contributors.
```
anaprietonem, line 1: minor - could we please update to 2025 pls

training/src/anemoi/training/train/train.py
```diff
@@ -522,6 +523,9 @@ def train(self) -> None:
             check_val_every_n_epoch=getattr(self.config.diagnostics, "check_val_every_n_epoch", 1),
         )
 
+        if hasattr(self.config.model, "compile"):
+            self.model = mark_for_compilation(self.model, self.config.model.compile)
```
anaprietonem:
```suggestion
            self.model = mark_for_compilation(self.model, self.config.model_dump(by_alias=True).model.compile)
```

Otherwise the code fails when Pydantic is off as we are not converting that config into a omegaconf.dictconfig.DictConfig

training/src/anemoi/training/config/model/transformer.yaml
```diff
@@ -77,6 +77,33 @@ attributes:
...
+#       limit has been hit), leading to possibly better performance but
+#       more recompilations
+# mode (str): Different compilation modes, allowing you to trade off
+#       compilation time versus potential performance. See the
+#       torch.compile documentation for list of possible modes
+# fullgraph (bool): When True, torch.compile will error when it hits a
+#       section of code it can't compile. When False, it will fallback to
+#       non-compiled ("eager") execution for those lines.
+# options (dict): a dict of further options which can be passed to torch.compile
+compile:
```
anaprietonem, line 101: Could people compile whole blocks rather than layers? What would happen if we have multiple encoders, - we would compile layers across both right?

models/src/anemoi/models/utils/compile.py
```diff
@@ -0,0 +1,102 @@
+# (C) Copyright 2024 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import logging
```
anaprietonem, line 10: What about moving this script/functionality to anemoi-models so that we can in the future consider reusing it for inference? I also have some concerns if the compile is living some trace associated to anemoi-training and hence 'pollute' the checkpoint with that dependency.

models/tests/utils/test_compile.py
anaprietonem: Similar comment - can we consider moving this to anemoi models?

japols (commented):
Nice work :) should we also make the recompile limit configurable? pretty easy to hit in my experience with many mapper chunks.

training/src/anemoi/training/config/model/transformer.yaml
cathalobrien: Yeah i dont see why you couldnt compile whole blocks by passing the block.
maybe that would fuse more of the code into one big kernel. is that easy to test?

models/src/anemoi/models/utils/compile.py
cathalobrien: sure i dont see why not

models/tests/utils/test_compile.py
cathalobrien: done

training/src/anemoi/training/train/train.py
cathalobrien: ok i'll remove it, but i think it is really spammy when you have O(100) ranks to print each rank

training/src/anemoi/training/train/train.py
cathalobrien: sure, good idea

models/src/anemoi/models/utils/compile.py
cathalobrien: right now it just doesnt compile if the libraries aren't new enough. I think if it throws an error then we should increase the minimum version of torch and torch geometric, but i think thats a tough sell

training/src/anemoi/training/config/model/transformer.yaml
ssmmnn11: I think it does currently not work with our  communication routines which are embedded in the blocks. But could be retested.

models/src/anemoi/models/utils/compile.py
cathalobrien: done

anaprietonem (approved):
LGTM, thanks for addressing the comments Cathal!

PR #182, opened 2025-03-11
feat: Kcrps 

ssmmnn11:
## Description

AIFS-CRPS

@mishooax, @MartinLeutbecher, @jakob-schloer, @Rilwan-Adewoyin, @mc4117, @JesperDramsch

to test use debug_ens config. Work in progress ...


training/src/anemoi/training/data/dataset.py
```diff
@@ -89,6 +85,9 @@ def __init__(
         self.reader_group_rank = 0
         self.reader_group_size = 1
 
+        self.seed_comm_num_groups = 1
```
mchantry: Maybe a comment to explain a bit what they do?

mchantry:
To do, open issues about what is supported, e.g. no CRPS msgGNN

training/src/anemoi/training/data/datamodule.py
```diff
@@ -220,3 +212,30 @@ def val_dataloader(self) -> DataLoader:
 
     def test_dataloader(self) -> DataLoader:
         return self._get_dataloader(self.ds_test, "test")
+
+
+class AnemoiEnsDatasetsDataModule(AnemoiDatasetsDataModule):
```
mchantry: I wonder about making this a separate file.
`ensdatamodule.py` ?

training/src/anemoi/training/config/debug_ens.yaml
```diff
@@ -0,0 +1,129 @@
...
+  batch_size:
+    training: 2
+    validation: 2
+    test: 2
+  limit_batches:
+    training: 8
+    validation: 8
+
+training:
+  forecaster: anemoi.training.train.forecaster.GraphEnsForecaster # refactor to use _target_ etc?
```
mchantry: This is called `task` in the interpolator PR. Thinking about future downscaling, I think `task` maybe a more futureproof name. Thoughts?

mchantry: How significant is the refactor to use `_target_`?

training/src/anemoi/training/train/train.py
```diff
@@ -154,18 +155,35 @@ def graph_data(self) -> HeteroData:
...
             "data_indices": self.data_indices,
             "graph_data": self.graph_data,
+            "truncation_data": self.truncation_data,
             "metadata": self.metadata,
             "statistics": self.datamodule.statistics,
             "supporting_arrays": self.supporting_arrays,
         }
 
-        model = GraphForecaster(**kwargs)
+        forecaster = get_class(self.config.training.forecaster)
```
mchantry: Replace by `model_class` as used in interpolator PR?

training/src/anemoi/training/train/train.py
```diff
@@ -154,18 +155,35 @@ def graph_data(self) -> HeteroData:
...
+        if self.config.hardware.files.truncation is not None:
+            truncation_data["down"] = load_npz(self.config.hardware.files.truncation)
+
+        if self.config.hardware.files.truncation_inv is not None:
+            truncation_data["up"] = load_npz(self.config.hardware.files.truncation_inv)
+
+        return truncation_data
+
+    @cached_property
+    def model(self) -> Any:
```
mchantry: Should `Any` be `pl.LightningModule`

training/src/anemoi/training/data/datamodule.py
mchantry: Should this live in a different file?
`ensdatamodule.py`

training/src/anemoi/training/train/forecaster.py
```diff
@@ -718,24 +720,389 @@ def validation_step(self, batch: torch.Tensor, batch_idx: int) -> None:
...
             optimizer,
             lr_min=self.lr_min,
             t_initial=self.lr_iterations,
-            warmup_t=self.warmup_t,
+            warmup_t=self.lr_warmup,
         )
         return [optimizer], [{"scheduler": scheduler, "interval": "step"}]
+
+
+class GraphEnsForecaster(GraphForecaster):
```
mchantry: I would propose this is moved to `ensforecaster.py`.
We can imagine this list of "forecasters" growing, and the file becoming very large.

mchantry: @ssmmnn11 thinking of a future with diffusion, how would that sit into the namespace? Would it use `GraphEnsForecaster`? Would it need a new Forecaster?

training/src/anemoi/training/losses/kcrps.py
```diff
@@ -0,0 +1,209 @@
...
+        -------
+        kCRPS value
+            The point-wise kernel CRPS, shape (batch_size, 1, latlon).
+        """
+        ens_size = preds.shape[-1]
+        mae = torch.mean(torch.abs(targets[..., None] - preds), dim=-1)
+
+        assert ens_size > 1, "Ensemble size must be greater than 1."
+
+        if ens_size == 1:
```
mchantry: Why have this and the assert above?

training/src/anemoi/training/config/debug_ens.yaml
ssmmnn11: If everyone is happy with the current form then we can leave it as it is. I mainly wanted other opinions on this.

training/src/anemoi/training/data/dataset.py
ssmmnn11: renamed to sample_comm_group...

training/src/anemoi/training/train/train.py
```diff
@@ -154,18 +155,35 @@ def graph_data(self) -> HeteroData:
...
         """Provide the model instance."""
         kwargs = {
             "config": self.config,
             "data_indices": self.data_indices,
             "graph_data": self.graph_data,
+            "truncation_data": self.truncation_data,
```
anaprietonem, line 186: With this I guess we make sure the truncation data matrix is available at inference time right? My other question is in the same way we track the datasets paths/files in the catalogue, it could be worth tracking the files too. Have you discussed this already with Baudouin, @ssmmnn11 ? (in terms of keeping full traceability and for back up of the files)

anaprietonem:
Just to flag, that this PR will need some docs update. The APIs will be updated automatically with sphinx, but it would be good to have some info to explain the new model interface and some of the major changes. Thank you!

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -33,8 +33,19 @@ swa:
...
+optimizer:
+  zero: False # use ZeroRedundancyOptimizer ; saves memory for larger models
+  kwargs:
+    betas: [0.9, 0.95]
+
+# select forecaster
+forecaster: anemoi.training.train.forecaster.GraphForecaster
+
+# select strategy
+strategy:
```
anaprietonem, line 46: In the init of the strategy we include the `read_group_size` why don't specify this at config level?

training/src/anemoi/training/config/small_ens_test.yaml
```diff
@@ -0,0 +1,170 @@
...
+
+data:
+  resolution: o96
+
+datamodule:
+  _target_: anemoi.training.data.datamodule.AnemoiEnsDatasetsDataModule
+
+hardware:
+  paths:
+    truncation_matrices: /ec/res4/hpcperm/nesl/inter_mat
```
anaprietonem: This should be just `truncation:` right? So it matches with the schemas and also with debug_ens.yaml

training/src/anemoi/training/train/forecaster.py
ssmmnn11: will split. regrading diffusion: it does not rely on ensemble members os I think we can then just go with GraphDiffForecaster etc.

training/src/anemoi/training/config/small_ens_test.yaml
```diff
@@ -0,0 +1,170 @@
...
+    data: /home/mlx/ai-ml/datasets/
+    grids: /home/mlx/ai-ml/grids/
```
anaprietonem: Grids are no longer needed to pass in the configs so this need to be deleted to also match with the pydantic schemas

training/src/anemoi/training/train/train.py
ssmmnn11: yes, same as graph data. I have not spoken to Baudouin about that yet

training/src/anemoi/training/config/small_ens_test.yaml
ssmmnn11: well spotted. this was forgotten when refactoring

training/src/anemoi/training/train/train.py
JPXKQX: I think we only need to store this in the `supporting_arrays` property

training/src/anemoi/training/config/debug_ens.yaml
JPXKQX: Would it make sense to also move the task specific args under task, i.e. `multistep_input`?

training/src/anemoi/training/config/debug.yaml
HCookie, line 1: A number of the config files seem to be left over from testing, and need to be removed. Some even contain local paths

HCookie (changes requested):

training/src/anemoi/training/train/forecaster/ensforecaster.py
```diff
@@ -0,0 +1,396 @@
...
+        batch: torch.Tensor,
+        batch_idx: int,
+        validation_mode: bool = False,
+    ) -> tuple:
+        """Training / validation step."""
+        del batch_idx
+
+        batch[0] = self.allgather_batch(batch[0])
+        if len(batch) == 2:
+            batch[1] = self.allgather_batch(batch[1])
```
theissenhelen, line 296: Under what circumstances do we have len(batch) ==2?

training/src/anemoi/training/distributed/strategy.py
```diff
@@ -147,48 +216,236 @@ def setup(self, trainer: pl.Trainer) -> None:
...
+        -> these are still divided by the total number of GPUs in DDP as if each rank would see a full set of inputs
+        note: the trainable parameters are added before the split across GPUs and are therefore not rescaled.
+        """
+        for name, param in self.model.named_parameters():
+            if param.requires_grad is True and "trainable" not in name:
+                param.register_hook(lambda grad: grad * float(self.model_comm_group_size))
+
 
-        return model_comm_group_id, model_comm_group_rank, model_comm_num_groups
+class DDPEnsGroupStrategy(DDPStrategy):
```
evenmn, line 272: Why do we even need a separate DDP strategy for ensembles? Can't we just reuse this for deterministic training, but with num_gpus_per_ensemble=num_gpus_per_model (effectively setting ensemble size to 1)?

training/src/anemoi/training/distributed/strategy.py
```diff
@@ -147,48 +216,236 @@ def setup(self, trainer: pl.Trainer) -> None:
...
-            Reader_group id, Reader_group rank, Reader_group root (global rank)
         """
-        reader_group_id = model_comm_group_rank // read_group_size
-        reader_group_rank = model_comm_group_rank % read_group_size
-        reader_group_size = read_group_size
-        reader_group_root = (self.global_rank // read_group_size) * read_group_size
+        super().__init__(**kwargs)
+        self.model_comm_group_size = num_gpus_per_model
+        self.read_group_size = read_group_size
+        self.ens_comm_group_size = num_gpus_per_ensemble
```
evenmn, line 291: As a user I find it more user friendly to specify the number of members to run in parallel rather than number of gpus per ensemble. Could we perhaps redefine 
```python
num_gpus_per_ensemble = num_gpus_per_model * num_members_in_parallel
```
and specify `num_members_in_parallel` in the config? Similar applies to `ensemble_size_per_device`, which should be renamed to `num_members_per_model` to keep a consistent naming convention

evenmn (changes requested):
I'm currently running some tests where I compare stretched grid results obtained from this PR to results obtained from a similar training in aifs-mono. As training a full stretched-grid model is time consuming (at least with the transfer learning steps discussed in https://arxiv.org/abs/2409.02891), I will instead train a model with MEPS+ERA5 (N320) from scratch with a limited number of iterations (100 epochs/13000 iterations) and compare it to a similar model trained in aifs-mono.

I will of course also do a full stretched-grid training, but this PR should not depend on it to be merged in.

My impressions so far is that training with stretched grid works out-of-the-box (though the results are not yet verified). I made a couple of suggestions on how to improve the user-friendliness, but apart from that everything seems fine to me.

training/src/anemoi/training/distributed/strategy.py
ssmmnn11: I wonder if this might be confusing - what happens if you set ensemble members to 4, but run the deterministic model? For now I would prefer to leave it like this, and we can do some more iterations in the future if we think it is necessary. But I also expect that we will have more strategies in the future.

training/src/anemoi/training/distributed/strategy.py
ssmmnn11: I guess the tricky bit is:
- there can be multiple members on one GPU
- all members are run in parallel - they can be on different devices.

I agree that the current way is not the most user friendly, but to substantially improve it more work would be required. My preference would be to go ahead as it is and improve in the future.

training/src/anemoi/training/train/forecaster/ensforecaster.py
ssmmnn11: This comes into play when we train with pert. initial conditions.

training/src/anemoi/training/train/train.py
ssmmnn11: what do we have to do here?

training/src/anemoi/training/config/debug_ens.yaml
ssmmnn11: good point. will do that in a follow up

anaprietonem (approved):

PR #186, opened 2025-03-12
feat(mlflow): Make direct endpoint calls robust

gmertes:
## Description

There are two parts in training that talk to the mlflow endpoint direct: the authentication layer, and the health check that verifies if the authentication succeeded and we can reach the server.

These send requests to the mlflow endpoint directly, and if it fails the code will crash.

This PR adds a retry mechanic, where it will retry those requests if the failure is caused by a recoverable HTTP response (bad gateway, service unavailable, etc). This will make training less prone to crashes from intermittent network issues.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

There is a new indirect dependency on `multiurl` (anemoi.utils robust calls that under the hood) , but this is an ECMWF package and is very lightweight with only the following dependencies: "requests", "tqdm", "pytz", "python-dateutil". All of these are already in the dependency tree.

### DIFF
--- a/training/src/anemoi/training/diagnostics/mlflow/auth.py
+++ b/training/src/anemoi/training/diagnostics/mlflow/auth.py
@@ -20,11 +20,11 @@ from getpass import getpass
 from typing import TYPE_CHECKING
 
 import requests
-from multiurl.http import robust
 from requests.exceptions import HTTPError
 
 from anemoi.utils.config import load_config
 from anemoi.utils.config import save_config
+from anemoi.utils.remote import robust
 from anemoi.utils.timer import Timer
 
 REFRESH_EXPIRE_DAYS = 29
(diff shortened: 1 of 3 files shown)


HCookie (approved):

training/src/anemoi/training/diagnostics/mlflow/auth.py
```diff
@@ -217,7 +226,12 @@ def _request(self, path: str, payload: dict) -> dict:
         }
 
         try:
-            response = requests.post(f"{self.url}/{path}", headers=headers, json=payload, timeout=30)
```
floriankrb, line 220: Making this robust may lead to multiple posts. It looks safe to do here.

floriankrb: If I understand corectly, "path" is always "refresh_token" and could (should?) be hard coded here?

gmertes: Yes, it is safe to post multiple times. 

The path is a leftover from when we used to have more than one path. It could be hardcoded, but it's also okay to leave it in case another path ever comes back.

PR #193, opened 2025-03-17
fix: add reference in integration test workflow

VeraChristina:
## Description

The integration test workflow fails due to a missing reference in the 'uses' attribute pointing to the action in reusable workflows. This PR adds a reference to fix the syntax.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -8,7 +8,7 @@ jobs:
   integration_tests:
     runs-on: ubuntu-latest
     steps:
-      - uses: ecmwf/reusable-workflows/nightly-pytest
+      - uses: ecmwf/reusable-workflows/nightly-pytest@main
         with:
           repository: ${{github.repository}}
           ref: ${{github.ref}}


.github/workflows/integration-tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   integration_tests:
```
theissenhelen, line 8: The workflow should have a proper name as well e.g. nightly-integration-tests

VeraChristina: You mean change integration-tests in line 8 to nightly-integration-tests?
The overall workflow has the name nightly-ci-integration-tests

theissenhelen: No, I meant that you actually assign it a name, like we do in the other workflows.
https://github.com/ecmwf/anemoi-core/blob/main/.github/workflows/label-public-pr.yml#L1C1-L7C1

VeraChristina: That's why I am confused. It looks to me like it has a name?

VeraChristina: I think it does not show up in the github ui because of the missing reference so that github can't parse the file?

theissenhelen: Oh, apologies, you are right. But for some reason it does not show up under this name here https://github.com/ecmwf/anemoi-core/actions/workflows/integration-tests.yml

VeraChristina: Yes, I also noticed that and I think it might be related to my mistake in the syntax, so hopefully also fixed by this fix

theissenhelen (approved):

PR #204, opened 2025-03-19
fix: changed profiler throughput calculation

cathalobrien:
## Description

This PR changes how the profiler calculates throughput, to make it easier to identify when your model is dataloader bound.

1. avg_training_throughput includes both dataloading and compute. It is based on the average time for an epoch divided by the number iterations in an epoch
2. avg_training_compute_throughput is just the compute happening on device. It is the average time for what pytorch lightning marks as "run_training_batch"

We dont look at the progress bar to get training throughput, since it was unclear where the numbers came form. I left the progress 

The upshot of this change can be seen below. The original run is dataloader bound. Notice how compute_throughput is much higher than dataloader_throughput. As a result overall throughput is low.
```
avg_training_throughput  0.06149 
avg_training_compute_throughput  0.46934 
avg_training_dataloader_throughput    0.07403 
```
In the next example I am no longer dataset-limited. My dataloader throughput is much higher and no longer drags down my overall throughput. As a result overall_throughput is now bound by computation on the GPU, as seen by the similar numbers for overall and compute throughput
```
avg_training_throughput 0.51719 
avg_training_compute_throughput 0.55784 
avg_training_dataloader_throughput    7.90282
```

## Type of Change

-   [X] Bug fix (non-breaking change which fixes an issue)
-   [ ] Documentation update

## Code Compatibility

-   [X] I have performed a self-review of my code

-   [X] I have tested the changes on a single GPU
-   [X] I have tested the changes on multiple GPUs / multi-node setups
-   [X] I have run the Benchmark Profiler against the old version of the code

### Dependencies

-   [X] I have ensured that the code is still pip-installable after the changes and runs

### Documentation

-   [X] My code follows the style guidelines of this project
-   [X] I have added comments to my code, particularly in hard-to-understand areas


training/src/anemoi/training/diagnostics/profilers.py
```diff
@@ -532,28 +532,33 @@ def get_speed_profiler_df(self, progressbar: _tqdm) -> pd.DataFrame:
...
-        training_rates_array = np.array(progressbar.training_rates)
-        speed_metrics["training_avg_throughput"] = training_rates_array.mean()
-        speed_metrics["training_avg_throughput_per_sample"] = training_rates_array.mean() / batch_size_tr
-
-        validation_rates_array = np.array(progressbar.validation_rates)
-        speed_metrics["validation_avg_throughput"] = validation_rates_array.mean()
-        speed_metrics["validation_avg_throughput_per_sample"] = validation_rates_array.mean() / batch_size_val
-
-        # Calculate per_sample metrics
+        num_iters_per_epoch = len(self.time_profiler.recorded_durations["run_training_batch"]) / len(
```
floriankrb, line 534: everything here is asking `self.time_profiler.recorded_durations` to do something.
Should this code be moved to the object in `self.time_profiler`? 

This would allow, for instance, to have another object for self.time_profiler.
Or to disable time profiling, we would have a dummy self.time_profiler.

floriankrb: After looking more closely, I think this code actually fails when self.time_profiler is the DummyProfiler.

training/src/anemoi/training/diagnostics/profilers.py
```diff
@@ -532,28 +532,33 @@ def get_speed_profiler_df(self, progressbar: _tqdm) -> pd.DataFrame:
...
-            speed_metrics["avg_training_dataloader_throughput"] / batch_size_tr
-        )
 
-        speed_metrics["avg_validation_dataloader_throughput"] = (
-            1 / np.array(self.time_profiler.recorded_durations["[_EvaluationLoop].val_next"]).mean()
-        )
-        speed_metrics["avg_validation_dataloader_throughput_per_sample"] = (
-            speed_metrics["avg_validation_dataloader_throughput"] / batch_size_val
-        )
+        if batch_size_tr > 1:
```
anaprietonem, line 550: Do we need an if statement? if batch_size_tr is 1 the division won't affect the results right?

training/src/anemoi/training/diagnostics/profilers.py
```diff
@@ -532,28 +532,33 @@ def get_speed_profiler_df(self, progressbar: _tqdm) -> pd.DataFrame:
...
+            speed_metrics["avg_training_throughput_per_sample"] = average_throughput / batch_size_tr
+            speed_metrics["avg_training_compute_throughput_per_sample"] = average_compute_throughput / batch_size_tr
+            speed_metrics["avg_training_dataloader_throughput_per_sample"] = (
+                speed_metrics["avg_training_dataloader_throughput"] / batch_size_tr
+            )
+
+        validation_rates_array = np.array(progressbar.validation_rates)
```
anaprietonem, line 557: So we would be having a different definition of the throughput at training and validation? I like you suggestion or not relaying on the progress bar to measure this and rather derive the statistics from the time profilers but I would propose then to be consisten in training and validation time

training/src/anemoi/training/diagnostics/profilers.py
anaprietonem: If the speed profiler is not dependant on the time profiler we could add an assert that the speed profiler shouldn't run if the time profiler is not enabled

training/src/anemoi/training/diagnostics/profilers.py
```diff
@@ -532,28 +532,33 @@ def get_speed_profiler_df(self, progressbar: _tqdm) -> pd.DataFrame:
...
-        speed_metrics["avg_validation_dataloader_throughput"] = (
-            1 / np.array(self.time_profiler.recorded_durations["[_EvaluationLoop].val_next"]).mean()
-        )
-        speed_metrics["avg_validation_dataloader_throughput_per_sample"] = (
-            speed_metrics["avg_validation_dataloader_throughput"] / batch_size_val
-        )
+        if batch_size_tr > 1:
+            speed_metrics["avg_training_throughput_per_sample"] = average_throughput / batch_size_tr
+            speed_metrics["avg_training_compute_throughput_per_sample"] = average_compute_throughput / batch_size_tr
+            speed_metrics["avg_training_dataloader_throughput_per_sample"] = (
```
anaprietonem, line 553: Might be I am missing something, but if the definition of the dataloader throughput is not changing, then where is the difference coming from the number on the PR description?

anaprietonem (commented):
Good work @cathalobrien, thanks for PR! these new definitions of throughput would be more robust and remove dependency with the progress bar. I just left some comments for clarification and if we do these changes, we would also need to update the docs so they reflect the new speed profiler behaviour!

training/src/anemoi/training/diagnostics/profilers.py
cathalobrien: yeah, when batch size is 1 the results are the same. This if statement makes it so the per_sample results aren't printed when batch size is 1.

training/src/anemoi/training/diagnostics/profilers.py
cathalobrien: Yeah I agree it would be nice. But pytorch lightning doesn't have a premade timer for "run_validation_batch" like it does for training. I will try to add one to the anemoi source code

HCookie:
@cathalobrien Still wanting to merge?

PR #207, opened 2025-03-20
fix(graphs): edge attribute from source/target

JPXKQX:
## Description

This PR drops the dependency of `torch_geometric < 2.5` for `anemoi-graphs`.
## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/graphs/src/anemoi/graphs/edges/attributes.py
+++ b/graphs/src/anemoi/graphs/edges/attributes.py
@@ -139,10 +139,7 @@ class BaseEdgeAttributeFromNodeBuilder(BaseBooleanEdgeAttributeBuilder, ABC):
             raise AttributeError(f"{self.__class__.__name__} class must set 'node_idx' attribute.")
 
     def compute(self, x_i: torch.Tensor, x_j: torch.Tensor) -> torch.Tensor:
-        return (x_j, x_i)[self.node_idx][self.node_attr_name]
-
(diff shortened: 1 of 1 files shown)


graphs/src/anemoi/graphs/edges/attributes.py
```diff
@@ -139,10 +139,7 @@ def __init__(self, node_attr_name: str) -> None:
             raise AttributeError(f"{self.__class__.__name__} class must set 'node_idx' attribute.")
 
     def compute(self, x_i: torch.Tensor, x_j: torch.Tensor) -> torch.Tensor:
-        return (x_j, x_i)[self.node_idx][self.node_attr_name]
-
-    def forward(self, x: tuple[NodeStorage, NodeStorage], edge_index: Adj, size: Size = None) -> torch.Tensor:
-        return self.propagate(edge_index, x=x, size=size)
+        return (x_j, x_i)[self.node_idx]
```
HCookie, line 143: How does this relate to the dependency change?

JPXKQX: It is related to `MessagePassing.propagate`, which expects a tuple of tensors. In this edge attribute builder, we were passing a `dict[str, tensor]`, and the internal methods were setting this `x=(None, None)` as the hint was not "recognised".

HCookie (approved):

PR #211, opened 2025-03-21
make activation checkpointing configurable

ssmmnn11:
Effort to make activation checkpointing configurable. With more GPU memory, small models will be able to run without full checkpointing and therefore we can increase throughput. 

Also sets the processor default chunking values to more sane defaults.

### DIFF
--- a/training/src/anemoi/training/schemas/training.py
+++ b/training/src/anemoi/training/schemas/training.py
@@ -80,6 +80,17 @@ class LR(BaseModel):
     "Number of warm up iteration. Default to 1000."
 
 
+class ActivationCheckpointing(BaseModel):
+    """Activation checkpointing configuration."""
+
+    encoder: bool = Field(example=False)
+    "Enable activation checkpointing for the encoder."
+    decoder: bool = Field(example=False)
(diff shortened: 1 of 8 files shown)


training/src/anemoi/training/train/train.py
```diff
@@ -189,6 +189,16 @@ def model(self) -> GraphForecaster:
                 freeze_submodule_by_name(model, submodule_name)
                 LOGGER.info("%s frozen successfully.", submodule_name.upper())
 
+        if self.config.training.activation_checkpointing.encoder:
```
theissenhelen, line 243: I think we could get rid of this bit by using the model_config in the encoder_processor_decoder.

training/src/anemoi/training/schemas/training.py
```diff
@@ -111,6 +111,17 @@ class TargetForcing(BaseModel):
     "Use target time as a fraction between input boundary times as input."
 
 
+class ActivationCheckpointing(BaseModel):
```
theissenhelen, line 114: Even though it is for use in training, I wonder if makes sense to put that into the model configs instead. We would also avoid the duplication below.

PR #217, opened 2025-03-25
test: add integration tests for stretched and lam use cases

VeraChristina:
## Description
- Add two integration tests for small training cycles using
   - a stretched grid
   - a local area model
- Add tests for config validation based on the same configs
- Update file names for the global use case for consistency -- i.e. using the name of the template used (config.yaml)
- Download the test data at beginning of test: This is faster and significantly reduces test time on some platforms (e.g. on github runners ~12 minutes vs almost two hours). Test run here: https://github.com/ecmwf/anemoi-core/actions/runs/14399980675

## Code Compatibility

-   [x] I have run the new integration tests locally, and they pass

### Documentation

----
📚 Documentation preview 📚: https://anemoi-training--217.org.readthedocs.build/en/217/

----
📚 Documentation preview 📚: https://anemoi-graphs--217.org.readthedocs.build/en/217/

----
📚 Documentation preview 📚: https://anemoi-models--217.org.readthedocs.build/en/217/


VeraChristina:
Tagging @anaprietonem because we wanted to revisit the file naming discussed on the first integration tests PR

VeraChristina:
marking as draft while adding the lam use case to this PR as well

theissenhelen (dismissed):
Happy with these changes!

training/tests/integration/config/test_lam.yaml
```diff
@@ -0,0 +1,52 @@
+# Modifications for the local area model template "lam.yaml"
+dataloader:
+  dataset:
+    cutout:
+      - dataset: ${hardware.paths.data}/${hardware.files.dataset}
+        thinning: 25 # ???-value in lam template
```
anaprietonem, line 6: Is that a general value ? @JPXKQX views on this?

training/tests/integration/config/test_lam.yaml
```diff
@@ -0,0 +1,52 @@
...
+  validation:
+    start: "2017-01-08"
+    end: "2017-01-10"
+
+hardware:
+  paths:
+    data: https://object-store.os-api.cci1.ecmwf.int/ml-tests/test-data/samples/anemoi-integration-tests
+  files:
+    dataset: regional-use-cases/cerra-rr-an-oper-0001-mars-5p5km-2017-2017-6h-v2-hmsi-testing.zarr
+    forcing_dataset: regional-use-cases/aifs-ea-an-oper-0001-mars-o96-2017-2017-6h-v8-testing.zarr
```
anaprietonem: For the forcing dataset, how long does this train currently takes? I believe for the global use case we were using a o48 dataset so wonder if that could make things faster

training/docs/contributing.rst
```diff
@@ -278,12 +278,12 @@ remain reliable and maintainable. Our approach includes:
...
 For an example, see `training/tests/integration/test_training_cycle.py`.
 The test uses a configuration based on the template
-`training/src/anemoi/training/config/basic.py`, i.e. the basic global
+`training/src/anemoi/training/config/config.yaml`, i.e. the basic global
 model. It applies testing-specific modifications to reduce batch_size
 etc. as detailed in
-`training/tests/integration/test_training_cycle.yaml`. It furthermore
-applies use-case-specific modifications as detailed in
-`training/tests/integration/test_basic.yaml` to provide the location of
+`training/tests/integration/config/testing_modifications.yaml`. It
```
anaprietonem, line 289: Regarding the naming:
- the 'test_config.yaml' is the one related to the global deterministic model right? Could we more meaningful to call it `test_global.yaml`. 
- Related to this, in the future we would have uses cases related to probabilistic model or tests for the temporal interpolator. Any view on how to scale the naming conventions for those? Just wondering it could be good to do it already now with test_[coverage]_[model_task]_[forecaster_class]. Where coverage would be global/lam/stretched grid, model_task could be forecaster/interpolator and class would refer to single or probabilistic

anaprietonem (commented):
Great work both! I left some comments regarding views on scaling the naming convention and pinned Mario as he might has some views on the settings for LAM & stretched grid

training/tests/integration/conftest.py
```diff
@@ -35,20 +43,32 @@ def set_working_directory() -> None:
...
     overrides = request.param
-    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="test_basic"):
+    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="test_config"):
         template = compose(
-            config_name="debug",
+            config_name="config",
             overrides=overrides,
         )  # apply architecture overrides to template since they override a default
-        use_case_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/test_basic.yaml")
+        use_case_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/test_config.yaml")
```
MeraX: Is `Path.cwd()` relative to the working directory when executing `pytest`?

training/tests/integration/conftest.py
```diff
@@ -35,20 +43,32 @@ def set_working_directory() -> None:
...
         return cfg
 
 
 @pytest.fixture
-def testing_modifications_with_temp_dir(tmp_path: Path) -> OmegaConf:
-    testing_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/testing_modifications.yaml")
-    temp_dir = str(tmp_path)
-    testing_modifications.hardware.paths.output = temp_dir
-    return testing_modifications
+def stretched_config(testing_modifications_with_temp_dir: OmegaConf) -> None:
```
MeraX: Is the return type correct?

training/tests/integration/config/test_lam.yaml
JPXKQX: This value is used to reduce the resolution of regional datasets. If we are using a smaller or coarser dataset, we can drop this parameter or use `thinning: 5`

training/docs/contributing.rst
VeraChristina: - I agree that test_global.yaml would be more explicit.
- I generally think that the naming of tests and test configs should follow the naming conventions of the code and the config templates provided. (Is there a reason to not rename config.yaml to global.yaml in `training/src/anemoi/training/config/`?)

training/tests/integration/config/test_lam.yaml
VeraChristina: Good point! @dietervdb-meteo mentioned another reason to use an o48 forcing dataset, namely to test if different grid resolutions for the dataset and the forcing_dataset lead to problems when building the graph. 

Currently, the bigger bottleneck for how long the tests take seems to depend heavily on data access from different set-ups. (It takes ~3mins on atos, and >30mins on the github runner.) I am looking into using the functionality in utils to make this faster (and into replacing the cerra dataset with a coarser one). Will look into o48-vs-o96 once the test times are more consistent across different set-ups

training/tests/integration/conftest.py
VeraChristina: Good spot! Will correct this

training/tests/integration/conftest.py
VeraChristina: Yes, it is. Hydra takes a relative path as well. I have added the fixture `set_working_directory()` that is run at the beginning of each test to navigate to the root of the repo in order to avoid problems with the "relative set-up"

VeraChristina:
> Thank you for this inspiring PR. I have to minor questions, posted individually, and I'm wondering whether this test is supposed to run on a CPU?

Thanks, Marek!
As already discussed last week (so just for the record), the tests are currently running on CPU but we are planning to move to runners with GPU. The long test time on github runners seems to be related to how we fetch data. (Running locally, both added training cycle tests take ~3 mins). Currently looking into this, will update later this week

training/tests/integration/conftest.py
```diff
@@ -27,28 +30,103 @@ def set_working_directory() -> None:
...
+def _download_datasets(config: OmegaConf, forcing_dataset: bool = False) -> list[str]:
+    url_dataset = config.hardware.files.dataset + ".tgz"
+    name_dataset = Path(config.hardware.files.dataset).name
+    tmp_path_dataset = get_test_archive(url_dataset)
+
+    if forcing_dataset:
+        url_forcing_dataset = config.hardware.files.forcing_dataset + ".tgz"
+        name_forcing_dataset = Path(config.hardware.files.forcing_dataset).name
+        tmp_path_forcing_dataset = get_test_archive(url_forcing_dataset)
+
```
theissenhelen: Here we might need to think about a more general version as in more than just dataset and forcing_dataset. For ocean or waves for instance we might have additional datasets that would not get captured here.

training/tests/integration/test_training_cycle.py
```diff
@@ -24,10 +25,77 @@
...
+
+
+@pytest.mark.longtests
+def test_training_cycle_stretched(stretched_config: Callable[[bool], DictConfig]) -> None:
+    config = stretched_config(download_data=True)
+    AnemoiTrainer(config).train()
+    shutil.rmtree(config.hardware.paths.data)
+
+
+def test_config_validation_stretched(stretched_config: Callable[[bool], DictConfig]) -> None:
```
theissenhelen: It could makes sense to have a separate file for testing the config validation.

VeraChristina: Do you mean a separate fixture based on the same files?

training/tests/integration/conftest.py
VeraChristina: I have updated this to take a list of datasets to be downloaded (i.e. dict attributes of hardware.paths.data) and to return a temporary dir and a list of relative paths for the datasets (such that we can update the config accordingly)

training/tests/integration/test_training_cycle.py
VeraChristina: I have separated this out into two fixtures, one that downloads the dataset and one that doesn't (for each use case) -- as discussed, for readability

training/tests/integration/config/test_lam.yaml
VeraChristina: - I have updated the tests to download the test data at the beginning of the test (instead of during training). This seems to solve the github runner issue (20 mins for entire workflow) and generally reduces test time
- To do this, I reduced the size of the regional dataset (by using less parameters, not by reducing grid resolution as that caused issues with the graph creation)

Locally, the four training cycle tests together now take 5 mins. So I don't think we need to reduce grid resolution at this point.

training/docs/contributing.rst
anaprietonem: - Probably legacy, I think changing config to 'global.yaml' and debug to 'debug_globa.yaml'
Screenshot 2025-04-09 at 11 56 20 would be a good idea

theissenhelen: I think once we have decided on naming for the configs, we can adjust the ones in the test folder accordingly.

training/tests/integration/conftest.py
```diff
@@ -27,28 +29,133 @@ def set_working_directory() -> None:
     os.chdir(repo_root)
 
 
+@pytest.fixture
+def testing_modifications_with_temp_dir(tmp_path: Path) -> OmegaConf:
+    testing_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/testing_modifications.yaml")
+    temp_dir = str(tmp_path)
+    testing_modifications.hardware.paths.output = temp_dir
```
theissenhelen: I think this could be one line

training/docs/contributing.rst
VeraChristina: Thanks, Ana, for the additional context.
I agree with Helen that we can update the naming of test, once we have decided on naming conventions for configs in src (and updated those).  I think the latter is a discussion to be had in a bigger group, and out of scope for this PR.
Helen has already put up [an issue](https://github.com/ecmwf/anemoi-core/issues/243) to align naming conventions for configs.

training/tests/integration/config/test_lam.yaml
VeraChristina: Given that the test time is okay, I'd suggest to address this separately in a follow-up PR. I will add this point to a follow-up issue about the graph for the stretched grid that Mario and I want to look into.

anaprietonem (approved):
The PR LGTM

training/tests/integration/conftest.py
```diff
@@ -27,28 +29,133 @@ def set_working_directory() -> None:
...
+
+
+def _download_datasets(config: OmegaConf, list_datasets: list[str]) -> tuple[str, list[str]]:
+    tmp_paths = []
+    dataset_names = []
+
+    for dataset in list_datasets:
+        url_dataset = config.hardware.files[dataset] + ".tgz"
+        name_dataset = Path(config.hardware.files[dataset]).name
+        tmp_path_dataset = get_test_archive(url_dataset)
```
MeraX, line 152: I was wondering if you had considered using a [zarr zip store](https://zarr.readthedocs.io/en/stable/user-guide/storage.html#zip-store) to store the whole dataset in a single file. Anemoi-datasets supports zipped zarr directly which might reduce the number of required workarounds.

training/tests/integration/test_training_cycle.py
```diff
@@ -16,18 +16,42 @@
...
 
 
+@skip_if_offline
 @pytest.mark.longtests
-def test_training_cycle_architecture_configs(architecture_config: DictConfig) -> None:
-    AnemoiTrainer(architecture_config).train()
-    shutil.rmtree(architecture_config.hardware.paths.output)
+def test_training_cycle_architecture_configs(architecture_config_with_data: DictConfig) -> None:
+    AnemoiTrainer(architecture_config_with_data).train()
+    shutil.rmtree(architecture_config_with_data.hardware.paths.data)
```
MeraX: I have a bad vibe about `shutil.rmtree`. It could cause serious issues if something screws up the path. Also, if AnemoiTrainer fails, there is no cleanup. Thus I would recommend using a python context.

Perhaps a `yield` fixtures in pytest could help here. https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization

MeraX: E.g.
```
@pytest.fixture
def testing_modifications_with_temp_dir(tmp_path: Path) -> OmegaConf:
    testing_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/testing_modifications.yaml")
    testing_modifications.hardware.paths.output = str(tmp_path)

    with tempfile.TemporaryDirectory() as output_dir:
        testing_modifications.hardware.paths.output = output_dir
        yield testing_modifications
```

Perhaps such config transformation could also be made available as general building block for tests... but that's perhaps something for a later round of factoring.

training/tests/integration/test_training_cycle.py
```diff
@@ -16,18 +16,42 @@
...
 
 
 def test_config_validation_architecture_configs(architecture_config: DictConfig) -> None:
     BaseSchema(**architecture_config)
+
+
+@skip_if_offline
+@pytest.mark.longtests
```
MeraX, line 37: I'm wondering why those fixtures live in different scopes and if `anemoi.utils.testing.skip_slow_tests` is intended to replace `skip_if_offline`...

MeraX (changes requested):
Oh and apologies for being a nuisance, but the minimum `anemoi-utils` version for this branch does not yet support the testing utilities. See https://github.com/ecmwf/anemoi-utils/tree/0.4.13/src/anemoi/utils

training/tests/integration/test_training_cycle.py
MeraX: To be honest, I don't fully understand the concept of `skip_if_offline`, but I've just noted the presence of the `data_dependent` tag and was wondering if they are different. Perhaps nothing to solve in this PR, but maybe worth a consolidation issue...

VeraChristina: On the first point, good spot, thanks! I'll update this when I know which version of utils we need -- we might need a small fix in the current get_test_archive (to get rid of the shutil.rmtree workaround)

On the second point, the two set-ups (in utils and in integration tests for training) evolved simultaneously. Planning to align/consolidate this separately. I'll put up an issue such that it is tracked

training/tests/integration/test_training_cycle.py
VeraChristina: In principle, the clean up is handled in the set-up in utils (unless the test crashes or is manually aborted). This was my workaround for an issue in the downloading utilities, but I agree it should be removed -- will update when I have a less temporary fix

training/tests/integration/conftest.py
VeraChristina: Thanks for the suggestion! Looking into this

training/tests/integration/test_training_cycle.py
VeraChristina: I have removed the local clean up. The clean up is handled by the testing utilities in anemoi-utils. We can revisit the second point (making config transformation available as building blocks) as part of the [consolidation ](https://github.com/ecmwf/anemoi-core/issues/257) of testing set-ups

training/tests/integration/conftest.py
VeraChristina: My first impression is that it's non-trivial to get this to work. Happy to revisit this!

training/tests/integration/test_training_cycle.py
VeraChristina: Have updated minimum required anemoi-utils version. Here's the issue about consolidating testing set-ups: https://github.com/ecmwf/anemoi-core/issues/257

training/tests/integration/conftest.py
MeraX: This is how I packed a zarr (not using the `zarr` package)
```
D="$PWD"
cd my.zarr/
zip -0 -r "$D/my.zarr.zip" .
cd "$D"
```

VeraChristina: This looks close to what I tried, but not exactly the same. I'll give it a try, thanks!

theissenhelen (approved):

PR #219, opened 2025-03-26
docs(graphs): update graph docs to template (docathon)

VeraChristina:
## Description

- Update the graph documentation to the anemoi template following the docathon meta discussion.
- Replace the contributing guidelines with reference to main contributing guidelines.
- Fix link to installing section in anemoi training docs to point to installing section in graphs instead.

----
📚 Documentation preview 📚: https://anemoi-graphs--219.org.readthedocs.build/en/219/


aaron-hopkinson (changes requested):
Thanks for taking this on. Looks good to me (aside from one minor concern re: "command line usage" vs "tool")

Another minor suggestion, could you fix the broken link on L15-16 of `index.rst` so that it points to the index of the main docs page with intersphinx, please?

I think this should be:
```
of the packages within the :ref:`anemoi framework <anemoi-docs:index>`.
```

(Can't do an inline suggestion because it's too far away from a change...)

graphs/docs/cli/introduction.rst
```diff
@@ -1,7 +1,7 @@
 .. _cli-introduction:
 
 ##################
-Command line tool
+Command line usage
```
aaron-hopkinson: It's `tool` in inference (side bar) and datasets (side bar and main page). Is it worth renaming here? (I can see an argument that this covers general usage, and the individual commands/interfaces are covered below...)

Just thinking about consistency vs effort :)

VeraChristina: Good point. Will revert to stay consistent with other packages

aaron-hopkinson (approved):
Looks good - thank you!

PR #236, opened 2025-04-08
ci: hpc gpu integration test

theissenhelen:
## Description

This PR adds a workflow to run integration tests on a gpu on the HPC. It runs nightly at 22 pm. 
## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Documentation


.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  schedule:
+    - cron: "0 23 * * *"  # every day at 22pm on default(main) branch
+  workflow_dispatch:
+
+ # Trigger after public PR approved for CI
+  pull_request_target:
+    types: [labeled]
+
+jobs:
```
VeraChristina, line 10: Do we want to add a job that checks if the latest commit was in the past 24h, as we do in the action run in the current nightly workflow (compare https://github.com/ecmwf/reusable-workflows/blob/main/nightly-pytest/action.yml)?

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,40 @@
...
+
+ # Controls when the workflow will run
+on:
+  schedule:
+    - cron: "0 23 * * *"  # every day at 22pm on default(main) branch
+  workflow_dispatch:
+
+ # Trigger after public PR approved for CI
+  pull_request_target:
+    types: [labeled]
```
VeraChristina: I am not familiar with this set-up. Is it to be able to run the test on PRs coming in from forked repos? Do we then need to check out the forked repo to run the test on that branch?

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,40 @@
...
+    if: ${{ !github.event.pull_request.head.repo.fork && github.event.action != 'labeled' || github.event.label.name == 'approved-for-ci' }}
+    steps:
+      - uses: ecmwf-actions/reusable-workflows/ci-hpc-generic@v2
+        with:
+          template: |
+            set -eux
+            module load python3
+            REPO_NAME=anemoi-core
+            REPO_BRANCH=${{ github.head_ref || github.ref_name }}
+            git clone -b $REPO_BRANCH https://${{ secrets.GH_REPO_READ_TOKEN }}@github.com/ecmwf/${REPO_NAME}.git
```
VeraChristina, line 38: Related to the comment above. As far as I can tell this requires the branch to exist on our repo, which it would not in case the PR comes from a fork? Maybe I just misunderstand the pull_request_target set-up, though.

VeraChristina (commented):
This looks good to me, many thanks! 

Just two small points raised in the comments:
- Do we want to add a check for the latest commit as in the current nightly workflow?
- Small questions about the pull_request_target set-up that I am not familiar with

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,40 @@
+name: nighty-integration-tests-hpc-gpu
```
VeraChristina: very minor, typo: nightly instead of nighty

.github/workflows/integration-tests-hpc.yml
aaron-hopkinson, line 38: I think you could speed up/reduce the footprint of the checked out code by adding the `--depth 1` flag, e.g.
```
git clone --depth 1 -b $REPO_BRANCH https://${{ secrets.GH_REPO_READ_TOKEN }}@github.com/ecmwf/${REPO_NAME}.git
```

.github/workflows/integration-tests-hpc.yml
theissenhelen: We want to ensure that external code does not automatically run on our infrastructure. For now this worklow runs nightly, but this will likely change at some point in the future. In case someone changes this workflow accordingly, this will check whether a PR stems from an external contributor and whether it has been approved for ci.

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,40 @@
...
+
+ # Controls when the workflow will run
+on:
+  schedule:
+    - cron: "0 23 * * *"  # every day at 22pm on default(main) branch
```
aaron-hopkinson: 22 or 23 here? (Comment different from cron spec)

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,40 @@
...
+            pip install -e ./training[all,tests]
+            pip install -e ./models[all,tests]
+            pip install -e ./graphs[all,tests]
+            python3 -m pytest training/tests/integration --longtests
+            deactivate
+            rm -rf $REPO_NAME
+          sbatch_options: |
+            #SBATCH --job-name=anemoi_core_pytest
+            #SBATCH --time=00:10:00
+            #SBATCH --qos=nf
```
aaron-hopkinson: PR title says GPU tests - should this be `ng` rather than `nf`?

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,40 @@
...
+  workflow_dispatch:
+
+ # Trigger after public PR approved for CI
+  pull_request_target:
+    types: [labeled]
+
+jobs:
+  pytest:
+    runs-on: hpc
+    if: ${{ !github.event.pull_request.head.repo.fork && github.event.action != 'labeled' || github.event.label.name == 'approved-for-ci' }}
```
aaron-hopkinson, line 16: Relating to @VeraChristina's question about forks, my reading of this is that we don't want to run these tests on PRs coming from forks (regardless of how they're labelled). Is that right? If so, it makes sense to me 😄

.github/workflows/integration-tests-hpc.yml
theissenhelen: absolutely right

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,55 @@
...
+            REPO_NAME=anemoi-core
+            REPO_BRANCH=${{ github.head_ref || github.ref_name }}
+            git clone --depth 1 -b $REPO_BRANCH https://${{ secrets.GH_REPO_READ_TOKEN }}@github.com/ecmwf/${REPO_NAME}.git
+            cd $REPO_NAME
+            python3 -m venv .venv
+            source .venv/bin/activate
+            pip install --upgrade pip
+            pip install -e ./training[all,tests]
+            pip install -e ./models[all,tests]
+            pip install -e ./graphs[all,tests]
```
gmertes: How would you feel about using `uv` here? 😀 For automated builds, I've been using uv to greatly speed up the pip install and it's been working great. Even with the copy mode I'm seeing up to 10x reduction in installation time on Atos. Installed in the venv so you don't need a system installed uv.

You still use the regular `python3 -m venv` to create the venv, so it's also still using the python from the module, but to install the packages you do:

```
pip install uv
uv pip install -e ./training[all,tests] -e ./models[all,tests] -e ./graphs[all,tests]
```

I would also suggest to do the pip install in 1 line like the above, so that the dependency resolver can load all dependencies at the same time. I believe this will give a more meaningful error in case of a dependency conflict - it should be clearer which source package is responsible for the conflict.

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,55 @@
...
+  integration-tests:
+    runs-on: hpc
+    needs: check_date
+    if: ${{ needs.check_date.outputs.should_run != 'false' && (!github.event.pull_request.head.repo.fork && github.event.action != 'labeled' || github.event.label.name == 'approved-for-ci') }}
+    steps:
+      - uses: ecmwf-actions/reusable-workflows/ci-hpc-generic@v2
+        with:
+          template: |
+            set -eux
+            module load python3
```
gmertes: Should we pin the version here, e.g. `python3/3.12.9-01` ?

.github/workflows/integration-tests-hpc.yml
```diff
@@ -0,0 +1,55 @@
...
+            pip install --upgrade pip
+            pip install -e ./training[all,tests]
+            pip install -e ./models[all,tests]
+            pip install -e ./graphs[all,tests]
+            python3 -m pytest training/tests/integration --longtests
+            deactivate
+            rm -rf $REPO_NAME
+          sbatch_options: |
+            #SBATCH --job-name=anemoi_core_pytest
+            #SBATCH --time=00:10:00
```
gmertes: I get that the tests should finish within 10 min, but should we bump this to have a bit of buffer to account for any system-slowness (slow disks, slow network etc). 30 min?

.github/workflows/integration-tests-hpc.yml
theissenhelen: Very happy do to that instead , but then we should do the same in the other workflows to be consistent..

.github/workflows/integration-tests-hpc.yml
theissenhelen: Yes, I agree. Especially considering that once we merge the LAM and stretched-grid use case we'll need more than 10 mins.

.github/workflows/integration-tests-hpc.yml
theissenhelen: yes, we should probably do that.

.github/workflows/integration-tests-hpc.yml
MeraX: And regardless of uv, i would suggest considering installing all core packages in one pip call.

.github/workflows/integration-tests-hpc.yml
MeraX: If I may add my experience from ICON: Nothing is more frustrating than a test that is killed just because of the time limit. Because the only option you are usually left with is restarting the whole job which and waiting even longer.

Nevertheless, it is of course crucial to keep an eye on the average test execution time.

gmertes: 1 hour?

theissenhelen: I have put it to 1 hr.

anaprietonem (dismissed):
Great work Helen!

anaprietonem (approved):

PR #238, opened 2025-04-08
feat: update config and doc to allow using dataset names along with path

floriankrb:
## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/training/src/anemoi/training/config/dataloader/native_grid.yaml
+++ b/training/src/anemoi/training/config/dataloader/native_grid.yaml
@@ -50,14 +50,24 @@ grid_indices:
 # See https://anemoi-datasets.readthedocs.io
 # ============
 
-dataset: ${hardware.paths.data}/${hardware.files.dataset}
+dataset:
+
+  dataset: ???
+  # Either a path to a dataset, a dataset name or a nested dictionary
+  # when using dataset names, make sure the search path is set in the anemoi-datasets config
+  # as described in the anemoi-datasets documentation
+  # https://anemoi.readthedocs.io/projects/datasets/en/latest/datasets/using/configuration.html
+
+  frequency: ${data.frequency}
+  drop: []
+  # start: ...
+  # end: ...
+  # select: []
 
 training:
   dataset: ${dataloader.dataset}
   start: null
   end: 2020
-  frequency: ${data.frequency}
-  drop:  []
 
 validation_rollout: 1 # number of rollouts to use for validation, must be equal or greater than rollout expected by callbacks
 
@@ -65,12 +75,8 @@ validation:
   dataset: ${dataloader.dataset}
   start: 2021
   end: 2021
-  frequency: ${data.frequency}
-  drop:  []
 
 test:
   dataset: ${dataloader.dataset}
   start: 2022
   end: null
-  frequency: ${data.frequency}
-  drop:  []


HCookie:
@floriankrb Is this a PR you are still looking to merge?

training/tests/integration/config/test_config.yaml
```diff
@@ -1,12 +1,9 @@
 # Modifications for the basic template "config.yaml"
-hardware:
-  files:
-    dataset: anemoi-integration-tests/training/datasets/aifs-ea-an-oper-0001-mars-o48-1979-19-6h-v6-testset.zarr
-  paths:
-    data: https://object-store.os-api.cci1.ecmwf.int/ml-tests/test-data/samples
```
MeraX, line 6: Where would this path be set in the future?

training/tests/integration/config/test_diffusion.yaml
```diff
@@ -1,15 +1,16 @@
...
   paths:
-    data: https://object-store.os-api.cci1.ecmwf.int/ml-tests/test-data/samples
     truncation: null
   files:
-    dataset: anemoi-integration-tests/training/datasets/aifs-ea-an-oper-0001-mars-o96-2017-2017-6h-v8-testing.zarr
     truncation: null
     truncation_inv: null
     graph: null
 
+
```
MeraX, line 10:
```suggestion
```

training/docs/user-guide/hydra-intro.rst
```diff
@@ -72,19 +72,17 @@ added to the training command like so:
...
 -  ``hardware.paths.output``: Location of output directory
 
--  ``hardware.files.dataset``: Filename(s) of datasets used for training
-
 -  ``hardware.files.graph``: If you have pre-computed a specific graph,
    specify its filename here. Otherwise, a new graph will be constructed
    on the fly and written to the filename given.
 
+-  ``dataloader.data``: Description of the dataset, as provided to
+   `anemoi.datasets.open_dataset`
```
MeraX, line 84: Would it be possible to reference to the respective documentation here?

And is it actually `dataloader.data` or `dataloader.dataset_defaults`?

training/src/anemoi/training/config/dataloader/native_grid.yaml
```diff
@@ -50,27 +50,33 @@ grid_indices:
...
+  drop: []
+  # start: ...
+  # end: ...
+  # select: []
 
 training:
-  dataset: ${dataloader.dataset}
+  dataset: ${dataloader.dataset_defaults}
   start: null
   end: 2020
```
MeraX, line 70: What is the relation of `start`, `end` , etc. specified here to the values specified in `dataloader.dataset_defaults`?

Do settings here (like `end=2020`) selectively overwrite settings made in the respective `dataset`?

MeraX (commented):
I'm slightly confused about the content of this PR. I neither see any change in the Python code to reflect the proposed feature (not even in Pydantic) nor in the dependencies, in case this feature is fully implemented elsewhere.

training/tests/integration/config/test_config.yaml
anaprietonem: This path is not really needed in core in any case, as the functionality to get the test data lives in anemoi-utils and there you can see this path is defined https://github.com/ecmwf/anemoi-utils/blob/da380be71d78274d72bd0a3859ef00b1c80e9469/src/anemoi/utils/testing.py#L25

training/docs/user-guide/hydra-intro.rst
anaprietonem: Yes I am still working based on the feedback from ATS regarding wanting to keep the hardware paths and files for datasets.  From a chat with Florian, we might want to rethink this further and keep the files and paths but making the path optional so if None is passed that's read from the `settings.toml`

Any case, I will update the docs as part of the changes, thanks for the nudge!

training/tests/integration/config/test_config.yaml
MeraX: Thanks for pointing this out. So this PR depends on an older update of anemoi-utils, which is probably already reflected in the dependencies.

Issue #240, opened 2025-04-09
[Improvement] Mlflow - run_id just known by rank0


anaprietonem:
**Is your feature request related to a problem? Please describe.**
Current mlflow logger implementation means that 'rank0' tasks will have access to run_id. This is 'imposed/given' by PTL MlflowLogger implementation (see https://lightning.ai/docs/pytorch/stable/_modules/lightning/pytorch/loggers/mlflow.html#MLFlowLogger):
```
    @property
    @rank_zero_experiment
    def experiment(self)
```
For now we have decided to not modify this and rather adapt our code to make it work. See 'def run_id' and 'def update_paths' in train.py. This fact translates into the following behaviour depending on the 'type of run' we submit.
<img width="1339" alt="Screenshot 2024-04-04 at 09 09 24" src="https://github.com/ecmwf-lab/aifs-mono/assets/91897203/677c1edb-9cd3-43ee-81bf-58b4524651cd">

This could lead to rank0 tasks having a diverging behaviour compared with the rest of the tasks0 and could be prone to bugs difficult to debug. Could be worth reviewing this design and thinking if there is a way to distribute/broadcast the run_id to all other tasks

anaprietonem:
Potential solution/idea - Use the strategy/torch distributed to broadcast the run id from rank 0 to all other ranks, and then update the paths. This solution works in multinode and multi-gpu cases, as well as if we run an srun interactive session and then execute 'aifs-train' or other commands (tested using srun -c 64 --mem=64G --partition=gpu --gpus-per-node=1^C-ntasks-per-node=1 -t 02:00:00 --pty bash) or simply if we use 1 gpu.

Screenshot 2024-07-23 at 11 30 35
Using the launcher to do the broadcasting and setting up the strategy environment ensures that the code works both in sbatch and interactive sessions, and also follow the way it's is done in pytorch lightning since to submit the training/validation and testing runs it uses:
        call._call_and_handle_interrupt(
            self, self._fit_impl, model, train_dataloaders, val_dataloaders, datamodule, ckpt_path
        )

<img width="761" alt="Screenshot 2024-07-23 at 11 30 35" src="https://github.com/user-attachments/assets/bd68f0ea-fc57-4ede-85ff-f6e797b7b34a">

See logs example with 1 full node - https://mlflow.copernicus-climate.eu/#/experiments/35/runs/48aeab1d25554842b4157c790510c8b2/artifacts
[...]

PR #164, opened 2025-02-28
feat: Add the possibility to train a model with a dry MLflow run ID

jjlk:
## Description
If several trainings are chained using a scheduler, a nice feature would be to be able to:
* Create a MLflow run ID using the API outside of the training CLI.
* Propagate this run_id through the different trainings and to external registries.
* Get the lineage in the MLflow server.

First bullet point is handled with a new CLI to generate a new run ID and to save it on disk and in the register:
```
>$ anemoi-training mlflow prepare --config-name=dev_mlflow
2025-02-28 16:24:06 INFO Access token refreshed: 0.2 seconds.
2025-02-28 16:24:06 INFO Creating new run_id: 00da1078efe4467b81f75587df3696ae
2025-02-28 16:24:06 INFO Saving mlflow_id in file in mlflow_run_id.yaml.
```

Second bullet point is achieved by:
* Introducing a new internal variable, called dry_run_id, to run a training without an existing checkpoint.
* Add a use case for the run ID setup to handle a case when we have a forked run_id + a run ID generated before.

Third bullet point is achieved by:
* Add a use case in the MLFlow logger to deal with the fact that we can have a run ID and a forked run ID.

### Standard training
Creation of a training run ID:
```
>$ anemoi-training mlflow prepare --config-name=dev_mlflow
2025-03-04 14:36:59 INFO Access token refreshed: 0.2 seconds.
2025-03-04 14:37:00 INFO Creating new run_id: 4d7f151592534836b0ca8507f2d78a21
2025-03-04 14:37:00 INFO Saving run id in file in mlflow_run_id.yaml. 
```
Training:
```
$ anemoi-training train --config-name=dev_mlflow hardware.paths.output=/perm/ecm6116/shared/anemoi/training/ training.run_id="4d7f151592534836b0ca8507f2d78a21"
[2025-03-04 14:39:49,139][anemoi.training.train.train][INFO] - Config validated.
[2025-03-04 14:39:49,139][anemoi.training.train.train][INFO] - Run id: 4d7f151592534836b0ca8507f2d78a21
[2025-03-04 14:39:49,154][anemoi.training.diagnostics.logger][INFO] - AnemoiMLFlow logging to https://mlflow.ecmwf.int/
[2025-03-04 14:39:49,160][anemoi.training.diagnostics.mlflow.logger][INFO] - MLflow token authentication enabled for https://mlflow.ecmwf.int/
...
[...]

PR #242, opened 2025-04-09
fix: Checkpoint path check for multiple tasks/GPUs training

jjlk:
## Description

Fix bug introduced in https://github.com/ecmwf/anemoi-core/pull/164 when running a training with multiple GPUs/tasks. This a quick fix and a more long term and robust solution might be investigated: https://github.com/ecmwf/anemoi-core/issues/240. 

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/training/src/anemoi/training/train/train.py
+++ b/training/src/anemoi/training/train/train.py
@@ -425,6 +425,8 @@ class AnemoiTrainer:
 
     def _get_dry_run_id(self) -> None:
         """Check if the run ID is dry, e.g. without a checkpoint."""
+        # The Path casting is needed because in some multiple-gpu case
+        # ranks > 0 checkpoint paths are Python strings.
         if Path(self.config.hardware.paths.checkpoints).is_dir():
             self.dry_run_id = False
         else:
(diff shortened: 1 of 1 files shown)


jjlk:
@anaprietonem @gmertes 
I initially went for the most straightforward solution using the os module but the pre-commit module did not accept it:
```
warning: Selection `CPY` has no effect because preview is not enabled.
training/src/anemoi/training/train/train.py:429:12: PTH112 `os.path.isdir()` should be replaced by `Path.is_dir()`
    |
427 |     def _get_dry_run_id(self) -> None:
428 |         """Check if the run ID is dry, e.g. without a checkpoint."""
429 |         if os.path.isdir(self.config.hardware.paths.checkpoints):
    |            ^^^^^^^^^^^^^ PTH112
430 |             self.dry_run_id = False
431 |         else:
```
What do you think?

gmertes:
You can use `Path(self.config.hardware.paths.checkpoints).is_dir()`

anaprietonem (approved):

PR #253, opened 2025-04-10
test: aicon metadata

MeraX:
## Description

This tests the presence of all metadata that DWD uses for inference and refactors the aicon tests to leverage pytest `yield` fixtures which improve the handling of temporary files.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

-   [na] I have tested the changes on a single GPU
-   [na] I have tested the changes on multiple GPUs / multi-node setups
-   [na] I have run the Benchmark Profiler against the old version of the code
-   [na] If the new feature introduces modifications at the config level, I have made sure to update Pydantic Schemas and default configs accordingly

### Dependencies

### Documentation

-   [na] I have updated the documentation and docstrings to reflect the changes


training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py
```diff
@@ -32,55 +34,82 @@
...
+    except KeyError:
+        keys = "".join(f"[{k!r}]" for k in keys)
+        raise KeyError("missing metadata" + keys) from None
 
 
-def test_config_validation_aicon() -> None:
-    BaseSchema(**aicon_config())
+@typechecked
+def test_config_validation_aicon(aicon_config_with_grid: DictConfig) -> None:
+    BaseSchema(**aicon_config_with_grid)
```
VeraChristina: Do you want to download the grid file for the config validation, or were you planning to use aicon_config_with_tmp_dir here?

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py
```diff
@@ -32,55 +34,82 @@
...
+        ("data_indices", "data", "input", "full"),
+        ("data_indices", "data", "output", "full"),
+        ("data_indices", "model", "input", "forcing"),
+        ("data_indices", "model", "input", "full"),
+        ("data_indices", "model", "input", "prognostic"),
+        ("data_indices", "model", "output", "full"),
+        ("dataset", "shape"),
+    )
+
+    assert torch.is_tensor(trainer.graph_data["data"].x), "data coordinates not present"
```
VeraChristina: I would have a preference to separate this part out -- just to be able to see with one glance whether it's the training cycle or the metadata test that failed. (Reminder for me to look into more verbose test output so that we actually see this immediately)
That would mean initialising the trainer twice (and also downloading the grid file) but I don't think it adds much test time compared to the training cycles.

VeraChristina (commented):
Looks good to me. Two small points in the comments below.

Just to clarify the aim here: The idea is to keep track of changes in the metadata interface such that you can adapt your inference code accordingly? I.e. more of an early notification (plus noticing when important information is lost, not just moved) rather than not expecting changes to the interface?

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py
MeraX: Thank you for pointing out this error. Indeed, the actual reason for splitting the fixture was to address this issue.

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py
MeraX: Thank your for this suggestion. I split the test. 

The metadata test actually does not need a trained model, so there is no need to train it twice. Thus there will je just the overhead of downloading the grid, which doesn't take long, and setting up the model, which is a slight overhead. None the less, the split makes the test results much more obvious.

In addition, I refactored the `assert_metadatakeys` to test for all keys and summarise all errors.

MeraX:
> Looks good to me. Two small points in the comments below.
> 
> Just to clarify the aim here: The idea is to keep track of changes in the metadata interface such that you can adapt your inference code accordingly? I.e. more of an early notification (plus noticing when important information is lost, not just moved) rather than not expecting changes to the interface?

The purpose is for early notification. Necessary changes of the interface are welcome. I added an according note to the code.

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py
MeraX: > (Reminder for me to look into more verbose test output so that we actually see this immediately) 

-> Perhaps like https://github.com/ecmwf/reusable-workflows/pull/19 ?

VeraChristina: Thanks a lot!

VeraChristina (approved):
Thanks for clarifying! The test run was also successful -- this looks good to me.

PR #254, opened 2025-04-10
chore: Define minimum versions for all dependencies.

### LINKED ISSUE #266: anemoi-graphs depends on anemoi-datasets with the extra "data" which does not exists
### What maintenance does this project need?

anemoi-graphs has the dependency "anemoi-datasets[data]>=0.3.3" with extra "data" although anemoi-datasets does not define any extra named "data".

### Organisation

Météo-France

MeraX:
## Description

Define minimum versions for all dependencies and update some minimal package versions are necessary to pass all tests with lowest package versions. Keeping the dependencies up to date helps when gradually updating Anemoi packages.

## Intent and Issue Number

Keep dependencies up to date.
Closes #266 

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation


models/pyproject.toml
```diff
@@ -46,7 +46,7 @@ dependencies = [
   "anemoi-utils>=0.1.9",
   "einops>=0.6.1",
   "hydra-core>=1.3",
-  "torch>=2.2",
```
mchantry, line 49: Is this increment based on testing?
Given that training requires models, would it make sense to update in the training toml?

MeraX: This increment is based on testing.  With torch 2.2, `AttributeError: module 'torch.nn' has no attribute 'attention'` is raised in multiple tests.

I'm not sure about the torch version required by `models`. But since `models` requires `training`, it inherits the `models` dependencies. I would leave it up to the software engineers to decide, whether `training` needs to declare its dependency on `torch` (and others) in this case at all.

tjhunter (commented):
thanks for tightening the bounds. My main high level comment: since your are updating the lower bounds, could you also put some sensible upper bounds? For example `scikit-learn>=1.5` -> `scikit-learn>=1.5,<2` . This will prevent future breakages.

models/pyproject.toml
mchantry: Let's leave it as this for now.

mchantry (dismissed):

mchantry (approved):
Many thanks, great contribution.

PR #256, opened 2025-04-11
feat(models): adding leaky boundings

pinnstorm:
## Description

Adding leaky bounding equivalents to `bounding.py` for use with model variables. Potential fix for AIFS Single "Rain Pox" issue. May lead to better back-propagation of gradients when adding bounding as a final layer in the model.

## Type of Change

-   [ ] Documentation update

## Issue Number

This issue was motivated by the "rain pox" seen in the AIFS Single in production where we get grid points predicting unphysical high rainfall in isolation from any other weather systems. Adding LeakyHardTanh bounding to soil moisture (instead of HardTanh) is a potential fix for this issue.

## Code Compatibility

### Dependencies

### Documentation


models/src/anemoi/models/layers/bounding.py
```diff
@@ -19,6 +19,35 @@
 from anemoi.models.data_indices.tensor import InputTensorIndex
 
 
+def leaky_hardtanh(
```
anaprietonem: Since this in an activation function could be worth to just have a `models/src/anemoi/models/layers/activations.py`?

models/src/anemoi/models/layers/bounding.py
```diff
@@ -227,6 +286,10 @@ class FractionBounding(HardtanhBounding):
     total_var : str
         A string representing a variable from which a secondary variable is derived. For
         example, in the case of convective precipitation (Cp), total_var = Tp (total precipitation).
+    statistics : dict, optional
```
anaprietonem, line 261: These are just cause they were missing from the docs right?

models/src/anemoi/models/layers/bounding.py
pinnstorm: Good idea! Have added that now 🙏

models/src/anemoi/models/layers/bounding.py
pinnstorm: Yes exactly just doing as I'm told by pre-commit 😅

PR #260, opened 2025-04-15
fix: dry run forking

anaprietonem:
## Description

With the current implementation of the dry_run, there was a bug introduced with forking by which the resume of the training checkpoint was skipped since  `Path(self.config.hardware.paths.checkpoints).is_dir()` was false for rank 0. This PR addresses the problem by:
- controlling the dry_run property with an mlflow tag so that if a run is created using `mflow prepare` command the dry_run tag is set
- For resuming a dry_run - the code checks the parent run and skip the loading of the checkpoint. 
- For forking a run and using a dry_run - training will still load a trained checkpoint from the forked run
- For resuming a resumed dry_run - training will still load a trained checkpoint from the forked run

## Type of Change

-   [X] Bug fix (non-breaking change which fixes an issue)
-   [ ] Documentation update

## Code Compatibility

-   [X] I have performed a self-review of my code

-   [X] I ran the [complete Pytest test](https://anemoi.readthedocs.io/projects/training/en/latest/dev/testing.html) suite locally, and they pass
-   [X] I have tested the changes on a single GPU
-   [X] I have tested the changes on multiple GPUs / multi-node setups

### Dependencies

-   [X] I have ensured that the code is still pip-installable after the changes and runs
-   [X] I have not introduced new dependencies in the inference portion of the pipeline

### Documentation

-   [X] My code follows the style guidelines of this project
-   [X] I have added comments to my code, particularly in hard-to-understand areas


anaprietonem:
with 1 node 2 gpus:
Whiteout prepml
  - new run
    - use that as run_id
    - use that as fork_run_id

With prepml
  1. mlflow prepare (run 2265)
    2. use that as run_id
    3. use that as run_id (resume)
    4.  use that as run_id (resume)
  2. mlflow prepare (runid 1234)
  3. use run_id 1234 and fork_run_id 2265

jjlk (approved):
Fine for me!

PR #262, opened 2025-04-15
fix: Small changes to make loss refactor work with CRPS training

jakob-schloer:
## Description

Fixes the crps training with the refactored losses and scalers. 

## Type of Change


training/src/anemoi/training/losses/kcrps.py
```diff
@@ -88,7 +88,7 @@ def forward(
 
         kcrps_ = einops.rearrange(kcrps_, "bs v latlon -> bs latlon v")
 
-        kcrps_ = self.scale(kcrps_, scalar_indices, without_scalars=without_scalars)
+        kcrps_ = self.scale(kcrps_.unsqueeze(1), scaler_indices, without_scalers=without_scalers)
```
jakob-schloer, line 91: The scalers currently assume that the loss always has 4 dimensions, i.e. [batch, ensembles, latlon, vars]. The crps computes a value over members and is thus only of dimension [batch, latlon, vars].
My changes here are a hack but I hope spark discussion on how to deal with it properly.

mc4117 (approved):

mc4117 (approved):

PR #268, opened 2025-04-23
Remove remapper

sahahner:
## Description

**Problem:**
For the remapping (for example mwd to sin(mwd) and cos(mwd)), we need to create space for the new columns where the variables are mapped to
Suboptimal solution at the moment:
- create a new batch
- this reallocates also the columns that are not touched
- we need memory for two batches! this does not necessarily cause problems at o96 and n320 but leads to problems at 9km scale since the batch takes up ~8GB

**Solution:**
Anemoi-inference now gives the flexibility to allow remappings compared to a few months ago when this was more complex. During dataset building the remapping will be done by anemoi-transform https://github.com/ecmwf/anemoi-transform/pull/31.

This works, only the writing to fdb needs some tidying up, so we will not need to do any remapping in anemoi-models anymore therefore eliminating the memory issues.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/aifs-dev/issues/25

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

Not merged before changes are in main of anemoi-inference

Possible reviewers: @japols


mchantry:
Given ATS until Monday to say something.

sahahner:
At the moment, the monomapper can be called via `anemoi.models.preprocessing.monomapper.Monomapper` and `anemoi.models.preprocessing.remapper.Remapper`, which refers to monomapper. I think we should keep only one option to reduce readability.

sahahner:
> At the moment, the monomapper can be called via `anemoi.models.preprocessing.monomapper.Monomapper` and `anemoi.models.preprocessing.remapper.Remapper`, which refers to monomapper. I think we should keep only one option to reduce readability.

After checking with @OpheliaMiralles, the monomapper reference will disappear. The monomapper will be called remapper. Distinguishing between multi and mono mapper will disappear.

models/src/anemoi/models/preprocessing/remapper.py
```diff
@@ -12,36 +12,139 @@
...
+from anemoi.models.preprocessing.mappings import inverse_boxcox_converter
+from anemoi.models.preprocessing.mappings import log1p_converter
+from anemoi.models.preprocessing.mappings import noop
+from anemoi.models.preprocessing.mappings import sqrt_converter
+from anemoi.models.preprocessing.mappings import square_converter
 
 LOGGER = logging.getLogger(__name__)
 
 
 class Remapper(BasePreprocessor, ABC):
```
JesperDramsch: Should this remain defined as an ABC?

JesperDramsch (commented):
Looks good from the Data Index perspective. 

I only left one comment / question on the remapper.

models/src/anemoi/models/preprocessing/remapper.py
sahahner: good catch. thank you

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -208,27 +208,27 @@ def _calculate_input_dim(self, model_config):
...
 
     def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
-        self.num_input_channels = len(data_indices.internal_model.input)
-        self.num_output_channels = len(data_indices.internal_model.output)
+        self.num_input_channels = len(data_indices.model.input)
+        self.num_output_channels = len(data_indices.model.output)
         self.num_input_channels_prognostic = len(data_indices.model.input.prognostic)
-        self._internal_input_idx = data_indices.internal_model.input.prognostic
-        self._internal_output_idx = data_indices.internal_model.output.prognostic
+        self._internal_input_idx = data_indices.model.input.prognostic
```
sahahner, line 214: I purposely chose not to change the name of the model attribute `self._internal_input_idx ` to something without _internal_, as this naming convention still comes from times before the remapper: https://github.com/ecmwf-lab/aifs-mono/blob/3d055ef6e1bb1e0cc901675145a66b8d0fe91cff/aifs/model/gnn.py#L48

I think, _internal_ in this context means that these indices are on the model level and not on the data level.
Any thoughts on this?

japols:
Nice work! I checked the memory usage and the savings of not remapping out-of-place during training line up with the ones reported in [this](https://github.com/ecmwf/aifs-dev/issues/25) issue.

PR #277, opened 2025-04-25
fix: Fix inference with kcrps 

jakob-schloer:
## Issue

When running anemoi-inference the `predict_step` function of the model interface is called, which in turn calls the forward function of the model. The forward function of `AnemoiEnsModelEncProcDec` however takes more arguments which needs to be passed from `anemoi-inference`. 

## Solution:
Introduction of a new `AnemoiEnsModelInterface`, which passes the arguments correctly.
For that, I've introduced a `model.interface` keyword in the model configs. This is a breaking change since the keyword is required. 

## Type of Change

-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Code Compatibility

-   [ ] I have trained a deterministic and kcrps model and preformed inference

### Documentation

### Notes
The arguments also need to be passed by `anemoi-inference` for that I've added a small PR https://github.com/ecmwf/anemoi-inference/pull/212

### DIFF
--- a/models/src/anemoi/models/interface/__init__.py
+++ b/models/src/anemoi/models/interface/__init__.py
@@ -129,3 +129,93 @@ class AnemoiModelInterface(torch.nn.Module):
             y_hat = self(x, model_comm_group)
 
         return self.post_processors(y_hat, in_place=False)
+
+
+class AnemoiEnsModelInterface(AnemoiModelInterface):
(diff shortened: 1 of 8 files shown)


ssmmnn11 (approved):
looks good :-)

PR #281, opened 2025-04-28
fix: Adapt predict_step in model interface to pass on arguments for model classes

jakob-schloer:
## Issue

When running anemoi-inference the `predict_step` function of the model interface is called, which in turn calls the forward function of the model. The forward function of `AnemoiEnsModelEncProcDec` however takes more arguments which needs to be passed from `anemoi-inference`. 

## Solution:
Passing on the relevant keywords coming from `anemoi-inference`.

## Alternative Solution
Interface classes: https://github.com/ecmwf/anemoi-core/pull/277

## Type of Change

-   [ ] Documentation update

## Code Compatibility

-   [x] I have trained a deterministic and kcrps model and preformed inference

### Documentation

### Notes
The arguments also need to be passed by `anemoi-inference` for that I've added a small PR https://github.com/ecmwf/anemoi-inference/pull/212

### DIFF
--- a/models/src/anemoi/models/interface/__init__.py
+++ b/models/src/anemoi/models/interface/__init__.py
@@ -126,6 +126,11 @@ class AnemoiModelInterface(torch.nn.Module):
             # batch, timesteps, horizonal space, variables
             x = batch[:, 0 : self.multi_step, None, ...]  # add dummy ensemble dimension as 3rd index
 
-            y_hat = self(x, model_comm_group)
+            if "fcstep" in kwargs:  # AnemoiEnsModelEncProcDec
+                y_hat = self(x, kwargs["fcstep"], model_comm_group=model_comm_group)
+            elif "target_forcing" in kwargs:  # AnemoiModelEncProcDecInterpolator
+                y_hat = self(x, kwargs["target_forcing"], model_comm_group=model_comm_group)
+            else:
+                y_hat = self(x, model_comm_group=model_comm_group)
 
         return self.post_processors(y_hat, in_place=False)


models/src/anemoi/models/interface/__init__.py
```diff
@@ -126,6 +126,11 @@ def predict_step(
             # batch, timesteps, horizonal space, variables
             x = batch[:, 0 : self.multi_step, None, ...]  # add dummy ensemble dimension as 3rd index
 
-            y_hat = self(x, model_comm_group)
+            if "fcstep" in kwargs:  # AnemoiEnsModelEncProcDec
+                y_hat = self(x, kwargs["fcstep"], model_comm_group=model_comm_group)
+            elif "target_forcing" in kwargs:  # AnemoiModelEncProcDecInterpolator
+                y_hat = self(x, kwargs["target_forcing"], model_comm_group=model_comm_group)
+            else:
+                y_hat = self(x, model_comm_group=model_comm_group)
```
jakob-schloer: An alternative would be passing on *args and **kwargs from predict step  and handle everything in `anemoi-inference`, i.e. 
`y_hat = self(x, *args, model_comm_group=model_comm_group, **kwargs)`

gmertes: There is no reason to have an if here and we should not do it this way.

The signature of `self` should be `self(x, model_comm_group, **kwargs)` , and you just pass through the kwargs. In inference, we pass a bunch of kwargs and then the derived implementation can select the ones it wants and ignore the others.

HCookie: Args/kwargs is a much more scalable solution

jakob-schloer: That works for the `AnemoiEnsModelEncProcDec` where we could change `fcstep` to be explicit and have a default value, but for the TimeInterpolator this is harder because the `target_interpolator` would require a default.

jakob-schloer: I agree :) However, currently we get more passed from `anemoi-inference`, 
```
y_pred = self.predict_step(self.model, input_tensor_torch, fcstep=s, step=step, date=date)
```

I think it wouldn't be nice to have to add **kwargs to all the model forward functions.

gmertes: That's what I mean: we need a `**kwargs` in `self()` so you can can pass arbitrary arguments to it.

jakob-schloer:
@gmertes @HCookie this avoids the if statements and makes the forward functions pick what attributes they need.

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -277,7 +277,7 @@ def _run_mapper(
             use_reentrant=use_reentrant,
         )
 
-    def forward(self, x: Tensor, model_comm_group: Optional[ProcessGroup] = None) -> Tensor:
+    def forward(self, x: Tensor, model_comm_group: Optional[ProcessGroup] = None, **kwargs) -> Tensor:
```
mchantry: Could you add `*` after the x tensor, to ensure that model_comm_group is being called as a kwarg everywhere please?

mchantry (approved):
Great contribution, thanks.

gmertes (approved):
LGTM

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -277,7 +277,7 @@ def _run_mapper(
             use_reentrant=use_reentrant,
         )
 
-    def forward(self, x: Tensor, model_comm_group: Optional[ProcessGroup] = None) -> Tensor:
+    def forward(self, x: Tensor, *, model_comm_group: Optional[ProcessGroup] = None, **kwargs) -> Tensor:
```
MeraX, line 280: Be careful: This line is a breaking change, which is incompatible with `anemoi-training==0.4.0`. Please consider bumbing the minor version number.

MeraX: ... Nonetheless, I believe this kind of change is quite beneficial for the code readability. :+1: 

We've just noticed that if you `pip install anemoi-training=0.4.0` (or even `pip install anemoi-training`) at the moment, pip will install `anemoi-models=0.5.1` which includes this incompatible change leading to non-functional installation of anemoi. (This could also be partially responsible for breaking the downstream CI right now.)

PR #287, opened 2025-04-30
fix(graphs,directions): fix error with direction computation

JPXKQX:
## Description

The method `direction_vec` is fixed, and a test for this bug has been added.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation


graphs/src/anemoi/graphs/edges/directional.py
```diff
@@ -93,8 +93,7 @@ def compute_directions(source_coords: torch.Tensor, target_coords: torch.Tensor)
 
     # Compute the direction from the rotated vector to the north pole.
     direction = direction_vec(rotated_source_coords_xyz, north_pole)
-    normed_direction = direction / torch.norm(direction, dim=1).unsqueeze(-1)
```
anaprietonem, line 96: why are we no longer returning the normalised vector?

graphs/src/anemoi/graphs/edges/directional.py
```diff
@@ -93,8 +93,7 @@ def compute_directions(source_coords: torch.Tensor, target_coords: torch.Tensor)
...
 
     # All 3rd components should be 0s
-    assert torch.max(torch.abs(normed_direction[:, 2])) < 1e-9, "Rotation should be aligned with the north pole"
-    return normed_direction[:, :2]
+    assert torch.max(torch.abs(direction[:, 2])) < 1e-9, "Rotation should be aligned with the north pole"
```
anaprietonem: would it make sense to use the same tolerance here and in the test?

graphs/src/anemoi/graphs/edges/directional.py
JPXKQX: `direction_vec` was already returning the normalised vector.

graphs/src/anemoi/graphs/edges/directional.py
JPXKQX: I think they represent 2 different things. In the `direction_vec` function, we are checking that it is not near the reference point, and the second is about the direction vector being 2d. Both are used to prevent numerical instability, but from different places.

anaprietonem (approved):

Issue #131, opened 2025-02-12
Resetting the cached nan locations of imputer fails when used together with asynchronous plotting


sahahner:
### What happened?

This recent PR https://github.com/ecmwf/anemoi-core/pull/127 allows the model to be validated and used in inference with changed nan locations. This is a very important feature since changes in nan locations are to be expected in inference.

Nevertheless, https://github.com/ecmwf/anemoi-core/pull/127 resets the cached `self.nan_locations` of the imputer to `None`. This leads to errors when asynchronous plotting is used.

I propose to use a copy of the `nan_locations` in the forward pass, that can be reset to `None` in validation/inference mode.

### What are the steps to reproduce the bug?

Use the imputer together with asynchronous plotting.

### Version

main

### Platform (OS and architecture)

atos

### Relevant log output

```shell

```

### Organisation

ECMWF

PR #290, opened 2025-05-01
131 model inference flag and resetting the cached nan locations of imputer

sahahner:
## Description

This recent PR https://github.com/ecmwf/anemoi-core/pull/127 allows the model to be validated and used in inference with changed nan locations. This is a very important feature since changes in nan locations are to be expected in inference.

Nevertheless, https://github.com/ecmwf/anemoi-core/pull/127 resets the cached self.nan_locations of the imputer to None, also in validation mode as there is no inference flag. This leads to errors when asynchronous plotting is used.

I propose to use a copy of the nan_locations in the forward pass, that can be reset to None in validation/inference mode.
Also we introduce an inference flag, that is true when running inference. Only then the nan_locations will be reset to None.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes issue https://github.com/ecmwf/anemoi-core/issues/131

## Code Compatibility

### Dependencies

### Documentation


models/src/anemoi/models/interface/__init__.py
```diff
@@ -72,13 +73,37 @@ def __init__(
...
+        """Set the inference mode of the model.
+        Parameters
+        ----------
+        inference : bool
+            If True, the model is in inference mode.
+        """
+        self._inference = inference
+        if hasattr(self, "pre_processors") and self.pre_processors is not None:
+            for name in self.pre_processors.processors:
+                self.pre_processors.processors[name].inference = inference
```
gmertes: Not a criticism of Sara's work but a general comment: having to propagate flags like this to each part of the model is a hassle and does not scale well.

We should consider passing a context object to all parts of the model, it can be just a dictionary, that can then be used to access attributes like this. Not in this PR, but in a follow-up.

models/src/anemoi/models/interface/__init__.py
```diff
@@ -72,13 +73,37 @@ def __init__(
...
         self.data_indices = data_indices
+        self.inference = inference
         self._build_model()
 
+    @property
+    def inference(self):
+        return self._inference
+
+    @inference.setter
+    def inference(self, inference: bool):
```
gmertes: I originally suggested the setter and getter with a boolean, but we could also consider something akin to `model.eval()` , like`model.inference()`. The downside is that it is less clear how to then switch it back to "non-inference-mode", but I don't think that use-case exists?

Keen to hear other's input.

gmertes (commented):
I tested to code and can confirm that this approach does not cause issues with backwards or forward compatibility. Old model pickles that did not have this flag still load and run with the new code, and vice versa. This should not be a breaking change.

And just to clarify: after this PR is merged, we will make the change in inference to set the inference flag before running the forecast.

models/src/anemoi/models/interface/__init__.py
sahahner: I was thinking about `self.inference_mode` as well. That might be clearer in comparison to` self.inference`.

ssmmnn11: I think it would be good to make this anemoi specific, like anemoi_inference or similar

models/src/anemoi/models/interface/__init__.py
ssmmnn11: very good idea

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,36 +104,45 @@ def _create_imputation_indices(
...
     def transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Tensor:
         """Impute missing values in the input tensor."""
         if not in_place:
             x = x.clone()
 
-        # Reset NaN locations outside of training for validation and inference.
-        if not self.training:
-            self.nan_locations = None
+        # work with copy of cached nan_nanlocations to avoid modifying the cached one
+        nan_locations = self.nan_locations
```
ssmmnn11: what type is nan_locations here? Is this actually creating a copy or will this be a reference?

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -141,7 +151,7 @@ def transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Tensor:
             # for all variables that are imputed and part of the model output, set the loss weight to zero
```
ssmmnn11: is this only at NaN location?

models/src/anemoi/models/interface/__init__.py
sahahner: Since we are inside the Anemoi framework, I don't know whether it is actually necessary to add anemoi to the flag.
Would `self.inference_mode` be fine for you?

models/src/anemoi/models/preprocessing/imputer.py
sahahner: This needs to be a copy. We cannot change the self.nan_locations. This is what causes problems on main at the moment. Let me specify that in a future commit.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Yes. At the fixed NaN locations we want the weight to be zero.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Should actually be `.clone()`! Thanks for the pointer.
Otherwise, we might run into trouble with multiple GPU inference runs.

ssmmnn11: How does the issue manifests in multi GPU (sharded?) inference runs?

models/src/anemoi/models/preprocessing/imputer.py
ssmmnn11: ok, the comment confused me. Can you write something like "... at NaN location.."? So not the full "variable" is set to zero.

models/src/anemoi/models/interface/__init__.py
gmertes: I also think `anemoi_inference` is less descriptive, I agree with `inference_mode`

ssmmnn11: I am happy with both - the important thing is that it is easy to discover what this changes. What is the plan for that? making it easy to search for + all realted settings would be a start -> no false positives.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: I double-checked, and the `self.nan_locations` are not changed. I had written a test that checked that already.
For the asynchronous plotting, the plots were still being computed (and imputers were being applied), while the next forward pass would set the self.nan_locations to None. I don't know whether this behaviour might occur in some way in inference some day as well.
Either way, the self.nan_locations must not be changed, as the fixed nan_locations from training are used in the postprocessor, see description of the intended behaviour in the description of the PR. This will need to be added to the documentation.

ssmmnn11: This seems now quite complicated, and a big contribution seem to be the plotting. Ideally we would want the training to be decoupled from the plotting. Is there another way to implement NaN handling? How expensive is it to get the nan mask? do we have to cache it? could it be carried along with the batch instead?

sahahner: I agree, it does not look like the most elegant solution. 
Nevertheless, as far as I see it, we need to cache the NaN mask, because we need it for the postprocessor. Also, since we assume the NaN locations are fixed, we have decided to make it a model parameter that we need in inference for the postprocessing to put the NaNs back into place.
At one time, we had been thinking about carrying the NaN mask with the batch, but this change needs to come from the dataset itself, where a new (static) field would need to be introduced for every variable with NaNs. So I would propose the solution to consider carrying the NaN mask around for another PR

sahahner: I changed the order, which, in my opinion, improves the readability. Please have a look and feel free to counterpropose another way to do it. :)

sahahner:
Another approach has been chosen in issue https://github.com/ecmwf/anemoi-core/issues/370. Once this is closed, this PR will be obsolete. At the moment, we leave it in draft mode to have it for comparison.

PR #292, opened 2025-05-01
ci: refactor test pr workflow

### LINKED ISSUE #284: CI: matrix jobs "local-changes" for "Test PR" fail on unrelated pull requests
### What happened?

There were tests failing on [this PR](https://github.com/ecmwf/anemoi-core/pull/253), that were unrelated to the changes in the PR but instead related to [this change](https://github.com/ecmwf/anemoi-core/commit/0ea06423e4979084b3afe70c30e43bb5dc5f88d5) merged into main the day before. It's only the "local-changes" matrix jobs that are failing. 
The reason is that some of the core-package versions were out of sync. Indeed, on main: pytests succeeded for editable installs of all packages, and they failed for an editable install of only training.

We fixed this by releasing graphs and bumping the version of graphs required in training to get those tests to pass on main again.

However, it seems suboptimal that the local-changes matrix jobs fail on unrelated PRs (depending on whether these change the same packages or not). Let's discuss how this workflow is intended to work and how we should amend it. One possible option is to only run this on release branches to catch versions that are out of sync

- Make a change to anemoi-graphs and anemoi-training, where the latter depends on the former.
- Merge both in one PR -- local-changes matrix jobs will not fail because they install both training and graphs from main
- Then create a PR that only modifies training (in an unrelated place) -- local-changes matrix jobs will fail because they install training from main but graphs from pypi 

### Version

the current workflow "Test PR"

VeraChristina:
## Description

Update the Test PR workflow
- On each PR only test on editable installs of all three packages (previously known as "all" matrix jobs)
- For release PRs add a workflow that runs the previous "local-changes" matrix jobs, i.e. an editable install of only the package to be released

This fixes the problem outlined in issue #284
## Type of Change

Change to the ci to fix a ci bug

## Issue Number

Closes #284

## Code Compatibility

### DIFF
--- a/.github/workflows/python-release-tests.yml
+++ b/.github/workflows/python-release-tests.yml
@@ -1,11 +1,11 @@
-name: Test anemoi-training for Release
+name: Test Release
 
 on:
   pull_request:
     types: [opened, synchronize, reopened]
-    # filter for PRs that change the changelog to only run the tests on release PRs. (Other PRs should not change the changelog)
+    # filter for PRs that change the manifest to only run the tests on release PRs. (Other PRs should not change the manifest)
     paths:
-      - "training/CHANGELOG.md"
+      - "release-please-manifest.json"
 
 jobs:
   pytest:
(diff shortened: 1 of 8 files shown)


.github/workflows/python-release-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  quality:
+    uses: ecmwf/reusable-workflows/.github/workflows/qa-precommit-run.yml@v2
+    with:
+      skip-hooks: "no-commit-to-branch"
+
+  checks:
```
gmertes: I would rename this to `pytest` instead of `checks`. This name shows up in the list of actions that are running under the PR, it makes it clearer what's happening.

And the same in the PR workflow.

.github/workflows/python-release-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.user.login == 'DeployDuck'}}
```
gmertes: Putting this check down here means that the workflow and quality job will run on every PR. I would add a path filter at the top so that the whole workflow only runs on release PRs, like so:

```
on:
  pull_request:
    types: [opened, synchronize, reopened]
    paths:
      - ".release-please-manifest.json"
```

.github/workflows/python-release-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+
+
+    - name: Install packages
+      if: ${{steps.release-package.outputs.training_release == 'true' || steps.release-package.outputs.graphs_release == 'true' || steps.release-package.outputs.models_release == 'true' }}
+      run: |
+        if [[ ${{ steps.release-package.outputs.training_changed }} == 'true' ]]; then
+          pip install -e ./training[all,tests]
+        elif [[ ${{ steps.release-package.outputs.graphs_release }} == 'true' ]]; then
+          pip install -e ./graphs[all,tests]
+        elif [[ ${{ steps.release-package.outputs.models_release }} == 'true' ]]; then
```
gmertes: Did you use the correct variable names here? Should this be `training_release` instead of `training_changed` ?

gmertes (approved):

Issue #294, opened 2025-05-01
Nightly Failures Tracker


VeraChristina:
This issue tracks failures of the nightly integration tests. Comments will be updated automatically.

PR #295, opened 2025-05-01
ci: add notify-team step for integration test workflow

VeraChristina:
## Description

Adds a step to the nightly integration test workflow that adds a comment to https://github.com/ecmwf/anemoi-core/issues/294 and tags anemoisecurity for broader visibility of failed nightly runs. Currently, only the person that created the workflow gets notified.

The first comment on the issue (which I edited to avoid confusion) was created by this test run, which simulated a failure: https://github.com/ecmwf/anemoi-core/actions/runs/14778582874


.github/workflows/integration-tests.yml
```diff
@@ -17,3 +17,22 @@ jobs:
...
+
+  notify-team:
+    needs: integration_tests
+    if: failure()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add or update failure comment
+        uses: peter-evans/create-or-update-comment@v3
+        with:
+          issue-number: 294
```
theissenhelen, line 29: Will this create a new comment or update the existing one on the issue?

theissenhelen (dismissed):

.github/workflows/integration-tests.yml
VeraChristina: Add a new one -- so that we actually get the notification. I'll update the name for clarity

theissenhelen (approved):

PR #300, opened 2025-05-05
fix: Adapt ensemble configs to work with loss refactor. 

jakob-schloer:
## Description

The default configs for the ensemble training were not working with the new loss refactor. This PR creates a `config/training/ensemble.yaml` that sets default values for training with ensemble members.

A new `ensemble.yaml` file was required because the `default.yaml` would set `mse` as a `training.validation_metrics`, which can be extended by other metrics but cannot be fully overwritten (mse would always be part of the validation_metrics). The MSE metric, however, is not compatible with the `GraphEnsForecaster` model. 

## Type of Change

## Code Compatibility

-   [x] I have tested inference on a trained model

### Dependencies

### DIFF
--- a/training/src/anemoi/training/config/debug_ens.yaml
+++ b/training/src/anemoi/training/config/debug_ens.yaml
@@ -6,10 +6,10 @@ defaults:
 - hardware: example
 - graph: encoder_decoder_only
 - model: transformer_ens
-- training: default
+- training: ensemble
 - _self_
 
-config_validation: True
+config_validation: False
 
 ### This file is for local experimentation.
 ##  When you commit your changes, assign the new features and keywords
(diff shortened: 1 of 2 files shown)


anaprietonem:
thanks for this fix Jakob!

training/src/anemoi/training/config/debug_ens.yaml
```diff
@@ -6,10 +6,10 @@ defaults:
 - hardware: example
 - graph: encoder_decoder_only
 - model: transformer_ens
-- training: default
+- training: ensemble
 - _self_
 
-config_validation: True
+config_validation: False
```
anaprietonem: I'd recommend, keeping the config_validation to default `True` value

jakob-schloer: Of course, sorry :)

training/src/anemoi/training/config/training/ensemble.yaml
```diff
@@ -0,0 +1,128 @@
...
+  ignore_nans: False
+  alpha: 1.0
+
+# Validation metrics calculation,
+# This may be a list, in which case all metrics will be calculated
+# and logged according to their name.
+# These metrics are calculated in the output model space, and thus
+# have undergone postprocessing.
+validation_metrics:
+  # loss class to initialise, can be anything subclassing torch.nn.Module
```
anaprietonem, line 86: Is there any reason why we go from 2 set of validation metrics to just one? (ie discarding the config section with alpha 0.95)

jakob-schloer: It's the difference between the almost fair kernel crps and kernel crps. I thought it was cleaner to show only one.

anaprietonem: thanks for the clarification!

anaprietonem (approved):
LGTM!

PR #305, opened 2025-05-07
Change output format in validation step to fix issue in plotting

mc4117:
## Type of Change

## Issue Number

Whilst investigating callbacks @sahahner and I found an issue in the PlotLoss callback which meant that the rollout step was being interpreted incorrectly as the batch. This small fix fixes this issue.

## Code Compatibility

### Dependencies


HCookie (approved):
So long as all plots are created and look fine, happy to merge

PR #306, opened 2025-05-07
test: restart training

VeraChristina:
## Description

Adds an integration test in training which
- runs a short training cycle with the basic configuration, then
- restarts the training from the run_id created in the short training cycle

There are a couple of asserts to check if the checkpoints are in the expected place (and if it's the expected number of run_ids and checkpoints) not so much for testing the restart but for checking the integrity of the test itself

## Type of Change

New integration test in training

## Code Compatibility


training/tests/integration/test_training_cycle.py
```diff
@@ -51,3 +52,32 @@ def test_training_cycle_lam(lam_config_with_data: DictConfig) -> None:
...
+@skip_if_offline
+@pytest.mark.longtests
+def test_restart_training(gnn_config_with_data: DictConfig) -> None:
+
+    AnemoiTrainer(gnn_config_with_data).train()
+
+    cfg = gnn_config_with_data
+    output_dir = Path(cfg.hardware.paths.output + "checkpoint")
+
+    if not output_dir.exists():
```
theissenhelen: That could be an assert statement instead. I don't think test should raise errors.

training/tests/integration/test_training_cycle.py
```diff
@@ -51,3 +52,32 @@ def test_training_cycle_lam(lam_config_with_data: DictConfig) -> None:
...
+        msg = f"Checkpoint directory not found at: {output_dir}"
+        raise FileNotFoundError(msg)
+
+    run_dirs = [item for item in output_dir.iterdir() if item.is_dir()]
+    if len(run_dirs) != 1:
```
theissenhelen: see above

theissenhelen (changes requested):
Very good work. The only thing I would change is using asserts rather than raising errors in the tests.

training/tests/integration/test_training_cycle.py
VeraChristina: Good point. Will update!

theissenhelen (approved):

PR #308, opened 2025-05-08
ci: only run notify-team step on cron jobs in integration tests

VeraChristina:
## Description
Only run the notify-team step on cron jobs in the integration tests workflow. For manually triggered tests on branches the AnemoiSecurity group doesn't need to be notified. The default notification (to the person who triggered the workflow) is enough.


anaprietonem (approved):
Thanks Vera, LGTM!

PR #311, opened 2025-05-09
fix(training): kcrps losses handling NaNs for validation

sahahner:
## Description

All our losses have the option `ignore_nans` as we want to ignore NaN fields if the losses are used as validation metrics. 
This option is still missing for the kcrps losses. 

Instead of using `torch.sum` and `torch.mean` use the sum and avg functions of the loss class to allow or not allow NaNs depending on what has been selected in the config file.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/training/src/anemoi/training/losses/kcrps.py
+++ b/training/src/anemoi/training/losses/kcrps.py
@@ -57,7 +57,7 @@ class KernelCRPS(BaseLoss):
             The point-wise kernel CRPS, shape (batch_size, 1, latlon).
         """
         ens_size = preds.shape[-1]
-        mae = torch.mean(torch.abs(targets[..., None] - preds), dim=-1)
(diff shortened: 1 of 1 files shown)


ssmmnn11:
Thank you for the pull request. 

There is always a question how to handel NaNs. I think the best way is to do it like in the other losses, in the spatial aggregation part. Here, the NaN "handling" happens in the point wise crps score computation, which leads to undefined behaviour if different ensemble members had NaNs at different places - though this is currently not the case. 

The loss scaling is currently broken after https://github.com/ecmwf/anemoi-core/pull/52. My suggestion is that we wait until this is fixed (hopefully soon) and then we can implement NaN handling in a consistent way in all losses.

PR #315, opened 2025-05-12
feat(model): postprocessors for leaky boundings

sahahner:
## Description

Introduce Postprocessors for the leaky bounding functions that push the model output values into the desired ranges. 

The leaky bounding functions make sure that gradients are not zero for big areas of the global fields (See PR https://github.com/ecmwf/anemoi-core/pull/256). This improves convergence. Nevertheless, some outputs might then be physically inconsistent. In combination with the introduced postprocessors, this is not an issue anymore.

For example, for sea ice albedo, the convergence is improved when combining leaky hardtan in combination with a postprocessor (orange, green, blue lines in plot), ensuring no negative values are found in the model output for sea ice albedo. In comparison, the model shows worse convergence with hardtan bounding function (red, purple lines)

Screenshot 2025-05-12 at 14 51 36

When using only leaky bounding, some predicted fields might still have unphysical forecasts.
For example, in this sea ice concentration forecast (should be in range [0,1]) leaky bounding without postprocessing is applied. The plot on the right shows a forecasts after a few training steps. The colobar goes slightly below zero, so there have been negative forecasts.

Screenshot 2025-07-01 at 14 56 34

Additional fixes:
- In the bounding class `NormalizedReluBounding`, the custom minimum values `self.norm_min_val` are now registered as a buffer, to avoid copying tensors to the model's device in every forward pass.
- Ensuring three different values in TwoSlopeNorm to avoid errors/failures.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

-   [ ] If the new feature introduces modifications at the config level, I have made sure to update Pydantic Schemas and default configs accordingly -> still open

### Dependencies

### Documentation


training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -588,6 +588,9 @@ def plot_flat_sample(
         data[0] = input_
         data[4] = pred - input_
         data[5] = truth - input_
+        # ensure vcenter is between minimum and maximum error
+        vcenter = max(np.nanmin(combined_error), 0.0)
+        vcenter = min(np.nanmax(combined_error), vcenter)
         norm_error = TwoSlopeNorm(vmin=np.nanmin(combined_error), vcenter=0.0, vmax=np.nanmax(combined_error))
```
lzampier: vcenter=vcenter ;)

lzampier: And this prevents failures if all three values are zero. 

norm_error = TwoSlopeNorm(vmin=min(-0.00001,np.nanmin(combined_error)), vcenter=vcenter, vmax=max(0.00001,np.              nanmax(combined_error)))

lzampier (approved):
I tested these modifications in training and they do the job: https://mlflow.ecmwf.int/#/experiments/245/runs/e881610196c44c88b2a270313d59d5dd.

Below are some suggestions for the callback part.

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -0,0 +1,354 @@
...
+            )
+
+        # Replace values
+        for postprocessor, idx_dst in zip(self.postprocessorfunctions, index):
+            if idx_dst is not None:
+                x[..., idx_dst] = postprocessor(x[..., idx_dst])
+        return x
+
+
+class CustomRelu(torch.nn.Module):
```
anaprietonem: activations functions should live under `/models/layers/activations.py` for consistency

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -0,0 +1,354 @@
...
+
+import torch
+
+from anemoi.models.data_indices.collection import IndexCollection
+from anemoi.models.preprocessing import BasePreprocessor
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Postprocessor(BasePreprocessor):
```
anaprietonem, line 23: Have you tested if this running fine with Pydantic? Quick look I'd say this might require updating the schemas so we have something like https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/schemas/data.py listing the post-processors. Similar note it would be great to have a couple of lines for docs! and thanks for the contribution Sara!

anaprietonem: I might being slow, but what's the difference between a post-processor and a bounding layer?

sahahner: The bounding layers are applied before the training loss function. The postprocessors are applied after and, therefore, not considered in the gradient calculation.
This is the reason why we want to use leaky bounding, which do not set gradients to zero for big areas of predictions (see issue linked above), and correct possibly non-physical values with the postprocessors after the loss function.

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -0,0 +1,354 @@
...
+
+        # Validate normalizer input
+        if self.normalizer not in {"none", "mean-std", "min-max", "max", "std"}:
+            raise ValueError(
+                "Normalizer must be one of: 'none', 'mean-std', 'min-max', 'max', 'std' in NormalizedReluBounding."
+            )
+
+        super().__init__(config, data_indices, statistics)
+
+    def _create_imputation_indices(
```
anaprietonem: Why does this function need to be redefined for each post-processor new class? I think there should be a way to abstract away some of this complexity into the base `Postprocessor` and then call the super method to avoid repeating code

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -0,0 +1,354 @@
...
+    0:
+        - x
+    3.14:
+        - q
+    ```
+    Thresholds are in un-normalized space. If normalizer is specified, the threshold values are not normalized.
+    This is necessary if in config file the normalizer is specified before the postprocessor, e.g.:
+    ```
+    data:
+        processors:
```
anaprietonem, line 147: Is the idea to have just a dictionary of processors and then here include both pre and post-processors? Thinking if there is a certain type of operation that could be applied both as a pre and post-processor how will the code and config handle it

models/src/anemoi/models/preprocessing/postprocessor.py
anaprietonem: Thanks Sara, now it's much clear. Have you done any tests comparing how bounding vs post-processors behaves? Maybe @gabrieloks looked at this when initially experimenting with the bounding. Also do we have to think or constrain the case where we apply both? (like a leakyReLU as a bounding and as a post-processor?)

models/src/anemoi/models/layers/bounding.py
```diff
@@ -144,25 +144,28 @@ def __init__(
...
             elif self.normalizer[ii] == "max":
                 max_stat = self.statistics["max"][stat_index]
-                self.norm_min_val[ii] = min_val[ii] / max_stat
+                norm_min_val[ii] = min_val[ii] / max_stat
             elif self.normalizer[ii] == "std":
                 std = self.statistics["stdev"][stat_index]
-                self.norm_min_val[ii] = min_val[ii] / std
+                norm_min_val[ii] = min_val[ii] / std
+        # register the normalized min values as a buffer to ensure they are moved to the correct device
+        self.register_buffer("norm_min_val", norm_min_val)
```
anaprietonem, line 168: Nice one @sahahner !

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -0,0 +1,354 @@
...
+        """
+        return torch.nn.functional.relu(x - self.threshold) + self.threshold
+
+
+class NormalizedReluPostprocessor(Postprocessor):
+    """Postprocess with a ReLU activation and customizable thresholds.
+
+    Expects the config to have keys corresponding to customizable thresholds as lists of variables to impute.:
+    ```
+    nornmalizer: 'mean-std'
```
anaprietonem: small typo should be -> 'normalizer'

models/src/anemoi/models/preprocessing/postprocessor.py
sahahner: All processors have a pre- and postprocessor, for example normalizer and denormalizer. All preprocessors have a transform (preprocessor) and an inverse transform (postprocessor) function. In the case of the postprocessors introduced by this PR, the preprocessor is the identity function.

models/src/anemoi/models/preprocessing/postprocessor.py
gabrieloks: A nice test for bounding vs post-processors was TP in AIFS 0.2.1 vs AIFS 1.0.0 +. With the post-processor you get rid of the negative "unphysical" values but with a bounding layer you change the dynamics of training, and facilitate the learning of the zero forecast. AIFS 1.0.0 + is much better at predicting non-rain, while AIFS 0.2.1 was overdoing light rain quite heavily

models/src/anemoi/models/preprocessing/postprocessor.py
sahahner: should be clearer now

models/src/anemoi/models/preprocessing/postprocessor.py
sahahner: Above you can see an example of hard bounding vs leaky bounding in combination with a postprocessor. The convergence is better and therefore the validation loss is a lot lower when using the leaky bounding with postprocessors. The postprocessors ensure that the output range of the variable is the same and physical consistent.
I think it is on the user side to combine leaky bounding functions and postprocessors accordingly.

theissenhelen: I have adjusted the schemas to fit the config model. The schemas diverge a bit from our usual setup to allow more flexibility for the keys - in this case the keys are actually numbers.

sahahner: I had to change a few things to work with my config files. Would you mind giving it another look to make sure I didn't break anything, please?

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -591,7 +591,12 @@ def plot_flat_sample(
         data[4] = pred - input_
         data[5] = truth - input_
         combined_error = np.concatenate(((pred - input_), (truth - input_)))
-        norm_error = TwoSlopeNorm(vmin=np.nanmin(combined_error), vcenter=0.0, vmax=np.nanmax(combined_error))
+        # ensure vcenter is between minimum and maximum error
+        norm_error = TwoSlopeNorm(
```
anaprietonem, line 595: where is the (+/-) 0.00001 coming from?

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def __init__(
+        self,
+        config=None,
+        data_indices: Optional[IndexCollection] = None,
+        statistics: Optional[dict] = None,
+    ) -> None:
+        super().__init__(config, data_indices, statistics)
+
+    def _create_postprocessing_indices(
```
anaprietonem: what's the difference between this class and the one from the base class? do we need to repeat it here? 
from a quick look it seems you're adding:
```
        self.masking_variable = self.remap
        self.masking_variable_training_output = name_to_index_training_output.get(self.masking_variable, None)
        self.masking_variable_inference_output = name_to_index_inference_output.get(self.masking_variable, None)
       
   ```
   but the rest is the same, you could call super() from the base class and then just add those below

training/src/anemoi/training/diagnostics/plots.py
sahahner: For some variables, we ran into the case where vmin and vcenter were the same value (0). Then this `twoslopenorm` breaks, as the values need to be different. So we make sure that vmin is always smaller than vcenter and vmax is larger.

models/src/anemoi/models/preprocessing/postprocessor.py
theissenhelen: The schemas look good to me now.

models/src/anemoi/models/preprocessing/postprocessor.py
sahahner: I want to follow the structure of all the other processors and not make `name_to_index_training_output` a variable of the class. we are using it later

models/src/anemoi/models/preprocessing/postprocessor.py
anaprietonem: could you add that to the docstring of the 'basepost-processor' class? (the fact that no forward_transform is overwritten)

models/src/anemoi/models/preprocessing/postprocessor.py
anaprietonem: What's the motivation for not having 'name_to_index_training_output' as 'self. name_to_index_training_output'?  Then it could be reused and avoid repeating bits of code across the post-processors.  If we find bugs in the indices creation, then by not being able to leverage the class in the base class that means having to fix the issues across many classes and this method easily diverding

sahahner: These temporary variables have been removed now, accessing the data_indices directly. The code is not longer repeated for the postprocessors. Feel free to have a look.

models/src/anemoi/models/preprocessing/postprocessor.py
sahahner: Done.

anaprietonem (approved):
This PR LGTM, and it's passing the integration tests. Thanks for the work and for addressing my comments @sahahner !

PR #321, opened 2025-05-15
fix: Use transforms: Variable for ExtractVariableGroupAndLevel

HCookie:
## Description

Use anemoi.transforms Variable for `ExtractVariableGroupAndLevel`.

Fixes the issue with param's not being recognised under the new name when renamed in the open_dataset call.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation

----
📚 Documentation preview 📚: https://anemoi-training--321.org.readthedocs.build/en/321/

----
📚 Documentation preview 📚: https://anemoi-graphs--321.org.readthedocs.build/en/321/

----
📚 Documentation preview 📚: https://anemoi-models--321.org.readthedocs.build/en/321/


training/src/anemoi/training/config/training/default.yaml
```diff
@@ -88,9 +88,20 @@ validation_metrics:
...
 # A default group is required and is appended as prefix to the metric of all variables not assigned to a group.
+# Variables are assigned to a group by their param if contained in the metadata, else by their name.
+
+# If more complex grouping is required, groups can be defined as a dictionary, such that all
+# keys must be evaluate to True.
+# .e.g.:
+#   variable_groups:
+#     default: sfc
+#     pl:
+#        is_pressure_level: True
```
mc4117, line 102: I would maybe add an example with is_model_level too

mc4117:
Look's good to me!

mc4117:
Here are some mlflow runs showing same performance pre loss scaling PR and post fix

Batch size 4
https://mlflow.ecmwf.int/#/metric?runs=[%229d26d9b95dba4555802b2ca5852cf680%22,%22467e94edf0cb40f8a61ec549fb32445e%22]&metric=%22epoch%22&experiments=[%2254%22]&plot_metric_keys=%5B%22train_mse_loss_step%22%2C%22train_wmse_step%22%5D&plot_layout={%22autosize%22:true,%22xaxis%22:{},%22yaxis%22:{}}&x_axis=relative&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=[]&last_linear_y_axis_range=[]

Batch size 2
https://mlflow.ecmwf.int/#/compare-runs?runs=[%227996170d02294357bcbec9b7f3007237%22,%226589ceede8de407da89250a3125b959a%22]&experiments=[%2254%22]

training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -9,6 +9,40 @@
 
 from __future__ import annotations
 
+import logging
+from functools import lru_cache
+
+from omegaconf import DictConfig
+from omegaconf import OmegaConf
+
+from anemoi.transform.variables import Variable
```
anaprietonem, line 19: Thanks for this @HCookie ! From a quick look and reading the docs update "`anemoi.transforms.variable.Variable` for the metadata attributes that are available." I was wondering if this functionality should rather live in datasets or utils so it can be used across the framework and hence avoid introducing transform as a direct dependency in training

HCookie: As the grouping stuff is hyper specific to training, I'm not sure there is a need to abstract it further. Although I do agree, that a direct dependency is not ideal, but where we put this a dependancy will be established.

training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -26,24 +60,115 @@ class ExtractVariableGroupAndLevel:
...
+                        variable_name,
+                        group_name,
+                        group_spec,
+                    )
+                    return group_name
+
+        return self.default_group
+
+    def get_param(self, variable_name: str) -> str:
+        """Get the parameter of a variable.
```
anaprietonem: parameter and variable name are referring to the same thing? If so we could rename this or at least mention that

HCookie: What do you mean? The param is part of the variable name?

anaprietonem: the `_crack_variable_name` returns a variable name as the first element, so in that case param and variable name are the same? However for the MARS data you extract the param from the variable name. I am trying to understand then what's a param and what's a variable name

training/src/anemoi/training/utils/variables_metadata.py
anaprietonem: Is the logic to put `anemoi.transforms.variable.Variable` in transforms because the metadata is driven by the source used and we expect sources to live in transform? If not I'd push this to be on datasets or better in anemoi-utils where we already have some features related to metadata tracking and hence could be potentially reused at inference time. What do you think?

mc4117: @b8raoult

training/src/anemoi/training/utils/variables_metadata.py
HCookie: variable_name is the training representation i.e. `q_500`, param is `q`

training/src/anemoi/training/utils/variables_metadata.py
HCookie: Not sure, I just used it where it was

training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -9,6 +9,40 @@
...
+    variable_name : str
+        Variable reference which corresponds to the variable name without the variable level
+    variable_level : str
+        Variable level, i.e. pressure level or model level
+    """
+    split = variable_name.split("_")
+    if len(split) > 1 and split[-1].isdigit():
+        return variable_name[: -len(split[-1]) - 1], int(split[-1])
+
+    return variable_name, None
```
HCookie:
```suggestion
def _crack_variable_name(variable_name: str) -> tuple[str, str | None]:
    """Attempt to crack the variable name into parameter name and level.
    
    If cannot split, will return variable_name unchanged, and None

    Parameters
    ----------
    variable_name : str
        Name of the variable.

    Returns
    -------
    parameter : str
        Parameter reference which corresponds to the variable_name without the variable level.
        If cannot be split, will be variable_name unchanged.
    variable_level : str | None
        Variable level, i.e. pressure level or model level.
        If cannot be split, will be None.
    """
    split = variable_name.split("_")
    if len(split) > 1 and split[-1].isdigit():
        return variable_name[: -len(split[-1]) - 1], int(split[-1])

    return variable_name, None
```

training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -26,24 +60,115 @@ class ExtractVariableGroupAndLevel:
...
+        ----------
+        variable_name : str
+            Name of the variable.
+
+        Returns
+        -------
+        param : str
+            Parameter of the variable.
+        """
+        if variable_name in self.metadata_variables:
```
HCookie:
```suggestion
    def get_param(self, variable_name: str) -> str:
        """Get the parameter from a variable_name.
        
        Tries to use the metadata, but if not given
        will attempt to crack the name. If cannot
        crack will be the variable_name unchanged.

        Parameters
        ----------
        variable_name : str
            Name of the variable.

        Returns
        -------
        param : str
            Parameter of the variable.
            Either from the metadata or cracked
            name.
        """
        if variable_name in self.metadata_variables:
```

training/src/anemoi/training/utils/variables_metadata.py
HCookie: I updated the docs, hope that helps

HCookie: Can this be resolved?

mc4117:
I have confirmed that this PR works when the dataset was not created by mars and has different metadata (specifically https://anemoi.ecmwf.int/datasets/metno-meps-archive-det-opendap-2p5km-2023-2023-6h-v1)

training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -55,29 +99,104 @@ def get_group_and_level(self, variable_name: str) -> tuple[str, str, int]:
...
+            Either from the metadata or cracked
+            name.
+        """
+        if variable_name in self.metadata_variables:
+            # if metadata is available: get variable name and level from metadata
+            return self.metadata_variables[variable_name].param
+
+        return _crack_variable_name(variable_name)[0]
+
+    def get_level(self, variable_name: str) -> str | None:
```
anaprietonem, line 163: this could be an hidden method and same for `get_param` and `get_group`

anaprietonem (approved):
@HCookie I left a comment as I think some of the methods of `ExtractVariableGroupAndLevel` should be hidden but it's a minor suggestion. Otherwise LGTM and will be raised for approval at ATS, thanks for the work!

training/src/anemoi/training/utils/variables_metadata.py
HCookie: They could be 'hidden' methods, but may prove useful in the future to have exposed?

anaprietonem (approved):
Great work @HCookie  and @mc4117 ! good to merge

PR #325, opened 2025-05-19
test: add integration test for the crps ensemble

mc4117:
## Description
Adding integration tests for the CRPS ensemble.

## Type of Change

### DIFF
--- a/training/tests/integration/test_training_cycle.py
+++ b/training/tests/integration/test_training_cycle.py
@@ -63,6 +63,7 @@ def test_training_cycle_ensemble(ensemble_config_with_data: DictConfig) -> None:
 def test_config_validation_ensemble(ensemble_config: DictConfig) -> None:
     BaseSchema(**ensemble_config)
 
+
 @skip_if_offline
 @pytest.mark.longtests
 def test_restart_training(gnn_config_with_data: DictConfig) -> None:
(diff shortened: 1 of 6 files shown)


training/tests/integration/test_training_cycle.py
```diff
@@ -54,6 +54,16 @@ def test_config_validation_lam(lam_config: DictConfig) -> None:
     BaseSchema(**lam_config)
 
 
+@skip_if_offline
```
VeraChristina, line 57: I think we'll need an additional flag here to skip the test if running without gpu (since the transformer requires flash attention which requires gpu), something like:
@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is not available")

VeraChristina (changes requested):
This looks great, many thanks! 
I think the main blocker before merging is that the ensemble test requires flash attention to be installed in the hpc workflow (and a flag to skip the test if running without gpu, see comments)

training/tests/integration/test_training_cycle.py
VeraChristina: Update: If possible, we'd like to run this test with a configuration that doesn't require flash attention

mc4117: I have changed the test to now run with graphtransformer so we no longer need flash attention

mc4117:
This has now passed the nightly integration tests

PR #327, opened 2025-05-21
feat: Log variable scaling in mlflow

jakob-schloer:
## Description

The variable scaling of the loss function is not trivial anymore since many scalers can be multiplied. To keep track of that, the final variable scaling is logged to mlfow hyperparameters (see image) 

image

## Type of Change

## Code Compatibility

### Documentation

### DIFF
--- a/training/src/anemoi/training/train/forecaster/forecaster.py
+++ b/training/src/anemoi/training/train/forecaster/forecaster.py
@@ -123,7 +123,7 @@ class GraphForecaster(pl.LightningModule):
             scalers=self.scalers,
             data_indices=self.data_indices,
         )
-        print_variable_scaling(self.loss, data_indices)
+        self._scaling_values = print_variable_scaling(self.loss, data_indices)
 
         self.metrics = torch.nn.ModuleDict(
             {
@@ -525,3 +525,10 @@ class GraphForecaster(pl.LightningModule):
         )
 
         return [optimizer], [{"scheduler": scheduler, "interval": "step"}]
+
+    def setup(self, stage: str) -> None:
+        """Lightning hook that is called after model is initialized but before training starts."""
+        # Log variable scaling on rank 0
+        # The two ifs should be separate, but are combined due to pre-commit hook
+        if stage == "fit" and self.trainer.is_global_zero:
+            self.logger.log_hyperparams({"variable_scaling": self._scaling_values})
--- a/training/src/anemoi/training/losses/utils.py
+++ b/training/src/anemoi/training/losses/utils.py
@@ -8,18 +8,23 @@
 # nor does it submit to any jurisdiction.
 
 
-import logging
+from __future__ import annotations
 
-from anemoi.models.data_indices.collection import IndexCollection
-from anemoi.training.losses.base import BaseLoss
+import logging
+from typing import TYPE_CHECKING
+
 from anemoi.training.utils.enums import TensorDim
+
+if TYPE_CHECKING:
+    from anemoi.models.data_indices.collection import IndexCollection
+    from anemoi.training.losses.base import BaseLoss
 
 LOGGER = logging.getLogger(__name__)
 
 
-def print_variable_scaling(loss: BaseLoss, data_indices: IndexCollection) -> None:
+def print_variable_scaling(loss: BaseLoss, data_indices: IndexCollection) -> dict[str, float]:
     """
-    Log the final variable scaling for each variable in the model.
+    Log the final variable scaling for each variable in the model and return the scaling values.
 
     Parameters
     ----------
@@ -27,9 +32,20 @@ def print_variable_scaling(loss: BaseLoss, data_indices: IndexCollection) -> Non
         Loss function to get the variable scaling from.
     data_indices : IndexCollection
         Index collection to get the variable names from.
+
+    Returns
+    -------
+    Dict[str, float]
+        Dictionary mapping variable names to their scaling values.
     """
     variable_scaling = loss.scaler.subset_by_dim(TensorDim.VARIABLE.value).get_scaler(len(TensorDim)).squeeze()
     log_text = "Final Variable Scaling: "
+    scaling_values = {}
+
     for idx, name in enumerate(data_indices.internal_model.output.name_to_index.keys()):
-        log_text += f"{name}: {variable_scaling[idx]:.4g}, "
+        value = float(variable_scaling[idx])
+        log_text += f"{name}: {value:.4g}, "
+        scaling_values[name] = value
+
     LOGGER.debug(log_text)
+    return scaling_values


sahahner:
Hi Jakob, I think this is a very useful PR. Thank you very much for opening it. Would it be feasible as well to list the variable scaling in a normalised way? I am wondering whether it would be helpful to divide the per-variable scalers by the sum of all scalers. On the other hand, this might not be the best solution when adding new variables. Any thoughts?

jakob-schloer:
> Would it be feasible as well to list the variable scaling in a normalised way? I am wondering whether it would be helpful to divide the per-variable scalers by the sum of all scalers.

I agree that it would be nice to have the percentage scaling, but this might also be a bit misleading because the variables are not all normalized exactly to std=1 (depending on the normalization set).

training/src/anemoi/training/train/forecaster/forecaster.py
```diff
@@ -525,3 +525,10 @@ def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[dict]]
         )
 
         return [optimizer], [{"scheduler": scheduler, "interval": "step"}]
+
+    def setup(self, stage: str) -> None:
+        """Lightning hook that is called after model is initialized but before training starts."""
+        # Log variable scaling on rank 0
+        # The two ifs should be separate, but are combined due to pre-commit hook
+        if stage == "fit" and self.trainer.is_global_zero:
+            self.logger.log_hyperparams({"variable_scaling": self._scaling_values})
```
jakob-schloer: Any opinions on the keyword? Maybe "loss_scaling_variable" is better.

sahahner: Can we also add a resulting or final to indicate it is not necessarily set by the user? For example:
`final_variable_loss_scaling.XXX`

anaprietonem:
Hey @jakob-schloer thanks for the PR! could you link an example mlflow run to have a better look? Also just to flag server performance in terms of rendering the runs can be impacted by the number of hparams we log so we need to be mindful there. In your screenshot we are talking about ~100 variables logged ?

jakob-schloer:
> Hey @jakob-schloer thanks for the PR! could you link an example mlflow run to have a better look? Also just to flag server performance in terms of rendering the runs can be impacted by the number of hparams we log so we need to be mindful there. In your screenshot we are talking about ~100 variables logged ?

Yes, the number of variables values are added as hparams. Here is an mlflow example run: 
https://mlflow.ecmwf.int/#/experiments/221/runs/6f1b5dd4597742709171a7ed442dd33a

training/src/anemoi/training/train/forecaster/forecaster.py
jakob-schloer: @sahahner I went for `variable_loss_scaling.XXX` but dropped the `final_`.

sahahner: I looked at your mlflow example and I think it is sufficiently clear now that these values are the resulting scaler values.

HCookie (approved):

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -593,3 +596,21 @@ def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[dict]]
         )
 
         return [optimizer], [{"scheduler": scheduler, "interval": "step"}]
+
+    def setup(self, stage: str) -> None:
+        """Lightning hook that is called after model is initialized but before training starts."""
+        # The two ifs should be separate, but are combined due to pre-commit hook
+        if stage == "fit" and self.trainer.is_global_zero:
+            # Log variable scaling on rank 0
```
jakob-schloer: I'm not sure, I like the max_variables_to_log. It would probably be better to have a `max_hparam` in the MLflow class, but then we would have to decide which hparams to drop and which not.

training/src/anemoi/training/config/diagnostics/evaluation.yaml
```diff
@@ -57,6 +57,7 @@ log:
     expand_hyperparams: # Which keys in hyperparams to expand
       - config
     http_max_retries: 35
+    max_num_variable_scalings: 200
```
jakob-schloer: If the keyword is not in the config, it would default to 200 using `hasattr(self.config.diagnostics.log.mlflow, "max_num_variable_scalings")`

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -593,3 +596,21 @@ def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[dict]]
...
+            max_variables_to_log = (
+                self.config.diagnostics.log.mlflow.max_num_variable_scalings
+                if hasattr(self.config.diagnostics.log.mlflow, "max_num_variable_scalings")
+                else 200
```
anaprietonem: this 200 could be a constant defined for mlflow under the training/diagnostics/mlflow/logger.py as that's just relevant for this

training/src/anemoi/training/train/tasks/base.py
anaprietonem: There is now a max `max_params_length` that takes care of limiting the length of parameters coming from the provenance/lineage tracking (so all the metadata) and the config entries. We could move the

```
    config_params = OmegaConf.to_container(convert_to_omegaconf(config), resolve=True)

    logger.log_hyperparams(
        config_params,
        expand_keys=config.diagnostics.log.mlflow.expand_hyperparams,
    )

```
 lines in the get_mlflow_logger function to this setup call and then stack together the config and self._scaling_values_log in one list so MAX_PARAMS_LENGTH would cover both. Thoughts?
 
 could still be good that the if statement in the setup also check that mlflow is enabled. This limit just apply to this logger

anaprietonem:
@jakob-schloer could be good if you can update the branch with main! hopefully conflicts should be minor. If you want to discuss any of my comments feel free to drop me a message or we can have a quick call! thanks again for this work!

training/src/anemoi/training/train/tasks/base.py
jakob-schloer: Very nice idea. I've combined the hyperparameters from the config and the loss_scaling now in the `setup` function and log them there. This way the `max_params_length` is respected and an additional keyword in the config is not needed.

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -761,3 +765,22 @@ def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[dict]]
...
+
+    def setup(self, stage: str) -> None:
+        """Lightning hook that is called after model is initialized but before training starts."""
+        # The conditions should be separate, but are combined due to pre-commit hook
+        if stage == "fit" and self.trainer.is_global_zero and self.logger is not None:
+            # Log hyperparameters on rank 0
+            hyper_params = OmegaConf.to_container(convert_to_omegaconf(self.config), resolve=True)
+            hyper_params.update({"variable_loss_scaling": self._scaling_values_log})
+            # Expand keys for better visibility
+            expand_keys = OmegaConf.select(
```
anaprietonem, line 777: Didn't know you could do this with OmegaConf, just curious was this for readability or  did you encounter any issues if calling?
```
            self.logger.log_hyperparams(
                hyper_params,
                expand_keys=self.config.diagnostics.log.mlflow.expand_hyperparams,
            )
```

anaprietonem (approved):
LGTM Jakob, left a comment just to clarify. Otherwise good to be merged, thanks!

training/src/anemoi/training/train/tasks/base.py
jakob-schloer: I've only added that logic in case someone is not using an mlflow logger. In that case, the keyword in the config might not exist, and expand_keys would default to `['config]`.

PR #329, opened 2025-05-22
add flash attention option to SDPA

cathalobrien:
## Description

this PR improves the performance and memory usage of the transformer when the flash attention library isnt installed

currently if the flash attention library isnt installed, we fall back to torches SDPA (scaled dot product attention) function. This has function has multiple algorithms. We currently hardcode the use of the MATH backend, a c++ implementation of the attention algorithm. This is expensive in memory as it must materialise the full seq_len ^2 array.

This PR just lets the PR use the "Flash Attention" algorithm to SDPA, if it is available. If not, it falls back to the MATH backend, same behavior as before. This gave a 10x improvement to running the transformer model on CPU. and it enables larger models to be run as it is more memory efficient.

## Type of Change

-   [ ] Documentation update

## Code Compatibility

-   [x] I have run the Benchmark Profiler against the old version of the code

### Dependencies

### Documentation

### DIFF
--- a/models/src/anemoi/models/layers/attention.py
+++ b/models/src/anemoi/models/layers/attention.py
@@ -252,7 +252,7 @@ class SDPAAttentionWrapper(nn.Module):
         if window_size is not None and (self.mask is None or tuple(self.mask.shape) != (sequence_len, sequence_len)):
             self.update_mask(sequence_len, window_size=window_size, device=query.device)
 
-        with torch.nn.attention.sdpa_kernel(backends=[torch.nn.attention.SDPBackend.MATH]):
+        with torch.nn.attention.sdpa_kernel(backends=[torch.nn.attention.SDPBackend.FLASH_ATTENTION, torch.nn.attention.SDPBackend.MATH]):
             out = self.attention(
                 query,
                 key,


anaprietonem:
@cathalobrien when is the SDPA algorithm available? will the MATH backend used then in the case of the transformer CI runners?

models/src/anemoi/models/layers/attention.py
```diff
@@ -252,7 +252,9 @@ def forward(
         if window_size is not None and (self.mask is None or tuple(self.mask.shape) != (sequence_len, sequence_len)):
             self.update_mask(sequence_len, window_size=window_size, device=query.device)
 
-        with torch.nn.attention.sdpa_kernel(backends=[torch.nn.attention.SDPBackend.MATH]):
+        with torch.nn.attention.sdpa_kernel(
+            backends=[torch.nn.attention.SDPBackend.FLASH_ATTENTION, torch.nn.attention.SDPBackend.MATH]
+        ):
             out = self.attention(
```
anaprietonem, line 258: could be good to add some logging to show which of the two algorithms is running

cathalobrien: Would be nice but there doesnt seem to be an easy way to query which algorithm is used. But at runtime, it will try flash attention first, and then if FA can't be used will print a warning why and fallback to MATH

anaprietonem:
The configs are under `training/tests/integration/config/` and https://github.com/ecmwf/anemoi-core/blob/main/training/tests/integration/config/testing_modifications.yaml contains the modifications we use across LAM, stretched grid and global. This is the script run for the training cycle https://github.com/ecmwf/anemoi-core/blob/main/training/tests/integration/test_training_cycle.py. 

But pinning @VeraChristina @theissenhelen as they might have suggestions on how we can include this in the configs!

VeraChristina:
This is great, thanks! 
For context, the integration tests in training are using pytest (not the CLI which we are planning to test in the system-level tests that span the anemoi pipeline), with the relevant pieces being the files Ana linked. 

To include the transformer architecture in the tests, we'll want to change the parametrization of the following fixture in training/tests/integration/conftest.py like so:

@pytest.fixture(
    params=[
        ["model=gnn"],
        ["model=graphtransformer"],
        ["model=transformer", "graph=encoder_decoder_only", "model.processor.attention_implementation=scaled_dot_product_attention"],
    ],
)
def architecture_config_with_data(
...

(This assumes that SDPA also works on GPU (?), and that we test that since we currently don't install flash attention in the CI.) However, my experience with setting up these tests is that they require some debugging because configs, which aren't yet included in tests, tend to diverge from the code. Very happy for you to have a look but equally happy for this to be merged and the transformer test being added in a separate PR

cathalobrien:
> This is great, thanks! For context, the integration tests in training are using pytest (not the CLI which we are planning to test in the system-level tests that span the anemoi pipeline), with the relevant pieces being the files Ana linked.
> 
> To include the transformer architecture in the tests, we'll want to change the parametrization of the following fixture in training/tests/integration/conftest.py like so:
> 
> @pytest.fixture( params=[ ["model=gnn"], ["model=graphtransformer"], ["model=transformer", "graph=encoder_decoder_only", "model.processor.attention_implementation=scaled_dot_product_attention"], ], ) def architecture_config_with_data( ...
> 
> (This assumes that SDPA also works on GPU (?), and that we test that since we currently don't install flash attention in the CI.) However, my experience with setting up these tests is that they require some debugging because configs, which aren't yet included in tests, tend to diverge from the code. Very happy for you to have a look but equally happy for this to be merged and the transformer test being added in a separate PR.
> 
> Also just for context, here are the two workflow files:
> 
>     * for the hpc workflow: https://github.com/ecmwf/anemoi-core/blob/main/.github/workflows/integration-tests-hpc.yml
> 
>     * for the github runner workflow: https://github.com/ecmwf/anemoi-core/blob/main/.github/workflows/integration-tests.yml

Hey Vera, Sure i'll add a transformer test now

training/src/anemoi/training/config/model/transformer.yaml
```diff
@@ -57,7 +57,6 @@ processor:
   window_size: 512
   dropout_p: 0.0 # GraphTransformer
   attention_implementation: flash_attention # flash_attention, scaled_dot_product_attention
-  qk_norm: False # Transformer and GraphTransformer only
```
VeraChristina, line 33: Why do we remove this? It now seems to fail the pydantic validation

cathalobrien: Before the field was there twice, which would lead to an error at runtime (see the file before this commit: https://github.com/ecmwf/anemoi-core/blob/2cea7db51d5c5ef63bb4b9c266deb05fb2acf66f/training/src/anemoi/training/config/model/transformer.yaml#L60 and https://github.com/ecmwf/anemoi-core/blob/2cea7db51d5c5ef63bb4b9c266deb05fb2acf66f/training/src/anemoi/training/config/model/transformer.yaml#L64). So I removed one copy of the duplicated field. strange that it would lead to pedantic validation failing?

VeraChristina: Fair enough, I missed that in the github view. Strange indeed! Let me pull again and make sure I am testing the right code

VeraChristina: My bad -- they pass locally now

models/src/anemoi/models/layers/attention.py
anaprietonem: Thanks @cathalobrien , could we at least add a comment about the logic in the code and to explain those kernels as just available for the SPDA?

anaprietonem:
@VeraChristina @cathalobrien is this in a position to be reviewed & merged, or are any other checks/tests we want to run?

anaprietonem:
Any news/views about this PR - @cathalobrien @VeraChristina ?

cathalobrien:
it was failing locally for some reason. I'll retry them this week and hopefully we can get it merged

anaprietonem:
looking at this further:

```
import torch
from torch.nn.functional import scaled_dot_product_attention
query = torch.rand(2, 8, 10944, 1, dtype=torch.float16, device="cuda")
key = torch.rand(2, 8, 10944, 1, dtype=torch.float16, device="cuda")
value = torch.rand(2, 8, 10944, 1, dtype=torch.float16, device="cuda")
mask = torch.rand(2, 8, 10944, 1, dtype=torch.float16, device="cuda")

from torch.backends.cuda import (
    can_use_efficient_attention,
    can_use_flash_attention,
    is_flash_attention_available,
    SDPAParams,
)



q = torch.randn(10)
k = torch.randn(10)
v = torch.randn(10)
m = torch.randn(10)
s = SDPAParams(q, k, v, m, 0.1, True, False)


with torch.nn.attention.sdpa_kernel(
    backends=[torch.nn.attention.SDPBackend.MATH]
):
    print("Using flash attention math")
    sdpa_params = SDPAParams(
        query, key, value, mask, 0.1, False, False
    )
    scaled_dot_product_attention(query, key, value, mask, 0.1, False) 
    print('math backend')
    
with torch.nn.attention.sdpa_kernel(
    backends=[torch.nn.attention.SDPBackend.FLASH_ATTENTION]
):
    print("Using flash attention")
    sdpa_params = SDPAParams(
        query, key, value, None, 0.1, False, False
    )
    print(can_use_flash_attention(sdpa_params,debug=True))
    scaled_dot_product_attention(query, key, value,None, 0.1, False) #fails
```

the above example shows that the FA backed doesn't work for non-null masks so we can't use this option in our settings. We will then close this and look at installing Flash attention and test the transformer architecture using that on gpu only.

Issue #332, opened 2025-05-26
Mllfow logging optimisation


anaprietonem:
### What maintenance does this project need?

Current metrics and artifacts settings default are broad and lead to a big amount of information (some redundant) to be logged to mlflow. We should revisit the defaults and propose changes to the base settings to reduce and optimise the amount of information logged

PR #333, opened 2025-05-26
fix: update defaults for mlflow logging

anaprietonem:
## Description

PR to update the default settings regarding tracking of system metrics. New default is to set those to `False` so that if the wants to track them, they will need to activate the system metrics by updating this to True. 

Default interval for logging system metrics was every 10 seconds. This lead to a large amount of information being logged into the mlflow server. 

https://github.com/ecmwf/anemoi-core/issues/332

##  Additional notes ##


training/src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -461,7 +461,7 @@ def experiment(self) -> MLFlowLogger.experiment:
         return super().experiment
 
     @rank_zero_only
-    def log_system_metrics(self) -> None:
+    def log_system_metrics(self, interval: int = 10) -> None:
```
gmertes: Maybe also do 100 here for consistent defaults?

training/src/anemoi/training/config/diagnostics/evaluation.yaml
```diff
@@ -50,14 +50,16 @@ log:
     tracking_uri: ???
     experiment_name: 'anemoi-debug'
     project_name: 'Anemoi'
-    system: True
+    system:
+      enabled: False
+      interval: 100 #Every interval seconds, the system metrics will be collected. By default interval=10.
```
gmertes: Is it needed to say that the default here is 10 ? It could be a bit confusing ( is it 10 or 100? ). Doesn't really matter what the default of the underlying function is if we are changing it anyway?

gmertes (commented):
Nice, LGTM. Can you rename the PR (chore is ignored) so that it gets included in the release notes. Maybe `feat(config)` ?

anaprietonem:
@gmertes  I renamed the PR to 'fix'. And following up from our conversation, I have simplified the logic so that the interval is kept to 100 second for system metrics and the default is not to log them

gmertes (approved):
Thanks! Should reduce the logging overhead significantly.

Issue #224, opened 2025-04-03
Missing check on variable ordering when doing transfer learning


icedoom888:
### What happened?

When loading a checkpoint anemoi will NOT load any model weights that have a different shape than the current network structure.
This is informative if the number of variable sis different as it will not load most weights from the parent checkpoint.

Unfortunately no issue is raised if the ordering of the dataset variables is different.
This will cause the model to get to training and the losses become NaNs without throwing any issues.

### What are the steps to reproduce the bug?

Load a checkpoint with the same number of variables.
Use the "selected_vars" to specify an order of variables:

```
selected_vars:
  - 10u
  - 10v
  - 2d
  - 2t
  - cos_julian_day
  - cos_latitude
  - cos_local_time
  - cos_longitude
  - insolation
  - lsm
  - msl

  - q_100
  - q_1000
  - q_150
  - q_200
  - q_250
  - q_300
  - q_400
  - q_50
  - q_500
  - q_700
  - q_850
  - q_925

  - sin_julian_day
  - sin_latitude
  - sin_local_time
  - sin_longitude
  - sp

  - skt
  - tp
  
  - t_100
  - t_1000
  - t_150
  - t_200
  - t_250
  - t_300
  - t_400
  - t_50
  - t_500
  - t_700
  - t_850
  - t_925

  - u_100
  - u_1000
  - u_150
  - u_200
  - u_250
  - u_300
  - u_400
  - u_50
  - u_500
  - u_700
  - u_850
  - u_925

  - v_100
  - v_1000
  - v_150
  - v_200
  - v_250
  - v_300
  - v_400
  - v_50
  - v_500
  - v_700
  - v_850
  - v_925

  - z
  - z_100
  - z_1000
  - z_150
  - z_200
  - z_250
  - z_300
  - z_400
  - z_50
  - z_500
  - z_700
  - z_850
  - z_925
```

### Version

graphs-0.5.0

### Platform (OS and architecture)

Linux balfrin-ln003 5.14.21-150400.24.81_12.0.87-cray_shasta_c #1 SMP Sun Dec 17 12:59:08 UTC 2023 (e30c7c1) x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell

```

### Organisation

MeteoSwiss

anaprietonem:
[...]

PR #334, opened 2025-05-26
fix: dataset_order

mc4117:
Currently if load_weights_only=True, the order of the variables in the dataset pre and post loading the checkpoint is not being checked. This change fixes this bug

This has already been flagged in issue #224 


models/src/anemoi/models/data_indices/collection.py
```diff
@@ -127,6 +130,81 @@ def todict(self):
     def representer(dumper, data):
         return dumper.represent_scalar(f"!{data.__class__.__name__}", repr(data))
 
+    def _compare_variables(self, ckpt_name_to_index: dict[str, int], data_name_to_index: dict[str, int]) -> None:
```
floriankrb: If this becomes public (used by anemoi-training), you should remove the '_' : `_compare_variables` -> `compare_variables`.

Also, you will need to choose between a method `data_indices.compare_variables(name_to_index)` or a function `compare_variables(checkpoint_name_to_index, data_name_to_index)`. 
I would vote for a function as there are not many "self" in this piece of code.

anaprietonem: @floriankrb any views on where such function should live?

training/src/anemoi/training/train/train.py
```diff
@@ -199,18 +199,19 @@ def model(self) -> pl.LightningModule:
...
-            if hasattr(self.config.training, "transfer_learning"):
-                # Sanify the checkpoint for transfer learning
-                if self.config.training.transfer_learning:
-                    LOGGER.info("Loading weights with Transfer Learning from %s", self.last_checkpoint)
-                    model = transfer_learning_loading(model, self.last_checkpoint)
-                else:
-                    LOGGER.info("Restoring only model weights from %s", self.last_checkpoint)
-                    model = model_task.load_from_checkpoint(self.last_checkpoint, **kwargs, strict=False)
-
+            ckpt_data_idx = torch.load(self.last_checkpoint, weights_only=False)["hyper_parameters"][
```
anaprietonem: What options do we have to avoid reloading the checkpoint here again? I was wondering if we could define as part of the forecaster a `on_load_checkpoint` (https://lightning.ai/docs/pytorch/stable/common/lightning_module.html#lightningmodule) where we get the information about the data_indices. The main difference would be that is a method of the lightning module not the callback. So it _should_ be triggered when doing `model_task.load_from_checkpoint`.

models/src/anemoi/models/data_indices/collection.py
mc4117: I have changed the name and changed it to a function

training/src/anemoi/training/train/train.py
mc4117: Thanks! I have edited the code to use on_load_checkpoint, so the checkpoint only gets loaded once

training/src/anemoi/training/train/train.py
```diff
@@ -215,19 +215,19 @@ def model(self) -> pl.LightningModule:
...
-                    LOGGER.info("Restoring only model weights from %s", self.last_checkpoint)
-                    model = model_task.load_from_checkpoint(self.last_checkpoint, **kwargs, strict=False)
-
+            # Sanify the checkpoint for transfer learning
+            if self.config.training.transfer_learning:
+                LOGGER.info("Loading weights with Transfer Learning from %s", self.last_checkpoint)
+                model = transfer_learning_loading(model, self.last_checkpoint)
             else:
                 LOGGER.info("Restoring only model weights from %s", self.last_checkpoint)
+                kwargs.pop("data_indices")
```
anaprietonem, line 226: could we add a comment about the logic for popping the 'data_indices'?

training/tests/unit/diagnostics/callbacks/test_variable_order.py
```diff
@@ -44,215 +45,142 @@ def name_to_index_rename_permute() -> dict:
 @pytest.fixture
 def fake_trainer(mocker: Any, name_to_index: dict) -> AnemoiTrainer:
     trainer = mocker.Mock(spec=AnemoiTrainer)
```
anaprietonem, line 49: Could it be possible to have a test with `load_weights_only` True, and another one with `transfer learning`? Asking, cause I am assume the ones here currently are based on the case on resuming/forking runs, but let me know if that's not the case

mc4117: I have added a test to check that on_load_checkpoint works as expected

mc4117:
@icedoom888 We think this fix should fix your issue. Would you be able to test this with the transfer_learning flag and let us know so we can merge the PR?

training/src/anemoi/training/diagnostics/callbacks/sanity.py
```diff
@@ -77,9 +51,14 @@ def on_validation_epoch_start(self, trainer: pl.Trainer, _: pl.LightningModule)
         """
         data_name_to_index = trainer.datamodule.ds_valid.name_to_index
 
-        self._compare_variables(data_name_to_index)
+        if hasattr(trainer.model.module, "_ckpt_model_name_to_index"):
+            self._model_name_to_index = trainer.model.module._ckpt_model_name_to_index
```
fprill, line 55: Hi Mariana (@mc4117), first of all: thanks for this fix! It has been in the code now since June but I hadn't noticed so far.

But I have one comment/question. The `CheckVariableOrder` instance now explicitly depends on the model being wrapped by `DDPStrategy`. When using `DDPStrategy`, `trainer.model` refers to the `DistributedDataParallel` wrapper. This wrapper has an attribute, `.module`, which would not otherwise be present. This seems to be the only issue in Anemoi that prevents us from running a single-GPU interactive Jupyter notebook. Do you think we could find a way around using the `.module` attribute?

mc4117: Hi Florian @fprill , I think the line where this name gets instantiated is here https://github.com/ecmwf/anemoi-core/blob/2b6c2055c0155b6cf5e58355f0358245dafd7705/training/src/anemoi/training/train/tasks/base.py#L305. Do you not have a GraphModule in your set-up? Or does it get called in a different way?

fprill: Hi, thanks for your reply! I guess I didn't express myself clearly... I think that in Anemoi the attribute `trainer.model.module` does not exist in certain environments, because `.module` is created only by distributed wrappers like `anemoi.training.distributed.strategy.DDPGroupStrategy`.  Therefore, e.g. when running Anemoi in a Jupyter notebook, you get  `object has no attribute 'module'`. 
My point here was that this dataset check is the only place in Anemoi that accesses `.module` and therefore prevents users from running Anemoi in Jupyter notebooks.

mc4117:
Ah I understand a bit better now! Where would you suggest this check goes instead? When we were doing this PR, we struggled a bit to find a good place to put it. Are there other parts of the code that are not instantiated when running in the interactive notebook?

mc4117:
You mean add a comment to the change log?

PR #335, opened 2025-05-27
refactor(models): bounding function with buffer

sahahner:
## Description
The minimum values used in the NormalizedReluBounding are now saved as a buffer. Therefore, we can avoid to move them to the device manually.

On main, the minimum values are moved manually in the forward function to the device.

-

##  Additional notes ##
-


sahahner:
moved to https://github.com/ecmwf/anemoi-core/pull/315

PR #339, opened 2025-05-28
fix(graphs): fix device mismatch error

JPXKQX:
## Description

This PR resolves an issue with mismatches between different devices that was raised during integration tests using the GPU.

Error trace raised by the integration tests.
```
    def get_raw_values(self, nodes: NodeStorage, **kwargs) -> torch.Tensor:
        assert self.mask_node_attr_name in nodes, f"Node attribute '{self.mask_node_attr_name}' not found in nodes."
        attr_values = super().get_raw_values(nodes, **kwargs)
        mask = nodes[self.mask_node_attr_name].squeeze()
>       return attr_values * mask
E       RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

No issue created

##  Additional notes ##

### DIFF
--- a/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
+++ b/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
@@ -214,7 +214,7 @@ class MaskedPlanarAreaWeights(PlanarAreaWeights):
     def get_raw_values(self, nodes: NodeStorage, **kwargs) -> torch.Tensor:
         assert self.mask_node_attr_name in nodes, f"Node attribute '{self.mask_node_attr_name}' not found in nodes."
         attr_values = super().get_raw_values(nodes, **kwargs)
-        mask = nodes[self.mask_node_attr_name].squeeze()
+        mask = nodes[self.mask_node_attr_name].squeeze().to(self.device)
         return attr_values * mask
 
 


graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -214,7 +214,7 @@ def __init__(
     def get_raw_values(self, nodes: NodeStorage, **kwargs) -> torch.Tensor:
         assert self.mask_node_attr_name in nodes, f"Node attribute '{self.mask_node_attr_name}' not found in nodes."
         attr_values = super().get_raw_values(nodes, **kwargs)
```
aaron-hopkinson:
```suggestion
        attr_values = super().get_raw_values(nodes, **kwargs).to(self.device)
```

I'm not sure the proposed change fixes the issue.

When I run the failing acceptance test myself with `pytest training/tests/integration -k test_training_cycle_stretched --longtests` I still get a failing test.

If I look at `mask.device` and `attr_values.device` (without your change), I get that `mask` is on `cuda`, and `attr_values` is on `cpu` (`self.device` is `cuda`) – so I _think_ the correct change is to leave mask where it is (as the `.to` has no effect), and instead move `attr_values`.

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -214,7 +214,7 @@ def __init__(
...
-        mask = nodes[self.mask_node_attr_name].squeeze()
+        mask = nodes[self.mask_node_attr_name].squeeze().to(self.device)
```
aaron-hopkinson:
```suggestion
        mask = nodes[self.mask_node_attr_name].squeeze()
```

aaron-hopkinson (changes requested):

aaron-hopkinson (approved):
Integration tests pass for me on a GPU. Merging

Issue #346, opened 2025-06-02
Decoder keyword `initialise_data_extractor_zero` cannot be used with boundings.


jakob-schloer:
### What happened?

When the `config.model.decoder.initialise_data_extractor_zero = True` the weights of the last layer are initialized to zero which leads to zeros in the diagnostic output fields after initialization (prognostic fields are residual and thus are not zero). In combination with some bounding, e.g. ReLuBounding, the gradients are zero and the fields are never learned. 

### What are the steps to reproduce the bug?

In the config set: 

```
data:
  diagnostic:
    - tp

model: 
  decoder:
    ...
    initialise_data_extractor_zero = True

  bounding: #These are applied in order
    - _target_: anemoi.models.layers.bounding.ReluBounding #[0, infinity)
       variables:
         - tp
```

Tp will always be zero, see mlflow run: 
[uniform_decoder_initialization](https://mlflow.ecmwf.int/#/experiments/266/runs/71c69667271e4e34bd04b7601f6795e8/artifacts)
[zero_decoder_initialization](https://mlflow.ecmwf.int/#/experiments/266/runs/f3d8bad407b44b2d9e38f0a70c5aa0b9/artifacts)

### Suggestion for fix

Create a big warning that you cannot use zero-weight initialization with bounding, potentially in Pydantic. 

### Version

v0.5.1

### Platform (OS and architecture)

ATOS

### Organisation

ECMWF

PR #349, opened 2025-06-03
feat!: Explicit weight initialization.

jakob-schloer:
## Description
The weight initialization is often hidden in the submodules, or torch defaults are assumed. To make this more explicit and be able to easily try other weight initialization, an `_init_weights()` function is added to the highest torch.nn.Module level.

```
class Model(nn.Module):
    def __init__(self, ...):
        ...
        self.apply(self._init_weights)

    def _init_weights(self, m):
        """Initialize weights for modules."""
        if isinstance(m, nn.Linear):
            nn.init.trunc_normal_(m.weight, std=0.02)
            if m.bias is not None:
                nn.init.constant_(m.bias, 0)
       ...
```

Examples for potential training improvement:
 - ViTs typically use truncated normal initialization e.g. https://github.com/microsoft/Swin-Transformer
- GraphAttention typically uses xavier uniform initialization (https://arxiv.org/abs/1710.10903)

https://github.com/ecmwf/anemoi-core/issues/346 

##  TODOs

- [x] test weight initialization for AnemoiModelEncProcDec
- [ ] test weight initialization for AnemoiEnsModelEncProcDec
- [ ] test weight initialization for AnemoiModelEncProcDecInterpolator
- [ ] test weight initialization for AnemoiModelEncProcDecHierarchical
- [ ] add tests
- [ ] try good default values  
- [ ] include`custom_weight_initializations` beyond zero initialization

### DIFF
--- a/models/src/anemoi/models/models/encoder_processor_decoder.py
+++ b/models/src/anemoi/models/models/encoder_processor_decoder.py
@@ -23,6 +23,10 @@ from torch_geometric.data import HeteroData
 
 from anemoi.models.distributed.shapes import get_shape_shards
 from anemoi.models.layers.graph import NamedNodesAttributes
+from anemoi.models.layers.graph import TrainableTensor
+from anemoi.models.layers.normalization import AutocastLayerNorm
+from anemoi.models.layers.utils import apply_zero_init
+from anemoi.models.layers.utils import mark_module_for_zero_init
 from anemoi.utils.config import DotDict
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -134,6 +138,51 @@ def __init__(
...
+                nn.init.constant_(m.trainable, 0)
+        elif isinstance(m, nn.Embedding):
+            # Initialize embeddings with truncated normal initialization
+            nn.init.trunc_normal_(m.weight, std=0.02)
+        elif isinstance(m, (nn.Conv1d, nn.Conv2d, nn.Conv3d)):
+            # GraphAttention typically uses xavier uniform initialization (https://arxiv.org/abs/1710.10903)
+            nn.init.xavier_uniform_(m.weight)
+            if m.bias is not None:
+                nn.init.constant_(m.bias, 0)
+
```
roxas6path, line 185: So interesting PR  @jakob-schloer - all looks good to me. 
Your current implementation essentially fixes default behaviour for specific types of Layers. A couple questions / things to clarify / get your opinion on?

1) Context: I understand self.apply effects all children parameters of the parent nn.Module, and the _skip_init property must be added to each submodule by the user. 

Suggestion: The modules to be skipped should be defined in the model config in a similar fashion to how you allow modules to be named for the custom_weight_initialization: section of the model.yaml

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -134,6 +138,51 @@ def __init__(
             ]
         )
 
+        # Weight initialization
+        LOGGER.info("Initializing weights!")
+        self.apply(self._init_weights)
+
+        # Zero initialization for specific modules if needed
```
roxas6path, line 145: Let me know what you think: @jakob-schloer 

*Context* : Your _init_weights function allows a user to opt-out of a pre-defined set of *optimal* initialisation strategies based on the weight type.

This custom_weight_initialisation strategy currently allows a user to define a custom initialisation strategy for any given module/parameter. 
Context: Currently I can see you have only written support zero_initialisation? I'm assuming since it is the one we that existed already in our model? But based on #346 that you raised I believe zero_initialisation should be removed from our model?

*Suggestion*: If zero_init is no longer a default or will not be in use, then I suggest this custom_weight_initialisation is written to support any init strategy from the torch.nn.init. e.g. via the use of \_target_\ and kwargs appropriate for that specific torch init strategy. This zero init case could be covered with: \_target_\: torch.nn.init.zeros_  in the config

Rilwan-Adewoyin (commented):
Couple comments - overall PR lgtm

models/src/anemoi/models/models/encoder_processor_decoder.py
jakob-schloer: Thanks for having a look already @Rilwan-Adewoyin.
I agree that the `custom_weight_initialization` requires more work. I like your idea to use `__target__` instantiation. I will give it a try in the next days.

models/src/anemoi/models/models/encoder_processor_decoder.py
jakob-schloer: In a model with skip connections, it makes sense to initialize the last layer with zero, which is also the gold standard. My approach would allow the zero_initialization of the last layer but throw an error if bounding is set using pydantic (probably in a different PR).

jakob-schloer:
> @jakob-schloer Is this a PR you are still looking to merge?

I still believe it would be good to have the weight initialization explicit and try how different initialization impact training performance. I however don't think, I will have time to work on that in the next months. We could close the Draft PR, but keep the branch.

mchantry:
I think leaving as an open draft is good. Easier to find that way.

PR #352, opened 2025-06-04
fix: config validator checks for conflict with bounding and graphtran…

### LINKED ISSUE #346: Decoder keyword `initialise_data_extractor_zero` cannot be used with boundings.

theissenhelen:
## Description
This PR introduces a new validator in the config to check whether the config.model.decoder.initialise_data_extractor_zero = True is not used with any bounding. 

Closes #346 

This is mainly a concern for the GraphTransformer in the decoder, see issue #346.

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/schemas/base_schema.py
+++ b/training/src/anemoi/training/schemas/base_schema.py
@@ -89,6 +89,21 @@ class BaseSchema(BaseModel):
             raise PydanticCustomError("logger_path_missing", msg)  # noqa: EM101
         return self
 
+    @model_validator(mode="after")
+    def check_bounding_not_used_with_data_extractor_zero(self) -> BaseSchema:
+        """Check that bounding is not used with zero data extractor."""
+        if (
+            self.model.decoder.target_ == "anemoi.models.layers.mapper.GraphTransformerBackwardMapper"
+            and self.model.decoder.initialise_data_extractor_zero
+            and self.model.bounding is not None
+        ):
+            error = "bounding_conflict_with_data_extractor_zero"
+            msg = "Boundings cannot be used with zero data extractor."
+            raise PydanticCustomError(
+                error,
+                msg,
+            )
+
     def model_dump(self, by_alias: bool = False) -> dict:
         dumped_model = super().model_dump(by_alias=by_alias)
         return DictConfig(dumped_model)


jakob-schloer (dismissed):
Looks good to me. I ran it once with conflicting settings and got a nice error.

training/src/anemoi/training/schemas/base_schema.py
```diff
@@ -89,6 +89,24 @@ def check_log_paths_available_for_loggers(self) -> BaseSchema:
             raise PydanticCustomError("logger_path_missing", msg)  # noqa: EM101
         return self
 
+    @model_validator(mode="after")
+    def check_bounding_not_used_with_data_extractor_zero(self) -> BaseSchema:
+        """Check that bounding is not used with zero data extractor."""
+        if (
+            self.model.decoder.target_ == "anemoi.models.layers.mapper.GraphTransformerBackwardMapper"
```
gmertes: If this symbol is ever moved or renamed and we forget to update this string comparison, this check will silently pass all the time. How should we handle that? Ideally we should check if this symbol actually exists, but we also don't want to import it if it's not being used. 🤔

Don't let this block the PR, it looks good to merge. Just something to think about.

PR #354, opened 2025-06-04
feat(graphs): Add target-centric EdgeBuilders 

matschreiner:
## Description
This PR introduces support for constructing KNN and CutOffRadius edge indices using a target-centric search while preserving the edge direction from source → target. In addition to the existing KNNEdges and CutOffRadiusEdges (which search per‐source), it adds two new edge‐builder classes:

`ReversedKNNEdges`: For each target node, find its k nearest source nodes, then create edges from each of those sources into the target.

`ReversedCutOffRadiusEdges`: For each target node, find all source nodes within a given radius, then create edges from each of those sources into the target.

Both new classes have unit tests and documentation.

This is a new feature that allows to build KNN and CutOffRadius edge indices based on the target nodes but still with the edge direction going from source -> target

Initially, we developed this feature to support research on blurred skip connections, but later discovered that a different approach would be needed for that research. Since the functionality is implemented, tested, and documented, it would be a shame to binning it. :) 

----
📚 Documentation preview 📚: https://anemoi-training--354.org.readthedocs.build/en/354/

----
📚 Documentation preview 📚: https://anemoi-graphs--354.org.readthedocs.build/en/354/

----
📚 Documentation preview 📚: https://anemoi-models--354.org.readthedocs.build/en/354/

### DIFF
--- a/graphs/docs/graphs/edges/knn.rst
+++ b/graphs/docs/graphs/edges/knn.rst
@@ -25,3 +25,16 @@ YAML configuration:
 
    The ``KNNEdges`` method is recommended for the decoder edges, to
    connect all target nodes with the surrounding source nodes.
+
+There is also a reversed KNN edges variant that will connect each target
+node to its k nearest source nodes, but the resulting edge direction is
+still from the source nodes to the target nodes.
+
+.. code:: yaml
+
+   edges:
+     -  source_name: source
+        target_name: target
+        edge_builders:
+        - _target_: anemoi.graphs.edges.ReversedKNNEdges
+          num_nearest_neighbours: 3 :contentReference[oaicite:5]{index=5}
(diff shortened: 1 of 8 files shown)


graphs/docs/graphs/edges/knn.rst
```diff
@@ -25,3 +25,16 @@ YAML configuration:
...
+still from the source nodes to the target nodes.
+
+.. code:: yaml
+
+   edges:
+     -  source_name: source
+        target_name: target
+        edge_builders:
+        - _target_: anemoi.graphs.edges.ReversedKNNEdges
+          num_nearest_neighbours: 3 :contentReference[oaicite:5]{index=5}
```
JPXKQX, line 40:
```suggestion
          num_nearest_neighbours: 3
```

JPXKQX:
Hi Mathias, thanks for opening the PR. This will need further testing, as I don't think this will work when you specify source/target_mask_attr_names

PR #356, opened 2025-06-05
Fix pre postprocessors in rollout training

sahahner:
##  Additional notes ##


sahahner:
> One general comment: I think we can see that our current pre / post processor design might not be best suited for in-rollout imputation. Instead of having to implement "if in_advance_input" it would be better to have the option to only call the imputer -> replace NaNs without normalisation / de-normalisation. We might want to call these in a different context in the future, so more flexibility here would be good.

Yes, that would indeed be practical. How would you solve this for our current setup where we can have several imputers and not only one?

ssmmnn11:
> > One general comment: I think we can see that our current pre / post processor design might not be best suited for in-rollout imputation. Instead of having to implement "if in_advance_input" it would be better to have the option to only call the imputer -> replace NaNs without normalisation / de-normalisation. We might want to call these in a different context in the future, so more flexibility here would be good.
> 
> Yes, that would indeed be practical. How would you solve this for our current setup where we can have several imputers and not only one?

We could split imputation and normalisation for a start. Then we could call them independently instead of only having pre_processing etc.

sahahner:
That would indeed allow us to avoid an if statement in the transform functions. I am not sure whether pulling the imputer(s) out of the pre/postprocessors object of the model is the most elegant solution, since for many users they are probably not part of the default configuration.

A regard: do we run into the danger, that we will accumulate more pre/postprocessors objects in the model over time?

With this PR I was originially also thinking about applying the postprocessors in the rollout step to resemble the behaviour of an inference predict step. A flag would give us the chance to implement this later, but the application of the imputer should be in main soon, so I am happy to focus on that for now.

ssmmnn11:
> That would indeed allow us to avoid an if statement in the transform functions. I am not sure whether pulling the imputer(s) out of the pre/postprocessors object of the model is the most elegant solution, since for many users they are probably not part of the default configuration.
> 
I don’t see how this would be problematic for them

> A concern: do we run into the danger, that we will accumulate more pre/postprocessors objects in the model over time?
> 
I don’t think so. Why do you think that would happen?

> With this PR I was originially also thinking about applying the postprocessors in the rollout step to resemble the behaviour of an inference predict step. A flag would give us the chance to implement this later, but the application of the imputer should be in main soon, so I am happy to focus on that for now.

My main point would be to make the structure more explicit instead of having the flags inside of the objects.

PR #363, opened 2025-06-05
fix: SphericalAreaWeights

### LINKED ISSUE #362: SphericalAreaWeights not working as intended for high resolution stretched grid
### What happened?

SphericalAreaWeights do not work as intended with a high resolution stretched grid setup (n320 global, 2.5km regional resolution). 

Image

This can be fixed by not casting the latitude / longitude coordinates to float32 (keeping them as float64). I would guess this bug happens because SphericalVoronoi's internal precision depends on that of the input data. Correct result when not casting to float32:

Image

Create a graph using cutout with the default stretched grid graph config and 
```yaml
  dataset:
    cutout:
      - dataset: aifs-meps-2.5km-2020-2023-6h-v7
        trim_edge: 50
      - dataset: aifs-ea-an-oper-0001-mars-n320-1979-2022-6h-v6
    min_distance_km: 0
    adjust: all
```

### Version

v0.6.0

Leonardo

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

MetNorway

havardhhaugen:
## Description
Fixes #362 

SphericalAreaWeights needs float64 precision input to work as intended for high resolution graphs

#362 

##  Additional notes ##

### DIFF
--- a/graphs/src/anemoi/graphs/generate/transforms.py
+++ b/graphs/src/anemoi/graphs/generate/transforms.py
@@ -32,6 +32,28 @@ def cartesian_to_latlon_rad(xyz: np.ndarray) -> np.ndarray:
     return np.array((lat, lon), dtype=np.float32).transpose()
 
 
+def latlon_rad_to_cartesian_np(locations: np.ndarray, radius: float = 1) -> np.ndarray:
+    """Convert planar coordinates to 3D coordinates in a sphere.
+
+    Parameters
+    ----------
+    locations : np.ndarray of shape (N, 2)
+        The 2D coordinates of the points, in radians.
+    radius : float, optional
+        The radius of the sphere containing los points. Defaults to the unit sphere.
+
+    Returns
+    -------
+    np.ndarray of shape (N, 3)
+        3D coordinates of the points in the sphere.
+    """
+    latr, lonr = locations[..., 0], locations[..., 1]
+    x = radius * np.cos(latr) * np.cos(lonr)
+    y = radius * np.cos(latr) * np.sin(lonr)
+    z = radius * np.sin(latr)
+    return np.stack((x, y, z), axis=-1)
+
+
 def latlon_rad_to_cartesian(locations: torch.Tensor, radius: float = 1) -> torch.Tensor:
     """Convert planar coordinates to 3D coordinates in a sphere.
 
--- a/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
+++ b/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
@@ -97,7 +97,7 @@ class PlanarAreaWeights(BaseNodeAttribute):
     """
 
     def get_latlon_coordinates(self, nodes: NodeStorage) -> torch.Tensor:
-        return nodes.x
+        return nodes.x.to(torch.float64)
 
     def _compute_mean_nearest_distance(self, points: np.ndarray) -> float:
         """Compute mean distance to nearest neighbor for each point.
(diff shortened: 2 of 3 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=363) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/anemoi-core?pullRequest=363) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/anemoi-core?pullRequest=363) it.</sub>

JPXKQX:
Thanks Havard! I think it could make sense to have something similar to the node & edge attribute builders, where we have a class attribute, `dtype`, but I still have some concerns. Would it make sense to have it as an attribute of the GraphCreator (or the config file), any thoughts?

I think it could make sense to set float64 as the default dtype. I wouldn't cast the `x` attribute in the `update_graph()` function as it may raise MismatchErrors for users of the low-level API

havardhhaugen:
> Thanks Havard! I think it could make sense to have something similar to the node & edge attribute builders, where we have a class attribute, `dtype`, but I still have some concerns. Would it make sense to have it as an attribute of the GraphCreator (or the config file), any thoughts?
> 
> I think it could make sense to set float64 as the default dtype. I wouldn't cast the `x` attribute in the `update_graph()` function as it may raise MismatchErrors for users of the low-level API

I added a get_latlon_coordinates method to SphericalAreaWeights, which casts to float64. This seemed better than implementing something for the base class since I could not see any of the other NodeAttributes classes using nodes.x.

JPXKQX:
> 
> I added a get_latlon_coordinates method to SphericalAreaWeights, which casts to float64. This seemed better than implementing something for the base class since I could not see any of the other NodeAttributes classes using nodes.x.

I agree, it seems like a better solution. Do you think it would make sense to abstract some of the functionality to a `BaseAreaWeights` class? I was thinking of something like the following, but I am not sure if the extra changes are worth it

```
class BaseAreaWeights(BaseAreaWeights, ABC):
    def get_latlon_coordinates(self, nodes: NodeStorage) -> torch.Tensor:
        return nodes.x
    
    @abstractmethod
    def compute_area_weights(self, latlons: np.ndarray) -> np.ndarray: ...

    def get_raw_values(self, nodes: NodeStorage, **kwargs) -> np.ndarray:
        latlons = self.get_latlon_coordinates(nodes).cpu().numpy()
        area_weights = self.compute_area_weights(latlons)
        return torch.from_numpy(area_weights)
```

havardhhaugen:
> > I added a get_latlon_coordinates method to SphericalAreaWeights, which casts to float64. This seemed better than implementing something for the base class since I could not see any of the other NodeAttributes classes using nodes.x.
> 
> I agree, it seems like a better solution. Do you think it would make sense to abstract some of the functionality to a `BaseAreaWeights` class? I was thinking of something like the following, but I am not sure if the extra changes are worth it
> 
> ```
> class BaseAreaWeights(BaseAreaWeights, ABC):
>     def get_latlon_coordinates(self, nodes: NodeStorage) -> torch.Tensor:
>         return nodes.x.to(torch.float64)
>     
>     @abstractmethod
>     def compute_area_weights(self, latlons: np.ndarray) -> np.ndarray: ...
> 
>     def get_raw_values(self, nodes: NodeStorage, **kwargs) -> np.ndarray:
>         latlons = self.get_latlon_coordinates(nodes).cpu().numpy()
>         area_weights = self.compute_area_weights(latlons)
>         return torch.from_numpy(area_weights)
> ```

I like that suggestion. Having to use .cpu() .numpy() inside specific methods is somewhat awkward so it would be good if we can move that to a more generic function.

havardhhaugen:
Let me know what you think about duplicating the coords_to_points function for numpy. I think I prefer this to generalizing the existing function to both torch and np with an if-statement, but maybe you have a better solution.

JPXKQX (approved):
The integration tests have passed. Thank you, Havard, for the fix!

PR #367, opened 2025-06-09
Feat/ai land extension

NRaoult:
Currently, the land component of the IFS (ecLand) is emulated using a simple MLP coded outside of Anemoi-core. 

To be able to utilise the full Anemoi workflow when emulating the land surface, we are adding a feature that would simplify the training in the anemoi-core, allowing for MLP-like training. This means effectively by-passing the graph structure by letting points be their own neighbours. This allows us to keep the single column structure of ecLand (no latent information between grids cells) and have a resolution agnostic model (i.e., if we train the model weights on an o96 dataset we can directly apply this to n320, as long as trainable_parameters is set to 0).

Next steps include by-passing the encoder/decoder structure by adding the option to have "0" num_heads.


training/pyproject.toml
```diff
@@ -56,7 +56,7 @@ dependencies = [
   "pyshtools>=4.13",
   "pytorch-lightning>=2.1",
   "timm>=0.9.2",
-  "torch>=2.2",
+  "torch>=2.2,<=2.6",
```
mchantry: Is this related to the feature?

training/src/anemoi/training/losses/__init__.py
```diff
@@ -26,5 +27,6 @@
     "MAELoss",
     "MSELoss",
     "RMSELoss",
+    "SmoothL1Loss",
```
mchantry: Ideally I think the loss should be brought in as a standalone PR.

models/src/anemoi/models/models/encoder_processor_decoder.py
mchantry, line 1: Why are there changes in this file?

models/src/anemoi/models/models/processor.py
```diff
@@ -0,0 +1,265 @@
...
+
+from anemoi.models.distributed.shapes import get_shape_shards
+from anemoi.models.layers.graph import NamedNodesAttributes
+from anemoi.utils.config import DotDict
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AnemoiModelProc(nn.Module):
+    """Message passing graph neural network."""
```
mchantry: What type of models does this support? Purely local MLPs applied to each gripoint? Would be nice for the description to be expanded.

training/src/anemoi/training/config/debug_ailand.yaml
```diff
@@ -0,0 +1,52 @@
...
+- datamodule: single
+- diagnostics: evaluation
+- hardware: example
+- graph: ailand
+- model: graphtransformer
+- training: default
+- _self_
+
+
+config_validation: False
```
theissenhelen: My recommendation would be to set this to true, so that we can ensure that the default configs and schemas are functional for users who want to check their config.

training/src/anemoi/training/losses/__init__.py
JPXKQX: Does something like https://github.com/ecmwf/anemoi-core/pull/398 make sense? @pinnstorm

models/src/anemoi/models/models/encoder_processor_decoder.py
JPXKQX: We should remove this, a separated PR was opened to delete these repeated lines

training/pyproject.toml
JPXKQX: I don't think so

models/src/anemoi/models/models/processor.py
JPXKQX: This class supports any of the previous processors (gnn, graphtransformer, transformer) and the new `PointMLP` at the data resolution, as it doesn't have an encoder/decoder.

training/src/anemoi/training/config/debug_ailand.yaml
JPXKQX: Done!

pinnstorm:
> Currently, the land component of the IFS (ecLand) is emulated using a simple MLP coded outside of Anemoi-core.
> 
> To be able to utilise the full Anemoi workflow when emulating the land surface, we are adding a feature that would simplify the training in the anemoi-core, allowing for MLP-like training. This means effectively by-passing the graph structure by letting points be their own neighbours. This allows us to keep the single column structure of ecLand (no latent information between grids cells) and have a resolution agnostic model (i.e., if we train the model weights on an o96 dataset we can directly apply this to n320, as long as trainable_parameters is set to 0).
> 
> Next steps include by-passing the encoder/decoder structure by adding the option to have "0" num_heads.

We can see some results from training a PointWiseMLP here:
https://mlflow.ecmwf.int/#/experiments/270/runs/c615f13596454d95ae6d7666fc54b90f/artifacts

This model specification will allow us to fine-tune a land surface model to point flux tower observations and potentially improve land-model fluxes will still being able to apply this model to a global grid

models/src/anemoi/models/models/processor.py
ssmmnn11: Still says Message passing graph neural network

JPXKQX: Any suggestions? It can still be a messa gepassing gnn if you choose `processor: gnn`. What about a graph neural network?

ssmmnn11: I would recommend to use the encoder_processor_decoder structure with suitable encoder processor decoder instead.

models/tests/layers/mapper/test_graphinterpolation_mapper.py
mchantry, line 1: Empty file?

JPXKQX: It seems that this empty file was accidentally included in the main repository in a previous PR (#436).

training/docs/user-guide/models.rst
```diff
@@ -99,6 +100,21 @@ coarser than the resolution of the base data.
...
+.. note::
+
+   The Transformer does not require a subgraph.
+
+Point-wise MLP
+==============
+
+The Point-wise MLP applies the same multilayer perceptron independently
+to each node. Because it ignores neighboring nodes (no message passing),
+the processor is location-independent.
```
ssmmnn11: I think it would be better to state that is only point-wise, results are not conditioned on other notes. Location-independent is not well defined. If you provide positional information, it will learn from these.

models/src/anemoi/models/layers/block.py
```diff
@@ -63,6 +63,35 @@ def forward(
...
+        super().__init__()
+        layers = [
+            layer_kernels.Linear(num_channels, hidden_dim),
+            layer_kernels.LayerNorm(hidden_dim),
+            layer_kernels.Activation(),
+        ]
+        if num_channels != hidden_dim:
+            layers.append(layer_kernels.Linear(hidden_dim, num_channels))
+
+        if dropout_p is not None and dropout_p > 0:
```
theissenhelen, line 81: It might better to raise an error for `dropout <0` instead of letting it fall through.

models/src/anemoi/models/layers/block.py
```diff
@@ -63,6 +63,35 @@ def forward(
...
 
 
+class PointWiseMLPProcessorBlock(BaseBlock):
+    """Point-wise block with MLPs."""
+
+    def __init__(self, *, num_channels: int, hidden_dim: int, layer_kernels: DotDict, dropout_p: float = 0.0):
+        super().__init__()
+        layers = [
+            layer_kernels.Linear(num_channels, hidden_dim),
+            layer_kernels.LayerNorm(hidden_dim),
```
ssmmnn11, line 75: Why do we layernorm after linear and before activation? Is this a intended pattern? :-)

models/src/anemoi/models/layers/block.py
JPXKQX: Thanks, check added now! I think the purpose of setting `dropout_p > 0` here was to avoid instantiating a Dropout(0).

ssmmnn11, line 81: please note that dropout is currently not safe to use for sharded models.

training/src/anemoi/training/config/model/point_wise.yaml
```diff
@@ -0,0 +1,98 @@
...
+    bias: False
+
+processor:
+  _target_: anemoi.models.layers.processor.PointWiseMLPProcessor
+  trainable_size: ${model.trainable_parameters.hidden2hidden}
+  sub_graph_edge_attributes: ${model.attributes.edges}
+  num_layers: 16
+  num_chunks: 2
+  mlp_hidden_ratio: 4 # GraphTransformer or Transformer only
+  num_heads: 16 # GraphTransformer or Transformer only
```
ssmmnn11: I don't hink you have heads here

training/src/anemoi/training/config/model/point_wise.yaml
```diff
@@ -0,0 +1,98 @@
...
+  qk_norm: False
```
ssmmnn11: qk norm is only relevant for the transformer graphtransformer

training/src/anemoi/training/config/model/point_wise.yaml
JPXKQX: You are right! It is removed now

models/src/anemoi/models/layers/block.py
JPXKQX: Do you have any comments on this? @pinnstorm

training/docs/user-guide/models.rst
JPXKQX: The documentation has now been updated

models/src/anemoi/models/layers/block.py
pinnstorm: This was just one of the patterns we had tested with the simple external MLP that had given us good results for land-surface model emulation. But don't think the exploration was exhaustive so if we want something that conforms more with the other MLPs we have here sure that is fine too

models/src/anemoi/models/layers/block.py
JPXKQX: What would you recommend here?

anaprietonem: I have added now an assert for this and included a unit test to check that if sharding and dropout are used an error is raised. Thoughts?

models/src/anemoi/models/schemas/processor.py
```diff
@@ -79,3 +79,14 @@ def check_valid_extras(self) -> Any:
...
+
+class PointWiseMLPProcessorSchema(PointWiseModelComponent):
+    target_: Literal["anemoi.models.layers.processor.PointWiseMLPProcessor"] = Field(..., alias="_target_")
+    "Transformer processor object from anemoi.models.layers.processor."
+    num_layers: NonNegativeInt = Field(example=16)
+    "Number of layers of Transformer processor."
+    num_channels: NonNegativeInt = Field(example=128)
+    "Number of channels."
+    dropout_p: NonNegativeFloat = Field(example=0.1)
+    "Dropout probability used for multi-head self attention, default 0.0"
```
ssmmnn11: Remove multi-head self attention?

JPXKQX: Removed!

ssmmnn11 (approved):
I think looks good.

Issue #368, opened 2025-06-13
Plotting Bug: Plotting logic breaks when precip related fields are not diagnostic


roxas6path:
### What happened?

The logic for plotting in plot_flat_sample breaks when precip related fields are not diagnostic.

Specifically: combined_error is not created if `vname in precip_and_related_fields`, but combined_error is required if a param is not a diagnostic.

[anemoi-core code link](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/diagnostics/plots.py#L569C5-L585C1)

In Interpolator workflows we may usually have precip related fields as prognostic not diagnostic

### What are the steps to reproduce the bug?

make combined_error be created independently of whether `vname in precip_and_related_fields` is True of False.

### Version

training: v0.5.0

### Platform (OS and architecture)

ATOS

### Relevant log output

```shell

```

### Organisation

ECMWF

PR #369, opened 2025-06-13
fix: Plotting error when precip related params are not diagnostic 

Rilwan-Adewoyin:
## Description

#368 

#368 

#368

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/config/interpolator.yaml
+++ b/training/src/anemoi/training/config/interpolator.yaml
@@ -15,7 +15,7 @@ data:
   frequency: 1h
   resolution: o96
 
-  diagnostics: []
+  diagnostics: [] # Default behaviour is to interpolate between observed fields, not predict unseen fields
 
 #dataloader:
   # model_run_info: #Add for non-analysis training
--- a/training/src/anemoi/training/config/interpolator.yaml
+++ b/training/src/anemoi/training/config/interpolator.yaml
@@ -15,6 +15,8 @@ data:
   frequency: 1h
   resolution: o96
 
+  diagnostics: []
+
 #dataloader:
   # model_run_info: #Add for non-analysis training
   #  start: 2020-02-05T12:00:00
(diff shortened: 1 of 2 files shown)


training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -15,6 +15,8 @@ data:
   frequency: 1h
   resolution: o96
 
+  diagnostics: []
```
mc4117: I was wondering why we need to add this to the config here and what role it plays. I don't see it used in the code?

roxas6path: This stops any params being used as diagnostics. 
Previously, default settings from forecastor were used to make tp and cp diagnostics. But base behaviour for interpolator should be to interpolate all fields available to it as opposed to predicting unseen fields.

Furthermore, having fields as diagnostics in time-interpolation leads to some minor issues in anemoi-infererence for time-interpolator : https://github.com/ecmwf/anemoi-inference/pull/227#issuecomment-2970211716

mc4117: Thanks for clarifying! Could you put a comment about this in the config file?

roxas6path: So addressed in the most recent commit

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -577,12 +579,14 @@ def plot_flat_sample(
...
         combined_data = np.concatenate((input_, truth, pred))
         # For 'errors', only persistence and increments need identical colorbar-limits
-        combined_error = np.concatenate(((pred - input_), (truth - input_)))
+
         norm = Normalize(vmin=np.nanmin(combined_data), vmax=np.nanmax(combined_data))
 
         norms[1] = norm
         norms[2] = norm
 
+    combined_error = np.concatenate(((pred - input_), (truth - input_)))
```
mc4117: Can you put this line under the if statement? I'm just a bit worried that it won't work if sum(input)=0. Otherwise looks good to me!

roxas6path: Good point, moved it under the if statement in most recent commit 
fe32be8198c4402598f98bf47c306f2eaeba701b

Issue #370, opened 2025-06-16
Handling and Imputing NaN values in input and output fields


sahahner:
### Is your feature request related to a problem? Please describe.

Some variables are not well-defined at certain locations, and the corresponding data typically contains NaN values at these positions. These NaNs must be handled and replaced with numerical values before being fed into the model, as the model cannot process NaNs. However, NaNs should be reintroduced in the forecasted fields after inference.

**Context**:
-	Surface ocean fields contain NaNs over land, and most land fields contain NaNs over ocean.
-	Fields from the deep ocean or certain upper-air variables that cross orography may contain additional NaNs not accounted for by the land-sea mask.
-	In general, the spatial pattern of NaNs remains constant in time. However, slight changes can occur, for instance through updates to the operational model, such as land-sea mask modifications.
-	In reanalysis datasets, NaN locations are fixed and do not change over time.

**Requirements**:
-	The model must not receive NaN values as input; all NaNs must be imputed with appropriate values prior to model ingestion.
-	The pattern of NaN locations may change slightly between operational model versions, and the preprocessing should account for this.
-	After the loss function is applied, NaNs should be restored in the model output. This also applies to rollout training, which is currently not handled correctly (a PR will address this soon, [branch](https://github.com/ecmwf/anemoi-core/compare/main...fix-pre-postprocessors-in-rollout-training) with a draft).
-	If a field is not well-defined over certain areas, the model weights should not be updated with respect to forecasts in these areas, therefore the model does not learn to forecast at those positions.

**Non-goal**
-	Big changes in NaN positions are not expected. If they occur, dynamic imputers should be used: see [#89](https://github.com/ecmwf/anemoi-models/pull/89). 
o	Changes in NaN locations due to changes in land-sea-mask should always be small.
o	What is big/significant? If NaN positions change “too much”, should there be an assert? Are we expecting this case in some cases?
-	If the presence or absence of values (i.e., the NaN pattern) itself needs to be forecasted, a separate variable representing the NaN mask should be introduced (see the Aurora paper for an example, specifically for wave fields under sea ice).

Current Implementation: Where NaN Handling Matters
-	Preprocessor: Imputes NaNs with user-defined fill values (configurable per variable). Ensures that no NaNs are passed to the model.
-	Loss function: Positions that are always NaN receive zero weight in the loss, so the model is not penalized for predictions at undefined points.
-	Postprocessor: After computing the loss, NaNs are reapplied to the model outputs at the correct positions.
[...]

PR #371, opened 2025-06-16
feat: Extend DelayedScalers into arbitary UpdatingScalers

HCookie:
## Description
Removes `DelayedScalers` and add `UpdatingScalers`. 
Allows for more fine control as to when and how the `scalers` are updated.

Part of #370 

##  Additional notes ##


training/src/anemoi/training/losses/scalers/loss_weights_mask.py
```diff
@@ -23,25 +23,11 @@
 LOGGER = logging.getLogger(__name__)
 
 
-class NaNMaskScaler(BaseDelayedScaler):
+class NaNMaskScaler(BaseUpdatingScaler):
 
     scale_dims: tuple[TensorDim] = (TensorDim.GRID, TensorDim.VARIABLE)
 
-    def __init__(self, norm: str | None = None, **kwargs) -> None:
```
sahahner, line 26: I have tested this PR in training: If this init function is removed, the `__init__` function of the `base_scaler` should allow `**kwargs` to support this scaler's initialisation.

sahahner:
Thank you very much for this PR. I like the solution with the callbacks as it is flexible and can be reused for other scalers.

I see you have tried to cover many different cases. Nevertheless, the NaN_weights need to be updated after `y_pred = self(x)` (after the forward pass, before the training loss calculation). None of the provided callbacks cover that case. 
Would it be possible to include that case, please? It seems to be non-standard callback event, so we might need to come up with a name. What do you think about `on_after_forward`? (pytorch lightning for example support `on_after_backward`).

Nevertheless, I do feel that this PR increases the complexity of the forecaster by calling a few more functions without any added functionality. We might want to remove the callbacks that are actually not used and leave it for future PRs that require this functionality. Either way, I might not be aware of other use cases that are planned to be included.



Thank you very much for this PR — I like the callback-based solution, as it’s flexible and potentially reusable for other scalers.

I see you've tried to cover many cases, which is great. However, for https://github.com/ecmwf/anemoi-core/issues/370 NaN_weights need to be updated after the forward pass (`y_pred = self(x)`) and before the training loss is computed. None of the current callbacks seem to handle this case. Would it be possible to add a callback event for this? Since it’s not a standard one, we might need to define a new name — what do you think about `on_after_forward`? (Similar to how PyTorch Lightning uses `on_after_backward`.)

That said, I feel this PR adds some complexity by introducing multiple callbacks that aren’t currently used. We might want to consider removing the unused ones for now and adding them later when there’s a clear need. Of course, there might be future use cases in mind that I do not know about.

For context, I was able to build https://github.com/ecmwf/anemoi-core/pull/378 on top of this for now after adjusting the position of `ON_TRAIN_BATCH_END`.

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
HCookie: What for? What kwargs are passed?

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -84,19 +94,121 @@ def get_scaling(self) -> SCALER_DTYPE:
...
+            Scaling values as a numpy array.
+        """
+        if self._cached_scaling_values is not None:
+            LOGGER.debug("Using cached scaling values for %s.", self.__class__.__name__)
+            return self._cached_scaling_values
+
+        return self.initial_scaling_values()
+
+    def get_scaling(self) -> SCALER_DTYPE:
+        """Get scaling values based on the initial scaling values callback."""
```
anaprietonem, line 131: What's the motivation for the caching of the scaling values?

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -84,19 +94,121 @@ def get_scaling(self) -> SCALER_DTYPE:
         return scale_dims, scaler_values
 
 
-class BaseDelayedScaler(BaseScaler):
-    """Base class for delayed Scalers.
+class AvailableCallbacks(StrEnum):
+    INITIAL_SCALING_VALUES = "initial_scaling_values"
```
anaprietonem: To my understanding scalers take default values and depending on the config values those can change for certain grid point, variables or vertical levels. What we are thinking now is those also might change according to certain requirements imposed by pre-processors like the imputers. I guess then I struggle to see the need for this hooks and how do they connect with the pre-processors

anaprietonem: Okey @sahahner kindly explained the fact that we need updates on training and validation as the mask has the batch dimension and those can change. I guess simplifying it so we just use the hooks we need as @sahahner  mentioned above could be good, but I wonder is there is no way to do directly on the forecaster? On the rollout_step directly . So directly doing what the `NaNMaskScaler` is doing on `on_training_start` on the forecaster method to update the scalers. That's method is called on the rollout_step so the update would happen both for training and validation?

HCookie: I tried to keep fairly general to capture future use cases. I'm happy to remove the superfolous callbacks for now and just keep start, and rollout_step ones

training/src/anemoi/training/losses/scalers/base_scaler.py
HCookie: They will be called for the get on each callback, but only some will update them, hence caching

training/src/anemoi/training/losses/scalers/base_scaler.py
anaprietonem: I appreciate the nice work you're doing here and that the callback is smart trick. But I think the imputer/handling of the nans is a special case so I'd be in favour of reducing complexity as much as possible so in the future we can identify the changes of https://github.com/ecmwf/anemoi-core/issues/370 much easier.  A few things from the current design that are still unclear to me are:
- the `get_scaling_values` of the BaseUpdatingScaler or any class that inherits from it - where is that used?
- We discussed that the scalers need to be updated per batch for training and validation but the NaNMaskScaler just has the 'on_training_start' implemented. 
- In https://github.com/ecmwf/anemoi-core/pull/378 - the pre_processor.loss_mask_training is being recomputed for each batch under the transform that gets called in the forecaster rollout_step. So what I was wondering is why it would be enough to call the update method there two?
```
        for pre_processor in model.pre_processors.processors.values():
            if hasattr(pre_processor, "loss_mask_training"):
                loss_weights_mask = loss_weights_mask * pre_processor.loss_mask_training.cpu().numpy()
```
since the result of `pre_processor.loss_mask_training` would be already taking the right shape depending if it's training or validation, so then we could simplify the logic

sahahner: > * We discussed that the scalers need to be updated per batch for training and validation but the NaNMaskScaler just has the 'on_training_start' implemented.

This change is implemented in #378 where we implement the logic of the nan calculation of the fly.

anaprietonem: Yes, maybe my comment was not clear. What I mean is looking at 378 and the changes at 371 - I think there is a mismatch in the logic - because of callback for update is 'on_training_start' and the update of the nan mask is done on the rollout step per batch.  But this also to me, means we could simply some of the logic here.

training/src/anemoi/training/losses/scalers/base_scaler.py
sahahner: I see the motivation behind the caching, but it seems to be causing me errors as these cached scaling values are not on the device. Could we access the buffered scalers here?

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
sahahner: There was an error in my config.

sahahner: I had another look and indeed this line seems to pass `group_config` to the updating scaler: https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/losses/scalers/scalers.py#L30. The other methods override the __init__ which is why i think just the nan_mask fails.

training/src/anemoi/training/losses/scalers/base_scaler.py
sahahner: I am not sure I understand the comment. `on_training_start` will actually not be implemented anymore for the loss mask.

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -84,19 +94,121 @@ def get_scaling(self) -> SCALER_DTYPE:
...
+    """Base class for updating scalers.
 
-    The delayed scalers are only initialised when creating all the scalers, but its value is
-    computed during the first iteration of the training loop. This delayed scalers are suitable
-    for scalers requiring information from the `model.pre_processors`.
+    The updating scalers have a variety of callback methods associated with them,
+    which are called during the training loop. These methods allow the scalers to
+    update their values based on the current state of the model and the training data.
+
+    The callback methods are expected to return a np.ndarray of scaling values,
```
sahahner, line 105: what if output is the default which is None? -> no update on scaler

training/src/anemoi/training/losses/scalers/base_scaler.py
HCookie: Caching has been removed

training/src/anemoi/training/train/forecaster/forecaster.py
```diff
@@ -234,11 +235,19 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
     def on_load_checkpoint(self, checkpoint: torch.nn.module) -> None:
         self._ckpt_model_name_to_index = checkpoint["hyper_parameters"]["data_indices"].name_to_index
 
-    def define_delayed_scalers(self) -> None:
+    def update_scalers(self, callback: AvailableCallbacks) -> None:
         """Update delayed scalers such as the loss weights mask for imputed variables."""
```
anaprietonem: docstring update - no longer delayed?

training/src/anemoi/training/train/forecaster/forecaster.py
```diff
@@ -234,11 +235,19 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
-        for name, scaler_builder in self.delayed_scaler_builders.items():
-            self.scalers[name] = scaler_builder.get_delayed_scaling(model=self.model)
-            self.loss.update_scaler(scaler=self.scalers[name][1], name=name)
+        for name, scaler_builder in self.updating_scalars.items():
+            scaler = scaler_builder.update_scaling_values(callback, model=self.model)
+            if scaler is None:
+                continue
+
```
anaprietonem: why scalers[1]? could be good to add some comments about what is going on inside each loop

training/src/anemoi/training/train/forecaster/forecaster.py
HCookie:
```suggestion
        """Update scalers, calling the defined function on them, updating if not None"""
```

training/src/anemoi/training/train/forecaster/forecaster.py
```diff
@@ -234,11 +235,19 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            if name in self.loss.scaler:
+                self.loss.update_scaler(scaler=scaler[1], name=name)
+
+            for metric in self.metrics.values():
+                if name in metric.scaler:
+                    metric.update_scaler(scaler=scaler[1], name=name)
```
HCookie:
```suggestion
        for name, scaler_builder in self.updating_scalars.items():
            scaler = scaler_builder.update_scaling_values(callback, model=self.model)
            if scaler is None: # If scalar is None, no update to be applied
                continue

            if name in self.loss.scaler: # If scalar in loss, update it
                self.loss.update_scaler(scaler=scaler[1], name=name) # Only update the values

            for metric in self.metrics.values(): # If scalar in metrics, update it
                if name in metric.scaler:
                    metric.update_scaler(scaler=scaler[1], name=name) # Only update the values
```

HCookie (commented):
@anaprietonem Thoughts?

training/src/anemoi/training/train/forecaster/interpolator.py
```diff
@@ -99,8 +100,15 @@ def _step(
 
         # Delayed scalers need to be initialized after the pre-processors once
         if self.is_first_step:
-            self.define_delayed_scalers()
+            self.update_scalars(callback=AvailableCallbacks.ON_TRAINING_START)
             self.is_first_step = False
+        self.update_scalars(
+            callback=(
```
anaprietonem: I think we thought that the valid/training abstraction was not needed, and just keep the training start and batch one. So this would need update?

training/src/anemoi/training/train/forecaster/forecaster.py
```diff
@@ -412,6 +423,10 @@ def rollout_step(
 
             yield loss, metrics_next, y_pred
 
+        self.update_scalers(
```
anaprietonem: why are we updating the scalar after the rollout step again?

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -84,19 +94,87 @@ def get_scaling(self) -> SCALER_DTYPE:
         return scale_dims, scaler_values
 
 
-class BaseDelayedScaler(BaseScaler):
-    """Base class for delayed Scalers.
+class AvailableCallbacks(StrEnum):
+    INITIAL_SCALING_VALUES = "initial_scaling_values"
```
anaprietonem: why we we have 'INITIAL_SCALING_VALUES'  as part of the callback?

training/src/anemoi/training/train/forecaster/forecaster.py
```diff
@@ -412,6 +423,10 @@ def rollout_step(
...
+            callback=(AvailableCallbacks.ON_BATCH_START),
+        )
```
HCookie:
```suggestion
```

training/src/anemoi/training/losses/scalers/base_scaler.py
HCookie:
```suggestion
```

training/src/anemoi/training/train/forecaster/ensforecaster.py
```diff
@@ -238,8 +239,15 @@ def rollout_step(
...
-            self.define_delayed_scalers()
+            self.update_scalers(callback=AvailableCallbacks.ON_TRAINING_START)
             self.is_first_step = False
+        self.update_scalers(
+            callback=(
+                AvailableCallbacks.ON_TRAIN_BATCH_START
+                if not validation_mode
+                else AvailableCallbacks.ON_VALID_BATCH_START
+            ),
+        )
```
HCookie:
```suggestion
        self.update_scalers(callback=AvailableCallbacks.ON_BATCH_START)
```

training/src/anemoi/training/train/forecaster/interpolator.py
```diff
@@ -99,8 +100,15 @@ def _step(
...
+                AvailableCallbacks.ON_TRAIN_BATCH_START
+                if not validation_mode
+                else AvailableCallbacks.ON_VALID_BATCH_START
+            ),
+        )
```
HCookie:
```suggestion
        self.update_scalers(callback=AvailableCallbacks.ON_BATCH_START)
```

training/src/anemoi/training/train/forecaster/forecaster.py
HCookie: Removed

training/src/anemoi/training/train/forecaster/interpolator.py
HCookie: Corrected

anaprietonem (commented):
This PR looks good. @HCookie addressed the comments and I think the logic is now simplified, keeping a trade-off between flexibility and complexity. I think before merging it would still need to be documented the new logic for updating scalars (could use the imputer as an example). 

I think the new logic to use None when no update is done it's also neat, and could open the door to simplify further the scalers abstraction (ie baseScalars could also be seen as 'updating' ones with None updates). This could be revisit in the future

training/docs/modules/losses.rst
```diff
@@ -203,6 +203,49 @@ If multiple groups are defined for a variable, the first group in the
 `variable_groups` is used. If the variable is not in any group, it is
 assigned to the default group.
 
+Custom Scalars
```
anaprietonem, line 206: Very nice!

anaprietonem (dismissed):
Good work @HCookie and nice docs!

anaprietonem (approved):
Changes were addressed and tests are passing! good work Harrison, good to be merged!

PR #378, opened 2025-06-20
feat(models): nan locations in imputer calculated on the fly

sahahner:
- calculate nan locations in imputer on the fly in the preprocessor
- cache nan locations to be reused in the postprocessor
- since nan locations can change for different samples, they are three-dimensional now.
  - before: (grid, variable)
  - after: (batchsize, grid, variable)
  - in preprocessor/transform: we need to even keep time dimension as nan locations can change over time. for forecasted fields use nan locations from first timestep
  - diagnostic fields: the logic, to take the nan mask from the input is not well defined for diagnostic variables in inference as they are not present in inference. 
- covers: InstantImputer, ConstantImputer and CopyImputer
- Additionally: added tests for CopyImputer 

TODO:
- check runtime/memory implications of covering batch (and time) dimension in nan-location-mask
- figure out multigpu after https://github.com/ecmwf/anemoi-core/issues/383: multi GPU and sharding works. inference is also fine as long as pre and postprocessors are both applied on sharded or unsharded batch

----
📚 Documentation preview 📚: https://anemoi-training--378.org.readthedocs.build/en/378/

----
📚 Documentation preview 📚: https://anemoi-graphs--378.org.readthedocs.build/en/378/

----
📚 Documentation preview 📚: https://anemoi-models--378.org.readthedocs.build/en/378/


training/src/anemoi/training/losses/scalers/loss_weights_mask.py
```diff
@@ -25,17 +25,33 @@
...
-        Before first application of loss function, the mask is replaced.
+        Always when calling the imputer, the NaN positions are updated.
+        Before every application of training loss function, the mask is replaced.
         """
-        loss_weights_mask = np.ones((1, 1))
+        loss_weights_mask = np.ones((1, 1, 1))
+        # iterate over all pre-processors and check if they have a loss_mask_training attribute
+        for pre_processor in model.pre_processors.processors.values():
+            if hasattr(pre_processor, "loss_mask_training"):
+                loss_weights_mask = loss_weights_mask * pre_processor.loss_mask_training.cpu().numpy()
```
sahahner: Todo: avoid moving tensors between devices.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
 
             LOGGER.debug(f"Imputer: replacing NaNs in {name} with value {self.replacement[-1]}")
 
-    def _expand_subset_mask(self, x: torch.Tensor, idx_src: int) -> torch.Tensor:
-        """Expand the subset of the mask to the correct shape."""
-        return self.nan_locations[:, idx_src].expand(*x.shape[:-2], -1)
+    def _expand_subset_mask(self, x: torch.Tensor, idx_src: int, nan_locations: torch.Tensor) -> torch.Tensor:
+        """Expand the subset of the nan location mask to the correct shape.
+
+        The mask is only saved for the first dimension (batch) and the last two dimensions (grid, variable).
```
ssmmnn11: Do we need the ensemble dimension?

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
...
+        x : torch.Tensor
+            Input tensor with shape (batch, time, ..., grid, variable)
+
+        Returns
+        -------
+        torch.Tensor
+            Tensor with NaN locations of shape (batch, time, ..., grid)
+        """
+        # The mask is only saved for the first two dimension (batch, timestep) and the last two dimensions (grid, variable)
+        # For the rest of the dimensions we use the first dimension
```
ssmmnn11: Could you explain what you mean by "we use the first dimension"? :-)

I think we have (batch, time, ensemble, grid, channel).

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
 
             LOGGER.debug(f"Imputer: replacing NaNs in {name} with value {self.replacement[-1]}")
 
-    def _expand_subset_mask(self, x: torch.Tensor, idx_src: int) -> torch.Tensor:
-        """Expand the subset of the mask to the correct shape."""
-        return self.nan_locations[:, idx_src].expand(*x.shape[:-2], -1)
+    def _expand_subset_mask(self, x: torch.Tensor, idx_src: int, nan_locations: torch.Tensor) -> torch.Tensor:
```
ssmmnn11: Will need work to make compatible with sharding

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
...
     def transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Tensor:
         """Impute missing values in the input tensor."""
         if not in_place:
             x = x.clone()
 
-        # Reset NaN locations outside of training for validation and inference.
-        if not self.training:
-            self.nan_locations = None
+        # recalculate NaN locations every forward pass and cache for backward pass
+        # choose correct index based on number of variables which are different for training and inference
```
ssmmnn11, line 192: saving for backward in autograd would be done like ctx.something = value. But I don't think this is what we are doing here? Are you saving it for the inverse transform?

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
...
+        if x.shape[-1] == self.num_training_input_vars:
```
ssmmnn11, line 193: Is it save to distinguish training and inference like that? could they have the same shape? Though I think we might have checks like that in other parts of the code as well.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
...
+            # training input
+
+            # get nan locations from training input
+            nan_locations = self.get_nans(x)
 
-        # Initialise mask if not cached.
-        if self.nan_locations is None:
+            # cache nan locations from training input, reduce time dimension by selecting the first timestep
```
ssmmnn11: can we say "save" here? I would use cache in a different context.

ssmmnn11: Reduce usually means "takes a collection of values and combines them into a single value" ; I think here you are selecting a time step instead. Is the assumption that all time steps will be the same?

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
...
+            self.nan_locations = nan_locations[:, 0]
 
-            # Get NaN locations
-            self.nan_locations = self.get_nans(x)
+            # data indices for training input
+            index = self.index_training_input
 
-            # Initialize training loss mask to weigh imputed values with zeroes once
+            # set training loss mask to match shape of training input
```
ssmmnn11, line 210: can we just zero out NaN location values in predictions and targets? Then we can remove the loss mask?

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -103,50 +103,114 @@ def _create_imputation_indices(
...
+            # for all variables that are imputed and part of the model output, set the loss weight to zero at NaN location
+            for idx_src, idx_dst in zip(index, self.index_inference_output):
                 if idx_dst is not None:
-                    self.loss_mask_training[:, idx_dst] = (~self.nan_locations[:, idx_src]).int()
+                    self.loss_mask_training[..., idx_dst] = (~self.nan_locations[..., idx_src]).int()
 
-        # Choose correct index based on number of variables
-        if x.shape[-1] == self.num_training_input_vars:
-            index = self.index_training_input
         elif x.shape[-1] == self.num_inference_input_vars:
```
ssmmnn11, line 220: Is it save to assume that these are always different?

Also: can we move the nan_locations = self.get_nans(x) outside of the if - because that happens in training and inference the same way? You could put the if around the loss mask part only and then have a check before that raises the runtime error.

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
```diff
@@ -25,20 +25,39 @@
...
 
-    def on_training_start(self, model: AnemoiModelInterface) -> np.ndarray:
-        """Get loss scaling.
+    def initial_scaling_values(self) -> torch.Tensor | None:
+        """Get initial scaling values.
+
+        Returns
+        -------
+        torch.Tensor
+            Initial scaling values, default is a tenser of ones.
```
ssmmnn11: typo

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
```diff
@@ -25,20 +25,39 @@
...
+        """
+        # TODO(sara,harrison): should this in general (for all scalers) be a tensor and not a numpy array?
```
ssmmnn11: Discussed offline: I think it is important to solve that.

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
```diff
@@ -25,20 +25,39 @@
...
+        Before every application of training loss function, the mask is replaced.
         """
-        loss_weights_mask = np.ones((1, 1))
+        loss_weights_mask = self.get_scaling_values()
+        device = loss_weights_mask.device
         # iterate over all pre-processors and check if they have a loss_mask_training attribute
         for pre_processor in model.pre_processors.processors.values():
             if hasattr(pre_processor, "loss_mask_training"):
-                loss_weights_mask = loss_weights_mask * pre_processor.loss_mask_training.cpu().numpy()
+                loss_weights_mask = loss_weights_mask * pre_processor.loss_mask_training.to(device)
```
ssmmnn11: Everything should always be on device, otherwise there will be a performance penalty.

ssmmnn11 (changes requested):
much improved!

models/src/anemoi/models/preprocessing/imputer.py
sahahner: I assumed that at the same time step, all ensemble members have the same nan locations, so I just select the first of the ensemble dimensions. Is this assumption correct?

models/src/anemoi/models/preprocessing/imputer.py
sahahner: I mean the first dimension of n ensemble dimensions.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: I am indeed saving for the inverse transform.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: I think this question is beyond this PR. I agree that it is not necessarily elegant, and all preprocessors have to distinguish between the two cases. 
As far as I know, they have different shapes, because in training the batch contains all forcing, prognostic and diagnostic variables. In inference, though, we only pass the forcing and prognostic variables to the model and output only diagnostic and prognostic.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Yes, for loss weights=0 that would be the same behaviour. 
Nevertheless, I would like to keep the possibility to use non-zero weights (but different to non-nan locations) for nan locations in the future.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Yes, reduce is indeed not fitting here. 
We need to decide on one set of nan locations for the nan locations in the output. I have chosen the nan locations from the first timestep, but we can decide on something else here. This seemed like the most general solution, keeping in mind the different architectures we are supporting.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Yes, the nan locations can be calculated before the ifs.

What are you referring to as assumed to be always different?

models/src/anemoi/models/preprocessing/imputer.py
ssmmnn11: I don't know. If the NaN location is conditioned on the forecast, then no. But for fixed ones yes. I think it would be good to write down in the beginning / somewhere, what the assumptions are: NaNs do not change ... NaNs are the same for step ...

models/src/anemoi/models/preprocessing/imputer.py
ssmmnn11: Ok, we need to check that this holds for models trained without diagnostic variables. @gmertes

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Yes, I can add this comment for the ensemble dimension.

models/src/anemoi/models/preprocessing/imputer.py
ssmmnn11: "differ", was duplicated question from above. please ignore.

models/src/anemoi/models/preprocessing/imputer.py
ssmmnn11: ok, so broadcast along ensemble dimension.

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
sahahner: will be solved here: https://github.com/ecmwf/anemoi-core/pull/390

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Comment added for self.get_nans().

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
sahahner: See https://github.com/ecmwf/anemoi-core/pull/390

training/src/anemoi/training/losses/scalers/loss_weights_mask.py
sahahner: All `.to()` have been removed. 
`loss_weights_mask = loss_weights_mask * pre_processor.loss_mask_training` is on model device. `loss_weights_mask` is initialized on model device.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: This if is to distinguish between training and inference. I can confirm that this works with no diagnostic variables as well. In inference, the loss mask would be calculated although not needed, in case there are no diagnostic variables. In the best case, we would have a flag to distinguish between inference and training, but this does not exist.

models/src/anemoi/models/preprocessing/imputer.py
sahahner: Thanks to this PR https://github.com/ecmwf/anemoi-core/pull/391 the scalers can be applied to the sharded batch. Since the nan-locations are calculated on the fly and the batch is sharded from before the preprocessors until after the postprocessors, this approach works for sharding over the grid dimension.
See two example runs on n320 grid here: https://mlflow.ecmwf.int/#/compare-runs?runs=[%2223a23459e9244881a6fe76ef11bd92de%22,%229bf7771f59504c66bcc1050c371308f7%22]&experiments=[%22263%22]

Callbacks applying the postprocessors are also working now, as long as they are not running in asyncmode.

ssmmnn11:
> I have trained approximately 5 models each with the old imputer and the new imputer logic:
> 
> [mlflow](https://mlflow.ecmwf.int/#/metric?runs=%5B%22e5fd3ee43bef4a168c4978347e5c3784%22,%221e78855676db49c6bc8a3bf33a034d47%22,%22eea9c16090854f03bd62f09f69770844%22,%22b50593e4a03643688e45db441316bd70%22,%220177e619b5c949cba701f4634d8ddc66%22,%22f14f52b17b13438dbe15ff200e6f32b9%22,%22bb26507055f9438684dec10e07682e51%22,%22e5c2b93542ba43afb3d0bb2bec2fe03a%22,%22740c56520e294fbdb9e0a105edc36084%22,%22f3ac4a42283b4c4a9f06a1a063628a3c%22,%22ce77b98487be4f4e9741c9900de78bc6%22,%22e2b0cf52dce24c2786459b100b2cf8a2%22,%22ac89229fa64d499a9cf4fb806fae9711%22,%22b8320316d964412580ebf4155781a6da%22,%22b8cb291d0b2f4c41a7a09f546ef6a676%22,%2278a5181dffbe4fa398a83e724fedf947%22%5D&metric=%22val_mse_metric%2Fsfc_2d%2F1%22&experiments=%5B%22263%22%5D&plot_metric_keys=%5B%22epoch%22%5D&plot_layout=%7B%22autosize%22:true,%22xaxis%22:%7B%22autorange%22:true,%22range%22:%5B451.5744878767397,549.1660576152823%5D%7D,%22yaxis%22:%7B%22autorange%22:true,%22type%22:%22linear%22,%22range%22:%5B2.5435244161358814,4.24203821656051%5D%7D%7D&x_axis=relative&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=%5B%5D&last_linear_y_axis_range=%5B%5D)
> 
> key information on the **setup**: n320, multi-GPU, no sharded batch, batchsize 1 no plots!
> 
> The runtime difference of the nan calculation on the fly and batch-wise scaler update is barely (if even?) visible in comparison to the instabilities of the cluster.
> 
> When plots are enabled, the new setup is more expensive, as now we need to copy the post-processors with the changing nan locations to CPU at the beginning of every validation epoch. I had a look at it with @japols, but I couldn't find an adequate way around that under the current setup of the callbacks.

I think we need to change the approach and return de-normalised fields for plotting instead. I think there is one callback that calculates loss contributions, but also that we could compute inside the validation step and then return what is needed for plotting (minimal data).

How big are the pos-processors? how much data do you need to transfer? I would assume that these are much smaller than the actual fields which we transfer for plotting?

sahahner:
> 
> I think we need to change the approach and return de-normalised fields for plotting instead. I think there is one callback that calculates loss contributions, but also that we could compute inside the validation step and then return what is needed for plotting (minimal data).

Agreed. I think this PR showcases that this could be optimized. 

> How big are the pos-processors? how much data do you need to transfer? I would assume that these are much smaller than the actual fields which we transfer for plotting?

Inside the post-processor, we have the nan-mask field, which, after this PR, is of size (batchsize, grid, variable) and therefore bigger than before. This field needs to be gathered in case of sharded training and moved to CPU for every batch to be plotted as the mask might change over time. Before, we would copy just once in the first validation epoch. The extra cost of repeated copying of the mask of the old size that you were interested in is small but noticeable: [mlflow](https://mlflow.ecmwf.int/#/metric?runs=[%22f31e6839e9ec45ca8cca755342536e0a%22,%22bbec6d36769c4582a6d6e40df437ee88%22,%2264f8a0e48f8447c3944722f24bb896e2%22,%222f6b460c51df4b549cadeac7f461e9e5%22,%2229424dfa454f4efab386ea2628e60f2f%22]&metric=%22system%2Fgpu_0_pcie_tx_megabytes%22&experiments=[%22263%22]&plot_metric_keys=%5B%22epoch%22%5D&plot_layout={%22autosize%22:true,%22xaxis%22:{%22range%22:[522.232853381118,654.0531776405255],%22autorange%22:true},%22yaxis%22:{%22range%22:[2.7983014861995756,4.225053078556264],%22autorange%22:true,%22type%22:%22linear%22}}&x_axis=relative&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=[]&last_linear_y_axis_range=[]) 
Additionally, since the post-processor mask depends on batchsize now, we also need to apply the post-processor to the entire batch instead of just the sample to be plotted. [code](https://github.com/ecmwf/anemoi-core/pull/378/files#diff-1c46a92722a9bff5208ef437c364b25b174c395c8ee136eac64b18438044dd20R450)

If in the future, callbacks were to receive postprocessed batches directly, this overhead might become easier to manage.

Given that plotting callbacks are inherently costly, I’d recommend using them less frequently, especially for training larger models on high-resolution grids, to avoid unnecessary runtime increase.

ssmmnn11 (approved):
Looks in good shape now. many thanks!

PR #379, opened 2025-06-24
fix(graphs): throw error instead of raising warning when graph exists. 

matschreiner:
## Description

The warning currently thrown by `anemoi-graphs create` when a graph already exists is too subtle. If you don’t scroll up in a successful-looking output, you will miss that the graph creation failed. Here’s an **example output from a failed graph creation** from  my terminal: 

Screenshot from 2025-06-24 08-34-22

I began working on a new PR because I thought there was a bug elsewhere,  because I didn’t notice the warning.

To make failure more noisy, I’ve changed the behavior to raise an explicit error instead of a warning:

Screenshot from 2025-06-24 08-58-28

 ##

### DIFF
--- a/graphs/src/anemoi/graphs/create.py
+++ b/graphs/src/anemoi/graphs/create.py
@@ -162,7 +162,7 @@ class GraphCreator:
             torch.save(graph, save_path)
             LOGGER.info(f"Graph saved at {save_path}.")
         else:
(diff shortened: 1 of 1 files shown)


matschreiner:
@mchantry I'm not sure how to fix the failing check with labels, I don't have permission to do anything related to labels

PR #381, opened 2025-06-24
test: add test to restart training from existing checkpoint

VeraChristina:
## Description
This PR adds a test to restart training from an existing checkpoint.

It is not aimed to be a blocking test but a first step to help track changes that result in incompatibilities with existing checkpoints which could lead to problems with e.g. transfer learning.

Related to the discussion about transfer learning but not directly addressing the issue discussed.

### DIFF
--- a/graphs/src/anemoi/graphs/create.py
+++ b/graphs/src/anemoi/graphs/create.py
@@ -162,7 +162,7 @@ class GraphCreator:
             torch.save(graph, save_path)
             LOGGER.info(f"Graph saved at {save_path}.")
         else:
-            LOGGER.info("Graph already exists. Use overwrite=True to overwrite.")
+            raise ValueError(f"Graph already exists at {save_path}. Use overwrite=True to overwrite.")
 
     def create(self, save_path: Path | None = None, overwrite: bool = False) -> HeteroData:
         """Create the graph and save it to the output path.
--- a/training/docs/troubleshooting.rst
+++ b/training/docs/troubleshooting.rst
@@ -35,6 +35,28 @@ adjustments:
 These settings limit the data processed and use a single GPU, helping
 isolate issues related to data or parallelization.
 
+****************
+ PyTorch errors
+****************
+
+1. Error when opening checkpoints > 2 GB
+========================================
+
+If you're using a version of PyTorch < 2.6 with checkpoint files > 2 GB
+you may encounter an error such as
+
+.. code:: python
+
+   RuntimeError: PytorchStreamReader failed reading zip archive: invalid header or archive is corrupted
+
+This is due to a limitation with the zip implementation. If you're
+unable to move to a more recent version of PyTorch you can try to remove
+the metadata stored in the checkpoint using
+
+.. code:: bash
+
+   anemoi-utils metadata checkpoint.ckpt --remove --output output.ckpt
+
 ***********************************
  PyTorch Lightning Debugging Tools
 ***********************************
--- a/training/src/anemoi/training/diagnostics/callbacks/checkpoint.py
+++ b/training/src/anemoi/training/diagnostics/callbacks/checkpoint.py
@@ -180,7 +180,7 @@ class AnemoiCheckpoint(ModelCheckpoint):
             tmp_supporting_arrays = model.supporting_arrays
             model.supporting_arrays = None
 
-            # Make sure we don't accidentally modidy these
+            # Make sure we don't accidentally modify these
             metadata = tmp_metadata.copy()
             supporting_arrays = tmp_supporting_arrays.copy()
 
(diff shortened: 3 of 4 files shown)


training/tests/integration/conftest.py
```diff
@@ -222,3 +223,18 @@ def gnn_config_with_data(
     cfg = OmegaConf.merge(template, testing_modifications_with_temp_dir, use_case_modifications)
     OmegaConf.resolve(cfg)
     return cfg
+
+
+@pytest.fixture
+def gnn_config_with_checkpoint(gnn_config_with_data: OmegaConf) -> OmegaConf:
+    existing_ckpt = get_test_data("anemoi-integration-tests/training/checkpoints/testing-checkpoint-global-Jun-24.ckpt")
```
MeraX: Would this filename enable versioning it? I would imagine that in the future, a new checkpoint would be needed so having some kind of version number in the path could help to keep things sorted. E.g. one could includethe version number of anemoi-training that generated this checkpoint.

And out of curiosity, how is this checkpoint been generated? Is it simply the result of `test_training_cycle_without_config_validation`?

VeraChristina: I thought I'd include the creation date for purposes of "versioning". That's less precise than including a version of training (though maybe not given that we don't currently pin fixed versions of models/graphs, which I suspect might be more relevant) --  but since it's mostly for purposes of tracking how frequently we break things, I thought it might be good enough. Not a strongly held opinion though!

And yes, it's created by running the (non_existing) `test_training_cycle_gnn` with the existing fixture `gnn_config` (which does have config validation) on main.

MeraX: I would be fine with using the creation date. But I would suggest including the year and perhaps a numerical representation of the month. e.g. `testing-checkpoint-global-2025-06-24.ckpt`

mchantry: Do we want the current one being used to just be called "latest.ckpt" and then when it is replaced, the old one is move to a more descriptive name?

VeraChristina: > I would be fine with using the creation date. But I would suggest including the year and perhaps a numerical representation of the month. e.g. `testing-checkpoint-global-2025-06-24.ckpt`

Fair enough!

> Do we want the current one being used to just be called "latest.ckpt" and then when it is replaced, the old one is move to a more descriptive name?

I do like this -- it would also mean we don't need to do a PR when replacing a checkpoint, just update the object store. But how do we keep track of what the future name of the current "latest.ckpt" will be? (Unless we want to record the day when it broke (not when it was created)

VeraChristina: We could put a txt file in the object store and put the next name there, but that requires people to update it and there is no mechanism in place to remind them. (Unlike with a failing test because you forgot to update the path to the new checkpoint.)

mchantry: I was going to propose symlinks, but these don't seem to be supported by S3

mchantry: It's probably best to make it clear in the test then, and use the dated name as you have above. It means more code changes, but its more transparent.

MeraX: I would never modify files that are used in (released versions) of tests. Doing so would render tests unreproducible. If there were a breaking change requiring an updated checkpoint, updating one path in the code would probably the smallest change of the whole PR.

VeraChristina: Yes, probably best to keep track of the checkpoint history in the commit history instead of keeping track of it in a different place and way. -- I have renamed the checkpoint accordingly

training/tests/integration/conftest.py
```diff
@@ -176,3 +177,18 @@ def gnn_config(
     cfg = OmegaConf.merge(template, testing_modifications_with_temp_dir, use_case_modifications)
     OmegaConf.resolve(cfg)
     return cfg, dataset_urls[0]
+
+
+@pytest.fixture
+def gnn_config_with_checkpoint(gnn_config: OmegaConf, get_test_data: callable) -> OmegaConf:
+    cfg, dataset_url = gnn_config
+    existing_ckpt = get_test_data(
```
MeraX, line 185: Hi Vera, could you check if the implementation here is in line with your latest improvement in https://github.com/ecmwf/anemoi-core/commit/8a2b0c8ac05412944859b65485a5af7916ac5c9a ?

VeraChristina: Yes, it is using the get_test_data fixture to download the checkpoint, and the gnn_config fixture (which indirectly uses the get_test_archive fixture), both in line with moving to fixtures for handling tmp dirs in the downloading utilities

mchantry (approved):
Great contribution, thanks.

training/tests/integration/conftest.py
MeraX: Thank you Vera, I've overseen that `get_test_data` is passed as a fixture.

Great PR!

PR #385, opened 2025-06-25
fix(models): uneven channel sharding

ssmmnn11:
the channel dimension was not correct for a unequal partitioning of parameters


JPXKQX (approved):
It looks good to me. The integration tests are passing

PR #386, opened 2025-06-26
feat: checkpoint migrations

bdvllrs:
New features or refactoring of models might break model checkpoints.
As discussed during a @ecmwf/anemoitechnicalsubgroup meeting, this PR introduces a systematic way of migrating model checkpoints so that they can be always up-to-date with the anemoi version. This will avoid the hard choice of updating anemoi and retrain a checkpoint or keep an old anemoi version.

## The idea
The idea is to defer the responsability of keeping a working checkpoint to the person who introduces a breaking change.
This person only has to handle **his own** breaking change. This means that he can expect that checkpoint are working before.

This takes inspiration from database migration systems that apply changes to a database incrementally.

Specifically:
-  Checkpoints will have a new "migrations" key that store migrations that were already executed (if the key does not exist, it is assumed to not have any migration executed);
- when saving a checkpoint after training, the "migrations" key is populated with all existing migrations (no need to execute them);
-  if a PR introduces a change that breaks checkpoint, a new migration must be created with:
```bash
anemoi-models migration create my-migration-name
```
Here is an example migration file:
```python
from anemoi.models.migrations import CkptType
from anemoi.models.migrations import MigrationMetadata

metadata = MigrationMetadata(
    versions={
        "migration": "1.0.0",  # this tracks the version of the migration script. Might be useful if we want to refactor the API
        "anemoi-models": "0.8.1",  # this tracks the version of anemoi-models when this migration was created
    }
)


def migrate(ckpt: CkptType) -> CkptType:
    """Migrate the checkpoint"""
    ckpt["test"] = ckpt.pop("baz")
    return ckpt


def rollback(ckpt: CkptType) -> CkptType:
    """Rollback the migration"""
    ckpt["baz"] = ckpt.pop("test")
    return ckpt
```
This file contains some version metadata, a `migrate` function that should fix the introduced breaking change of the checkpoint, and a `rollback` function that does the opposite opperation.

## User interface
The user can use the command
```bash
anemoi-models migration sync ckpt_path [--steps STEPS]
```
to execute migrations of an old checkpoint. Only missing migrations will be executed.

Additional option:
- `--steps STEPS`: will execute only STEPS migration files. If negative, will rollback STEPS times

## The pipeline
- A new PR indroduces a change that breaks checkpoints
- A CI pipeline detects that checkpoints are broken by this PR
- A new migration file must be created and filled
- CI passes again
- The PR can be merged
- Users will migrate their checkpoint with `anemoi-models migrate`
 
## Roadmap
Here is the progress status of this PR (we may want to split this for the CI check?):
- [x] a script to generate new migrations
- [x] a system to select the migrations or rollbacks to execute
- [x] callback to store existing migrations when saving checkpoints
- [x] CI to detect that a PR broke a checkpoint and ask the developper to create a new migration. This can be a complement to #381.
- [x] add docs (see tests for now for examples)

## Open Questions
- Should the migrator be automatically ran if a checkpoint is out of date? Or should it always be done manually by the users? I tend to think that it should be manual, with anemoi logging the command to run to update the checkpoint when crashing.
- How to handle inference checkpoints? (probably regenerate the inference checkpoint from the complete model after migration)
- How to handle different model types? Should there be a set of migrations per models? Or a unique set of migrations but migrations also take some additional parameters to know the model to migrate? (although it can probably be infered from the checkpoint)

----
📚 Documentation preview 📚: https://anemoi-training--386.org.readthedocs.build/en/386/

----
📚 Documentation preview 📚: https://anemoi-graphs--386.org.readthedocs.build/en/386/

----
📚 Documentation preview 📚: https://anemoi-models--386.org.readthedocs.build/en/386/


models/src/anemoi/models/commands/migrate.py
```diff
@@ -0,0 +1,101 @@
...
+            default=None,
+            type=int,
+            help=(
+                "Relative number of steps to execute. Positive migrates, negative rollbacks. "
+                "Defaults to execute all migrations."
+                "Mutually exclusive with --n_migrations and --target"
+            ),
+        )
+        command_parser.add_argument(
+            "--n_migrations",
```
floriankrb: `n-migrations` instead of `n_migrations` please

bdvllrs: done in 665f3ef

bdvllrs:
Here are some additional thoughts on handling multiple architectures:
I updated `versions` to `metadata` in migration files, and we could record some information such as on what type of model this migration is useful for. E.g.:
```python
metadata = Metadata(
    versions={
        "migration": "1.0.0",  # this tracks the version of the migration script. Might be useful if we want to refactor the API
        "anemoi-models": "0.8.1",  # this tracks the version of anemoi-models when this migration was created
    },
    architecture="graphtransformer",
)
```
We can add a filtering system on the migrator, to only apply migrations that follows certain constrains:
```python
migrator = Migrator(only={"architecture": ["graphtransformer"]})
```

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+CkptType = MutableMapping[str, Any]
+
+
+# migration is the version of the migration module to allow future update of
+# the script and keep backward compatibility
+Versions = TypedDict("Versions", {"migration": str, "anemoi-models": str})
+
+
+@dataclass
+class Metadata:
```
gmertes: Could we come up with a better name, so that it is clear that this is migration-metadata? Metadata is already overloaded with the model's metadata and could cause some confusion.

gmertes (commented):
This is really cool stuff!

I do have one suggestion about the CLI: I suggest to consolidate CLI commands as much as possible. So instead of creating two distinct commands, put everything under 1 shared command and use subparsers for the different actions. Extra actions can then easily be added to the top level command.

I suggest something like:
```
anemoi-models migration create  (instead of create-migration)
anemoi-models migration apply   (instead of migrate)
```

Alternatively, `migrate` instead of `migration`.

We already use subparsers in many places, for example [here](https://github.com/ecmwf/anemoi-core/blob/fecc928470b3fd3662253b6cb575ad8c8b78c14a/training/src/anemoi/training/commands/mlflow.py#L24).

models/src/anemoi/models/migrations/__init__.py
bdvllrs: Sure! I renamed to MigrationMetadata

bdvllrs:
> One thought on the proposed pipeline. Some changes might break checkpoints to a degree where it does not make sense to patch them (an example would be some innovation that completely replaces parts of one of the model architectures), so requiring a migration file for PRs might be too strict? In such cases it would still be useful to somehow flag the change as breaking in the Migrator, and for the PR author to explain why & how.

Hi @havardhhaugen , this is a great idea! I added this feature with the "final" metadata in the migration. 
A final migration indicates that there is a change that cannot be migrated. It can be generated with:
```
anemoi-models migration create --final my-migration-name
```
and it looks like 
```python
from anemoi.models.migrations import CkptType
from anemoi.models.migrations import FinalMigrationException
from anemoi.models.migrations import MigrationMetadata

metadata = MigrationMetadata(
    versions={
        "migration": "1.0.0",
        "anemoi-models": "1.0.0",
    },
    final=True,
)


def migrate(ckpt: CkptType) -> CkptType:
    """Migrate the checkpoint."""
    raise FinalMigrationException


def rollback(ckpt: CkptType) -> CkptType:
    """Rollback the migration."""
    raise FinalMigrationException
```

The migrator will try to migrate up until the next final migration. If an explicit number of steps if provided, it will error out with a `FinalMigrationException`.

Similarly, too recent checkpoint cannot be rollbacked before a final migration.

bdvllrs:
@gmertes  Thanks for the comment!

> anemoi-models migration create  (instead of create-migration)
> anemoi-models migration apply   (instead of migrate)

I went for `anemoi-models migration create` and `anemoi-models migration sync` to follow the python API.


Additionnaly I removed all of the `--target` and `--n-migrations` options as I found they do not bring anything. It's easier to just pip install the desired version and call `anemoi-models migration sync`.

models/docs/cli/migration.rst
```diff
@@ -0,0 +1,66 @@
...
+ Inspect
+********
+
+To inspect the registered versions in the checkpoint, run:
+
+.. code:: bash
+
+   anemoi-models migration inspect PATH_TO_CKPT
+
+You can remove colors with the ``--no-color`` argument.
```
anaprietonem: could be worth adding an example of that command to the docs?
(and similar for the other commands)

models/docs/contributing.rst
```diff
@@ -18,3 +18,14 @@ anemoi-models include unit tests that can be executed locally using
 pytest. For more information on testing, please refer to the
 :ref:`general Anemoi testing guidelines
 <anemoi-docs:testing-guidelines>`.
+
+*******************************
+ Provide checkpoint Migrations
+*******************************
+
+If your changes break existing checkpoints, you must provide a
```
anaprietonem, line 26: Could be worth mentioning here how one can detect if the changes break the checkpoint - ie mention the tests that  checks that? @VeraChristina @mchantry  is this already documented or just in confluence for now?

Looking at the inspect command this one should be able to tell you so maybe that's the recommended way?

models/docs/cli/migration.rst
```diff
@@ -0,0 +1,66 @@
...
+********
+ Migrate
+********
+
+If you want to use your checkpoint with the currently installed version of anemoi-models,
+you can use:
+
+.. code:: bash
+
+   anemoi-models migration sync PATH_TO_CKPT
```
anaprietonem, line 51: Not related strictly to the PR but the CLI in training also have some utilities related to checkpoints - https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/commands/checkpoint.py to convert a training checkpoint to an inference one. It could be good in the future to think about unifying these two CLIs (note the training conversion requires pytorch lightning so that's why it lives in anemoi-training and not in models)

models/docs/create-migrations.rst
```diff
@@ -0,0 +1,319 @@
...
+
+.. code:: bash
+
+   anemoi-models migration create --final MIGRATION_NAME
+
+**************
+ Full example
+**************
+
+Here is a full example of a migration to fix `PR 433
```
anaprietonem, line 357: Thanks for this example! very nice

models/docs/create-migrations.rst
```diff
@@ -0,0 +1,319 @@
...
+       """
+       Migrate setup callback to be run before loading the checkpoint.
+
+       Parameters
+       ----------
+       context : MigrationContext
+          A MigrationContext instance
+       """
+
+To generate your script with the setup callbacks, use the
```
anaprietonem, line 232: it should be possible to create a single migration script that handles both. the case of a layer renaming and this one right?
I think it could be worth adding a bit more here in regards to when are setup callbacks required and when 'renaming' are just enough

models/src/anemoi/models/commands/migration.py
```diff
@@ -0,0 +1,372 @@
...
+        if args.with_setup:
+            imports_items.append("from anemoi.models.migrations import MigrationContext")
+        imports_items.append("from anemoi.models.migrations import MigrationMetadata")
+        imports = "\n".join(imports_items)
+
+        with open(args.path / name, "w") as f:
+            f.write(
+                dedent(
+                    """\
+                    # (C) Copyright 2024 Anemoi contributors.
```
anaprietonem: as these are new files it would be better to use the header for 2025 - ``# (C) Copyright 2025 Anemoi contributors.``

models/docs/create-migrations.rst
```diff
@@ -0,0 +1,319 @@
...
+
+.. code:: python
+
+   from anemoi.models.migrations import CkptType
+   from anemoi.models.migrations import MigrationContext
+   from anemoi.models.migrations import MigrationMetadata
+
+   metadata = MigrationMetadata(
+       versions={
+           "migration": "1.0.0",
```
anaprietonem, line 263: how those this versioning works? (ie when would someone do migration 1.0.1 or 1.1.0 or 2.0.0?

models/src/anemoi/models/commands/migration.py
```diff
@@ -0,0 +1,372 @@
...
+    return f"{now}_{name}.py"
+
+
+def maybe_plural(count: int, text: str) -> str:
+    if count >= 2:
+        return text + "s"
+    return text
+
+
+class Migration(Command):
```
anaprietonem, line 170: what happens if one try to run this CLI with ckpt that is the inference one? Should be have a catch or error message to remember users to just use this with a training ckpt?

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
+# (C) Copyright 2024 Anemoi contributors.
```
anaprietonem: can this be updated to 2025 pls?

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
...
+    -------
+    CkptType
+
+    """
+    import torch
+
+    pickle_module = get_unpickler(replace_attrs)
+    ckpt = torch.load(path, map_location="cpu", pickle_module=pickle_module, weights_only=False)
+    # TODO: remove this. Only for testing
+    if _ckpt_migration_key not in ckpt:
```
anaprietonem: just a reminder of this #TODO

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
...
+import cloudpickle
+
+MIGRATION_PATH = Path(__file__).parent
+
+_ckpt_migration_key = "migrations"
+
+LOGGER = logging.getLogger(__name__)
+
+
+class MissingMigrationException(BaseException):
```
anaprietonem: when is this error triggered? @aaron-hopkinson could this be tackling the aspect you raised at the community meeting?

models/docs/create-migrations.rst
```diff
@@ -0,0 +1,319 @@
...
+ Final migrations
+******************
+
+If the modifications are too complex, and it is decided that migrating
+old checkpoint should not be supported, you can create a "final"
+migration with:
+
+.. code:: bash
+
+   anemoi-models migration create --final MIGRATION_NAME
```
anaprietonem, line 351: I struggle a bit to understand this final feature - so just to check say I have a checkpoint from 3 months ago and I want to update it so I can run it with my latest main but that's not possible because the code has changed a lot - if I run `` anemoi-models migration create --final MIGRATION_NAME`` that will indicate me the  (least old) version of anemoi that I can use?

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
...
+        Returns
+        -------
+        list[str]
+            The names of registered migrations
+        """
+        if _ckpt_migration_key not in ckpt:
+            return []
+        return [Migration.from_serialized(migration) for migration in ckpt[_ckpt_migration_key]]
+
+    def register_migrations(self, ckpt: CkptType) -> CkptType:
```
anaprietonem: where are those migrations being stored? (is this in the checkpoint metadata or as part of the state dict?)

models/docs/contributing.rst
mchantry: I don't think it's documented but Vera can confirm. Adding a quick mention to this test would be a nice addition.

models/src/anemoi/models/migrations/1750845283_test.py
```diff
@@ -0,0 +1,29 @@
+# (C) Copyright 2024 Anemoi contributors.
```
anaprietonem: where is the numbers for these files coming from?

training/tests/integration/conftest.py
```diff
@@ -227,18 +228,27 @@ def gnn_config(testing_modifications_with_temp_dir: DictConfig, get_tmp_paths: G
...
+    migrator: Migrator,
     gnn_config: tuple[DictConfig, str],
     get_test_data: GetTestData,
 ) -> tuple[DictConfig, str]:
     cfg, dataset_url = gnn_config
     existing_ckpt = get_test_data(
         "anemoi-integration-tests/training/checkpoints/testing-checkpoint-global-2025-07-31.ckpt",
     )
+
+    _, new_ckpt, _ = migrator.sync(existing_ckpt)
```
anaprietonem, line 251: as it is right now - the test loads the checkpoint and then it tries to migrate and then save back the checkpoint migrated? So if the test fails the user should:
- update that migration script if possible
- if not possible - still provide a new checkpoint

Might be worth adding some comments about this - pinging @VeraChristina @mchantry for visibility

training/src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -195,6 +196,7 @@ def get_callbacks(config: DictConfig) -> list[Callback]:
         (
```
anaprietonem, line 197: add comment above (line 195) to also say which callback is this?

models/pyproject.toml
```diff
@@ -43,8 +43,10 @@ dynamic = [ "version" ]
 dependencies = [
   # Fix certain dependencies during development
   "anemoi-utils>=0.1.9",
+  "cloudpickle>=3.1",
```
anaprietonem: what about adding those as separate section like:

optional-dependencies.migrations = [
  "cloudpickle>=3.1",
  "rich>=14",
]

Users would need to then do ` pip install anemoi-models[migrations]` to get these but it's then easier to control the dependencies coming from the migration functionalities

models/docs/create-migrations.rst
```diff
@@ -0,0 +1,319 @@
...
+       ----------
+       context : MigrationContext
+          A context object with some utilities
+       """
+       context.move_attribute("anemoi.training.schemas.data.DataSchema", "anemoi.models.schemas.data_processor.DataSchema")
+       context.move_attribute(
+           "anemoi.models.schemas.data_processor.DataSchema", "anemoi.training.schemas.data.NormalizerSchema"
+       )
+
+
```
anaprietonem, line 419: could be worth adding here a note about the inference checkpoints? (about regenerating the inference checkpoint once the training one is migrated)

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
...
+        self._grouped_migrations: list[list[Migration]] = []
+        current_group: list[Migration] = []
+        for migration in migrations:
+            if migration.metadata.final:
+                self._grouped_migrations.append(current_group)
+                current_group = []
+            current_group.append(migration)
+        self._grouped_migrations.append(current_group)
+
+        self._raise_missing_migrations = raise_missing_migrations
```
anaprietonem: where is this used?

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
...
+            Whether to check if there are out of order migrations missing from the checkpoint
+        """
+
+        if migrations is None:
+            migrations = _migrations_from_path(MIGRATION_PATH, __name__)
+
+        # Compatibility groups. Checkpoints cannot be migrated past their
+        # own group. This is useful to indicate when migrating checkpoints is no longer
+        # supported.
+        self._grouped_migrations: list[list[Migration]] = []
```
anaprietonem: How could someone pass a migrations that are not in the current folder?

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
...
+        current_group: list[Migration] = []
+        for migration in migrations:
+            if migration.metadata.final:
+                self._grouped_migrations.append(current_group)
+                current_group = []
+            current_group.append(migration)
+        self._grouped_migrations.append(current_group)
```
anaprietonem: could we add a logger to print here a 'summary' of the migrations found?

models/src/anemoi/models/migrations/__init__.py
```diff
@@ -0,0 +1,691 @@
...
+        migrations : Sequence[Migration] | None, default None
+            List of migration to execute. If None, get migrations from the current folder.
+        raise_missing_migrations : bool
+            Whether to check if there are out of order migrations missing from the checkpoint
+        """
+
+        if migrations is None:
+            migrations = _migrations_from_path(MIGRATION_PATH, __name__)
+
+        # Compatibility groups. Checkpoints cannot be migrated past their
```
anaprietonem: what do you mean with a compatibiliy/checkpoint group? should we have a section about this on the docs?

models/src/anemoi/models/commands/migration.py
```diff
@@ -0,0 +1,372 @@
...
+
+        imports_items: list[str] = []
+        if not args.final:
+            imports_items.append("from anemoi.models.migrations import CkptType")
+        if args.with_setup:
+            imports_items.append("from anemoi.models.migrations import MigrationContext")
+        imports_items.append("from anemoi.models.migrations import MigrationMetadata")
+        imports = "\n".join(imports_items)
+
+        with open(args.path / name, "w") as f:
```
anaprietonem: Talking to @aaron-hopkinson, we could use jinja2 to create a template for this! What do you think @bdvllrs ?

models/src/anemoi/models/migrations/__init__.py
anaprietonem: what's the 'order' in which these migrations are applied?

models/docs/create-migrations.rst
bdvllrs: This is not to be changed by people. This just a number to track the version of the migration system. This is in case we change it in the future and still want to support migrations at a previous version, we need to know it in order to apply old code.

models/src/anemoi/models/migrations/__init__.py
bdvllrs: By the way, I'll need a new testing checkpoint with an additional {"migrations": []} in the checkpoint.

models/src/anemoi/models/migrations/__init__.py
bdvllrs: This is triggered when you try to execute a migration, but the implementation is missing in one of the scripts.
This allows having migration only scripts (no rollback) (and techically rollback only and no migrate, but I don't see how this is useful).

models/docs/create-migrations.rst
bdvllrs: Yes absolutely! I'll updates the docs to make it clear.

models/docs/create-migrations.rst
bdvllrs: I added a not in the docs that the metadata object is automatically generated and that contributors should not update it.

models/pyproject.toml
floriankrb: `rich` is already a dependency of anemoi-utils.
">=14" could be added there (if it is needed).

models/docs/create-migrations.rst
bdvllrs: Not exactly. This is not for end users, but for a contributor that introduced a feature that cannot be migrated (e.g. a new layer is added, and a full retrain is needed to compute the missing weights). Then he has to generate a "final" migration that acts as a separator (checkpoint trained before this migration cannot be migrated passed it).

Until now, the API you describe doesn't exist. I added in 3f447d4 the fisrt incompatible anemoi-models version information in the error message when you try to sync a checkpoint too old. Also the inspect command now also shows it.

models/src/anemoi/models/migrations/__init__.py
bdvllrs: it's in a new "migrations" key at the root of the checkpoint. So neither?

models/src/anemoi/models/migrations/1750845283_test.py
bdvllrs: It's in the anemoi-models migration create command. I blindly copied the copyright from another file. I updated everything to use 2025 now.

training/tests/integration/conftest.py
bdvllrs: Yes sounds good. But if not possible, create a final migration with `anemoi-models migration create --final some-name" and provide a new checkpoint

models/src/anemoi/models/migrations/1750845283_test.py
anaprietonem: My bad, I meant the '1750845283' and the ones used in the other files inside migrations? and thanks for updating the headers to 2025!

training/src/anemoi/training/diagnostics/callbacks/__init__.py
bdvllrs: I'm not sure what value the comments bring compared to the following code, but ok.

models/pyproject.toml
bdvllrs: done in dcd1f35e

models/docs/create-migrations.rst
bdvllrs: done in 2a504c16

models/src/anemoi/models/migrations/__init__.py
bdvllrs: good catch, this is old code. Removing.

models/src/anemoi/models/migrations/__init__.py
bdvllrs: All migrations go there. I'll remove the `--path` argument to `anemoi-models migration create` to avoid any confusion (I used this to generate test migrations).

Technically, one can still do it with `Migrator.from_path` (like the migrator fixture in the tests). Do you have a particular use case where this should be stored anywhere else?

models/src/anemoi/models/migrations/__init__.py
bdvllrs: Ok for the log. I didn't understand the following comment, sorry!

models/src/anemoi/models/migrations/__init__.py
bdvllrs: This is more of an implementation detail. It just groups together migrations that are compatible with one another (in between "final" migrations). This is to detect if the checkpoint is compatible with the current version, or if one need to use an older version of anemoi-models.

Is this explanation clearer? Should I rephrase with this the comment in the code?

models/src/anemoi/models/commands/migration.py
bdvllrs: Yes, it's probably better. I chose against it at the beginning because the generation script was easy, but it's getting too messy.

bdvllrs: done in 43c12f48

models/src/anemoi/models/commands/migration.py
bdvllrs: Added the error in 2a504c1.

models/docs/cli/migration.rst
bdvllrs: done in 9c645715

bdvllrs:
Thanks @anaprietonem  for your extensive review! I've answered back to your comments.

models/docs/create-migrations.rst
anaprietonem: Thanks for the clarification! and in those cases those people will still need to provide an update checkpoint for the CI test to pass correct? - nevermind saw your comment below

models/tests/.gitignore
```diff
@@ -0,0 +1 @@
+!checkpoints/
```
anaprietonem: if you don't add this, this should be tracked by default?

models/src/anemoi/models/migrations/__init__.py
anaprietonem: Okey that's clear now but I think updating the comment in the code could be good - a follow up question is then let' assume we have migration1, migration2, migration3(final), migration4, migration5(filnal) and migration 6
where 6 is newer and migration1 is the oldest. Then in which order is this applied? It's related to the question above about the order.

models/src/anemoi/models/migrations/__init__.py
anaprietonem: No I was just curious, i think it's better if that's removed cause I assume for debugging/understanding the flow knowing that the sync/create commands are looking at just one folder is clearer. Thanks!

models/src/anemoi/models/migrations/__init__.py
anaprietonem: Are those migrations keys then kept in the inference checkpoint? Asking also in terms of lineage and if it could be worth knowing what migrations where applied to the 'original training checkpoint' at inference time

models/src/anemoi/models/migrations/__init__.py
anaprietonem: Do you mean uploading the checkpoint to the s3 bucket for the CLI with the key {"migrations": []} stored on it right?

models/docs/cli/migration.rst
anaprietonem: maybe I didn't explain myself very well here - I meant a 'terminal output example' of the ``anemoi-model migration inspect`` command

models/src/anemoi/models/migrations/__init__.py
bdvllrs: Actually I'm realizing that I never use it, and use `IncompatbleMigrationException` istead in the code, so let's remove it.

models/src/anemoi/models/migrations/1750845283_test.py
bdvllrs: It's a timestamp of the time when `anemoi-models migration create` is triggered. This is used to keep them ordered.

There has been a concern brought up during an anemoi community meeting that this could be problematic when merging different PRs with migrations (the final order should depend on merging order, not on time of script creation).

Honestly, this is still an open question: should there be a CI test that checks that ordering is correct before merging?

models/src/anemoi/models/migrations/__init__.py
bdvllrs: - Assuming that the checkpoint has no migration registered: nothing is applied, the checpkpoint is too old to be migrated.
- Assuming that the checkpoint has migration3(final) and migration 4: nothing is applied, the checkpoint is still too old.
- Assuming the checkpoint has migration5(final) registered: migration 6 only will be applied.

Note that checkpoint do not store migrations before the last existing final migration.

models/tests/.gitignore
bdvllrs: I removed this and renamed `checkpoint` to `migrations` in the tests.

models/src/anemoi/models/commands/migration.py
```diff
@@ -0,0 +1,431 @@
...
+    return f"{now}_{name}.py"
+
+
+def maybe_plural(count: int, text: str) -> str:
+    if count >= 2:
+        return text + "s"
+    return text
+
+
+def new_migrations_from_main_branch():
```
anaprietonem, line 46: Nice solution @bdvllrs ! thanks for implementing this

models/src/anemoi/models/commands/migration.py
```diff
@@ -0,0 +1,431 @@
...
+            return
+
+        new_timestamp = int(last_upstream_name.partition("_")[0]) + 1
+        for k, name in enumerate(new_migrations):
+            path = MIGRATION_PATH / name
+            _, _, new_name = name.partition("_")
+            new_name = f"{new_timestamp + k}_{new_name}"
+            print(f"Renaming {name} to {new_name}.")
+            print(path, path.with_name(new_name))
+            # path.rename(path.with_name(new_name))
```
anaprietonem: I guess this can be removed?

models/src/anemoi/models/commands/migration.py
```diff
@@ -0,0 +1,431 @@
...
+        if last_upstream_name is None or not len(incorrect_order):
+            print("No migration to rename.")
+            return
+
+        new_timestamp = int(last_upstream_name.partition("_")[0]) + 1
+        for k, name in enumerate(new_migrations):
+            path = MIGRATION_PATH / name
+            _, _, new_name = name.partition("_")
+            new_name = f"{new_timestamp + k}_{new_name}"
+            print(f"Renaming {name} to {new_name}.")
```
anaprietonem, line 407: I noticed there is some mixing of print and LOGGER statements. I'd propose to replace the print statements and just use the LOGGER

models/src/anemoi/models/migrations/__init__.py
anaprietonem: I have updated the checkpoint on the s3 bucket - so `anemoi-integration-tests/training/checkpoints/testing-checkpoint-global-2025-07-31.ckpt` should have now a key called _ckpt_migration_key with an empty list inside!

models/src/anemoi/models/commands/migration.py
bdvllrs: Oops sorry, this was a debug print. Removing it and uncommenting the actual line that renames the files...

models/src/anemoi/models/commands/migration.py
bdvllrs: I mostly use print because it allows colors with rich. The only place I use the logger is for errors. I think the best here is to not use the logger for formatted outputs. I can remove the error handling with the logger and let the regular python exception if you prefer? Or print the error but in stderr?

bdvllrs: Given that I print the error in one command and log in the other, I decided to remove the logger completely.

anaprietonem:
> @anaprietonem I've updated the docs, hopefully things are more clear now. Here are some other changes that I've made to help you review:
> 
> * I've moved the scripts in `anemoi.models.migrations.scripts` so that they are in a separate folder
> * I've added a history of operations done in the checkpoint in the metadata under the "migrations.history" key. This will allow better debugging as it traces exactly what happenned to the checkpoint
> * I've added a "signature" key stored in the migration. This is a sha256 digest of the migration script (passed through the ast package so that it's indentation and comment independant). It will also raise a warning to the user if the signature of scripts registered in the checkpoint are different from the signature of the current script. I did this to allow us updating old scripts if we find there is a bug. People that already executed these script will then get a warning asking them to reapply the migrations.
> * the `rollback_setup` callback can automatically be inferred from the `migrate_setup` callback, so I did that.
> 
> What's missing now:
> 
> * the CI part to update the anemoi-models version in scripts.
> * update the `migration create` command to put a placeholder for the version.

Very nice Benjamin! thanks for addressing the comments and for the summary. I think this page looks very nice https://anemoi--386.org.readthedocs.build/projects/models/en/386/modules/migrations.html . Only thing is the two sections at the bottom are those expected to be autocompleted?

from the what's missing - the CI parts - that I guess refer to the branch from Gert and that's already merged - so are those two points still applicable?

bdvllrs:
> Very nice Benjamin! thanks for addressing the comments and for the summary. I think this page looks very nice https://anemoi--386.org.readthedocs.build/projects/models/en/386/modules/migrations.html . Only thing is the two sections at the bottom are those expected to be autocompleted?

Yes, it's supposed to be the API from automodule, but it seems like it's not generated, I'll look into why. (Maybe it needs a __all__ export?)

> from the what's missing - the CI parts - that I guess refer to the branch from Gert and that's already merged - so are those two points still applicable?

Yes, those are resolve already. From my side, everything is great!

The only thing I've realized: I'm checking the signature on the registered migrations, which means that checkpoint who were trained after the migration and did not ran it might get a warning if migrations change. I should probably do it from the metadata history field which actually tracks the migration scripts that were executed.

anaprietonem (approved):
Great work @bdvllrs! thanks for addressing the feedback, the PR LGTM!

PR #388, opened 2025-06-27
feat: multi-scale loss implementations (including multi-scale kcrps)

ssmmnn11:
Implement the multiscale loss, as proposed in
https://arxiv.org/abs/2506.10868

The MultiScaleLossWrapper wraps around already implemented loss capabilities and applies them to user-defined scales. The user defines "truncation matrices" to decrease the resolution at the individual scales. The following is an example loss config: 

```
training_loss:
  _target_: anemoi.training.losses.MultiscaleLossWrapper
  truncation_path: ${hardware.paths.truncation}
  filenames: ${hardware.files.truncation_loss}
  weights:
    - 1.0
    - 1.0
  keep_batch_sharded: ${model.keep_batch_sharded}

  per_scale_loss:
    _target_: anemoi.training.losses.kcrps.AlmostFairKernelCRPS
    scalers: ['node_weights']
    ignore_nans: False
    no_autocast: True
    alpha: 0.95
```

Of note, the corresponding filenames always require a `None` valued added to the truncation matrices.

The multi scale loss will now be treated as the default loss for the ensemble model. 

The aggregated loss of the various scale will be tracked in the validation metrics. A future PR will focus on implementing loss tracking for individual variables.

----
📚 Documentation preview 📚: https://anemoi-training--388.org.readthedocs.build/en/388/

----
📚 Documentation preview 📚: https://anemoi-graphs--388.org.readthedocs.build/en/388/

----
📚 Documentation preview 📚: https://anemoi-models--388.org.readthedocs.build/en/388/


JPXKQX (commented):
Thanks for the PR, Simon! I have two general questions,
1) Do you think this should be specific to the ensemble model? I am not sure if a multi-scale MSE would produce good results, but it could still be interesting to see it used as a validation metric.
2) What are your thoughts on where this functionality should be located (as part of the forecaster or as a new loss function)? are there any performance issues we should be aware of?

ssmmnn11:
> Thanks for the PR, Simon! I have two general questions,
> 
> 1. Do you think this should be specific to the ensemble model? I am not sure if a multi-scale MSE would produce good results, but it could still be interesting to see it used as a validation metric.

Probably does not make much sense for the single model. But could be tested, and having it as an option would not hurt.

> 2. What are your thoughts on where this functionality should be located (as part of the forecaster or as a new loss function)? are there any performance issues we should be aware of?

It has a performance impact if it is used - this is unavoidable. So we should be able to switch it off. It could be a new loss function. the way it is implemented now means it can be used with any loss function though - therefore I see it more as a wrapper of a loss function, or something like that.

training/src/anemoi/training/losses/__init__.py
```diff
@@ -26,6 +27,8 @@
     "LogCoshLoss",
     "MAELoss",
     "MSELoss",
+    "MultiscaleLoss",
```
anaprietonem: is there a `MutliScaleLoss` or just the `MultiscaleLossWrapper`?

training/src/anemoi/training/losses/multiscale.py
```diff
@@ -0,0 +1,230 @@
...
+from anemoi.models.distributed.graph import shard_channels
+from anemoi.models.distributed.shapes import apply_shard_shapes
+from anemoi.models.truncation import interpolate_batch
+from anemoi.models.truncation import make_truncation_matrix
+from anemoi.training.losses.base import BaseLoss
+
+LOGGER = logging.getLogger(__name__)
+
+
+class MultiscaleLossWrapper(BaseLoss):
```
anaprietonem, line 27: is it worth having a test to ensure when the wrapper is applied with just one scale the output of the loss is the same and the one of individual loss being wrapped around?

theissenhelen: Already waiting for a gpu to allow comparison.

training/src/anemoi/training/config/hardware/files/example.yaml
```diff
@@ -2,6 +2,7 @@ dataset: ???
 graph: ???
 truncation: null
 truncation_inv: null
+truncation_loss: [False]
```
jakob-schloer: Could we change the default to null or an empty list?

training/src/anemoi/training/config/training/ensemble.yaml
```diff
@@ -84,11 +99,35 @@ training_loss:
...
+    _target_: anemoi.training.losses.MultiscaleLossWrapper
+
+    truncation_path: ${hardware.paths.truncation}
+    filenames: ${hardware.files.truncation_loss}
+    keep_batch_sharded: ${model.keep_batch_sharded}
+    weights:
+     - 1.0
+     - 1.0
+
+    internal_loss:
```
jakob-schloer: Personally, I think the keyword `internal_loss` is a bit misleading. I associate that with something like a latent loss. I would probably rename the keyword to `per_scale_loss`, `scale_loss`, or `base_loss`. I don't have a strong opinion on that, though, and would leave that up to you.

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -324,11 +339,29 @@ def _step(
         """Training / validation step."""
         LOGGER.debug("SHAPES: batch.shape = %s", list(batch.shape))
 
-        loss = torch.zeros(1, dtype=batch.dtype, device=self.device, requires_grad=False)
+        loss = torch.zeros(1, dtype=batch[0].dtype, device=self.device, requires_grad=False)
+
+        if self.loss.name == "MultiscaleLossWrapper":
```
jakob-schloer: Do we have this if-block because we later want to track the losses for the different scales? In my opinion, we should avoid adding complexity to the Trainer class for the callbacks. Could we not move this logic to the `RolloutEval` class in `diagnostics.callbacks.evaluation`?

jakob-schloer (commented):
Great work! I've left a couple of minor comments. I also think that some more tests and a small section to the documentation would be great to have for that.

training/src/anemoi/training/config/training/ensemble.yaml
theissenhelen: Yes, I agree. My preference would probably be `per_scale_loss` as this explains the purpose best.

theissenhelen:
> > > What is scale in this context? Filtering on grid points for specific areas and re-weight the total loss? Because there is a FilteringLossWrapper that was used for filtering out some variables and could be used the same for grid points, then wrapped in a CombinedLoss to associate weight to each spatial area? But maybe I didn't get exactly and it is doing more than that
> > 
> > 
> > this is for scale aware training, we explain it here: https://arxiv.org/abs/2506.10868
> 
> I saw the paper, I just wonder how is this different than constraining the loss on the spatial dimension to a subset of gridpoints, do that for a set of truncation matrices and aggregate with weights. If it is the case, it could maybe be done with a CombinedLoss + FilteringLossWrapper, because that is how I worked with it on the variable dimension and it wouldn't change much to have truncation matrices instead of specific variable names. Did anyone try to implement it using existing code?

To compute the loss of one scale we need the field computed at the previous scale which is why cannot simply have a weighted sum as in the combined loss.

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -199,15 +209,49 @@ def _rollout_step(
...
+    ) -> tuple:
+        """Training / validation step."""
+        LOGGER.debug("SHAPES: batch.shape = %s", list(batch.shape))
+
+        loss = torch.zeros(self.loss.num_scales, dtype=batch[0].dtype, device=self.device, requires_grad=False)
+
+        metrics = {}
+        y_preds = []
+
+        for loss_next, metrics_next, y_preds_next, _ens_ic in self._rollout_step(
```
JPXKQX: `_ens_ic` is no longer returned by `_rollout_step()`

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -199,15 +209,49 @@ def _rollout_step(
...
+                metrics_next = self.calculate_val_metrics(
+                    y_pred_ens_group,
+                    y,
+                    self.grid_shard_slice,
+                    rollout_step,
+                )
+
+            yield loss, metrics_next, y_pred_ens_group if validation_mode else [], (x if validation_mode else None)
+
+    def _step(
```
JPXKQX: I think we could remove the entire `_step()` function. My only concern is that the loss will be created as a tensor of size `(1, )`, but I would expect it to be broadcasted later when it is added to a tensor of size `(num_scales, )`.

theissenhelen: The `_step()` method gets called by the `training_step` method of the base class, so I don't think we should remove it. The loss will return a tensor of size (num_sales) when using the `MultiscaleLossWrapper`.

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -199,15 +209,49 @@ def _rollout_step(
...
+            metrics_next = {}
+            if validation_mode:
+                metrics_next = self.calculate_val_metrics(
+                    y_pred_ens_group,
+                    y,
+                    self.grid_shard_slice,
+                    rollout_step,
+                )
+
+            yield loss, metrics_next, y_pred_ens_group if validation_mode else [], (x if validation_mode else None)
```
JPXKQX:
```suggestion
            yield loss, metrics_next, y_pred_ens_group if validation_mode else []
```
The 4th argument corresponds to the ICs, which are no longer passed

training/src/anemoi/training/train/tasks/ensforecaster.py
JPXKQX: My concern is about whether this is something specific to the ensforecaster. Another option would be to rename the `num_scales` attribute to `dims` (or other) and put it into the loss base class

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -127,16 +127,26 @@ def compute_loss_metrics(
...
+            squash=True,
+            grid_shard_slice=self.grid_shard_slice,
+            model_comm_group=self.model_comm_group,
+            grid_dim=self.grid_dim,
+            grid_shard_shape=self.grid_shard_shapes,
+        )
+
+        return loss, y_pred_ens if validation_mode else None
 
     def _rollout_step(
```
JPXKQX, line 151: I'm not sure what is specific to the `ensforecaster` in here. Do we need to reimplement the `_rollout_step()`? Is it only the self.grid_dim?

theissenhelen: The `_rollout_step` is an abstract method in `BaseRolloutGraphModule`

training/tests/unit/train/test_loss_function.py
```diff
@@ -510,3 +511,54 @@ def test_filtered_loss() -> None:
...
+    )
+
+    pred, target, loss_result = loss_inputs_multiscale
+    loss = multiscale_loss(pred, target)
+
+    assert isinstance(loss, torch.Tensor)
+    assert torch.allclose(loss, loss_result), "Loss should be equal to the expected result"
+
+
+def test_multiscale_loss_eqivalent_to_per_scale_loss() -> None:
```
JPXKQX:
```suggestion
def test_multiscale_loss_equivalent_to_per_scale_loss() -> None:
```

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -683,30 +690,48 @@ def validation_step(self, batch: torch.Tensor, batch_idx: int) -> None:
...
+            val_loss_scales, metrics, *args = self._step(batch, validation_mode=True)
+        val_loss = val_loss_scales.sum()
 
         self.log(
             "val_" + self.loss.name + "_loss",
             val_loss,
             on_epoch=True,
             on_step=True,
-            prog_bar=True,
+            prog_bar=False,
```
anaprietonem: just to clarify is this change required or just here from debugging?

anaprietonem:
Integration tests are passing https://github.com/ecmwf/anemoi-core/actions/runs/20340613399!

cathalobrien (approved):

training/src/anemoi/training/config/training/ensemble.yaml
JianjunLiu-NOAA: then, how to determine appropriate weights?

PR #390, opened 2025-06-30
fix: Improve device movement of scalers

HCookie:
## Description
Scalers were initialised as numpy arrays and unneccessary movement between devices was occuring. 
This simplifies the movement paradigm.

### DIFF
--- a/training/src/anemoi/training/losses/scalers/loss_weights_mask.py
+++ b/training/src/anemoi/training/losses/scalers/loss_weights_mask.py
@@ -12,12 +12,12 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING
 
-import torch
-
 from anemoi.training.losses.scalers.base_scaler import BaseUpdatingScaler
 from anemoi.training.utils.enums import TensorDim
 
 if TYPE_CHECKING:
+    import torch
+
     from anemoi.models.interface import AnemoiModelInterface
 
 LOGGER = logging.getLogger(__name__)
@@ -38,7 +38,7 @@ class NaNMaskScaler(BaseUpdatingScaler):
         super().__init__(norm=norm)
         del kwargs
 
-    def on_training_start(self, model: AnemoiModelInterface) -> torch.Tensor:
+    def on_training_start(self, model: AnemoiModelInterface) -> None | torch.Tensor:
         """Get loss scaling.
 
         Get  mask multiplying NaN locations with zero.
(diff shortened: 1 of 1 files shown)


ssmmnn11 (approved):
looks good.

training/src/anemoi/training/losses/scalers/base_scaler.py
```diff
@@ -185,7 +186,7 @@ def get_scaling(self) -> SCALER_DTYPE:
         scale_dims = tuple(x.value for x in self.scale_dims)
         return scale_dims, scalar_values
 
-    def get_callback_scaling_values(self, callback: AvailableCallbacks, **kwargs) -> SCALER_DTYPE:
+    def get_callback_scaling_values(self, callback: AvailableCallbacks, **kwargs) -> TENSOR_SPEC:
```
anaprietonem: how are the scalers expected to change across a training epoch or from training to validation?

sahahner: For the updating scalers, the value of the scalers can change from batch to batch.

ssmmnn11:
in aifs-mono, we had forecaster (LightningModule)-> has self.loss (loss is a nn.Module), loss has buffers: "self.register_buffer("weights", area_weights, persistent=True)" so when you create the forecaster and move it to the GPU, all this is moved to the GPU

can you implement the same for the scalers / scale tensors? during init, everything required is registered as buffer, then every update can happen on device. remove all .to() from the code.

sahahner:
Thank you Harrison for your work on this.
To me this is ready to be merged.

sahahner (approved):
Thank you Harrison for your work on this.
To me this is ready to be merged.

ssmmnn11 (changes requested):
Very good. Can you do some test runs and check runtimes?

training/src/anemoi/training/losses/scaler_tensor.py
```diff
@@ -633,11 +653,6 @@ def get_scaler(self, ndim: int, device: str | None = None) -> torch.Tensor:
             return complete_scaler.to(device)
```
ssmmnn11, line 653: if complete_scaler is None we would get torch.ones(1), which would then need to be moved on  device? Here, this would still happen in every call of the routine, is that correct? could you do either torch.ones(1, device=device) or torch.ones(1, device=some_reference_tensor)?

ssmmnn11 (changes requested):

PR #392, opened 2025-07-01
fix(models): remove redundant logic in enc-proc-dec instantiation

matschreiner:
## Description
This change removes duplicated instantiation logic in the model’s constructor — `multi_step`, `num_channels`, and `node_attributes` were previously set twice.

This is a cleanup of redundant logic in model initialization.


JPXKQX:
https://github.com/ecmwf/anemoi-core/pull/377

PR #394, opened 2025-07-01
test: Add integration tests for time interpolator

roxas6path:
## Description
Adding integration tests for the Time Interpolation related code

## Type of Change


training/tests/integration/config/test_interpolator.yaml
```diff
@@ -0,0 +1,20 @@
+# Modifications for the basic template "interpolator.yaml"
+hardware:
+  paths:
+    data: https://object-store.os-api.cci1.ecmwf.int/ml-tests/test-data/samples
+    truncation: null
+  files:
+    dataset: anemoi-integration-tests/regional-use-cases/aifs-ea-an-oper-0001-mars-o96-2017-2017-6h-v8-testing.zarr
```
VeraChristina: Would you do me a favour and update this to anemoi-integration-tests/training/datasets/aifs-ea-an-oper-0001-mars-o96-2017-2017-6h-v8-testing.zarr

-- Just because we're in the process of cleaning up the test data dir in the object store -- Thanks!

training/tests/integration/config/test_interpolator.yaml
```diff
@@ -0,0 +1,20 @@
...
+    truncation: null
+    truncation_inv: null
+    graph: null
+
+data:
+  frequency: 6h
+  timestep: 36h
```
VeraChristina: Just to confirm my understanding, we take 0h and 36h as input and predict the 6h steps in between?

VeraChristina (commented):
Thank you for adding this! It all looks good to me.

The only question I have is whether there's an easy way to speed it up? When running locally, the test takes quite a bit longer than e.g. the stretched grid test. (I guess that's to be expected with the interpolation task.) Would adding something like this to the test_interpolator.yaml work?

training:
    explicit_times:
       input: [0,4]
       target: [1,2,3]
 
 Possibly, there are good reasons we don't want to do that and are happy with the longer test run time -- just checking if this would still test all functionality and be faster.

training/tests/integration/config/test_interpolator.yaml
roxas6path: Yep, there should be no problem with moving the input and target times down to speed it up - I'll implement this in next commit

training/tests/integration/config/test_interpolator.yaml
```diff
@@ -0,0 +1,27 @@
+# Modifications for the basic template "interpolator.yaml"
+hardware:
+  paths:
+    data: https://object-store.os-api.cci1.ecmwf.int/ml-tests/test-data/samples
```
anaprietonem, line 4: is this the same server as anemoi-integration-tests? Since Anemoi is open source it would be better to try to avoid posting this paths.

roxas6path: I'm not sure - for this I simply used the same config dataset path settings thst have been used in the previously passed CRPS integration test PR.

training/tests/integration/config/test_interpolator.yaml
VeraChristina: Great, thanks! Could you also update the dataset path as commented above? The datasets are now moved

training/tests/integration/config/test_interpolator.yaml
VeraChristina: I think we can remove these paths here (or replace them by dummy/path to avoid hydra errors), but the path is still exposed in anemoi-utils. However, I thought that we put test data in this folder so that it can be accessible? So not sure if we need to keep this invisible

training/tests/integration/config/test_interpolator.yaml
roxas6path: Done in newest commit

training/tests/integration/config/test_interpolator.yaml
roxas6path: Yep this understanding is correct - config has now been updated though to do interpolation over a small time period to save time during integration tests. As you advised in another comment

VeraChristina (approved):
Thanks for adding this! LGTM

PR #395, opened 2025-07-01
feat(graphs): support to export edges to npz

JPXKQX:
## Description

This PR adds new functionality to `anemoi-graphs` so users are able to create NPZ files with the truncation matrices.

Example:
```
anemoi-graphs export_to_sparse recipe_graph.yaml truncation_matrices/ --edges-attributes-name gauss_weight --edges-name data down --edges-name down data
```

It allows to create the sparse matrices inside anemoi.

##  Additional notes ##

### DIFF
--- a/graphs/src/anemoi/graphs/export.py
+++ b/graphs/src/anemoi/graphs/export.py
@@ -9,12 +9,14 @@
 
 import os
 from pathlib import Path
+from typing import Iterable
 from typing import Tuple
-from typing import Union, Iterable
+from typing import Union
 
-import torch
-from anemoi.graphs.create import GraphCreator
 import scipy.sparse as sp
+import torch
+
+from anemoi.graphs.create import GraphCreator
 
 
 class GraphExporter:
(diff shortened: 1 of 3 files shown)


graphs/src/anemoi/graphs/export.py
```diff
@@ -0,0 +1,101 @@
...
+        num_target_nodes = self.graph[target_name].num_nodes
+        return num_source_nodes, num_target_nodes
+
+    @staticmethod
+    def get_sparse_matrix(edge_index, edge_attribute, num_source_nodes, num_target_nodes):
+        # Create sparse matrix
+        A = torch.sparse_coo_tensor(
+            edge_index, edge_attribute, (num_source_nodes, num_target_nodes), device=edge_index.device
+        )
+        return A.coalesce()
```
ssmmnn11, line 72: Do you have an option (when building graphs in general) to check for duplicated edges? like we can check for unconnected nodes?

ssmmnn11 (approved):
Very nice!

graphs/src/anemoi/graphs/export.py
JPXKQX: We only check this when the user specifies 2 or more edge builders for a subgraph. If this is the case, we delete the repeated ones

PR #398, opened 2025-07-03
feat(training)!: support torch losses

JPXKQX:
## Description

This PR adds a `TorchLoss` class to use any torch.nn class without needing to implement them inside Anemoi. This would allow us to delete the mae.py, huber.py, and the new smooth_l1.py proposed in #367 .

We would move from
```
training_loss:
   _target_: anemoi.training.losses.MAELoss
  ignore_nans: False
```
to
```
training_loss:
   _target_: anemoi.training.losses.TorchLoss
  loss: 
     _target_: torch.nn.L1Loss
  ignore_nans: False
```

## Tests

Two transformer models with 6h steps (o96, num_ch=1024) have been trained with the old and new classes to compare convergence. The experiments can be found in `MLFlow > aifs-debug` under the following names:
- PR398: TorchLoss(MSELoss)
- PR398: MSELoss

This PR reduces code duplication, as it avoids the need to reimplement functionality that is already available in Torch. It also prevents the need for future PRs to test new loss functions inside Anemoi.

### DIFF
--- a/training/src/anemoi/training/losses/__init__.py
+++ b/training/src/anemoi/training/losses/__init__.py
@@ -13,7 +13,6 @@ from .kcrps import KernelCRPS
 from .logcosh import LogCoshLoss
 from .loss import get_loss_function
 from .mse import MSELoss
-from .rmse import RMSELoss
 from .torch import TorchLoss
 from .weighted_mse import WeightedMSELoss
 
@@ -23,7 +22,6 @@ __all__ = [
     "KernelCRPS",
     "LogCoshLoss",
     "MSELoss",
-    "RMSELoss",
     "TorchLoss",
     "WeightedMSELoss",
     "get_loss_function",


training/src/anemoi/training/losses/torch.py
```diff
@@ -0,0 +1,48 @@
...
+        self.loss = loss
+        self.name = loss.__class__.__name__.lower().replace("loss", "")
+
+    def calculate_difference(self, pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        """Calculate the loss.
+
+        Parameters
+        ----------
+        pred : torch.Tensor
+            Prediction tensor, shape (bs, ensemble, lat*lon, n_outputs)
```
ssmmnn11: Minor: do we write lat*lon everywhere? better gridpoints?

mchantry:
@JPXKQX is this labelled draft for a reason?

training/src/anemoi/training/losses/torch.py
JPXKQX: naming update!

JPXKQX:
Should we remove mse, mae, rmse and huber_loss from our code base?

anaprietonem:
> Should we remove mse, mae, rmse and huber_loss from our code base?

Sounds good, but I'd say then for loading old checkpoints if we deprecate/replace anemoi.training.losses.MSELoss with torch.nn.MSELoss, you might need to provide a migration script

mchantry:
Can we see some short MLFlow runs to sanity check that you get the same convergence with the use of TorchLoss?

training/src/anemoi/training/losses/torch.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, loss: Callable, ignore_nans: bool = False, **_kwargs) -> None:
+        super().__init__(ignore_nans)
+        self.loss = instantiate(loss, reduction="none")
+
+    @property
+    def name(self) -> str:
+        """Used for logging identification purposes."""
+        return self.loss.__class__.__name__.lower().replace("loss", "")
+
+    def calculate_difference(self, pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
```
matschreiner, line 39: why is the this function not called calculate loss?

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -62,7 +62,9 @@ loss_gradient_scaling: False
 # loss function for the model
 training_loss:
   # loss class to initialise
-  _target_: anemoi.training.losses.MSELoss
+  _target_: anemoi.training.losses.TorchLoss
```
matschreiner, line 65: How come this is not pointing directly at torch.nn.MSELoss?

training/src/anemoi/training/losses/torch.py
```diff
@@ -0,0 +1,54 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TorchLoss(FunctionalLoss):
+    """Loss function."""
+
+    def __init__(self, loss: Callable, ignore_nans: bool = False, **_kwargs) -> None:
+        super().__init__(ignore_nans)
+        self.loss = instantiate(loss, reduction="none")
```
matschreiner, line 32: If I understand correctly, we inherit from FunctionalLoss here mainly to overwrite the init function so that a torch loss can be attached as an attribute. This feels like a bit of unnecessary inheritance gymnastics. What is the advantage of this compared to simply passing a callable directly to the init method of BaseLoss/FunctionalLoss? 

I think the name FunctionalLoss implies that it applies some mathematical function directly to calculate the loss, rather than requiring subclassing to implement calculate_difference.

training/src/anemoi/training/config/training/default.yaml
matschreiner: Ok, I've been reading more of the code now. 
Seeing this configuration I'd expect to just get the pure torch loss that I specify in a thin wrapper, but we are getting all the scaling functionality as well. 

I think it could be made clearer by configuring like this:

```
training_loss:
  _target_: anemoi.training.losses.ScalingLoss
  callable:
     _target_: torch.nn.MSELoss
  scalers: ['pressure_level', 'general_variable', 'node_weights']
  ignore_nans: False
```

matschreiner: So essentially this would also imply moving some of the current BaseLoss functionality. In my view, though, BaseLoss should define the interface between the training and the loss function, rather than implementing all the scaling and reduction logic. That logic should live in a child, and then if I just want a pure torch loss for my training I can point to that in the config file with 

```
training_loss:
   _target_: torch.nn.MSELoss
```

anaprietonem: Hey @matschreiner  thanks for the feedback! @HCookie as it is right now could someone just pass a loss without defining any scaler (omitting the lines 
  scalers: ['pressure_level', 'general_variable', 'node_weights']
  ignore_nans: False
in the configs?). Any thoughts on the decoupling Mathias is proposing? I'd still think the reduce could be part of the BaseClass as what you want is just a single loss value out of your loss computation

HCookie: I don't believe a pure loss can be given, particuarly with how the losses are used in the callbacks. We expect certain arguments and shapes of the result. I think the decoupling Matthias mentioned could work, we would end up with a deeper inheritance chain without muny branches or use tho. As said in another comment, clarifying the functional loss might be a useful step to take.

matschreiner: We’d only introduce one additional inheritance layer, but that layer would act as a clear interface definition for all losses. This abstract BaseLoss would define the expected interface for use in the callbacks — including things like `comm_groups`, `shapes`, etc. — but would not implement any of the reduction or scaling logic, since those are not necessary for computing a loss.

The problem with inheritance is when functionality is scattered across multiple layers and problems that should be solved with composition are solved with inheritance instead. In this case, the new layer would improve clarity by separating interface from implementation.

If we want the framework to be attractive for research, it should be easy to extend with custom components such as new losses.

anaprietonem: In the example you're giving Mathias, 
```
training_loss:
  _target_: anemoi.training.losses.ScalingLoss
  callable:
     _target_: torch.nn.MSELoss
  scalers: ['pressure_level', 'general_variable', 'node_weights']
  ignore_nans: False
````
and based on what you're proposing for a baseLoss class that doesn't implement the reduce - how would the reduce functionality of for example `torch.nn.MSELoss` interfere with the anemoi one, where in Anemoi we need to apply the scaling before the reduction and in the reduction also account for sharding?

matschreiner: I think it is fine to have the reduce and scaling functionality on a high level class that other losses that depend on that functionality can inherit from. The idea is just that at the top level, configuration should rely on abstractions instead of concrete implementations such that the framework can be extended without hacking the BaseClass implementations.

PR #399, opened 2025-07-03
refactor: BaseGraphModule and tasks introduced in anemoi-core

icedoom888:
## Description
As addressed in #270 : introducing a new base class for all anemoi-training tasks.
So far al classes had to inherit from `GraphForecaster`, even if not doing any forecasting task: eg interpolator.
Now all tasks inherit from `BaseGraphModule`, including `GraphForecaster`.

Makes inheritance easier and clearer for existing and upcoming tasks.

#270 

##  Additional notes ##


training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -19,7 +19,7 @@
 from anemoi.models.distributed.graph import gather_tensor
 from anemoi.training.utils.inicond import EnsembleInitialConditions
 
-from .forecaster import GraphForecaster
+from .base import BaseGraphModule
```
HCookie: Please use full paths for the imports

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -0,0 +1,208 @@
...
+# nor does it submit to any jurisdiction.
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING
+
+import torch
+from torch.utils.checkpoint import checkpoint
+
+from .base import BaseGraphModule
```
HCookie: Please use full paths for the imports

training/src/anemoi/training/train/tasks/interpolator.py
```diff
@@ -18,12 +18,13 @@
 from torch_geometric.data import HeteroData
 
 from anemoi.models.data_indices.collection import IndexCollection
-from anemoi.training.train.forecaster import GraphForecaster
+
+from .base import BaseGraphModule
```
HCookie: Please use full paths for the imports

HCookie (changes requested):
Missing functions

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -266,37 +336,6 @@ def set_reader_groups(
...
-
-        # get new "constants" needed for time-varying fields
-        x[:, -1, :, :, self.data_indices.model.input.forcing] = batch[
-            :,
-            self.multi_step + rollout_step,
-            :,
-            :,
-            self.data_indices.data.input.forcing,
-        ]
-        return x
```
HCookie, line 294: The ensforecaster also uses this function

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -156,10 +233,6 @@ def __init__(
         self.lr_iterations = config.training.lr.iterations
         self.lr_warmup = config.training.lr.warmup
         self.lr_min = config.training.lr.min
-        self.rollout = config.training.rollout.start
-        self.rollout_epoch_increment = config.training.rollout.epoch_increment
-        self.rollout_max = config.training.rollout.max
-
```
HCookie, line 162: Removing rollout makes sense for the base class, but some rollout logging is still in place. This will need to be moved/removed or abstracted.

HCookie (commented):
Sorry, other commenrs didnt pop up

training/src/anemoi/training/train/tasks/base.py
icedoom888: I have done something about this in the autoencoder PR, you can check it out!

training/src/anemoi/training/schemas/training.py
```diff
@@ -351,24 +350,37 @@ class BaseTrainingSchema(BaseModel):
...
+    "Rollout configuration."
     ensemble_size_per_device: PositiveInt = Field(example=1)
     "Number of ensemble member per device"
 
 
 class InterpolationSchema(BaseTrainingSchema):
-    model_task: Literal["anemoi.training.train.forecaster.GraphInterpolator"] = Field(..., alias="model_task")
+    model_task: Literal["anemoi.training.train.tasks.GraphInterpolator"] = Field(..., alias="model_task")
     "Training objective."
+    rollout: Rollout = Field(default_factory=Rollout)
```
OpheliaMiralles: If "explicit_times" are passed, rollout is not needed for the interpolator

icedoom888: Oh would you mind making this change then? I'm not super familiar with the interplay or structure

OpheliaMiralles: I can't edit only make suggestions

icedoom888: Ok i changed it!

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -31,7 +30,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-class GraphEnsForecaster(GraphForecaster):
+class GraphEnsForecaster(BaseGraphModule):
```
anaprietonem, line 33: shouldn't the 'GraphEnsForecaster' inherit from the `GraphForecaster`?  (then I'd believe we wouldn't need to repeat the advance input function)
@ssmmnn11 in case you'd like to chime in

ssmmnn11: I am not super keen on deep inheritance chains, in the diffusion model we would have something like
basemodule -> forecaster -> diffusionforecaster -> diffusiontendforecaster ...

Alternatively we can have a basemodule and then a more functional style, e.g. the ensforecaster uses function rollout_step_ens, the diffusion forecaster rollout_step_diffusion and so on (same for logging and everything else). For that, all the routines have to be factored out, and we need to do the same for init (not manually putting rollout_step= ... in each module etc., but have convenience functions instead). Then it could be reasonably quick to mix and match. But probably best to try this in a prototype and then iterate on that.

training/src/anemoi/training/train/tasks/base.py
HCookie: I mean't that you still had references to the rollout despite it not being set in this base class

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -29,9 +29,7 @@ model:
   grid_skip: 0 # Which of the input indices to use as residual connection, null if none.
 
 training:
-  model_task: anemoi.training.train.forecaster.GraphInterpolator
-
-# Instead of inferred using multistep and timeincrement, specify time indices explicitly
+  # Instead of inferred using multistep and timeincrement, specify time indices explicitly
   explicit_times:
```
mc4117, line 33: Was just wondering why we need to put the explicit times in again as it is already in the default file?

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -436,30 +404,14 @@ def on_after_batch_transfer(self, batch: torch.Tensor, _: int) -> torch.Tensor:
...
-            training_mode=True,
-            validation_mode=validation_mode,
-        ):
-            loss += loss_next
-            metrics.update(metrics_next)
-            y_preds.append(y_preds_next)
-
-        loss *= 1.0 / self.rollout
-        return loss, metrics, y_preds
+        pass
```
mc4117, line 410: I think I would leave some basic logic from _step in this class, as taking everything out of _step feels a bit extreme to me. Maybe a version with no rollout? But this is just my opinion so I'm happy to be overruled on this

training/src/anemoi/training/diagnostics/profilers.py
```diff
@@ -32,7 +32,7 @@
     from omegaconf import DictConfig
     from pytorch_lightning.utilities.types import STEP_OUTPUT
 
-    from anemoi.training.train.forecaster import GraphForecaster
+    from anemoi.training.train.tasks import GraphForecaster
```
mc4117: Does this mean that the profilers are hard-coded to only work with the GraphForecaster? Shouldn't it get this from the config?

training/src/anemoi/training/train/tasks/base.py
mc4117: Similarly I think checkpointing should be part of the base_class. I realise that is currently done in the rollout_step but maybe this could be refactored?

training/src/anemoi/training/train/tasks/ensforecaster.py
icedoom888: Couldn't really follow what you would introduce as a change here?

training/src/anemoi/training/config/interpolator.yaml
icedoom888: In my opinion it makes this new parameter more transparent in the config. 
Otherwise it would stay hidden in the default interpolator training config.
If you don't agree i can remove it.

training/src/anemoi/training/train/tasks/ensforecaster.py
anaprietonem: What I meant is that GraphEnsForecaster could inherit from 'GraphForecaster' rather than 'BaseGraphModule'. As it is currently in the main branch, so we don't need to repeat functions like advance_input in both the GraphForecaster and GraphEnsForecaster. However going for this means having more inheritance and strictly coupled classes. We can keep it as you have done right now, and in the future think about how to factor out as functions methods like 'advance_input' so they can be reused across

training/src/anemoi/training/train/tasks/base.py
icedoom888: @mc4117 I think this would defy the whole purpose of this abstract base class.
Not all modules that would inherit from this will necessarily require a similar step to what the forcatser does, similarly with the checkpointing.
Eg, take a look at my other pr: #252, here the step behaviour is different and the loss is computed between input and output. 
When inheriting from this, you implement the step and checkpointing as expected.

training/src/anemoi/training/train/tasks/ensforecaster.py
icedoom888: Oh i see, as you prefer!

training/src/anemoi/training/diagnostics/profilers.py
icedoom888: This was done by the automatic refactoring, turns out this is only used for typechecking so i changed it to import the `BaseGraphModule` instead

training/src/anemoi/training/train/tasks/ensforecaster.py
ssmmnn11: I think it would be better to do 

from anemoi.training.rollut_steps import rollout_step_deterministic # or something like that, similar for other methods

and then do 

self.rollout_step = rollout_step
self.some_routine = ...

this way we don't need a deep class structure, but we can still avoid code duplication.

training/src/anemoi/training/train/tasks/base.py
icedoom888: Just fixed this, will request a new review ;)

icedoom888:
@HCookie Looked closer to rollout occurrencies outsid eof forcaster, should be handled now ;)

training/src/anemoi/training/train/tasks/ensforecaster.py
MeraX: Simon, I could imagine, that multiple-inherence with short mixin-classes could become useful in this case? Otherwise, `rollout_step` would have to be a method (not a function) which might be confusion. Especially, when `rollout_step` would require multiple sub-methods, things would become cleaner, if all respective methods were defined in a single `Roleout` mixin class.

ssmmnn11: Maybe best for now to go with @anaprietonem  original suggestion and have GraphEnsForecaster inherit form GraphForecaster.

HCookie:
Integration tests as of this commit are passing.
https://github.com/ecmwf/anemoi-core/actions/runs/16468098350/job/46550367438

anaprietonem:
> Semantics police here: If the PR brings breaking changes then it is inherently not a refactor. It should be a feat or fix.
> 
> This is not just nit-picking: the side effect of naming this refactor is that release-please ignores it, and it would not show up in the release notes (it ignores commits that are irrelevant to the user like chore, ci, style, refactor, etc). And we definitely want breaking config changes to go in there.
> 
> Semantic-commits kind of lacks a term for this, since you're not really adding a new feature, but it's also not a refactor on account of the breaking changes :') But I would still go with feat since it is not a bug fix, and you are changing functionality.

@icedoom888 are you happy to rename this to feat?

training/src/anemoi/training/train/tasks/base.py
mc4117: I think then we can maybe revisit this in a further PR

training/src/anemoi/training/data/datamodule/singledatamodule.py
```diff
@@ -91,14 +91,18 @@ def relative_date_indices(self, val_rollout: int = 1) -> list:
...
+        rollout_max = getattr(rollout_cfg, "max", None)
+        rollout_start = getattr(rollout_cfg, "start", 1)
+        rollout_epoch_increment = getattr(rollout_cfg, "epoch_increment", 0)
+
+        # Fallback if max is None or rollout_cfg is missing
+        rollout_value = rollout_start
+        if rollout_cfg and rollout_epoch_increment > 0 and rollout_max is not None:
+            rollout_value = rollout_max
+
+        rollout = max(rollout_value, val_rollout)
```
HCookie: Can you add a warning if we fallback?

HCookie (changes requested):

HCookie:
Integration tests as of [25bd86f](https://github.com/ecmwf/anemoi-core/pull/399/commits/25bd86f9104bd6011720f245a9cc342951350aa6) are passing
https://github.com/ecmwf/anemoi-core/actions/runs/16591796855/job/46929042031

training/src/anemoi/training/data/datamodule/singledatamodule.py
HCookie: Once this is done, I am happy to approve and merge

icedoom888: Done!

HCookie:
As we requested comments at the community meeting, I will give it a couple hours before merging.

HCookie (approved):

PR #403, opened 2025-07-04
ci: memory and performance tests in CI

cathalobrien:
## Description
CI tests which run the profiler to test model performance and memory usage

This will alert us to memory usage and performance regressions on new features

##  Additional notes ##

Currently we have an integration test `test_benchmark_training_cycle` which runs an o48 GT model on 1 GPU.

It tests peak memory, av throughput and av time per batch. Currently we have a 5% tolerance for av time per batch and av throughput, and memory should batch exactly.

This test should be run while a branch is in development, to test the performance implications of the branch. it should be run at least once before merge. 

When a branch is merged into main, the tests could be run again with an `update_server=True` flag. At this point that branches values would become the new values of the server. *Note* the performance could degrade at this point, but hopefully we have run the tests before this point so we are aware of these performance changes.

```
Reference benchmark results:
--------------------
avThroughputIterPerS: 3.26iter/s (commit: 8baad, date: 2025-08-05)
avTimePerBatchS: 0.25s (commit: 8baad, date: 2025-08-05)
peakMemoryMB: 638.78MB (commit: 8baad, date: 2025-08-05)
--------------------

Local benchmark results:
--------------------
avThroughputIterPerS: 3.17iter/s (commit: 50d4f, date: 2025-08-05)
avTimePerBatchS: 0.26s (commit: 50d4f, date: 2025-08-05)
peakMemoryMB: 638.78MB (commit: 50d4f, date: 2025-08-05)
--------------------

Comparing local benchmark results against reference values from the server
PASS. Local value for avThroughputIterPerS is within 5% tolerance of the reference value (3.17iter/s local vs 3.26iter/s reference)
PASS. Local value for avTimePerBatchS is within 5% tolerance of the reference value (0.26s local vs 0.25s reference)
PASS. Local value for peakMemoryMB has improved compared to the reference value (638.78MB local vs 638.78MB reference)
```

## future extensions

- test on a larger source grid (how to get around downloading large datasets? can we read from /mlx if we test on Atos?)

- test the time spent on specific components (e.g. moving data between host and device). we dont have this granularity in the profiler currently, but these features are being worked on.

- log commit and date alongside performance on the server, to create a database of performance over time. This can be included in this PR

----
📚 Documentation preview 📚: https://anemoi-training--403.org.readthedocs.build/en/403/

----
📚 Documentation preview 📚: https://anemoi-graphs--403.org.readthedocs.build/en/403/

----
📚 Documentation preview 📚: https://anemoi-models--403.org.readthedocs.build/en/403/


training/tests/integration/config/benchmark.yaml
```diff
@@ -0,0 +1,32 @@
+# Modifications for the basic template "config.yaml"
+hardware:
+  files:
+  #TODO change to n320-o96 (transformer?)
```
anaprietonem: Yes we were discussing about this with @japols, maybe we could try to think for a different test where we just called the forward pass of the model with a matrix generated on the fly at n320/o1280?

training/tests/integration/conftest.py
```diff
@@ -194,7 +194,30 @@ def gnn_config(
...
+        ["model=graphtransformer"],
+    ],
+)
+def benchmark_config(
+    testing_modifications_with_temp_dir: OmegaConf,
+    get_tmp_paths: callable,
+) -> tuple[OmegaConf, str]:
+
+    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="benchmark"):
+        template = compose(config_name="config")
```
VeraChristina: I think you meant to use the graphtransformer, i.e.
```suggestion
        template = compose(config_name="config", overrides=["model=graphtransformer"])
```

training/tests/integration/conftest.py
```diff
@@ -194,7 +194,30 @@ def gnn_config(
...
     OmegaConf.resolve(cfg)
-    return cfg, dataset_urls[0]
+    return cfg
+
+
+@pytest.fixture(
+    params=[
+        ["model=graphtransformer"],
+    ],
+)
```
VeraChristina: Unless we're planning to run this test for several architectures/configurations, I don't think we need to parametrize the fixture. If we are planning to run it for multiple configurations, I'd suggest to reuse existing fixtures and add your benchmark config entries to those (similarly to how we add testing modifications, though later) -- happy to discuss if that seems helpful

VeraChristina (commented):
Thanks for adding this, this is great! 

I didn't have a look at the tests, just added two small comments on the fixtures. Happy to have a closer look later this week.

Also, I'd have a small preference to separate these tests out into another file, something like `test_benchmark.py`, just because I assume that they will keep growing. (Though this makes me think that we might also separate out the checkpoint tests, which we could all do in a separate PR, so certainly not blocking.)

training/tests/integration/conftest.py
cathalobrien: yes! thank you :)

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+
+# @skip_if_offline
+# add flag to save snapshot
+# add multi-gpu support
+# read from database on the s3 bucket, of a csv
+# complain if any of the values differ by more than 10%
+# can skip multi-gpu
+# add compute/nccl/memory breakdown from pytorch profiler
+@pytest.mark.longtests
+def test_benchmark_training_cycle(
```
MeraX, line 32: For the design of the error reporting of `test_benchmark_training_cycle` I would recommend doing the 3 checks first and then raising a joint Exception in case of any failure. This way, the failed test would be more verbose and could perhaps save additional testing cycles.

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+        raise ValueError(
+            f"Peak memory usage {peak_active_mem_mb:.3f}MB is greater than current benchmark peak of {results['peakMemoryMB']:.3f}MB",
+        )
+    print(
+        f"Peak memory usage of {peak_active_mem_mb:.3f}MB is equal to or less than current benchamrk peak of  {peak_active_mem_mb:.3f}MB",
+    )
+
+    throuhput_tolerance_percent = 5
+    throughput_upper_bound = results["avThroughputIterPerS"] * (100 + throuhput_tolerance_percent) / 100
+    if av_training_throughput < throughput_upper_bound:
```
MeraX: Did you consider also checking an upper bound to update the reference data in cases where performance has improved? Perhaps this could be covered by the monitoring you were mentioning in your initial post. Noting performance improvement there might be more convenient than having a failing PR pipeline.

MeraX (commented):
Hi Cathal, great work. The (compute) performance cover a new aspect and is a very valuable addition to the test suite.

Reading through the changes, it didn't become quite clear to me how the comparison data is version-controlled in case it requires an update.

> - log commit and date alongside performance on the server, to create a database of performance over time. This can be included in this PR
I support this idea!

training/tests/integration/test_benchmark.py
cathalobrien: great idea!

training/tests/integration/test_benchmark.py
cathalobrien: The tests wont fail if the throughput/memory usage improves. At the moment, the plan is to have a flag which, when passed, will always update the reference data. In this case, if I understand your point, we don't need an upper bound.

The idea here is that we can run our performance benchmarks multiple times while a PR is being developed. And then once is merged, then the reference data will be updated. This will happen regardless of how the scores change (e.g. we introduce a cool new feature but it increases memory slightly, we still want the feature).

Updating the reference data only after a merge also prevents clashing when developing multiple PRs in parallel, both of which change the performance.

MeraX: So would update the reference data without version controlling it? This would mean that my feature branch could fail after the merge of another PR to the main, as it would silently update the reference data. Of course, I could then try updating my branch to the main, but this could come with its own problems. Also updating a central reference in place, would make it hard to run tests for older releases of the code.
Thus I would always recommend assigning new ids to any reference data used in the testing instead of updating it in-place. Alternatively small reference data could be stored directly in the repo (like the expected values for other unit tests).

cathalobrien: nice feedback. I will add commit hashes to the performance database before this feature is completed. Once we have commit hashes, we can iterate through the database of past CI runs and find the latest commit in your branches history and compare the performance against that. 
There's a question about what to do in a miss in that case, I suppose fail and say "reference result not found. merge your branch with main if you want a reference result"

Adding reference data to the repo would be really nice, altough it does make the performance tests not portable, unless someone has the exact same system.

training/tests/integration/test_benchmark.py
cathalobrien: Done, 
```
Comparing local benchmark results against reference values from the server
FAIL. Local value for avThroughputIterPerS has degraded compared to the reference value (3.3iter/s local vs 3.481iter/s reference)
FAIL. Local value for avTimePerBatchS has degraded compared to the reference value (0.242s local vs 0.238s reference)
PASS. Local  value for peakMemoryMB is within tolerance of the reference value (638.750MB local vs 638.750MB reference)
The following tests failed: ['avThroughputIterPerS', 'avTimePerBatchS']
```

training/tests/integration/test_benchmark.py
cathalobrien: @MeraX commit hashes are now stored in the benchmark data. Now the local benchmark results will be compared against the latest matching commit on the server. 

```
Reference benchmark results:
--------------------
avThroughputIterPerS: 3.26iter/s (commit: 8baad, date: 2025-08-05)
avTimePerBatchS: 0.25s (commit: 8baad, date: 2025-08-05)
peakMemoryMB: 638.78MB (commit: 8baad, date: 2025-08-05)
--------------------

Local benchmark results:
--------------------
avThroughputIterPerS: 3.17iter/s (commit: 50d4f, date: 2025-08-05)
avTimePerBatchS: 0.26s (commit: 50d4f, date: 2025-08-05)
peakMemoryMB: 638.78MB (commit: 50d4f, date: 2025-08-05)
--------------------

Comparing local benchmark results against reference values from the server
PASS. Local value for avThroughputIterPerS is within 5% tolerance of the reference value (3.17iter/s local vs 3.26iter/s reference)
PASS. Local value for avTimePerBatchS is within 5% tolerance of the reference value (0.26s local vs 0.25s reference)
PASS. Local value for peakMemoryMB has improved compared to the reference value (638.78MB local vs 638.78MB reference)
```

training/src/anemoi/training/config/model/graphtransformer.yaml
```diff
@@ -56,6 +56,7 @@ decoder:
   qk_norm: False
   cpu_offload: ${model.cpu_offload}
   layer_kernels: ${model.layer_kernels}
+  shard_strategy: "edges"
```
anaprietonem, line 59: is this missing or did you spot that the code failed without it?

training/src/anemoi/training/diagnostics/profilers.py
```diff
@@ -632,15 +632,18 @@ def get_memory_profiler_df(self) -> pd.DataFrame:
...
+                time_rows_dict = {}
+                for row in time_rows:
+                    key, val = row.split(":")
+                    val = convert_to_seconds(val.strip())
+                    time_rows_dict[key] = val
+                self.time_rows_dict = time_rows_dict
+
+                memory_df = memory_df[~memory_df["Name"].isin(time_rows)]
+        except ValueError as e:
+            LOGGER.info(f"Error saving memory df: {e}")
```
anaprietonem: do you know when/why this being triggered?

training/tests/integration/config/benchmark.yaml
```diff
@@ -0,0 +1,58 @@
+# Modifications for the basic template "config.yaml"
+# TODO currentl have a hidden grid of 5 when runnign GT, change to 6 on n320
+hardware:
+  files:
+          #dataset:  aifs-ea-an-oper-0001-mars-n320-2020-2020-6h-v6.zarr #q50
+    dataset:  aifs-ea-an-oper-0001-mars-n320-1979-2023-6h-v8.zarr #q50
+  paths:
+    data:  /home/mlx/ai-ml/datasets/
```
anaprietonem: is there a way we could hide this path?

training/tests/integration/config/benchmark.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+dataloader:
+  num_workers:
+    training: 6
+    validation: 6
+  limit_batches:
+    training: 10
+    validation: 1
+  training:
```
anaprietonem: any reason why you chose an end date but not the start date? guess we could just also limit that to ensure the samples of the training batches is narrower

training/tests/integration/config/benchmark.yaml
```diff
@@ -0,0 +1,58 @@
...
+      enabled: false
+    snapshot:
+      enabled: true
+    system:
+      enabled: false #some error about metric key in extract_gpu_metrics being invalid
+  log:
+    mlflow:
+      enabled: true
+      offline: True
+      tracking_uri: ""
```
anaprietonem: you shouldn't need to pass this if the offline is activated?

training/tests/integration/config/benchmark.yaml
```diff
@@ -0,0 +1,58 @@
...
+    memory:
+      enabled: true
+    time:
+      verbose: true
+    model_summary:
+      enabled: false
+    snapshot:
+      enabled: true
+    system:
+      enabled: false #some error about metric key in extract_gpu_metrics being invalid
```
anaprietonem: can you post the error?

training/tests/integration/conftest.py
```diff
@@ -194,7 +194,66 @@ def gnn_config(
...
+#TODO change it so i can inherit from stretched_config etc. rather then rewriting
+@pytest.fixture(
+    params=[ #selects different test cases
+        #"graphtransformer_n320_1g",
+        #"gnn_n320_1g",
+        #"stretched",
+        "lam",
+    ],
+)
+def benchmark_config(
```
anaprietonem, line 247: I think it could be better to have a:
- benchmark_config for the global
- benchmark_lam_config
- benchmark_stretched_config
so that way we have a similar set up as in the 'normal integration tests' and we keep the functions clearer

training/tests/integration/conftest.py
```diff
@@ -194,7 +194,66 @@ def gnn_config(
...
+        template = compose(config_name=config_name, overrides=overrides)
+    
+    use_case_modifications = OmegaConf.load(top_level_yaml)
+
+    if test_case == "stretched" or  test_case == "lam":
+        #tmp_dir, rel_paths, dataset_urls = get_tmp_paths(use_case_modifications, ["dataset", "forcing_dataset"])
+        #dataset, forcing_dataset = rel_paths
+
+        #just remove the server paths from files
+        use_case_modifications.hardware.files.dataset = "cerra-rr-an-oper-0001-mars-5p5km-2017-2017-6h-v3-testing.zarr"
```
anaprietonem: could be worth having configs for those too? (could create a benchmark folder in our configs) or add those to https://github.com/ecmwf/anemoi-core/tree/main/training/tests/integration/config

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+        self.unit = unit
+        self.date = date
+        self.commit = commit
+        self.op = op
+        self.tolerance = tolerance
+
+    def __str__(self):
+        return f"{self.name}: {self.value:.2f}{self.unit} (commit: {self.commit[:5]}, date: {self.date})"
+
+    # header="testName,unit,date,commit,value"
```
anaprietonem: can this be removed?

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+
+@pytest.mark.longtests
+def test_benchmark_training_cycle(
+    benchmark_config: tuple[DictConfig, str], #cfg, benchmarkTestCase
+    get_test_archive: callable,
+    update_data=True,  # if true, the server will be updated with local values. if false the server values will be compared to local values
+    throw_error=True,  # if true, an error will be thrown when a benchmark test is failed
+) -> None:
+    cfg, testCase = benchmark_config
+    print(f"Benchmarking the configuration: {testCase}")
```
anaprietonem: better to use logger statements

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+        self.testCase = testCase
+        if self.testCase is not "":
+            self.store += f"/{self.testCase}"
+            if not local:
+                self.set_remote_path += f"/{self.testCase}"
+        if self.local:
+            LOGGER.debug(f"Using a local server under '{self.store}'")
+            subprocess.run(["mkdir", "-p", self.store], check=True)
+
+        self.artifactLimit=10 #How many commits artifacts will be saved at once.
```
anaprietonem: this could be a constant value we defined at top of the file for clarity

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+        else:
+            result_str += f"FAIL. Local value for {localValue.name} has degraded compared to the reference value "
+        result_str += f"({localVal:.2f}{localValue.unit} local vs {refVal:.2f}{referenceValue.unit} reference)"
+        print(result_str)
+
+        return passed
+
+    # trys to update a metric on a remote server, with a given benchmarkValue
+    #if overwrite is true, setValue wont try append. it will be like the exisitng file doesnt exist
+    def setValue(self, value: BenchmarkValue, overwrite=False):
```
anaprietonem: If we run this from many gpus of do we control that these commands just get executed by just one of the Processes?

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+            else:
+                LOGGER.debug("existing file found... appending")
+                with open(f"{self.store}/{value.name}", "a") as f:
+                    f.write(value.to_csv() + "\n")
+        else:
+            #Get existing csv
+            url = f"{self.store}/{value.name}"
+            print(f"Fetching benchmark data from {url}...")
+            try:
+                df = pd.read_csv(url)  # requires pandas 0.19.2, see comments for alternative
```
anaprietonem: is the pandas version we get with the current dependencies an issue?

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+    def __init__(self, testCase:str="", local=True):  # use a local folder to store data instead of a remote server
+        self.benchmarkValues = {}
+
+        self.local = local
+        if self.local:
+            self.store="./server"
+        else:
+            # TODO could unify these by getting via scp
+            # for reading the data we read over internet
+            # TODO should certainly not hardcode these
```
anaprietonem: these paths, are okey and already made public for the integration tests (see https://github.com/search?q=repo%3Aecmwf%2Fanemoi-core+https%3A%2F%2Fobject-store.os-api.cci1.ecmwf.int%2F&type=code) but  `data@anemoi.ecmwf.int` I would prefer if we can find a way to not have this here

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+# This prevents tests failing because someone pushed a performance improvement and a developer hasnt merged
+
+
+# example output
+#   isCommitInProject("34d9c6f4a3c7563d7a4a646e9d69544912932a13")=False
+#   isCommitInProject("34d9c6f4a3c7563d7a4a646e9d69544912932a18")=True
+# cd ..
+#   Not a git repository.
+#   isCommitInProject("34d9c6f4a3c7563d7a4a646e9d69544912932a18")=False
+def _isCommitInProject(commit):
```
anaprietonem: In anemoi-utils we have some functionality in the provenance scrip to deal with the git sha. Have you looked at it/ is there anything we could reuse https://github.com/ecmwf/anemoi-utils/blob/7c987258b8c3ccfc159175d6d8f5bf460f308499/src/anemoi/utils/provenance.py#L86?

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+    localBenchmarkResults = []
+    localBenchmarkResults.append(
+        BenchmarkValue(
+            name="avThroughputIterPerS",
+            value=av_training_throughput,
+            unit="iter/s",
+            date=yyyy_mm_dd,
+            commit=commit,
+            op=operator.ge,
+            tolerance=5,
```
anaprietonem: the tolerance could also be a constant defined at the top of the file

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+    for benchmarkValue in localBenchmarkResults:
+        print(benchmarkValue)
+    print("-" * 20 + "\n")
+
+    # either update the data on the server, or compare reference results against local results
+    if update_data:
+        print("Updating metrics on server")
+        for localBenchmarkValue in localBenchmarkResults:
+            benchmarkServer.setValue(localBenchmarkValue)
+        store_artifacts=True
```
anaprietonem: should this part of the config settings?

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,554 @@
...
+
+from anemoi.training.train.profiler import AnemoiProfiler
+
+os.environ["ANEMOI_BASE_SEED"] = "42"  # need to set base seed if running on github runners
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True" #reduce memory fragmentation
+
+LOGGER = logging.getLogger(__name__)
+
+
+class BenchmarkValue:
```
anaprietonem: I think it would be better move this to a utils script or similar, and just leave the tests on the test_benchmark.py

training/src/anemoi/training/config/model/graphtransformer.yaml
cathalobrien: its missing in main as of 11/08
https://github.com/ecmwf/anemoi-core/blob/613ea33a7b7b99c7b4109f90fa177b74b2d15bc4/training/src/anemoi/training/config/model/graphtransformer.yaml#L48

training/src/anemoi/training/diagnostics/profilers.py
cathalobrien: I suspect its to do with not running enough validation steps to produce a trace file

training/tests/integration/config/benchmark.yaml
cathalobrien: We could read from an env var, but that would be annoying

Isnt there something in anemoi utils about a default path for each system? I'll try remove the path and see if utils can find it

training/tests/integration/config/benchmark.yaml
cathalobrien: if you don't set it, the start date is set to the start of the dataset. I want to finalise the benchmarks (e.g. limit_batches 10 is just for debugging, will probably do something like 200) and then I'll set it

training/tests/integration/config/benchmark.yaml
cathalobrien: I will double check, At one point months ago pydantic complained

training/tests/integration/conftest.py
cathalobrien: yeah good idea! I've made that change now. under integration/tests/config there's a benchmark/ folder with a config for each test

training/tests/integration/conftest.py
cathalobrien: done

training/tests/integration/test_benchmark.py
cathalobrien: yes, good catch

training/tests/integration/test_benchmark.py
cathalobrien: understandable. I could refactor it so that benchmarkSever() (and the pytest) take a string which could either be a local path or a "ssh://[host]:[path]" string and based on that make a remote or local server

training/tests/integration/test_benchmark.py
cathalobrien: sure, done

training/tests/integration/test_benchmark.py
cathalobrien: haha good point, I think nothing.
I haven't tested multi-gpu yet. I will try that and see how many things break.

training/tests/integration/test_benchmark.py
cathalobrien: ah didn't see we mandate >1.3.2 
no, I'll remove this comment then

training/tests/integration/test_benchmark.py
cathalobrien: I think it makes sense to have tolerance vary per benchmark.
I would expect throughput to vary more then memory usage for instance

training/tests/integration/test_benchmark.py
cathalobrien: sure, done

training/tests/integration/test_benchmark.py
cathalobrien: my idea was that update_data would be passed as a flag to pytest. But yeah I suppose we could add artefact limit and update_data and store_artefacts into the config. Good idea

training/tests/integration/config/benchmark.yaml
anaprietonem: there is the `` ~/.config/anemoi/settings.toml `` (see example https://anemoi.ecmwf.int/datasets/aifs-ea-an-oper-0001-mars-n320-1979-2022-6h-v6?tab=locations) and Harrison introduced this https://github.com/ecmwf/anemoi-utils/blob/7c987258b8c3ccfc159175d6d8f5bf460f308499/docs/modules/config.rst#L9 last week so we could potentially look at using that env to set this path

training/tests/integration/test_benchmark.py
anaprietonem: yes agree with that, what I meant is that we could have at the top of file:
```
TOLERANCE_THROUGHPUT=5
TOLERANCE_TIMEBATCH=5
TOLERANCE_MEMORY=1
```

so it's a bit cleaner to also know what's being used for each case

training/docs/user-guide/benchmarking.rst
```diff
@@ -744,3 +744,35 @@ Improvements​​
...
 
 -  Support other components of Anemoi like ``anemoi-inference``
+
+***************************************
+ Benchmarking pytest
+***************************************
+
+Anemoi training now supports automatic and reproducible benchmarking as an integration test. Currently
+the following components are benchmarked: 'throughput, time per batch and total memory usage'.
+Each time a commit is merged into Anemoi, these benchmark tests are run. The score is then logged, alongside the date and commit, on a database.
```
mchantry: Is this correct? Maybe change to every night?

training/docs/user-guide/benchmarking.rst
```diff
@@ -744,3 +744,35 @@ Improvements​​
...
+In this way we can track the performance of anemoi over time and ensure that no pull requests unintentionally reduce performance.
+
+A suite of different anemoi configurations are tested. The current list of configurations is 
+  - global GraphTransformer model, n320 source resolution, 1 GPU
+  - global GNN model, n320 source resolution, 1 GPU
+  - LAM model
```
mchantry: Why less detail here than the previous two?

training/docs/user-guide/benchmarking.rst
mchantry, line 1: Nice documentation, thanks!

training/docs/user-guide/benchmarking.rst
```diff
@@ -744,3 +744,35 @@ Improvements​​
...
+For details on how this is implemented the code is found here: ``tests/integration/test_benchmark.py``
+
+Local Benchmarking
+===================
+
+The benchmarking tests can also be run locally.
+
+.. code:: bash
+   cd anemoi-core
+   python3 -m pytest training/tests/integration --longtests -s -k "test_benchmark_training_cycle"
```
MeraX:
```suggestion
   python3 -m pytest training/tests/integration --slow -s -k "test_benchmark_training_cycle"
```

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,592 @@
...
+        string += "-" * 20 + "\n"
+        if self.local:
+            string += f"(Server location: '{self.store}')\n"
+        else:
+            string += f"(Server location: '{self.remote_host}:{self.store}')\n"
+        return string
+
+    # trys to read a row from a csv stored on a server and create a benchmark value from that
+    # If a benchmark value is found, update list of benchmark values
+    def getValue(self, benchmarkName: str, forceGetFromServer: bool = False):
```
MeraX: just a minor detail, but I would recommend using doc strings.

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,592 @@
...
+        return f"{self.name}: {self.value:.2f}{self.unit} (commit: {self.commit[:5]}, date: {self.date})"
+
+    # header="testName,unit,date,commit,value"
+    def to_csv(self, include_header=False):
+        header = "testName,unit,date,commit,value"
+
+        result = f"{self.name},{self.unit},{self.date},{self.commit},{self.value}"
+        if include_header:
+            result = header + "\n" + result
+        return result
```
MeraX: You're wondering why there's a mix of self-written CSV code, the use of the `csv` package, and the use of `pandas` to work with CSV files.

training/tests/integration/test_benchmark.py
MeraX: Thank you. The code for handling the reference data of the SFTP server became quite lengthy now. I couldn't review it in all its details, but if it does what it's supposed to do, I'm happy.

training/docs/user-guide/benchmarking.rst
cathalobrien: thanks

training/tests/integration/test_benchmark.py
cathalobrien: added rank_zero_only wrapper for the benchmark code

training/tests/integration/test_benchmark.py
cathalobrien: done

training/tests/integration/conftest.py
```diff
@@ -227,8 +227,47 @@ def gnn_config(testing_modifications_with_temp_dir: DictConfig, get_tmp_paths: G
 
     cfg = OmegaConf.merge(template, testing_modifications_with_temp_dir, use_case_modifications)
     OmegaConf.resolve(cfg)
-    assert isinstance(cfg, DictConfig)
-    return cfg, dataset_urls[0]
+    return cfg
```
VeraChristina: Guessing this was changed unintentionally?
```suggestion
    return cfg, dataset_urls[0]
```

cathalobrien: yeah, good catch. cheers!

training/docs/user-guide/benchmarking.rst
```diff
@@ -744,3 +744,39 @@ Improvements​​
...
+Anemoi training now supports automatic and reproducible benchmarking as
+an integration test. Currently the following components are benchmarked:
+'throughput, time per batch and total memory usage'. Each time a commit
+is merged into Anemoi, these benchmark tests are run. The score is then
+logged, alongside the date and commit, on a database. In this way we can
+track the performance of anemoi over time and ensure that no pull
+requests unintentionally reduce performance.
+
+A suite of different anemoi configurations are tested. The current list of configurations is
+   -  global GraphTransformer model, n320 source resolution, 1 GPU
```
anaprietonem: @cathalobrien aren't all those going to be running with 2 gpus?

training/src/anemoi/training/utils/masks.py
```diff
@@ -94,10 +95,13 @@ def apply(
         torch.Tensor
             The masked tensor with fill_value in the positions where the mask is False.
         """
+        mask = self.mask[grid_shard_slice] if grid_shard_slice is not None else self.mask
+
         if isinstance(fill_value, torch.Tensor):
-            indices = (~self.mask).nonzero(as_tuple=True)[0].to(x.device)
```
anaprietonem: the changes shown here are due to the fact that this branch is using the LAM fix right? (bit confused otherwise)

training/pyproject.toml
```diff
@@ -84,6 +84,7 @@ optional-dependencies.profile = [
   "holistictraceanalysis>=0.2",
   "pandas>=1.3.2",
   "rich>=13.6",
+  "sshfs",
```
anaprietonem: should this rather on the test dependencies rather than on the profiler?

cathalobrien: yeah good idea

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,40 @@
...
+from anemoi.training.diagnostics.benchmark_server import benchmark
+from anemoi.training.train.profiler import AnemoiProfiler
+
+os.environ["ANEMOI_BASE_SEED"] = "42"  # need to set base seed if running on github runners
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"  # reduce memory fragmentation
+
+LOGGER = logging.getLogger(__name__)
+
+
+@pytest.mark.multigpu
```
anaprietonem, line 30: this should also be listed in the `pytest.ini` as other markers? @VeraChristina ?

training/src/anemoi/training/utils/masks.py
cathalobrien: yeah, I have merged main in now, thanks

training/docs/user-guide/benchmarking.rst
cathalobrien: I need to update the docs, thanks for the reminder

training/tests/integration/config/benchmark/graphtransformer.yaml
```diff
@@ -0,0 +1,70 @@
+# configuration for the "graphtransformed_2g_n320" benchmark
+config_validation: False # I want to disable validation
```
anaprietonem: why do we have config_validation disabled? I think for the rest of the test we have this activated

training/tests/conftest.py
```diff
@@ -49,3 +49,27 @@ def graph_with_nodes() -> HeteroData:
     graph["test_nodes"].test_attr = (torch.tensor(coords) ** 2).sum(1)
     graph["test_nodes"].mask = torch.tensor([True] * len(coords))
     return graph
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
```
anaprietonem, line 54: this might be related to my question of pytest.ini - but I'd assume this should be similar set up the same way the slow marker that's why I am asking

training/tests/integration/test_benchmark.py
VeraChristina: I think it's not necessary to avoid warnings as we register it in the pytest.configure -- but probably good for purpose of documentation

VeraChristina: Speaking of documentation, we should probably add the commands to run the benchmark tests to the docs

training/tests/integration/conftest.py
```diff
@@ -231,6 +231,50 @@ def gnn_config(testing_modifications_with_temp_dir: DictConfig, get_tmp_paths: G
...
+        overrides = []
+        base_config = "lam"
+    elif test_case == "ensemble_crps":
+        overrides = ["model=graphtransformer", "graph=multi_scale"]
+        base_config = "ensemble_crps"
+    else:
+        msg = f"Error. Unknown benchmark configuration: {test_case}"
+        raise ValueError(msg)
+
+    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="benchmark"):
```
anaprietonem, line 267: @cathalobrien for the lam and the stretched grid shouldn't we be loading the same base configs as the one used in the normal tests? 

in ```def lam_config(``` ->  

    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="test_lam"):
        template = compose(config_name="lam")

and  for the stretched grid:
```
with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="test_stretched"):
        template = compose(config_name="stretched")
        
 ```
 
 so I wonder if this might affecting some of the things you're seeing with the LAM?

File: training/tests/conftest.py (hunk 35 above)
Comment (VeraChristina) [reply training/tests/conftest.py:54]: For the slow marker we use a plugin (which does pretty much exactly this), as slow tests are a very standard thing to have, whereas for this I added a custom cli option

Comment (anaprietonem) [reply training/tests/conftest.py:54]: All good, I had missed that, thanks for the clarification Vera!

File: training/tests/integration/test_benchmark.py [hunk 37]
```diff
@@ -0,0 +1,40 @@
...
+) -> None:
+    """Runs a benchmark and then compares them against the values stored on a server."""
+    cfg, test_case = benchmark_config
+    LOGGER.info("Benchmarking the configuration: %s", test_case)
+
+    # Run model with profiler
+    reset_peak_memory_stats()
+    AnemoiProfiler(cfg).profile()
+
+    store: str = "ssh://data@anemoi.ecmwf.int:/home/data/public/anemoi-integration-tests/training/benchmarks"
```
anaprietonem: just a reminder - we agreed to discuss with Baudouin about having the path hardcoded here

training/tests/integration/test_benchmark.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.mark.slow
+def test_benchmark_training_cycle(
+    benchmark_config: tuple[DictConfig, str],  # cfg, benchmarkTestCase
+) -> None:
+    """Runs a benchmark and then compares them against the values stored on a server."""
+    cfg, test_case = benchmark_config
+    LOGGER.info("Benchmarking the configuration: %s", test_case)
+
+    # Run model with profiler
+    reset_peak_memory_stats()
```
anaprietonem, line 42: out of curiosity is this just to clean the gpus?

training/docs/user-guide/benchmarking.rst
```diff
@@ -744,3 +744,39 @@ Improvements​​
...
+   -  global GNN model, n320 source resolution, 1 GPU
+   -  LAM model
+   -  Stretched grid model
+
+You can also run the benchmark integration test while developing a PR to
+anemoi. For information on how to trigger integration tests see
+``docs/user-guide/contributing.html#integration-tests-and-use-cases``.
```
VeraChristina: Have we added some info on how to run the benchmark tests locally in the testing section? Also, this link doesn't seem to render

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,639 @@
...
+    LOGGER.info("Comparing local benchmark results against reference values from the server")
+
+    failed_tests = []
+    for local_benchmark_value in local_benchmark_results:
+        passed = benchmark_server.compare(local_benchmark_value)
+        if not passed:
+            failed_tests.append(local_benchmark_value.name)
+
+    if len(failed_tests) > 0:
+        maybe_raise_error(f"The following tests failed: {failed_tests}", throw_error)
```
anaprietonem: why this maybe_raise_error rather than just raising them?

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,639 @@
...
+
+    return branch == current_branch
+
+
+# example output
+#   isCommitInProject("34d9c6f4a3c7563d7a4a646e9d69544912932a13")=False
+#   isCommitInProject("34d9c6f4a3c7563d7a4a646e9d69544912932a18")=True
+#   cd .. # Not a git repository.
+#   isCommitInProject("34d9c6f4a3c7563d7a4a646e9d69544912932a18")=False
+def _is_commit_in_project(commit: str) -> bool:
```
anaprietonem, line 131: @cathalobrien  what would happen if someone tries to run this ci hpc tests locally from a branch with uncommitted code?

cathalobrien: if you run locally with uncommitted code, the benchmark server will find the benchmark result from your last commit and will compare against that

training/tests/integration/test_benchmark.py
cathalobrien: kind of. it cleans the logs of GPU memory.

We call cuda_memory_stats to get the peak memory used during a run. and we call this before running to clear the record of the peak memory from the previous run. this ensures we report the peak memory used during each run, and not the peak memory used by the run with the highest memory usage

training/docs/user-guide/benchmarking.rst
cathalobrien: good catch I'll fix the link and update the docs

training/tests/integration/conftest.py
cathalobrien, line 272: hey @anaprietonem, 
Yes the configs should be the same. I have the line
`config_name=base_config`
this resolves to "lam" for LAM and "stretched" for stretched grid
```
    elif test_case == "stretched":
        overrides = []
        base_config = "stretched"
    elif test_case == "lam":
        overrides = []
        base_config = "lam"
```

training/src/anemoi/training/diagnostics/benchmark_server.py
cathalobrien: yeah think that was just for developing, thanks

training/tests/integration/config/benchmark/graphtransformer.yaml
cathalobrien: haha sorry that's unclear from my comment.

I only want to run training, no validation afterwards. that's what my comment meant. If you have config_validation enabled, it enforces a minimum of 1 validation step

training/tests/integration/test_benchmark.py
cathalobrien: done, good idea

training/docs/user-guide/benchmarking.rst
cathalobrien: yeah changed. cheers

training/docs/user-guide/benchmarking.rst
cathalobrien: WIP. changed now

training/tests/integration/test_benchmark.py
cathalobrien: done, thanks for the tip

training/tests/integration/test_benchmark.py
anaprietonem: all clear! if we could add the ' this ensures we report the peak memory used during each run, and not the peak memory used by the run with the highest memory usage'  in the code could be nicer so in the future we don't delete or wonder what this line is doing

training/tests/integration/conftest.py
anaprietonem: indeed I got this wrong,  thanks for clarifying ! I resolved the comment

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,654 @@
...
+    def __init__(
+        self,
+        name: str,
+        value: float,
+        unit: str,
+        date: str,
+        commit: str,
+        op: Callable[[Any, Any], bool] = operator.le,
+        tolerance: int = 0,  # percentage
+    ):
```
anaprietonem, line 64: could be nice to have a docstring for this class

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,654 @@
...
+    failed_tests = []
+    for local_benchmark_value in local_benchmark_results:
+        passed = benchmark_server.compare(local_benchmark_value)
+        if not passed:
+            failed_tests.append(local_benchmark_value.name)
+
+    if len(failed_tests) > 0:
+        msg = f"The following tests failed: {failed_tests}"
+        artifacts = get_local_benchmark_artifacts(cfg.hardware.paths.profiler)
+        artifacts_tar = Path(f"./{test_case}_artifacts.tar.gz")
```
anaprietonem, line 729: is the idea that if a test fails the developer should go an check the artifacts to understand why? If so, we could add that to the info logging statement

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,654 @@
...
+        _tar_files(artifacts, artifacts_tar)
+        LOGGER.info("Profiling artifacts from failed run stored under: %s", artifacts_tar)
+        raise ValueError(msg)
+    # the tests have passed, possibly update the data on the server
+    update_data = update_data or _is_repo_on_branch("main")  # update if our branch is main
+    if update_data:
+        LOGGER.info("Updating metrics on server")
+        for local_benchmark_value in local_benchmark_results:
+            benchmark_server.set_value(local_benchmark_value)
+        if store_artifacts:
```
anaprietonem: this is always true, so do we need to have a if?

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,654 @@
...
+        ),
+    )  # added 1% tolerance here so it doesnt fail over a few stray kilobytes
+
+    return local_benchmark_results
+
+
+def get_local_benchmark_artifacts(profiler_path: str) -> list[Path]:
+    """Runs after a benchmark and returns a list of paths to artifacts produced by the profiler.
+
+    Currently it captures the pytorhc trace file and the memory snapshot
```
anaprietonem: small typo - pytorch trace

cathalobrien: good catch

training/src/anemoi/training/diagnostics/benchmark_server.py
cathalobrien: yeah good point, i removed it

training/src/anemoi/training/diagnostics/benchmark_server.py
cathalobrien: yeah. without this, the artifacts go under $TMPDIR/pytest/test-XXX and then are deleted once the test ends. So this saves them.
Then the user can retrieve the artifacts from the commit of the test they failed against from the server, and compare the traces. I tried to document this in the docs

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,654 @@
...
+
+        If a benchmark value has already been loaded, return it.
+        you can force to retrieve from the server with the force_get_from_server parameter.
+        If the benchmark store is remote, the CSV is downloaded.
+        Then the CSV is opend and the value retrieved.
+        If the name cant be found in the benchmark store, the function returns
+        trys to read a row from a csv and create a benchmark value from that
+        If a benchmark value is found, update list of benchmark values.
+        """
+        if not force_get_from_server and benchmark_name in self.benchmark_values:
```
anaprietonem: is `force_get_from_server` being used ?

cathalobrien: no haha, i'll remove

training/src/anemoi/training/diagnostics/benchmark_server.py
anaprietonem: I think this is fine as we hope people will be triggering those from github, but wondering if some has a new branch not yet pushed or committed, bet then we would hit the line 300:
```
            msg = "Error. Couldn't find an entry in the server sharing a commit with your branch.\n"
            msg += "Please consider merging 'main' to enable performance benchmarks"
            raise RuntimeError(msg)
 ```

File: training/docs/user-guide/benchmarking.rst [hunk 47]
```diff
@@ -744,3 +744,128 @@ Improvements​​
    able to profile other hardware like AMS GPUs or Graphcore IPUs
 
 -  Support other components of Anemoi like ``anemoi-inference``
+
+*********************
+ Benchmarking pytest
+*********************
+
```
anaprietonem, line 751: nicely documented! should we add something in regards to the reading of the datasets and then fact then this would just run from ATOS if someone wants to test them locally?

training/tests/integration/config/benchmark/graphtransformer.yaml
anaprietonem: oh right but that shouldn't be the case - as in this config_validation is related to the pydantic schemas not with anything related to training/validation?

training/tests/integration/config/benchmark/ensemble_crps.yaml
```diff
@@ -0,0 +1,75 @@
...
+  processor:
+    num_chunks: 2
+  num_channels: 256 #run a smaller model then other benchmarks since we only have 1 gpu per model here
+
+training:
+  max_epochs: 1
+  num_sanity_val_steps: 0
+  ensemble_size_per_device: 1
+
+diagnostics:
```
anaprietonem: there are some keys or even blocks repeated across all the benchmark configs (diagnotics, training) - would be an option to have a high-level config and then the modifications for each model (lam, crps, global, and ens)? Just thinking in the future it's easier to check the settings used and keep things more organised for the benchmarking

training/docs/user-guide/benchmarking.rst
cathalobrien: actually yeah thats a good point. I'll update the "running" documentation to say the local paths have to be updated

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,659 @@
...
+    ):  # use a local folder to store data instead of a remote server
+        self.benchmark_values = {}
+
+        self._parse_store_location(store)
+
+        # TestCase creates an optional subdir under BenchmarkServer to store the results
+        # So that you can store GNN_n320_1g and graphtransformer_n320_1g results under the same server
+        # If testcase is "" then no subdirs are created
+        self.test_case = test_case
+        if self.test_case != "":
```
anaprietonem: is this just checking that self.test_case is not empty string- can't just do ```if self.test_case:```

training/src/anemoi/training/diagnostics/benchmark_server.py
```diff
@@ -0,0 +1,659 @@
...
+            self.store = Path(f"{self.store}/{self.test_case}")
+
+        if not self.local:
```
anaprietonem: Looking to the code there are a few cases where we need to do 
```
if self.local:
    something related to self.store
else:
   something related to self.fs
 ```
 
 i think this seems like we could abstract this better as a:
 ```
 class Store(ABC):
 
     @abstractmethod
     def method 
     
     @abstractmethod
     def open
     
class RemoteStore(Store):
      def __init__:
        from sshfs import SSHFileSystem
        self.fs = SSHFileSystem(host, username=user)

     def open():
            self.fs.mkdir()
          
     def method():

 
 class LocalStore(Store):
     def open():
           self.store.mkdir(parents=True, exist_ok=True)
           
     def method():
     
  ```
   and then in the BenchmarkServer we could just initialise the store.

File: training/tests/integration/config/benchmark/ensemble_crps.yaml (hunk 48 above)
Comment (cathalobrien) [reply training/tests/integration/config/benchmark/ensemble_crps.yaml]: good idea, done

**Review by mchantry — CHANGES_REQUESTED** [review]
Really nice contribution thanks. I think having it/s & batch time is unnecessary duplication, which may confuse. So I would propose dropping the batch time.

File: training/src/anemoi/training/diagnostics/benchmark_server.py (hunk 50 above)
Comment (cathalobrien) [reply training/src/anemoi/training/diagnostics/benchmark_server.py]: good idea, done

File: training/src/anemoi/training/diagnostics/benchmark_server.py (hunk 49 above)
Comment (cathalobrien) [reply training/src/anemoi/training/diagnostics/benchmark_server.py]: thanks, done

File: training/src/anemoi/training/diagnostics/benchmark_server.py (hunk 42 above)
Comment (cathalobrien) [reply training/src/anemoi/training/diagnostics/benchmark_server.py:64]: ok, done

File: training/tests/integration/test_benchmark.py (hunk 22 above)
Comment (cathalobrien) [reply training/tests/integration/test_benchmark.py]: removed this keyword

File: training/tests/integration/config/benchmark/graphtransformer.yaml (hunk 34 above)
Comment (cathalobrien) [reply training/tests/integration/config/benchmark/graphtransformer.yaml]: fixed issue w pydantic not allowing 0 step training/validation/test

File: training/tests/integration/test_benchmark.py (hunk 38 above)
Comment (cathalobrien) [reply training/tests/integration/test_benchmark.py:42]: sure

Comment (cathalobrien) [reply training/tests/integration/test_benchmark.py:42]: done

File: training/src/anemoi/training/diagnostics/benchmark_server.py (hunk 41 above)
Comment (cathalobrien) [reply training/src/anemoi/training/diagnostics/benchmark_server.py:131]: good point.
This part can't really be hit, if they read from the server we update with the nightly CI tests

As soon as this branch is merged, nightly runs with the merged commit will be in the server.
Branches created **after** this merge will share this commit.
branches created **before** merge wont have the benchmark pytest code.

The only way you could hit this branch is if you were reading from a local server with newer commits then you have locally. i dont think thats very likely

**Comment by cathalobrien** [comment]
@copilot

> Really nice contribution thanks. I think having it/s & batch time is unnecessary duplication, which may confuse. So I would propose dropping the batch time.

done

**Review by mchantry — APPROVED** [review]
Great contribution, thanks.

File: training/docs/user-guide/benchmarking.rst [hunk 51]
```diff
@@ -744,3 +744,131 @@ Improvements​​
...
+benchmark is run over two GPUs to ensure the model sharding performance
+is captured in the automatic benchmarks. The tests take approximately 40
+minutes. A successful test looks like
+
+.. code::
+
+    11:54:52 INFO Reference benchmark results:
+   --------------------
+   avThroughputIterPerS: 0.41iter/s (commit: 1d58f, date: 2025-08-25)
+   avTimePerBatchS: 1.97s (commit: 1d58f, date: 2025-08-25)
```
mchantry: @cathalobrien does this need to be removed now?

cathalobrien: good catch. done

training/src/anemoi/training/config/model/graphtransformer.yaml
mchantry: To be fixed in #512

PR #405, opened 2025-07-04
fix: Update level retrieval logic

### LINKED ISSUE #404: Bug: Cannot find levelist in malformed mars metadata
### What happened?

(Not sure whether this belongs in training or datasets)
One of our datasets (https://anemoi.ecmwf.int/datasets/aifs-meps-2.5km-2020-2023-6h-v7) is missing levelist in the mars metadata. For instance:
```
  'z_925': {'mars': {'date': '20200205',
    'param': 'z_925',
    'step': 0,
    'time': '0000',
    'valid_datetime': '2020-02-05T00:00:00',
    'variable': 'z_925'}}}
```
This causes issues where the general variable and pressure level scalers are not applied correctly to pressure level variables. The issue boils down to `ExtractVariableGroupAndLevel` returning `variable_name = z_500` and `variable_level = None` instead of `z` and `500` due to the missing levellist in the metadata.

I don't think we have done anything weird when creating the dataset (can post the config if needed), and I also see that the Era datasets we use do not have this mars metadata at all (causing  `ExtractVariableGroupAndLevel` to use the correct default fallback). 

```
from anemoi.datasets import open_dataset
from anemoi.training.utils.variables_metadata import ExtractVariableGroupAndLevel
ds = open_dataset(paths["meps"]) #Available on EWC
md = ds.metadata()["variables_metadata"]
group_config={
    "default": "sfc",
    "pl": ["q", "t", "u", "v", "w", "z"]
}
variable_metadata_extractor = ExtractVariableGroupAndLevel(group_config, md)
group, var_name, var_level = variable_metadata_extractor.get_group_and_level("z_500")
print("group", group) #sfc, should be pl
print("var_name", var_name) #z_500, should be z
print("var_level", var_level) #None, should be 500
```

### Version

06dde94219119746215b767b846542ee31bbff63 

Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

HCookie:
## Description
Closes #404 
Some datasets have malformed metadata, and fail on the level retrieval

This updates the logic, including checks for surface level if attempting to use the variable metadata.

### DIFF
--- a/training/src/anemoi/training/utils/variables_metadata.py
+++ b/training/src/anemoi/training/utils/variables_metadata.py
@@ -135,6 +135,32 @@ class ExtractVariableGroupAndLevel:
 
         return self.default_group
 
+    def _is_metadata_trusted(self, variable_name: str) -> bool:
+        """Check if the metadata for a variable is trusted.
+
+        This checks if the variable has metadata and checks
+        for valid relations.
+
+        Parameters
+        ----------
+        variable_name : str
+            Name of the variable.
+
+        Returns
+        -------
+        bool
+            True if the metadata is trusted, False otherwise.
+        """
+        if variable_name not in self.metadata_variables:
+            return False
+
+        level = self.metadata_variables[variable_name].level
+        is_vertical_level = not self.metadata_variables[variable_name].is_surface_level
+
+        # If level is not None and is not a surface level, True
+        # If level is None and is a surface level, True
+        return is_vertical_level ^ (level is None)
+
     def get_param(self, variable_name: str) -> str:
         """Get the parameter from a variable_name.
 
@@ -154,8 +180,8 @@ class ExtractVariableGroupAndLevel:
             Either from the metadata or cracked
             name.
         """
-        if variable_name in self.metadata_variables:
-            # if metadata is available: get variable name and level from metadata
+        if self._is_metadata_trusted(variable_name):
+            # if metadata is available: get param from metadata
             return self.metadata_variables[variable_name].param
 
         return _crack_variable_name(variable_name)[0]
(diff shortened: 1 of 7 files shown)


training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -170,13 +170,19 @@ def get_level(self, variable_name: str) -> str | None:
...
-        variable_level : str | None
+        variable_level : int | None
             Variable level, checks the variable metadata, or attempts
             to crack the name, if not found None.
         """
         if variable_name in self.metadata_variables:
-            # if metadata is available: get variable name and level from metadata
-            return self.metadata_variables[variable_name].level
+            # if metadata is available: get level from metadata
+            level = self.metadata_variables[variable_name].level
```
mc4117: What happens if this attribute doesn't exist?

HCookie: It can't
https://github.com/ecmwf/anemoi-transform/blob/e6e9ac4904bfdf8267d4c768bd6baaa7da360b26/src/anemoi/transform/variables/variables.py#L50-L53

mc4117: Ah perfect!

havardhhaugen (changes requested):
The get_param method needs the same logic, otherwise it returns the full variable name rather than the reference name.

HCookie:
@havardhhaugen that'll be hard, as the param is in the metadata, and therefore 'trusted'....

havardhhaugen (dismissed):
That fixes it, I tested it on our dataset and the scalers look correct now.

aaron-hopkinson (dismissed):
LGTM

HCookie:
@havardhhaugen Can you check with a full training run?

mc4117 (dismissed):

havardhhaugen:
> LGTM once @havardhhaugen has confirmed it fixes his issue with a longer training run

Full training works as expected now, thanks for the fix on this!

PR #366, opened 2025-06-07
feat(models): Mapper edge sharding

japols:
## Description
Adds edge sharding to the GraphTransformerMapper, reducing communication time compared to previous "head sharding" strategy significantly at scale.

Head sharding required switching between grid- and head-sharding for each Mapper layer. The new edge sharding can be enabled with `config.model.encoder.shard_strategy = "edges"` and distributes the graph into smaller subgraphs across gpus, keeping a consistent grid-sharding scheme.
Destination nodes are equally split across gpus along the grid dimension together with their incoming edges and source nodes. This requires a "resharding" step of source nodes which is implemented as allgather + drop_unconnected_src_nodes.

Time (s) per iteration for training a n320 model across 4 gpus and o1280 across 16 gpus with different sharding strategies:
| Resolution | heads | edges | speedup |
|------------|-----------|-----------|---------|
| n320       | [0.6948](https://mlflow.ecmwf.int/#/experiments/45/runs/0fd4f50f668d4b60902e4c140dbc71e9/artifacts)   | [0.6341](https://mlflow.ecmwf.int/#/experiments/45/runs/e7a7bca2d5e1443e9abe900229a2e5e4/artifacts)   | 8.75% |
| o1280      | [4.7438](https://mlflow.ecmwf.int/#/experiments/45/runs/51069c0514814552abf865c086e5253f/artifacts)   | [3.6075](https://mlflow.ecmwf.int/#/experiments/45/runs/7e1fff0dac7b483c859689d06823f4f9/artifacts)   | 23.95% |

## What problem does this change solve?
Lots of time spent in communication when training on high res data, especially communication w.r.t. input/output grid. Now it's only a single all-gather + reduce-scatter instead of ~10 all-alls on the full grid per step.

##  Additional notes ##
In the future the "resharding" step could be implemented as a small halo exchange instead of the all gather, which would be faster and more memory efficient.

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

----
📚 Documentation preview 📚: https://anemoi-training--366.org.readthedocs.build/en/366/

----
📚 Documentation preview 📚: https://anemoi-graphs--366.org.readthedocs.build/en/366/

----
[...]

PR #406, opened 2025-07-07
feat(models): Checkpointed Mapper Chunking

japols:
## Description
Adds an improved chunking scheme for both inference and training when using the GraphTransformerMappers with edge sharding, drastically reducing peak memory usage of the Encoder/Decoder. This can be controlled via `model.encoder.num_chunks` and `model.decoder.num_chunks` or by setting the environment variable`ANEMOI_INFERENCE_NUM_CHUNKS` during inference.

This change allowed us to reduce the number of GPUs required per model from 8 (H200s) at 9 km global resolution down to just 2 GPUs/model, cutting both resource requirements and GPU-hours needed for training.

Memory profile of 1 (right) vs 4 (left) mapper chunks when training a 32km model across 4 gpus:
Screenshot 2025-02-06 at 13 51 54
Reducing peak memory from 22GB to 9GB.

High peak memory usage of Mappers reduced by chunking the computation into smaller checkpointed parts with little overlap.

##  Additional notes ##
Merging into edge sharding branch until [#366 ](https://github.com/ecmwf/anemoi-core/pull/366) is merged.


models/src/anemoi/models/layers/mapper.py
```diff
@@ -40,6 +42,10 @@
 
 LOGGER = logging.getLogger(__name__)
 
+# Number of chunks used in inference (https://github.com/ecmwf/anemoi-models/pull/46)
```
ssmmnn11: These are a bit confusing now. Can we label them better? Or remove what we think we won't use? I would only keep the new chunking, because it would be possible to use it even if the model is trained with heads sharding.

models/src/anemoi/models/layers/mapper.py
```diff
@@ -328,12 +336,56 @@ def pre_process_edge_sharding_wrapper(
...
 
-        x_src, x_dst, shapes_src, shapes_dst = self.pre_process(
-            (x_src, x_dst), shard_shapes, model_comm_group, x_src_is_sharded=True, x_dst_is_sharded=x_dst_is_sharded
-        )  # x_src is sharded since we dropped unconnected src nodes
+        if not x_dst_is_sharded:
+            x_dst = shard_tensor(x_dst, 0, shapes_dst, model_comm_group)
 
         return x_src, x_dst, edge_attr, edge_index, shapes_src, shapes_dst
 
+    def run_processor_chunk(
```
ssmmnn11: run_processor_chunk_edge_sharding? So that it is clear it is only used in edge sharding strategy.

models/src/anemoi/models/layers/mapper.py
```diff
@@ -328,12 +336,56 @@ def pre_process_edge_sharding_wrapper(
...
+            edge_index_chunk,
+            shapes,
+            batch_size,
+            chunk_size,
+            model_comm_group,
+        )
+
+        return self.post_process(x_dst_out, shapes[1], model_comm_group, keep_x_dst_sharded=True)
+
     def forward_with_edge_sharding(
```
ssmmnn11, line 388: having these two forwards is not super intuitive. Would it be possible to give a config option to the mapper when we instantiate, so it choses on forward or the other, if shard_strategy == ...: self = ...? Or similar to "model_task = get_class(self.config.training.model_task)". You would move the routines out of the mapper then. If too complicated we could to it in a follow up. 



encoder:
  _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
  sharding_strategy_forward: edge_sharding

ssmmnn11 (approved):
Very good. Some suggestions. Either now or later.

models/src/anemoi/models/layers/mapper.py
japols: I think having the possibility to chunk more aggressively in inference is still useful but I agree that we can remove the "old" mapper chunking.

models/src/anemoi/models/layers/mapper.py
japols: let's think about this in a follow up :)

ssmmnn11 (approved):
very good :-)

PR #413, opened 2025-07-10
ci: Use uv in testing workflow 🚀

gmertes:
## Description
Replace the `pip install` with `uv pip install` in the testing workflow. This workflow was already reasonably fast, but with uv the dependency installation time goes from 1-3 min (depending on the arch and python version) down to 30s.

I also renamed the workflow to the same name used in the other anemois, for consistency. 

### DIFF
--- a/.github/workflows/python-pull-request.yml
+++ b/.github/workflows/python-pull-request.yml
@@ -1,4 +1,4 @@
-name: Test PR
+name: Code Quality and Testing
 
 on:
   pull_request:
@@ -9,6 +9,9 @@ on:
   schedule:
     - cron: "9 2 * * 0" # at 9:02 on sunday
   workflow_dispatch:
+
+env:
+  UV_SYSTEM_PYTHON: 1
 
 jobs:
   quality:
@@ -26,7 +29,7 @@ jobs:
     timeout-minutes: 60
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
         fetch-tags: true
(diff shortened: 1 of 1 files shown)


.github/workflows/python-pull-request.yml
```diff
@@ -35,17 +38,16 @@ jobs:
...
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
+      with:
+        version: 0.7.19
 
     - name: Install packages
       run: |
         # Install all packages from local
-        pip install -e ./training[all,tests] -e ./graphs[all,tests] -e ./models[all,tests]
+        uv pip install -e ./training[all,tests] -e ./graphs[all,tests] -e ./models[all,tests] pytest pytest-md pytest-emoji
```
anaprietonem, line 50: what's `pytest pytest-md pytest-emoji` doing?

.github/workflows/python-pull-request.yml
```diff
@@ -35,17 +38,16 @@ jobs:
...
         python-version: ${{ matrix.python-version }}
-        cache: 'pip'
 
-    - name: Install test dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install pytest pytest-md pytest-emoji
+    - name: Install uv
+      uses: astral-sh/setup-uv@v6
+      with:
```
anaprietonem, line 44: this should be kept the same as the one in reusable workflows right? could be worth adding a comment to remind ourselves?

anaprietonem (approved):

.github/workflows/python-pull-request.yml
gmertes: These are used by the pytest action. The pytest is a bit redundant but it doesn't hurt to have it just in case,  pytest-md generates test reports and pytest-emoji to print emojis in the log depending on pass/fail 😀

Example:
image

.github/workflows/python-pull-request.yml
gmertes: I added a comment. Maybe a better long term solution is to also use the reusable-workflows one here, since this one is now doing the same thing just 3 times. But it needs some modifications to reusable-workflows to run more than 1 package.

Issue #190, opened 2025-03-14
Mlflow logging LR duplicate key issue with PostgreSQL DB


anaprietonem:
### What happened?

We are aware that when running anemoi-training using an mlflow server to log your results. If using online logging, some people are hitting the following error:
```
`mlflow.exceptions.RestException: BAD_REQUEST: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "metric_pk"
DETAIL:  Key (key, "timestamp", step, run_uuid, value, is_nan)=(lr-AdamW, 1741862947817, 60899, 843c1331fefa436bab56485cfc5bc16e, 0.000322870392593084, f) already exists.`

```

### What are the steps to reproduce the bug?

`anemoi-training train` with mlflow:enabled and offline:False

### Version

Latest

### Platform (OS and architecture)

ATOS & LUMI

### Relevant log output

```shell

```

anaprietonem:
Errors reported by colleagues at DMI (@dietervdb-meteo). The config used:
```
training:
  run_id:  ba67b02f88ff488c9365735c571a88e5
  max_epochs: null # set max_epochs or max_steps. Training stops at the first limit reached.
  max_steps: 150000 #NOTE: increase to 300,000 for high-accuracy full-scale run on LUMI

  lr:
    rate: 2.5e-4  #4 * 0.625e-4
    min: 1.2e-6 #4 * 3e-7
    iterations: ${training.max_steps}
    warmup_t: 1000
````

[mlflow_issue (1).log](https://github.com/user-attachments/files/19244281/mlflow_issue.1.log)

dietervdb-meteo:
Thanks for following up on this. This seems to happen and prematurely end training runs quite often recently.

stietsche:
Yes quite frequent but random for me as well. The key includes the time stamp so it can only be duplicate if two concurrent processes try to log the exact same thing at the exact same time. Probably needs a change to the logging somewhere to make sure only one process logs.

It's always the same RestException thrown by mlflow, looks like this:
`Training: |          | 12079/? [1:57:53<00:00,  1.71it/s, v_num=1154, train_fkcrps_step=7.410][2025-03-13 20:22:28,682][anemoi.training.diagnostics.mlflow.auth][INFO] - Your MLflow login token is valid until 2025-04-11 18:18:01 UTC
[...]

PR #414, opened 2025-07-10
fix: mlflowlogger duplicated key

anaprietonem:
Duplicate metric logged by mlflow to the database

https://github.com/ecmwf/anemoi-core/issues/190

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/diagnostics/mlflow/logger.py
+++ b/training/src/anemoi/training/diagnostics/mlflow/logger.py
@@ -16,7 +16,6 @@ import os
 import re
 import sys
 import time
-from collections import deque
 from pathlib import Path
 from threading import Thread
 from typing import TYPE_CHECKING
@@ -31,9 +30,10 @@ from pytorch_lightning.loggers.mlflow import _flatten_dict
 from pytorch_lightning.utilities.rank_zero import rank_zero_only
 from typing_extensions import override
 
+from anemoi.training.diagnostics.mlflow.utils import FixedLengthSet
 from anemoi.training.diagnostics.mlflow.utils import clean_config_params
+from anemoi.training.diagnostics.mlflow.utils import expand_iterables
 from anemoi.utils.mlflow.auth import TokenAuth
-from anemoi.utils.mlflow.utils import expand_iterables
 from anemoi.utils.mlflow.utils import health_check
 
 if TYPE_CHECKING:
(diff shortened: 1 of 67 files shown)


training/src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -458,8 +463,20 @@ def _get_mlflow_run_params(
     def experiment(self) -> MLFlowLogger.experiment:
         if rank_zero_only.rank == 0:
             self.auth.authenticate()
+        if self._run_id is None:
+            self._logged_metrics.clear()
```
gmertes: What is this here for? I think this attribute can be accessed many times during the experiment, does this check need to happen every time?

training/src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -458,8 +463,20 @@ def _get_mlflow_run_params(
     def experiment(self) -> MLFlowLogger.experiment:
```
gmertes, line 467: You could add the override decorator here too

training/src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -349,6 +352,8 @@ def __init__(
             run_id=run_id,
         )
 
+        self._logged_metrics = set()  # Track (key, step)
```
gmertes: Maybe add some more explanation to this comment, something like:
`track previously logged metrics to prevent duplicate logs`

training/src/anemoi/training/diagnostics/mlflow/logger.py
anaprietonem: So this is here to clear the cache. Ultimately the MlflowLogger from PTL, logs the info using `self.experiment.log_batch` so then by calling this in the experiment, we can make sure that you clear your metrics between batch and hence don't have a huge set of metrics being dragged along the training run. I had run a couple of 6 hr experiments, but I can submit this week one longer

gmertes: I'm not sure if I understand how this `if` statement relates to clearing the cache between batches. After the first metrics have been logged, `self._run_id` will be set for the rest of the experiment, no? So how does it clear the cache between batches? Is this there to clear the cache on the rank > 0 ? If yes then checking `rank_zero_only.rank` will be clearer.

Also if the intent is to clear the cache between batches, how certain are we that the duplicate metrics exist inside the same batch, and not between batches?

gmertes (commented):
Looks good! I have not tested with a real training, but I did some synthetic tests of the FixedLengthSet and the performance impact is negligible. 

FixedLengthSet could maybe go into the mlflow.utils (the one here, not the one in anemoi utils), but doesn't have to.

PR #424, opened 2025-07-21
fix: mlflow hp params limit

anaprietonem:
## Description
When logging to many hp_params into the mlflow server, the runs and the experiment takes long to load. We set a configurable limit to better control this, so if a run has many hp_params the code will crash and avoid that run being logged. This limit can be configured from the config, but if not limit is passed then the default value of 2000 will be used.

Additionally this PR modifies the W&B schema so the identity when W&B is not enabled, is update to None so it doesn't crash or complaint about the '?'

##  Additional notes ##


training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -247,15 +251,30 @@ class WandbSchema(BaseModel):
     entity: Union[str, None] = None
     "Username or team name where to send runs. This entity must exist before you can send runs there."
 
+    @root_validator(pre=True)
+    def clean_entity(cls: type[WandbSchema], values: dict[str, Any]) -> dict[str, Any]:  # noqa: N805
+        if values["enabled"] is False:
+            values["entity"] = None
+        return values
```
gmertes, line 257: Really nice! Should we do the same for the mlflow entry (the tracking_uri) ?

training/src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -644,9 +653,9 @@ def log_hyperparams_in_mlflow(
                 expanded_params = clean_config_params(expanded_params)
 
             LOGGER.info("Logging %s parameters", len(expanded_params))
-
-            if len(expanded_params) > MAX_PARAMS_LENGTH:
-                LOGGER.warning("Logging a large number of parameters to %s", len(expanded_params))
+            if len(expanded_params) > max_params_length:
+                error_msg = f"Too many params: {len(expanded_params)} > {max_params_length}"
+                raise ValueError(error_msg)
```
gmertes: What should the user do if this exception is raised? Can we add some tips to the error message on how to fix it?

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -277,6 +296,22 @@ class MlflowSchema(BaseModel):
     http_max_retries: PositiveInt = Field(example=35)
     "Specifies the maximum number of retries for MLflow HTTP requests, default 35."
 
+    @model_validator(mode="after")
+    def check_valid_extras(self) -> Any:
+        # Check for valid extra fields related to MultiHeadSelfAttention and MultiHeadCrossAttention
```
gmertes: Copy-pasted comment? 😀

anaprietonem: Indeed, any case I reviewed the logic and tested the suggestion and it was okey to simplify the logic

training/src/anemoi/training/diagnostics/mlflow/logger.py
anaprietonem: Good point, I have added a comment so hopefully this is now a more meaningful error

training/src/anemoi/training/schemas/diagnostics.py
anaprietonem: Added that !

anaprietonem:
Note - while testing this PR we noticed that logging_models into mlflow is currently not working. This is unrelated to the changes of this PR as mentioned here https://github.com/mlflow/mlflow/issues/15111

PR #208, opened 2025-03-21
Feature/variable filtering

OpheliaMiralles:
## Description

Provides ability to select variables on which to compute a specific loss. Related to issue #63.

## Type of Change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Issue Number

solves #63 

## Code Compatibility

-   [x] I have performed a self-review of my code

### Code Performance and Testing

-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I ran the [complete Pytest test](https://anemoi.readthedocs.io/projects/training/en/latest/dev/testing.html) suite locally, and they pass
-   [ ] I have tested the changes on a single GPU
-   [ ] I have tested the changes on multiple GPUs / multi-node setups
-   [ ] I have run the Benchmark Profiler against the old version of the code
-   [x] If the new feature introduces modifications at the config level, I have made sure to update Pydantic Schemas and default configs accordingly

### Dependencies

-   [ ] I have ensured that the code is still pip-installable after the changes and runs
-   [ ] I have tested that new dependencies themselves are pip-installable.
-   [x] I have not introduced new dependencies in the inference portion of the pipeline

### Documentation

-   [x] My code follows the style guidelines of this project
-   [x] I have updated the documentation and docstrings to reflect the changes
-   [x] I have added comments to my code, particularly in hard-to-understand areas

PR #426, opened 2025-07-21
feat: target indices

OpheliaMiralles:
## Description
This PR introduces support for a new category of input features called targets, which are used to compute the loss but are neither predicted by the model nor part of the diagnostic or prognostic features. It refines how input and output features are handled during training, especially in the context of loss computation and data indexing.

Currently, the feature classification in the Anemoi training pipeline assumes:

- Prognostic = features predicted by the model (i.e., diagnostic - forcing)
- Forcing = input-only features
- Diagnostic = output-only features 

However, we now have use cases where certain features:
- Are used as input
- Are required to compute the loss
- Are not predicted by the model

To handle this, we introduce a new category called targets, defined as features used to compute the loss but not part of the forecast state.

This update allows:
- Better separation of model inputs and loss-relevant features.
- More flexible handling of loss computation by explicitly reindexing predicted and target variables per loss.
- Improved interpretability of feature roles in the pipeline.

N/A (internal feature enhancement)

## Additional notes
Prognostic features are now defined as: prognostic = diagnostic - forcing - targets

A feature can simultaneously be a target and a forcing. Loss functions can be updated to explicitly reindex predictions and targets based on their respective configurations.
This improves modularity of the loss pipeline and prepares for future loss extensions.

----
📚 Documentation preview 📚: https://anemoi-training--426.org.readthedocs.build/en/426/

----
📚 Documentation preview 📚: https://anemoi-graphs--426.org.readthedocs.build/en/426/

----
📚 Documentation preview 📚: https://anemoi-models--426.org.readthedocs.build/en/426/


mchantry:
Hi Ophelia,
Could you please provide a concrete example of your use case, I am struggling to imagine one, but I am sure that one exists.

OpheliaMiralles:
> Hi Ophelia, Could you please provide a concrete example of your use case, I am struggling to imagine one, but I am sure that one exists.

see https://github.com/ecmwf/anemoi-core/pull/208#issuecomment-3056894767

OpheliaMiralles:
Set this as a draft, because as long as we restrict y to match index of the model output in the forecaster/interpolator etc, this will have no impact at all. It would work only if every loss is a filtering loss wrapper with the default targets being output indices...
`y = batch[:, self.imap[interp_step], :, :, self.data_indices.data.output.full]`
`y = batch[:, self.multi_step + rollout_step, ..., self.data_indices.data.output.full]`

OpheliaMiralles:
> I was wondering about targets -> in ML we usually refer to targets and predictions, so it could be a bit confusing to name something targets that is different from that (if I understood correctly). Maybe a more descriptive name?
> 
> I would also like to understand better how they impact the loss computation and how they influence the weight updates.

Actually target is used here with the ML reference, e.g. the data you compare the predicted values to in the loss computation.

models/src/anemoi/models/layers/bounding.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         self.name_to_index_stats = name_to_index_stats
 
     def _create_index(self, variables: list[str]) -> InputTensorIndex:
-        return InputTensorIndex(includes=variables, excludes=[], name_to_index=self.name_to_index)._only
+        return torch.Tensor(sorted(i for name, i in self.name_to_index.items() if name in variables)).to(torch.int)
```
sahahner: why is this sorting necessary?

OpheliaMiralles: I think I might have sorted because it was sorted in the original indices but I can remove

OpheliaMiralles:
> Thank you for this work. I am trying to understand the implications on the model output shape that goes into the loss function. Does this mean, that, for your usecase, the shape of the predicted tensor that goes into the loss function is different to that target tensor that goes into the loss function?

The target variables can be any subset of the total set of variables, so in my usecase there a more target variables than predicted variables. But I always compare tensors of the same shape when the loss computation is done, and it practice it means some predicted variables are compared to several target variables with different losses.

models/src/anemoi/models/layers/bounding.py
sahahner: Thank you, it would be good if the indices were of type InputTensorIndex.

OpheliaMiralles: The issue is that I don't have the information about forcing/prognostic/target/diagnostic needed to build the index as it does not only depend on includes and excludes anymore...

OpheliaMiralles: So should I leave it as it is? For this specific example I don't think it makes a difference to implement it without InputTensorIndex, and if there is a test about the boundings and it passes then it is probably the case. 
```suggestion
        return torch.Tensor([i for name, i in self.name_to_index.items() if name in variables]).to(torch.int)
```

models/src/anemoi/models/data_indices/index.py
```diff
@@ -46,49 +46,72 @@ def representer(dumper, data):
...
+            target=target,
+            diagnostic=diagnostic,
+            prognostic=self._prognostic,
             name_to_index=name_to_index_model_output,
         )
 
     def __repr__(self) -> str:
         return (
-            f"{self.__class__.__name__}(diagnostic={self._input}, forcing={self._output}, "
+            f"{self.__class__.__name__}(diagnostic={self.input}, forcing={self.output}, "
```
anaprietonem, line 114: was this even working before? 😧

models/src/anemoi/models/data_indices/collection.py
```diff
@@ -34,20 +34,94 @@ def __init__(self, config, name_to_index) -> None:
...
+            name_to_index=name_to_index_internal_data_input,
+        )  # internal after the remapping applied to data (training)
+        self.model = ModelIndex(
+            diagnostic=self.diagnostic,
+            forcing=self.forcing,
+            target=self.target,
+            name_to_index_model_input=name_to_index_model_input,
+            name_to_index_model_output=name_to_index_model_output,
+        )
+        self.internal_model = ModelIndex(
```
anaprietonem: where are the internal model and internal data indices used?

models/src/anemoi/models/data_indices/collection.py
```diff
@@ -34,20 +34,94 @@ def __init__(self, config, name_to_index) -> None:
         self.diagnostic = (
             [] if config.data.diagnostic is None else OmegaConf.to_container(config.data.diagnostic, resolve=True)
         )
+        self.target = (
+            [] if config.data.get("target", None) is None else OmegaConf.to_container(config.data.target, resolve=True)
+        )
+        defined_variables = set.union(set(self.forcing), set(self.diagnostic), set(self.target))
+        self.prognostic = [v for v in self.name_to_index.keys() if v not in defined_variables]
+        # config.data.remapped is an optional dictionary with every remapper as one entry
+        self.remapped = (
```
anaprietonem: Why do we need to account for the remapper in this case? I struggle to understand how's this related to the target functionality

models/src/anemoi/models/data_indices/tensor.py
```diff
@@ -14,102 +14,136 @@
...
-        self._removed = self._build_idx_from_includes(self.excludes)
-        self.prognostic = self._build_idx_prognostic()
-        self.diagnostic = NotImplementedError
-        self.forcing = NotImplementedError
+        ), f"Data indexing has invalid entries {[var for var in includes if var not in self.name_to_index]}, not in dataset."
+        self.prognostic = self._build_idx_from_list(prognostic)
+        self.diagnostic = self._build_idx_from_list(diagnostic)
+        self.forcing = self._build_idx_from_list(forcing)
+        self.target = self._build_idx_from_list(target)
+        self.full = self._build_idx_from_list(includes)
```
anaprietonem, line 53: shouldn't self.full refer to the excludes?

models/src/anemoi/models/data_indices/tensor.py
```diff
@@ -14,102 +14,136 @@
...
-        self.full = self._build_idx_from_excludes()
-        self._only = self._build_idx_from_includes()
-        self._removed = self._build_idx_from_includes(self.excludes)
-        self.prognostic = self._build_idx_prognostic()
-        self.diagnostic = NotImplementedError
-        self.forcing = NotImplementedError
+        ), f"Data indexing has invalid entries {[var for var in includes if var not in self.name_to_index]}, not in dataset."
+        self.prognostic = self._build_idx_from_list(prognostic)
+        self.diagnostic = self._build_idx_from_list(diagnostic)
+        self.forcing = self._build_idx_from_list(forcing)
```
anaprietonem, line 51: what was the motivation to pass prognostic and target here?  I don't see otherwise how self.target is being used in the tensors, so was wondering if we could we keep the include for input forcing + prognostic as  and for output prognostic + diagnostic?

models/src/anemoi/models/data_indices/collection.py
OpheliaMiralles: I think internal model / data indices are used in the remapper, probably in the loss computation to select variables and in anemoi-inference but @sahahner would know more

models/src/anemoi/models/data_indices/collection.py
sahahner: Hi Ophelia, 
Would you mind bringing your branch up to date with main, please, as there should be no more mentions of remapped in the data indices, as we removed this functionality.

models/src/anemoi/models/data_indices/index.py
OpheliaMiralles: I don't know but it was probably not tested (I never saw _input and _output anywhere)

models/src/anemoi/models/data_indices/collection.py
sahahner: see here: https://github.com/ecmwf/anemoi-core/pull/426#discussion_r2332322271

the internal_model and internal_data indices do not exist anymore on main, as they have been removed in this PR: https://github.com/ecmwf/anemoi-core/pull/268

models/src/anemoi/models/data_indices/collection.py
OpheliaMiralles: OK yes I hadn't noticed

models/src/anemoi/models/data_indices/tensor.py
OpheliaMiralles: Not sure I understood

models/src/anemoi/models/layers/bounding.py
sahahner: Before we resolve this, I would like to test a setup with a more complex collection of bounding and pre/prostprocessors to see if the changes in the indices impact this. At the moment this is not yes included in the tests on GPU.

models/src/anemoi/models/data_indices/tensor.py
anaprietonem: So I am wondering why we have to change the logic on the output and input tensor? from the changes you're proposing I can see we now add the target also to these tensors and get rid of the self._only and self._removed. Could you have kept the previous interface and then do include= forcing + prognostic for  inputs, and for output=prognostic+diagnostic, and exclude being then name_to_index - include?

(and mostly asking to better understand the changes - not to say we don't have to change things)

models/src/anemoi/models/data_indices/index.py
```diff
@@ -16,8 +16,8 @@ class BaseIndex:
     """Base class for data and model indices."""
 
     def __init__(self) -> None:
-        self.input = NotImplementedError
-        self.output = NotImplementedError
+        self.input = None
+        self.output = None
```
VeraChristina: I see that self.input and self.output were changed from NotImplementedError to None. Was this intentional?

Using NotImplementedError makes it clear that these attributes must be set by subclasses, whereas None suggests they’re optional or have a default. I think we might want to keep the NotImplementedError (or define them as abstract properties using abc), but maybe I am missing something?

models/src/anemoi/models/layers/bounding.py
sahahner: I checked and it runs fine: https://mlflow.ecmwf.int/#/experiments/5/runs/c431710656c0453a80c9652e3b6b75dc

Thanks for removing the sorted. The indices should be in the same order as name_to_index.

models/src/anemoi/models/data_indices/tensor.py
OpheliaMiralles: With this version excludes is never used, we only include variables we want and no variable is included by default

models/src/anemoi/models/layers/bounding.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         self.name_to_index_stats = name_to_index_stats
 
     def _create_index(self, variables: list[str]) -> InputTensorIndex:
-        return InputTensorIndex(includes=variables, excludes=[], name_to_index=self.name_to_index)._only
+        return torch.Tensor([i for name, i in self.name_to_index.items() if name in variables]).to(torch.int)
```
ssmmnn11: you can create with a dtype directly like this:

torch.tensor(
    [i for name, i in self.name_to_index.items() if name in variables],
    dtype=torch.int
)

models/tests/data_indices/test_collection.py
```diff
@@ -29,15 +29,32 @@ def data_indices():
...
 
+@pytest.fixture()
+def data_indices_with_target():
+    # mimicks example where we try to predict precipitation (tp) by comparing predictions with point measurements (tp_point) and radar data (tp_radar) and using topography (dem) as forcing
+    config = DictConfig(
+        {
+            "data": {
+                "forcing": ["tp_point", "tp_radar", "dem"],
+                "diagnostic": [],
+                "target": ["tp_point", "tp_radar"],
```
anaprietonem, line 40: Could we add an update to the docs to mention about targets? Also the pydantic schema for data https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/schemas/data.py should have now an optional entry for target

models/src/anemoi/models/data_indices/tensor.py
anaprietonem: Yes I guess this is related to my question below - why do we get rid of excludes? Was that because of the concept of targets or because you thought the original design was too rigid?

models/src/anemoi/models/data_indices/tensor.py
OpheliaMiralles: Ah, so maybe this one was a choice from me because I thought it was very convoluted to include/exclude some indices in the call to an Index to then redefine prognostic, diagnostic, etc from the include and exclude instead of passing directly the interesting sets of variables prognostic, diagnostic, forcing (that we want in the Index anyway) and the ones we include for Input or Output

models/src/anemoi/models/layers/bounding.py
OpheliaMiralles: Great OK thanks a lot!

models/src/anemoi/models/data_indices/tensor.py
OpheliaMiralles: Indeed it was mostly related to the lack of transparency in the design (i.e. prognostic variables were implicitely included but needed then to be excluded to compute forcings) which I found made things maybe overly complicated if we just want to include some variables in the inputs and some in the outputs

models/tests/data_indices/test_collection.py
OpheliaMiralles: How should I set up the schema? target indices are really optional, they don't need to be written in the config at all, but integration tests started failing when I added the schema, because the pydantic validation did not find target.

anaprietonem: Could you try `target: list[str] | None = Field(default_factory=[])` or `target: list[str] | None = Field(default=None)`

models/src/anemoi/models/data_indices/tensor.py
```diff
@@ -14,102 +14,136 @@
...
+        self.target = self._build_idx_from_list(target)
+        self.full = self._build_idx_from_list(includes)
+        self.excludes = list(set(self.name_to_index.keys()) - set(self.includes))
```
VeraChristina: Related to the discussion above, I think I can see why you prefer the more explicit set up instead of includes/excludes logic. Why do you set the excludes here? Does some of the downstream logic still depend on it?

OpheliaMiralles: Backward compatibility, because most tests still rely on "excludes"

anaprietonem: Could it then be worth keeping the checks in the equality tests?
````
            and self.includes == other.includes
            and self.excludes == other.excludes
 ````

anaprietonem (approved):
Thanks @OpheliaMiralles for addressing the comments and the nice work, LGTM!

PR #431, opened 2025-07-22
fix: introducing SLURMID based cuda device selection to fix cuda busy devices error

### LINKED ISSUE #430: CUDA-capable device(s) is/are busy or unavailable when loading or creating a graph during training
### What happened?

When running any training job with: ` num_gpus_per_node>1` (eg. num_gpus_per_node=4) the training fails throwing the following error:

```
...
File "/users/apennino/anemoi-core/training/src/anemoi/training/train/train.py", line 155, in graph_data
...
File "/users/apennino/anaconda3/envs/anemoi-fix/lib/python3.10/site-packages/torch/cuda/__init__.py", line 444, in __enter__
    self.prev_idx = torch.cuda._exchange_device(self.idx)
RuntimeError: CUDA error: CUDA-capable device(s) is/are busy or unavailable
```

The error arises in multiple situations:

1.  **Loading an existing graph.**
     The issue comes from the following:
     > ➤ torch.load(..., map_location=None) defaults to device used during saving 
     > If the file at graph_filename was saved on GPU 0, then all ranks will try to load CUDA tensors onto GPU 0 by default — even ranks 1, 2, 3 — which causes a CUDA error because multiple processes are trying to use GPU 0 simultaneously without proper coordination.
     > This is common in multi-GPU training when using torch.load without map_location.

2. **Creating a new graph.**
    When overwrite=True, we are recomputing the graph, and inside that logic, in multiple places we are calling:
    `nodes = graph[nodes_name].to(self.device)`
    But self.device is not properly set — it's likely defaulting to "cuda" (which implies cuda:0) instead of the correct local GPU rank.

    So just like before, all workers try to use cuda:0, and only one of them succeeds.

The bug is caused by different cluster configuration and the way default cuda device is selected.

### Version

models v0.8.1

Linux balfrin-ln004 5.14.21-150400.24.81_12.0.87-cray_shasta_c #1 SMP Sun Dec 17 12:59:08 UTC 2023 (e30c7c1) x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

MeteoSwiss

icedoom888:
## Description
For each process the local device is now determined from the SLURM environment variable if available during graph building/loading.

Issue described in #430. 

**Why self.device = "cuda" sometimes causes this CUDA busy/unavailable error**

> self.device = "cuda" alone means cuda:0 by default — the first GPU device.
> On single-GPU setups, or when your process is the only one running, this works perfectly fine.
> But on multi-GPU distributed environments (e.g., SLURM jobs with multiple GPUs per node), each process must select its own GPU device (e.g., cuda:0, cuda:1, cuda:2, etc.).
> If multiple processes blindly pick "cuda" (which equals "cuda:0"), they all try to use GPU 0, causing device contention and the "device busy or unavailable" error.

_Why does it work for some people?_

- Some people might be running single GPU jobs, so "cuda" works just fine.
- Others might have their environment set up differently — e.g., environment variables like CUDA_VISIBLE_DEVICES limiting visible GPUs per process, which implicitly isolates devices.
- Some setups explicitly set device IDs per process (like you should do with SLURM’s SLURM_LOCALID or LOCAL_RANK env vars).
- If people run on older PyTorch versions or different cluster configs, their distributed launches or device assignment could be handled automatically.

_Is this a PyTorch version issue?_

- No, this is not specific to PyTorch versions.
- The semantics of "cuda" meaning "cuda:0" haven’t changed.
- The difference is in how you launch your distributed jobs and assign devices.
- PyTorch's device management requires you to tell each process which GPU to use explicitly in multi-GPU/multi-node scenarios.

This PR fixes #430 

### DIFF
--- a/.github/workflows/integration-tests-hpc.yml
+++ b/.github/workflows/integration-tests-hpc.yml
@@ -41,7 +41,7 @@ jobs:
             source .venv/bin/activate
             pip install --upgrade pip
             pip install -e ./training[all,tests] -e ./models[all,tests] -e ./graphs[all,tests]
-            python3 -m pytest -v training/tests/integration --longtests
+            python3 -m pytest -v training/tests/integration --slow
             deactivate
             rm -rf $REPO_NAME
           sbatch_options: |
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -15,7 +15,7 @@ jobs:
           python-version: '3.10'
           install-package: "-e ./training[all,tests] -e ./graphs[all,tests] -e ./models[all,tests]"
           test-directory: "training/tests/integration"
-          pytest-options: "--longtests"
+          pytest-options: "--slow"
           manual: ${{github.event_name == 'workflow_dispatch' && 'true' || 'false'}}
 
   notify-team:
--- a/training/docs/contributing.rst
+++ b/training/docs/contributing.rst
@@ -244,7 +244,7 @@ available, then from the top-level directory of anemoi-core run:
 
 .. code:: bash
 
-   pytest training/tests/integration --longtests
+   pytest training/tests/integration --slow
 
 *********************************************
  Configuration handling in integration tests
--- a/training/pyproject.toml
+++ b/training/pyproject.toml
@@ -86,7 +86,7 @@ optional-dependencies.profile = [
   "rich>=13.6",
   "tabulate>=0.9",
 ]
-optional-dependencies.tests = [ "hypothesis>=6.11", "pytest>=8", "pytest-mock>=3" ]
+optional-dependencies.tests = [ "hypothesis>=6.11", "pytest>=8", "pytest-mock>=3", "pytest-skip-slow>=0.0.5" ]
 urls.Changelog = "https://github.com/ecmwf/anemoi-training/CHANGELOG.md"
 urls.Documentation = "https://anemoi-training.readthedocs.io/"
 urls.Homepage = "https://github.com/ecmwf/anemoi-training/"
--- a/training/pytest.ini
+++ b/training/pytest.ini
@@ -3,7 +3,7 @@ markers =
     data_dependent: marks tests depending on data (deselect with '-m "not data_dependent"')
     auth: marks tests that require authentication (deselect with '-m "not auth"')
     gpu: marks tests that require a GPU (deselect with '-m "not gpu"')
-    longtests: mark test as long-running (skipped unless --longtests is used)
+    slow: mark test as slow (skipped unless --slow is used)
 
 
 tmp_path_retention_policy = none
--- a/training/tests/conftest.py
+++ b/training/tests/conftest.py
@@ -22,30 +22,6 @@ from anemoi.training.data.datamodule import AnemoiDatasetsDataModule
 pytest_plugins = "anemoi.utils.testing"
 
 
-def pytest_addoption(parser: pytest.Parser) -> None:
-    parser.addoption(
-        "--longtests",
-        action="store_true",
-        dest="longtests",
-        default=False,
-        help="enable tests marked as longtests",
-    )
-
-
-def pytest_configure(config: pytest.Config) -> None:
-    """Register the 'longtests' marker to avoid warnings."""
-    config.addinivalue_line("markers", "longtests: mark tests as long-running")
-
-
-def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item]) -> None:
-    """Automatically skip @pytest.mark.longtests tests unless --longtests is used."""
-    if not config.getoption("--longtests"):
-        skip_marker = pytest.mark.skip(reason="Skipping long test, use --longtests to enable")
-        for item in items:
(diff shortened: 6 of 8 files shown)


MeraX:
> The semantics of "cuda" meaning "cuda:0" haven’t changed.
I would tend to disagree. `cuda` without a number refers to some default CUDA device.

If I'm running a job with `torchrun` using `to("cuda")`, the tensors are allocated on the respective GPU.
```
# test.py
x=torch.empty(1).to(f"cuda:{rank}")
print(f"{x.device=}")
```

`torchrun   --nproc_per_node=4  test.py` output:
```
x.device=device(type='cuda', index=1)
x.device=device(type='cuda', index=0)
x.device=device(type='cuda', index=3)
x.device=device(type='cuda', index=2)
```

icedoom888:
> There's a significant amount of duplicate code here, would it be better to import a function to do this? I think this PR would need testing across many infrastructures to ensure by fixing for one it doesn't break another.

@mchantry agreed, any suggestions on where the get_device function should be?

icedoom888:
It's quite pressing for us to get this merged asap as it's a crucial fix for our set up. 
I have to manually include and remove this in all my PR to test anything 🗡️

MeraX (dismissed):
> 
> Hey @MeraX , shouldn't this mean that cuda visible devices is 1 and cuda:0 still applies where 0 is just the index of the visible devices?
> 
> So for rank 3 for example: cuda_visible_devices = [3] -> cuda:0 means cuda_visible_devices[0] which still gets gpu 3

I'm sorry, yes. you are right. I've been trying the wrong thing. If my `torch_run` `LOCAL_RANK` is 3 and I'm allocating with `torch.empty(1).to(f'cuda')` the device is, as you initially suggested, `device(type='cuda', index=0)`.

Nonetheless, I can confirm, that the proposed changes still works on our system. (`num_gpus_per_node=2; num_gpus_per_model=2` still allocates memory on two different devices.)  I would propose testing this PR on other systems and merging it asap, if it is a pain point for MeteoSwiss.


But still I'm wondering: How are the tensors distributed to different GPUs on other multi-GPU systems with `cuda:0`? Are all other tensor devices derived from the four definitions you are touching with this PR?

mchantry:
> > There's a significant amount of duplicate code here, would it be better to import a function to do this? I think this PR would need testing across many infrastructures to ensure by fixing for one it doesn't break another.
> 
> @mchantry agreed, any suggestions on where the get_device function should be?

Maybe a new utils in distributed, or in the utils folder?

mchantry:
> > Hey @MeraX , shouldn't this mean that cuda visible devices is 1 and cuda:0 still applies where 0 is just the index of the visible devices?
> > So for rank 3 for example: cuda_visible_devices = [3] -> cuda:0 means cuda_visible_devices[0] which still gets gpu 3
> 
> I'm sorry, yes. you are right. I've been trying the wrong thing. If my `torch_run` `LOCAL_RANK` is 3 and I'm allocating with `torch.empty(1).to(f'cuda')` the device is, as you initially suggested, `device(type='cuda', index=0)`.
> 
> Nonetheless, I can confirm, that the proposed changes still works on our system. (`num_gpus_per_node=2; num_gpus_per_model=2` still allocates memory on two different devices.) I would propose testing this PR on other systems and merging it asap, if it is a pain point for MeteoSwiss.
> 
> But still I'm wondering: How are the tensors distributed to different GPUs on other multi-GPU systems with `cuda:0`? Are all other tensor devices derived from the four definitions you are touching with this PR?

On our system I believe each task is only given visibility to one GPU.

icedoom888:
@mchantry introduced the function as discussed. I opted for anemoi.graphs.utils given that it's used withing graphs only basically.

training/src/anemoi/training/train/train.py
```diff
@@ -148,8 +148,10 @@ def graph_data(self) -> HeteroData:
             )
 
             if graph_filename.exists() and not self.config.graph.overwrite:
+                from anemoi.graphs.utils import get_distributed_device
+
                 LOGGER.info("Loading graph data from %s", graph_filename)
-                return torch.load(graph_filename, weights_only=False)
+                return torch.load(graph_filename, map_location=get_distributed_device(), weights_only=False)
```
mchantry, line 155: @icedoom888 why has `map_location` been introduced here?

icedoom888: @mchantry This is what is causing issues when graph is not created but loaded, similar problem: the graph is saved with device specified as "cuda" and without this the model does not know where to load it

HCookie (approved):

icedoom888:
> @icedoom888 I've requested a review from Michael, who will test this on LUMI (AMD GPUs) to ensure this works on alternate hardware. After that, this can be merged.

Awesome thanks!

icedoom888:
@mpvginde @icedoom888 any updates on this?

mpvginde (approved):
Successfully tested on LUMI on using 8 GPUs. All GPUs build the graph.

Issue #421, opened 2025-07-15
Reference to anemoi.training.schemas in inference checkpoint pickle


gmertes:
### What happened?

Training a model with the debug config produces an inference checkpoint that still contains a target reference to `anemoi.training.schemas.data.PreprocessorTarget`. So when loading the checkpoint in a fresh inference environment, it crashes due to a missing anemoi-training. 

The reference is under the normalizer data_indices in `anemoi.models.data_indices.collection.IndexCollection.config.data.processors.normalizer._target_`

See the printout of the normalizer attributes below.

### What are the steps to reproduce the bug?

`anemoi-training train --config-name debug`

### Version

latest

### Platform (OS and architecture)

atos

### Relevant log output

```shell
"normalizer": {
  "anemoi.models.preprocessing.normalizer.InputNormalizer": {
    "data_indices": {
      "anemoi.models.data_indices.collection.IndexCollection": {
        "config": {
          "data": {
            "processors": {
              "normalizer": {
                "_target_": "anemoi.training.schemas.data.PreprocessorTarget(...)", <=======
                "config": {
                  "default": "mean-std",
                  "remap": {},
...
```

### Organisation

ECMWF

theissenhelen:
Yes, you are correct. It should be importing the schema from utils.

PR #433, opened 2025-07-23
fix!: for schemas of data processors

anaprietonem:
## Description
Pydantic schemas not resolving correcting and getting into the model's metadata and internal class names. Resolve this by:
- Moving data-processors schemas to models. Initially left in training since those are defined in the 'data' config. However those are ultimately model layers - note this change is a **breaking change**
- Test at the beginning of the training that there is no pydantic schemas in the model, otherwise crash
- Add test in the data processors to ensure all schemas work ( This is to mitigate the fact that default configs just include test for normalizers, so we do not test others like imputers or post-processors commonly used for Earth System Models)

Issue - https://github.com/ecmwf/anemoi-core/issues/421

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/utils/checkpoint.py
+++ b/training/src/anemoi/training/utils/checkpoint.py
@@ -119,7 +119,7 @@ def freeze_submodule_by_name(module: nn.Module, target_name: str) -> None:
 
 class LoggingUnpickler(pickle.Unpickler):
     def find_class(self, module: str, name: str) -> str:
-        if "anemoi.training.schemas" in module:
+        if "anemoi.training" in module:
             msg = (
                 f"anemoi-training Pydantic schemas found in model's metadata: "
                 f"({module}, {name}) Please review Pydantic schemas to avoid this."


anaprietonem:
@sahahner I have extended the tests, could you check these works fine and have a look?
@gmertes  I have now added a check in our ModelCheckpoint callback to catch this problem. Tested it in main and caught the problem fine. Should just be done at the beginning of the training.

sahahner (approved):
Thank you for correcting the schemas for the pre/postprocessors. They work fine for my config files, and the tests for the schemas look good to me.

training/src/anemoi/training/utils/checkpoint.py
```diff
@@ -113,3 +115,21 @@ def freeze_submodule_by_name(module: nn.Module, target_name: str) -> None:
         else:
             # Recursively search within children
             freeze_submodule_by_name(child, target_name)
+
+
+class LoggingUnpickler(pickle.Unpickler):
+    def find_class(self, module: str, name: str) -> str:
+        if "anemoi.training.schemas" in module:
```
gmertes: Should we also check for any `anemoi.training` reference just to be sure?

anaprietonem: Yes I wondered, I guess it should be fair to enforce that since that's checking the 'plain torch' model. Will update it

anaprietonem:
@gmertes I have update the check to do just 'anemoi.training' rather the schemas specifically. Updated the branch and tested it and seems fine (ie if I test this with the current main it crashes as expected).

gmertes (approved):
Very nice, thanks for fixing this bug!

PR #434, opened 2025-07-23
fix: Backward compatibility for inference

OpheliaMiralles:
## Description
Models from new anemoi versions have new attributes which are not expected for inference for older models (the forward call fails).

We can run inference with models where no A_down/qk_norm/layer_norm_mlp_dst were defined.

### DIFF
--- a/models/src/anemoi/models/layers/block.py
+++ b/models/src/anemoi/models/layers/block.py
@@ -702,8 +702,8 @@ class GraphTransformerMapperBlock(GraphTransformerBaseBlock):
         query, key, value, edges = self.get_qkve(x, edge_attr)
         if not hasattr(self, "shard_strategy"):
             query, key, value, edges = self.shard_qkve_heads(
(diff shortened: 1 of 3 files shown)


anaprietonem:
@OpheliaMiralles are you using inference with a different version of anemoi-model to the one used to train your model? Just trying to understand if this problem could be solved by making sure the right version of model is installed or you have use case where you need those versions to be different?

OpheliaMiralles:
> @OpheliaMiralles are you using inference with a different version of anemoi-model to the one used to train your model? Just trying to understand if this problem could be solved by making sure the right version of model is installed or you have use case where you need those versions to be different?
> 
> ( I click the closed button by accident - ignore please - it's reopened now)

I actually wanted to close it because indeed, we can solve it by pinning the anemoi version used for training. But I will just keep the branch for exploration purposes if this is OK, I'll delete it later.

PR #438, opened 2025-07-28
fix: Remove default pressure level variable_groups from training/*.yaml 

### LINKED ISSUE #435: Default Config Has Inconsistent Variable Group Definitions
### What happened?

As a minimal example, I'm trying to modify the default [stretched.yaml](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/config/stretched.yaml) configuration file to have the following training variable groups

```yaml
training:
  variable_groups:
    default: sfc
    pl:
      param: [u, v, w]
```

This gives me the error:

```
hydra.errors.ConfigCompositionException: In 'stretched': ConfigTypeError raised while composing config:
Cannot merge DictConfig with ListConfig
    full_key:
    object_type=dict
```

Similarly, if I modify the default [config.yaml](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/config/config.yaml) as follows:

```yaml
training:
  variable_groups:
    default: sfc
    pl: [u, v, w]
```

Then I get the same error. This appears to be caused by the fact that the defaults inside of [training/default.yaml](https://github.com/ecmwf/anemoi-core/blob/70ecdd9e46f5c9f46dc644cbb8d179392eab6910/training/src/anemoi/training/config/training/default.yaml#L111-L115) and [training/stretched.yaml](https://github.com/ecmwf/anemoi-core/blob/70ecdd9e46f5c9f46dc644cbb8d179392eab6910/training/src/anemoi/training/config/training/stretched.yaml#L103-L105) are different. This is not really a big issue, but I think it would nice to have some consistency.

I know this issue form does not ask for proposed solutions, but I'll just give my two cents that I think one possible modification could be to make all of the training yamls (training/default.yaml, training/stretched.yaml, etc) have the following default for variable_groups:

```yaml
# inside e.g. training/default.yaml
variable_groups:
  default: sfc
```

Then it's easy for users to modify it however they like, since there are plenty of good examples in the comments. This would make a breaking change for the default configuration though, and I'm not sure how big of an issue that is. I think at the very least it would be nice for all of the different configs to be the same (either `pl: {param: []}` or `pl: []`).

If you're open to it, I'd be happy to contribute a PR with a solution that's agreed upon, but no pressure. I just want to offer some help as a very happy and thankful user of Anemoi!

```
anemoi-training config generate
echo -e "training:\n  variable_groups:\n    default: sfc\n    pl: [u,v,w]" >> config.yaml
anemoi-training config dump --config-path=. --config-name=config --output=combined.yaml
```

### Version

70ecdd9e (current main branch)

GNU Linux x86_64

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

NOAA Physical Sciences Laboratory

timothyas:
## Description
Closes #435 

By removing any default configuration format for variable_groups within the training yamls.

Very minor bugfix, in that it allows the user to easily define variable groups using whatever notation they like, e.g. they can do

```yaml
training:
  variable_groups:
    pl: [u, v, w]
```

or

```yaml
training:
  variable_groups:
    pl:
      param: [u, v, w]
```

without any OmegaConf errors.

#435 

##  Additional notes ##

Users relying on the defaults will have to update their configurations accordingly.

----
📚 Documentation preview 📚: https://anemoi-training--438.org.readthedocs.build/en/438/

----
📚 Documentation preview 📚: https://anemoi-graphs--438.org.readthedocs.build/en/438/

----
📚 Documentation preview 📚: https://anemoi-models--438.org.readthedocs.build/en/438/

### DIFF
--- a/training/docs/user-guide/training.rst
+++ b/training/docs/user-guide/training.rst
@@ -384,6 +384,23 @@ level has a weighting less than 0.2), defined in class
       y_intercept: 0.2
       slope: 0.001
 
+.. note::
+
+   Note that by default, there is no pressure level scaling, because
+   there is no default variable grouping. In order to apply the pressure
+   level scaling, the ``pressure_level`` code block above needs to be
+   added to the ``training.scalers`` section of the configuration.
+   Additionally, ``variable_groups`` need to be defined that group all
+   pressure level variables together. For example by defining the
+   following in the training section of the config
+
+   .. code:: yaml
+
+      variable_groups:
+        default: sfc
+        pl:
+          is_pressure_level: True
+
 The loss is also scaled by assigning a weight to each node on the output
 grid. These weights are calculated during graph-creation and stored as
 an attribute in the graph object. The configuration option


mc4117:
Good idea! I agree this would make the definition of variable groups much simpler. However, I was just wondering, if you remove default groups does the code still run or do you get an error because the default is undefined?

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -111,8 +111,6 @@ validation_metrics:
 
 variable_groups:
   default: sfc
-  pl:
```
mc4117, line 114: Instead of removing these lines entirely, I would maybe keep them as a comment with some text, just so that there is an example of what the user can run

timothyas: Great suggestion! I brought them back as comments in 8a21f74

HCookie:
@timothyas Such is the joys of Omegaconf.
Commenting out the pressure level scalar seems fine, as I do agree that no default is better than a bad default. Feel free to make that modifcation to this PR. I would also suggest you add some docs / warnings to make it clear what is happening. This will also mean this PR becomes a breaking change.

timothyas:
That makes sense, thanks for the note @HCookie. I went ahead and took this path to what we discussed earlier: I removed the pressure scaler by default and added documentation to the scalers about what would need to be added to incorporate pressure scaling. Hopefully this helps evaluate this option, and I totally understand if you want to go the other route and change the format rather than break defaults.

PR #439, opened 2025-07-29
feat(models): dynamic mappers

JPXKQX:
## Description

This PR includes some mappers that support dynamic graphs. The main difference is that now the `graph` is passed as an argument to the `forward()` method.

The current mappers set the graph during its creation.

##  Additional notes ##


anaprietonem:
> I wonder if we actually need an extra dynamic mapper class.

As in Simon you mean we could directly refactor the existing mappers so all are dynamic by default? That would significantly reduce the code complexity ... so I like the idea. But it would also mea breaking/not-being to reuse old checkpoints?

mchantry:
Refactor planned as described in #552

PR #440, opened 2025-07-29
feat: introducing AnemoiBaseModel

icedoom888:
[Code Quality and Testing](https://github.com/ecmwf/anemoi-core/actions/workflows/python-pull-request.yml)
## Description
Following discussion in #270 and PR #399 , introducing a new base class on anemoi models to make easier integration of new tasks and models.

Makes inheritance easier and clearer for existing and upcoming models.

Introduction of `_build_networks` internal function makes network elements more evident and clear.

#270 

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--440.org.readthedocs.build/en/440/

----
📚 Documentation preview 📚: https://anemoi-graphs--440.org.readthedocs.build/en/440/

----
📚 Documentation preview 📚: https://anemoi-models--440.org.readthedocs.build/en/440/


icedoom888:
Integration tests passing: https://github.com/ecmwf/anemoi-core/actions/runs/16619506944

icedoom888:
@HCookie Could you take a look at this and give some feedback?

icedoom888:
Hey @mchantry, this is effectively a refactor, any reason it needs ATS approval?

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+from anemoi.models.distributed.shapes import apply_shard_shapes
+from anemoi.models.distributed.shapes import get_shard_shapes
+from anemoi.models.layers.graph import NamedNodesAttributes
+from anemoi.models.layers.mapper import GraphTransformerBaseMapper
+from anemoi.utils.config import DotDict
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AnemoiModelBase(nn.Module):
```
matschreiner: Is this specifically for graph based models, since the graph is an input? 
Maybe the name is too general then?

icedoom888: @matschreiner at the amoment anemoi-models supports exclusively graph based models!
Removing the graph dependency would implicate big changes in anemoi-training.

We could rename this in: `AnemoiGraphModelBase`

mchantry:
HPC integration tests run successfully.

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+        if model_comm_group is not None:
+            assert (
+                model_comm_group.size() == 1 or batch_size == 1
+            ), "Only batch size of 1 is supported when model is sharded across GPUs"
+
+            assert (
+                model_comm_group.size() == 1 or ensemble_size == 1
+            ), "Ensemble size per device must be 1 when model is sharded across GPUs"
+
+    def _run_mapper(
```
matschreiner: Could this be a member of `anemoi.models.layers.mapper.BaseMapper`?

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+            self.A_up = self._make_truncation_matrix(truncation_data["up"])
+            LOGGER.info("Truncation: A_up %s", self.A_up.shape)
+
+    @abstractmethod
+    def _build_networks(self, model_config: DotDict) -> None:
+        """Builds the networks for the model."""
+        pass
+
+    @abstractmethod
+    def _assemble_input(self, x, batch_size, grid_shard_shapes=None, model_comm_group=None):
```
matschreiner, line 134: Wouldn't you say that assemble i/o is an implementation detail of the forward passes of the children - what is the reason for enforcing it on the baseclass?

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+        self.num_channels = model_config.model.num_channels
+
+        self.node_attributes = NamedNodesAttributes(model_config.model.trainable_parameters.hidden, self._graph_data)
+
+        self._calculate_shapes_and_indices(data_indices)
+        self._assert_matching_indices(data_indices)
+
+        # build networks
+        self._build_truncation(self._truncation_data)
+        self._build_networks(model_config)
```
matschreiner, line 73: I've been thinking to implement dependency injection into Anemoi models for some time. It is a very well posed  pattern for this type of system where we want to be able to compose our models of interchangable components like transformers/GNNs/etc.. It is also working quite well with hydra since you are allowed to instantiate directly from the config files. It is still not 100% clear to me exactly how to do it, but if you are interested we could have a chat about it some day?

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+                self.A_up = self.A_up.to(x.device)
+                x = self._truncate_fields(x, self.A_up)  # back to high resolution
+
+            if grid_shard_shapes is not None:
+                # back to grid-sharding as before
+                x = gather_channels(x, shard_shapes, model_comm_group)
+
+        return x
+
+    def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
```
matschreiner, line 81: Could this function return something like an "index_schema" as a dot_dict instead of assigning attributes on the base model? Then everything related to shapes and indices would be contained on a single object which can be referenced when needed for instantiation of objects etc.?

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+            "x_dst_is_sharded": x_dst_is_sharded,
+            "keep_x_dst_sharded": keep_x_dst_sharded,
+        }
+
+        if isinstance(mapper, GraphTransformerBaseMapper) and mapper.shard_strategy == "edges":
+            return mapper(data, **kwargs)
+
+        return checkpoint(mapper, data, **kwargs, use_reentrant=use_reentrant)
+
+    def _build_boundings(
```
matschreiner: Would it make sense to have this in a separate module?

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
+# (C) Copyright 2024 Anemoi contributors.
```
anaprietonem: minor - but new files should use 2025 header (so 2024 -> 2025)

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+from anemoi.models.distributed.shapes import apply_shard_shapes
+from anemoi.models.distributed.shapes import get_shard_shapes
+from anemoi.models.layers.graph import NamedNodesAttributes
+from anemoi.models.layers.mapper import GraphTransformerBaseMapper
+from anemoi.utils.config import DotDict
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AnemoiGraphModelBase(nn.Module):
```
anaprietonem: since in training we have the 'BaseGraphModule' should this be 'BaseGraphModel'? (naming it's always the hardest 😅)(and I think docstring need updating)

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+            Model communication group, by default None
+        grid_shard_shapes : list, optional
+            Shard shapes of the grid, by default None
+
+        Returns
+        -------
+        Tensor
+            Output of the model, with the same shape as the input (sharded if input is sharded)
+        """
+        pass
```
anaprietonem, line 166: Thoughts on having the predict step implemented as part of the basemodel to get rid off the Interface? Currently both classes seems to have some overlap

models/src/anemoi/models/models/base.py
anaprietonem: Are you thinking @matschreiner that this function could live for example in the `layers/bouding.py` and import it as factory directly? I think that would make it easier to test the boundings, so I like that suggestion and otherwise the alternative would be to at least make this as staticmethod since it's not really accessing any self atttributes

models/src/anemoi/models/models/base.py
```diff
@@ -0,0 +1,300 @@
...
+                    cfg,
+                    name_to_index=data_indices.model.output.name_to_index,
+                    statistics=statistics,
+                    name_to_index_stats=data_indices.data.input.name_to_index,
+                )
+                for cfg in getattr(model_config.model, "bounding", [])
+            ]
+        )
+
+    def _build_truncation(self, truncation_data: dict) -> None:
```
anaprietonem: I think a similarly argument could be made for this (But I'd guess this would be in any case handled with the PR @matschreiner for the residual connections?)

models/src/anemoi/models/models/base.py
anaprietonem: For me, by having it on the base class it can be clearer to developers that when creating new anemoi-models the I/O methods needs to be defined. So it works as a 'template'

models/src/anemoi/models/models/base.py
anaprietonem: The input dim is something that sometimes requires a completely new definition as for example in the interpolator, by making something like `self.input_dim = self._calculate_input_dim()` then in the interpolator we can keep all the others settings and simply redefine the input_dim.

models/src/anemoi/models/models/base.py
matschreiner: I think that we should have a distinction between network architectures and models.
The network should own the forward pass and the trainable parameters, while the model should own the prediction and training logic and other stuff like fetching or owning graphs, processing diffusion times, stacking ensemble inputs, prediction step, and training steps, etc.

I think we are going down a path of mixing these responsibilities, which means that we wont easily be able to experiment with different architecures without rewriting the models or create new models without rewriting lots of the architecture.

models/src/anemoi/models/models/base.py
matschreiner: Yeah, could also be something like that :) But my initial thought woul be to have a class or module that can handle all things related to boundings.

models/src/anemoi/models/models/base.py
anaprietonem: I guess to me at the moment the network architectures are the blocks that lives under 'layers' while the models itself live under 'models'. With the diffusion model we already saw that we need how we needed a specific predict_step for that model and hence the generic one of the interface was no longer an option.  As it is right now the interface is just a very light wrapper around the 'predict_step' that we need to handle the instantiation of the pre/post processors and run inference in the right way. 

```
The network should own the forward pass and the trainable parameters, while the model should own the prediction and training logic and other stuff like fetching or owning graphs, processing diffusion times, stacking ensemble inputs, prediction step, and training steps, etc.
```

you reckon then we'd still need an abstraction layer between the models and the layers? I thought that by having the predict step as part of the `AnemoiGraphModelBase` is a step towards model's owning prediction and training logic.

matschreiner: So for example, in the diffusion model, there is additional logic required for the forward step — such as embedding the diffusion time and concatenating it to the input data, or preconditioning the corrupted data in EDM. This logic should IMO be separate from the encoder–processor–decoder backbone. If we embed both the model logic and the backbone in a single class, we give up on modularity and make it harder to swap out backbones in the future.

Essentially, the diffusion model, ensemble model, and all the other variants we have, are overarching models that all require a backbone as a component but implement different training and prediction logic etc.. If we mix the backbone with the model, for example by building the diffusion model directly into the encoder–processor–decoder architecture, we make it difficult to experiment with different backbones in the diffusion setting because we have to implement a new diffusion model each time. And similarly, any change to the backbone would then have to be implemented separately in each model class.

So I think that what we need to decide on about this PR is, are we defining a base class for backbones that can be reused across models or are we defining a base class for models? Or are we intentionally combining these responsibilities into a single class? I know that separating these concerns is extremely hard, but I think that for all the effort it takes it be worth it in the long run. 

Would it make sense to move this discussion to an issue so more people can join the discussion?

@mchantry

ssmmnn11: > So for example, in the diffusion model, there is additional logic required for the forward step — such as embedding the diffusion time and concatenating it to the input data, or preconditioning the corrupted data in EDM. This logic should IMO be separate from the encoder–processor–decoder backbone. If we embed both the model logic and the backbone in a single class, we give up on modularity and make it harder to swap out backbones in the future.
> 
> Essentially, the diffusion model, ensemble model, and all the other variants we have, are overarching models that all require a backbone as a component but implement different training and prediction logic etc.. If we mix the backbone with the model, for example by building the diffusion model directly into the encoder–processor–decoder architecture, we make it difficult to experiment with different backbones in the diffusion setting because we have to implement a new diffusion model each time. And similarly, any change to the backbone would then have to be implemented separately in each model class.
> 
> So I think that what we need to decide on about this PR is, are we defining a base class for backbones that can be reused across models or are we defining a base class for models? Or are we intentionally combining these responsibilities into a single class? I know that separating these concerns is extremely hard, but I think that for all the effort it takes it be worth it in the long run.
> 
> Would it make sense to move this discussion to an issue so more people can join the discussion?
> 
> @mchantry

I think it depends a bit on what you see / define as backbone. Currently you can switch out encoder, decoder, processor, so for example have a Transformer GNN as processor, or a sliding window transformer etc. Do you understand backbone as the whole, encoder, processor, decoder? In this case, if you would want to move the diffusion logic / ens perturbations out, then each backbone would need to support e.g. conditioning, so that you provide input, target, conditioning, and it then does the right thing with it. Like our current blocks, that handle conditioning via conditional layernorms, or a potential DiT block, which has similar properties. Or another architecture could just concatenate it to the inputs and edges etc. But for multi-modal training, would you then have encoders that give their output to the backbone encoder, that decodes and then gives that output to the additional decoders?

matschreiner: Yes, I think we should view the whole enc-proc-dec as the backbone. The models work differently and therefore their backbones also have to differ. In the ensemble we need a probabilistic backbone. Now we hardcode the way of providing stochastic predictions by injecting noise into the processor layer inside the ensemble, and thereby we tie all the ensemble-specific logic like CRPS and calculating uncertainties etc to this specific probabilistic backbone. This means that we can't experiment with other probabilistic backbones in the ensemble model. 

It is the same in the diffusion model, it is true that the backbone will requires different arguments and therefore has to be different. But hardcoding the enc-proc-dec and the wiring between them into the diffusion model gives it responsibilities it shouldn't have. And it makes it very hard to do experiments and research with the model.

models/src/anemoi/models/models/base.py
icedoom888: Yes as @anaprietonem said this is basically a guideline for developers ot know what to implement once inherinting fromthis class

models/src/anemoi/models/models/base.py
icedoom888: Sure we can branch this out in another PR for future work

models/src/anemoi/models/models/base.py
icedoom888: We can do either.
I would actually isolate in in the layers/bounding.py, as it makes it more organised and doesn't confuse people that don't want to use the bounding

models/src/anemoi/models/models/base.py
icedoom888: Agreed, i was trying to keep it as close as possible to the original naming not to confuse people ;)

models/src/anemoi/models/models/base.py
icedoom888: We should probbaly schedule a meetin gto reach a decision on this, @anaprietonem, @matschreiner do you have any time this/next week?

models/src/anemoi/models/models/base.py
icedoom888: Done!

models/src/anemoi/models/models/base.py
icedoom888: Done!

models/src/anemoi/models/models/base.py
matschreiner: Yes, right, I guess I'm just in doubt of why it should be enforced to implement assemble i/o, since you are free to implement the forward method in any way you like when inheriting from this base class?
Maybe there other functionality in Anemoi that requires the models to have something like this?

models/src/anemoi/models/models/base.py
icedoom888: I moved this to a truncation class, makes it much cleaner.

models/src/anemoi/models/models/base.py
anaprietonem: Hello, yep happy to have a chat! Can we extend the invite to  @JPXKQX and @ssmmnn11 too? Checking our calendars, what about Thursday 11:30 CEST or 16:30 CEST?

matschreiner: Thursdsay 1630 is fine for me! :)

icedoom888:
@anaprietonem thanks for taking a look at this, i reimplemented and refactor all truncation, bounding and mapper functions, to make it all more organised and smooth. Let me know what you think!

Integratio tests are passing [here](https://github.com/ecmwf/anemoi-core/actions/runs/18143542388) ;)

icedoom888:
@ssmmnn11 I also implemented the changes to the diffusion classes, could already be enough without the extra PR, could you take  alook?

models/src/anemoi/models/models/base.py
icedoom888: 16:30 works for me!

ssmmnn11:
> @ssmmnn11 I also implemented the changes to the diffusion classes, could already be enough without the extra PR, could you take alook?

I would like to change some things, should I do it in the branch and push or via a pull request?

models/src/anemoi/models/distributed/shapes.py
```diff
@@ -38,3 +38,10 @@ def apply_shard_shapes(tensor: Tensor, dim: int, shard_shapes_dim: list) -> list
         shard_shapes[i][dim] = shard_shape
 
     return shard_shapes
+
+
+def gather_shard_shapes(x, dim=0, shard_shapes_dim: int = None, model_comm_group: Optional[ProcessGroup] = None):
```
ssmmnn11: One concern: in almost all call sites the intent is unambiguous - we either get shard shapes or apply them. Only two places actually need “get or apply” behaviour. Replacing everything with a single gather_shard_shapes blurs those paths and makes the parallelisation model harder to follow.

Could you please:

Rename gather_shard_shapes to get_or_apply_shard_shapes and use it only in the two call sites that genuinely need to branch.

Restore the original split elsewhere: keep get_shard_shapes and apply_shard_shapes

icedoom888: Sure, i can revert that change.

ssmmnn11: thx!

icedoom888: It's done ;)

models/src/anemoi/models/layers/mapper.py
```diff
@@ -513,9 +513,9 @@ def forward(
         }
 
         if self.shard_strategy == "edges":
-            return self.forward_with_edge_sharding(**kwargs_forward)
+            return self.mapper_forward_with_edge_sharding(**kwargs_forward)
         else:  # self.shard_strategy == "heads"
-            return self.forward_with_heads_sharding(**kwargs_forward)
+            return checkpoint(self.mapper_forward_with_heads_sharding, **kwargs_forward, use_reentrant=False)
```
anaprietonem, line 518: why has the `use_reentrant=False` being added  here?

models/src/anemoi/models/models/base.py
anaprietonem: Thanks Alberto, that looks good. My two cents is the interface for bounding and truncation is different (as in you went for a class for the matrices and a function for the bounding). Could we update the bounding to also use a base class or the other way around?

models/src/anemoi/models/models/base.py
anaprietonem: thanks for fixing that Alberto!

models/src/anemoi/models/models/base.py
anaprietonem: But as it is right now the assemble I/O methods are just 'abstract methods' that then the children specify, so if a new model class doesn't need to specify that if could be left as it is right? (Maybe I am missing something from your comment)

models/src/anemoi/models/layers/mapper.py
```diff
@@ -852,6 +852,30 @@ def forward(
...
+        x: PairTensor,
+        batch_size: int,
+        shard_shapes: tuple[tuple[int], tuple[int]],
+        model_comm_group: Optional[ProcessGroup] = None,
+        x_src_is_sharded: bool = False,
+        x_dst_is_sharded: bool = False,
+        keep_x_dst_sharded: bool = False,
+        **kwargs,
+    ) -> PairTensor:
+        return checkpoint(
```
anaprietonem, line 866: these changes seems a bit unrelated to the AnemoiBaseModel? ( have you seen or run the GNN set up to check if the checkpointing was having any benefit?)

models/src/anemoi/models/layers/mapper.py
japols: I think @icedoom888 just moved the checkpointing logic to mapper.py, including `use_reentrant=False`, which is more consistent across edge/head sharding.

models/src/anemoi/models/layers/mapper.py
ssmmnn11: I am responsible for that change -> @icedoom888  moved run_mapper to the mapper.py, but then it is just like forward so I changed forward to have the checkpointing instead.

models/src/anemoi/models/layers/mapper.py
ssmmnn11: Because _run_mappers was moved from model to the mappers. However, here it basically duplicates forward, so I changed forward to have the checkpoint. use_reentrant is the "new" way to run gradient checkpoints - which is probably default by now and then we could remove the flag.

icedoom888: Hey @ssmmnn11 , i left it separated from forward because of cyclic calls within checkpoint, but i think like this should work!

ssmmnn11: forward wraps now mapper_forward, before run_mapper wrapped forward

icedoom888:
@ssmmnn11 thanks for the inputs yesterday, I noticed some tests are not working on this PR after some changes you introduced, could we maybe wix them up and try to finally merge this in? @anaprietonem I'm working on other PRs in models that are heavily dependant on this ;)

models/src/anemoi/models/models/base.py
matschreiner: Since they are abstract methods you have to define them in the children which means that we are forcing people towards a specific implementation which they actually don't need and may not want to follow. 
I don't think there are any good reason to tie peoples hands here

matschreiner: > But as it is right now the assemble I/O methods are just 'abstract methods' that then the children specific, so if a new model class doesn't need to specific that if could be left as it is right? (Maybe I am missing something from your comment)

The point of abstract methods is actually that you have to define them. The abstract method decorator is supposed to be used when inheriting from ABC - here it is just a noop. But if you had inherited from ABC here and a subclass of this didn't implement the abstract methods it would throw an error.

anaprietonem:
> @ssmmnn11 thanks for the inputs yesterday, I noticed some tests are not working on this PR after some changes you introduced, could we maybe wix them up and try to finally merge this in? @anaprietonem I'm working on other PRs in models that are heavily dependant on this ;)

Hey Alberto, indeed it would be good to merge it this week as there are already some PRs from other contributors waiting for this. I see my comments are now addressed, so I have marked them as resolved and approved.  @matschreiner anything from your side or are we good to go?

anaprietonem (approved):
Thanks Alberto for the addressing the comments, nice work. I have triggered the Integration tests https://github.com/ecmwf/anemoi-core/actions/runs/18273206927 if those are passing and no other comments from reviewers we are good to merge

icedoom888:
Thanks a lot @anaprietonem for looking again into this! Happy to see all the green checks 😉 
@mchantry most are currently still private, I didn't want to overflow the board 😆

models/src/anemoi/models/models/base.py
icedoom888: Actually this was done on purpose! 
In my opinion it's better to enforce some rules to the developers inheriting from the base model.
This forces developers to structure the code following the assemble i/o logic making the code more structured and readable by others that will have to interact with the new models. 
This can be bypassed by returning x anyways.

icedoom888:
> @icedoom888, was talking to @matschreiner to check if there was something from his side. This comment [#440 (comment)](https://github.com/ecmwf/anemoi-core/pull/440#discussion_r2404188625) is still opened. Any thoughts on how to address this?

Imo that is actually expected behaviour, do you have any preferences?

models/src/anemoi/models/models/base.py
matschreiner: You are forcing users to implement an interface that they don't need which is breaking with the design principle called interface segregation. 
I don't think that structuring the code with assemble i/o is the necessarily best pattern everywhere, but now it is enforced regardless.

But maybe we've discussed it enough now, I think you should go for your implementation :)

icedoom888: If we wanted to solve "interface segregation" we would need to refactor the whole anemoi 😆

anaprietonem: I think it's fine to merge as it is, as we touched on the other day there are few other changes that will follow up from this (AnemoiInterface and Backbones), so this might come up again and we can revisit then!

Issue #437, opened 2025-07-28
Issue when using warm_start


anaprietonem:
### What happened?

To start from a checkpoint in anemoi-training the condition doesn't account for passing a`warm_start` file path so this functionality is currently broken.  Warm_path needs to include a fork_run_id to work and that's not correctly checked in the code.

Proposed fix:
- Update the logic for starting from a checkpoint to account for passing a warm_start path
- Option to define a warm_start as an independent path different from the run_id/fork_run_id

### What are the steps to reproduce the bug?

Try to start a training with no fork_run_id or run_id, and a warm_path defined. The `last_checkpoint` will be None and training won't start from the expected checkpoint

### Version

Main

### Platform (OS and architecture)

ATOS

### Relevant log output

```shell

```

mtgarciag:
I have tried with changing line 79 in train.py to:
 `self.start_from_checkpoint = bool(self.config.training.run_id) or bool(self.config.training.fork_run_id) or bool(self.config.hardware.files.warm_start)`
to avoid using a run_id, but this does not work when using more than one gpu.

PR #444, opened 2025-07-30
feat(graphs): support field interpolation in recipe.yaml

JPXKQX:
## Description

This PR adds support for field interpolation in configuration recipes and validation of configuration files. The config can be validated during the graph creation with an extra "--validate/--no-validate" argument or in a specific command "anemoi-graphs validate recipe.yaml"

New feature. Field interpolation is added to graph recipes. Config validation is supported for anemoi-graphs.

#437 

##  Additional notes ##

### DIFF
--- a/graphs/src/anemoi/graphs/commands/validate.py
+++ b/graphs/src/anemoi/graphs/commands/validate.py
@@ -1,0 +1,117 @@
+# (C) Copyright 2024 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import logging
+import os, re
+from . import Command
+from anemoi.graphs.create import GraphCreator
+from anemoi.graphs.schemas.base_graph import BaseGraphSchema
+from omegaconf import DictConfig, OmegaConf
+from anemoi.utils.schemas import BaseModel
+
+LOGGER = logging.getLogger(__name__)
+
+class Validate(Command):
+    """Validate a graph."""
+
+    internal = True
+    timestamp = True
+
+    def add_arguments(self, command_parser):
+        command_parser.add_argument("config", help="Path to the configuration file (a YAML file).")
+        command_parser.add_argument(
+            "--mask_env_vars",
+            "-m",
+            help="Mask environment variables from config. Default False",
+            action="store_true",
+        )
+
+    def _mask_slurm_env_variables(self, cfg: DictConfig) -> None:
+        """Mask environment variables are set."""
+        # Convert OmegaConf dict to YAML format (raw string)
+        raw_cfg = OmegaConf.to_yaml(cfg)
+        # To extract and replace environment variables, loop through the matches
+        updated_cfg = raw_cfg
+        primitive_type_hints = extract_primitive_type_hints(BaseGraphSchema)
+
+        patterns = [
+            r"(\w+):\s*\$\{oc\.env:([A-Z0-9_]+)\}(?!\})",
+            r"(\w+):\s*\$\{oc\.decode:\$\{oc\.env:([A-Z0-9_]+)\}\}",
+        ]
+        replaces = ["${{oc.env:{match}}}", "${{oc.decode:${{oc.env:{match}}}}}"]
+        # Find all matches in the raw_cfg string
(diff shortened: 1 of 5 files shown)


graphs/src/anemoi/graphs/commands/validate.py
```diff
@@ -0,0 +1,125 @@
...
+from anemoi.graphs.create import GraphCreator
+from anemoi.graphs.schemas.base_graph import BaseGraphSchema
+from anemoi.utils.schemas import BaseModel
+
+from . import Command
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Validate(Command):
```
anaprietonem, line 28: We could look at abstracting this class in anemoi-utils and then reusing the command for training and graphs, as this ultimately validating the yaml and not caring about what that file is about

graphs/src/anemoi/graphs/schemas/base_graph.py
```diff
@@ -40,27 +44,41 @@ class EdgeSchema(BaseModel):
...
-            msg = "If overwrite is True, nodes and edges must be provided."
-            raise ValueError(msg)
-        return self
+
+    def model_dump(self, by_alias: bool = False) -> DictConfig:
+        dumped_model = super().model_dump(by_alias=by_alias)
+        return DictConfig(dumped_model)
+
+
+class UnvalidatedGraphSchema(PydanticBaseModel):
```
anaprietonem, line 63: could be worth having an `UnvalidatedBaseSchema` in utils and reusing that across training and graphs

graphs/src/anemoi/graphs/schemas/base_graph.py
```diff
@@ -40,27 +44,41 @@ class EdgeSchema(BaseModel):
...
+    "Nodes schema for all types of nodes (ex. data, hidden)."
+    edges: Any
+    "List of edges schema."
+
+    def model_dump(self, by_alias: bool = False) -> DictConfig:
+        dumped_model = super().model_dump(by_alias=by_alias)
+        return DictConfig(dumped_model)
+
+
+def validate_graph_schema(config: DictConfig) -> BaseGraphSchema:
```
anaprietonem, line 76: same for this

training/src/anemoi/training/schemas/base_schema.py
```diff
@@ -42,6 +44,26 @@
 LOGGER = logging.getLogger(__name__)
 
 
+class GraphSchema(BaseGraphSchema):
```
anaprietonem, line 47: why should this leave in training? If we go this route we should check that again the schema of the graph object doesn't get pickled into the checkpoint making anemoi-training a dependency of inference (there is a test to check this) but my preference any way would for this to live in graphs

graphs/src/anemoi/graphs/create.py
```diff
@@ -16,48 +16,64 @@
...
 
+from anemoi.graphs.schemas.base_graph import BaseGraphSchema
+from anemoi.graphs.schemas.base_graph import UnvalidatedGraphSchema
 from anemoi.utils.config import DotDict
 
 LOGGER = logging.getLogger(__name__)
 
 
+# Support previous version. This will be deprecated in a future release
+def map_edge_builder_to_new_schema(config: DictConfig) -> DictConfig:
```
anaprietonem, line 30: What's is this about?

anaprietonem (commented):
Thanks @JPXKQX  for this work. Happy to discuss some of the suggestions and how to organise this work. I think it could be good to have a test in graphs to confirm the validations works well similarly to the one we have in training

anaprietonem:
Btw i tested and the omegaConf interpolation works nicely, though I noticed I had to update the original yaml to address the from attributes: ${graph.attributes.nodes} -> attributes: ${attributes.nodes} which makes sense as that's the key for resolving from training. Any views on how to handle this?

graphs/src/anemoi/graphs/create.py
JPXKQX: This mapping was added to prevent breaking the configurations, as we were doing so very frequently at one point (~1 year ago)

graphs/src/anemoi/graphs/commands/validate.py
JPXKQX: yes, in fact, I copied this from `training`

training/src/anemoi/training/schemas/base_schema.py
JPXKQX: This includes not only the graph configuration, but also additional information such as the name of the data, hidden nodes, and whether to overwrite the graph

JPXKQX:
Alternative solution implemented in `feat/graph-subcommand`, where a subcommand graph is added to anemoi-training CLI to use the training configs directly.

PR #448, opened 2025-07-31
fix: Make batch normalization explicit.

jakob-schloer:
## Description
Fixes the inconsistency on the batch normalization between the forecaster and ensforecaster. The PR

- makes batch normalization explicit by moving it to the lightning hook `on_after_batch_transfer`
- removes the `training_mode` keyword since there also exists a `validation_mode` keyword 
- Updated all training task types including diffusion

## Test training of

- [x] Deterministic GraphForecaster (https://mlflow.ecmwf.int/#/experiments/250/runs/ac7222c71bfe4971879d01b15c313f86)
- [x] EnsGraphForecaster (https://mlflow.ecmwf.int/#/experiments/266/runs/672614b1db4a49838f3b95df9e5dbc97)
- [x] GraphDiffusionForecaster (https://mlflow.ecmwf.int/#/experiments/387/runs/8f833d545fb74a4983e721061b0bc6e1)
- [x] GraphDiffusionTendForecaster
- [ ] TimeInterpolatorForecaster

### DIFF
--- a/training/src/anemoi/training/train/tasks/diffusionforecaster.py
+++ b/training/src/anemoi/training/train/tasks/diffusionforecaster.py
@@ -283,7 +283,6 @@ class GraphDiffusionTendForecaster(GraphDiffusionForecaster):
             y_pred_state_full, y_state_full, grid_shard_slice_metrics = self._prepare_tensors_for_loss(
                 y_pred_state,
                 y_state,
-                False,
                 validation_mode,
             )
             metrics_next = self._compute_metrics(


training/src/anemoi/training/train/tasks/base.py
```diff
@@ -339,12 +339,11 @@ def compute_loss_metrics(
         y_pred: torch.Tensor,
         y: torch.Tensor,
         rollout_step: int = 0,
-        training_mode: bool = True,
         validation_mode: bool = False,
     ) -> torch.Tensor:
         is_sharded = self.grid_shard_slice is not None
 
-        sharding_supported = (self.loss_supports_sharding or not training_mode) and (
+        sharding_supported = (self.loss_supports_sharding or validation_mode) and (
```
ssmmnn11, line 371: I think training_mode (very confusing name) here is used here (I don't think elsewhere apart from the callbacks - but better to check):

https://github.com/ecmwf/anemoi-core/blob/5ce2ef2a12d42c413985ca4d6f1d81eaec7abbe7/training/src/anemoi/training/diagnostics/callbacks/plot.py#L450

I think validation_mode True does not mean we should set sharding_supported to True (in "training_mode we skip the loss computation - not in validation_mode").

training_mode should get probably a better name ; or we would like to implement the logic differently. 

I would be in favour of removing LongRolloutPlots

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -363,7 +362,7 @@ def compute_loss_metrics(
                 grid_shard_slice=grid_shard_slice,
                 group=self.model_comm_group,
             )
-            if training_mode
```
ssmmnn11: see comment above

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -316,7 +302,7 @@ def rollout_step(
                     validation_mode,
                     use_reentrant=False,
                 )
-                if training_mode
```
ssmmnn11, line 321: see comment above

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -332,6 +318,38 @@ def rollout_step(
...
+        dataloader_idx : int
+            Dataloader index
+
+        Returns
+        -------
+        torch.Tensor
+            Batch after transfer and preprocessing
+        """
+        # First handle batch gathering/sharding from parent class
+        batch = super().on_after_batch_transfer(batch, dataloader_idx)
```
ssmmnn11, line 337: does this work? in parent you have self.model.pre_processors(batch)

ssmmnn11 (changes requested):
I think a good idea. But we need to think about a couple of things regarding the training_mode weirdness. And can we merge it after the diffusion model?

training/src/anemoi/training/train/tasks/base.py
jakob-schloer: Well spotted.

I think we want to compute the loss during training and validation, but compute additional metrics only during validation. 

The `training_mode` (I agree very confusing name) is only used in the LongRolloutPlots callbacks, and I guess only for saving memory. However, the loss computation shouldn't require a lot of memory/computation during validation, as the gradients are not computed. 
But maybe I'm missing something.

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -332,6 +318,38 @@ def rollout_step(
...
+
+        # Apply preprocessing (normalization) to the ensemble batch
+        self.model.pre_processors(batch[0])  # normalized in-place
+
+        # If we have EDA perturbations, preprocess them too
+        if len(batch) == 2:
+            self.model.pre_processors(batch[1])
+
+        # Initialize delayed scalers after first preprocessing
```
jakob-schloer: Any opinions on whether this should be in the lightning hook or in the rollout_step?

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -381,6 +380,8 @@ def compute_loss_metrics(
     def on_after_batch_transfer(self, batch: torch.Tensor, _: int) -> torch.Tensor:
         """Assemble batch after transfer to GPU by gathering the batch shards if needed.
 
+        Also normalize the batch in-place if needed.
```
jakob-schloer, line 497: We could also move the normalization in the base trainer class. Can you think of any case where we don't want to normalize the batch in training?

training/src/anemoi/training/train/tasks/ensforecaster.py
ssmmnn11: In the hook I think.

training/src/anemoi/training/train/tasks/base.py
ssmmnn11: I think that is correct. Best to test how it impacts memory. But I also don’t know if this callback is actually used by anyone now. Can you remove the training_mode then everywhere and run a test?

jakob-schloer:
> I think a good idea. But we need to think about a couple of things regarding the training_mode weirdness. And can we merge it after the diffusion model?

But will the diffusion model then have the `training_mode` flag?

training/src/anemoi/training/train/tasks/base.py
jakob-schloer: Done! I ran a test for both forecaster and ensforecaster. Even the LongRolloutPlots still work for me :)

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -304,20 +290,15 @@ def rollout_step(
...
-                if training_mode
-                else None
+            loss, y_pred_ens_group = checkpoint(
+                self.gather_and_compute_loss,
+                y_pred,
+                y,
+                self.loss,
+                self.ens_comm_subgroup_size,
+                self.ens_comm_subgroup,
+                self.model_comm_group,
```
ssmmnn11, line 325: I think before you would only return the ensemble if in validation mode. now you also during the training step.

ssmmnn11:
> > I think a good idea. But we need to think about a couple of things regarding the training_mode weirdness. And can we merge it after the diffusion model?
> 
> But will the diffusion model then have the `training_mode` flag?

No, but I had to refactor loss computation and it would be easier to introduce this after the refactor.

training/src/anemoi/training/train/tasks/ensforecaster.py
jakob-schloer: Yes, just realized that now. Got confused with the keyword. It should be `return_pred_ens=validation_mode`.

training/src/anemoi/training/train/tasks/diffusionforecaster.py
```diff
@@ -105,11 +105,40 @@ def _compute_loss(
...
+        dataloader_idx : int
+            Dataloader index
+
+        Returns
+        -------
+        torch.Tensor
+            Batch after transfer and preprocessing
+        """
+        # First handle batch gathering/sharding from parent class
+        batch = super().on_after_batch_transfer(batch, dataloader_idx)
```
ssmmnn11: Is this correct here? We have GraphDiffusionForecaster(GraphForecaster), which here will be super.

training/src/anemoi/training/train/tasks/diffusionforecaster.py
```diff
@@ -335,9 +344,6 @@ def rollout_step(
         rollout : Optional[int], optional
             Number of times to rollout for, by default None
             If None, will use self.rollout
-        training_mode : bool, optional
-            Whether in training mode and to calculate the loss, by default True
-            If False, loss will be None
         validation_mode : bool, optional
             Whether in validation mode, and to calculate validation metrics, by default False
             If False, metrics will be empty
```
ssmmnn11, line 316: is the tendency diffusion forecaster broken now?

training/src/anemoi/training/train/tasks/diffusionforecaster.py
ssmmnn11: Instead of changing the on_after_batch_transfer in each class, I think it would be easier to understand if we introduce methods that are called in on_after_batch_transfer which are then used / overwritten by the child classes. And we could also reference them in the step / rollout step function (and say when and where they are called).

ssmmnn11 (changes requested):
Some suggestions :-)

training/src/anemoi/training/train/tasks/diffusionforecaster.py
jakob-schloer: That part was correct, but I forgot to update the GraphDiffusionTendForecaster. Fixed now.

training/src/anemoi/training/train/tasks/diffusionforecaster.py
jakob-schloer: My bad, I forgot to update the GraphDiffusionTendencyForecaster. Should be fixed now.

training/src/anemoi/training/train/tasks/diffusionforecaster.py
jakob-schloer: Good idea. Have a look now. I've moved the three steps in the on_batch_transfer operation to separate functions.

mchantry:
@jakob-schloer the integration tests are currently failing (sorry if you already know this)
```
E           TypeError: BaseGraphModule._prepare_tensors_for_loss() takes from 3 to 4 positional arguments but 5 were given

training/src/anemoi/training/train/tasks/diffusionforecaster.py:315: TypeError
```
https://github.com/ecmwf/anemoi-core/actions/runs/18100798416/job/51502969022#step:3:3978

PR #451, opened 2025-08-04
feat(models): add configurable residual connections in enc-proc-dec

matschreiner:
## Description

This PR refactors the truncation logic previously embedded within the encoder-processor-decoder model by extracting it into a dedicated TruncationMapper module.

It introduces a new truncation key to the BaseGraphSchema, which specifies the set of grid nodes to which the full grid will be projected during truncation.

The TruncationMapper builds on the new SparseProjector class, which constructs a sparse projection matrix from:
an edge_index, edge_length, the number of source nodes, and the number of destination nodes.

By default, the SparseProjector applies a Gaussian distance weighting and row normalization, ensuring that the projection weights sum to one.

The truncation logic, which was previously scattered throughout the enc-proc-dec model, is now fully encapsulated in the TruncationMapper, simplifying and streamlining the core model code.

This PR also introduces a new configurable key, residual, under the model configuration, which allows flexible specification of the behavior of the residual input x_skip via:

SkipConnection: direct pass-through of the most recent input frame (x_skip).

TruncationMapper: coarse-graining via sparse down/up projection.

NoConnection: disables residual input by returning a zero tensor.

Modularizes and centralizes the truncation logic.
Enables configurable residual strategies

----
📚 Documentation preview 📚: https://anemoi-training--451.org.readthedocs.build/en/451/

----
📚 Documentation preview 📚: https://anemoi-graphs--451.org.readthedocs.build/en/451/

----
📚 Documentation preview 📚: https://anemoi-models--451.org.readthedocs.build/en/451/


mchantry:
@matschreiner thanks for the contribution. Could you update this to be consistent with main please, as this will greatly help the ability to review.

ssmmnn11:
I have this pull request here:

https://github.com/ecmwf/anemoi-core/pull/388

can we make it work for this use case as well? that would be nice.

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -73,15 +70,9 @@ def __init__(
...
-        # these will be moved to the GPU when first used via sefl.interpolate_down/interpolate_up
-        self.A_down, self.A_up = None, None
-        if "down" in self._truncation_data:
-            self.A_down = self._make_truncation_matrix(self._truncation_data["down"])
-            LOGGER.info("Truncation: A_down %s", self.A_down.shape)
-        if "up" in self._truncation_data:
-            self.A_up = self._make_truncation_matrix(self._truncation_data["up"])
-            LOGGER.info("Truncation: A_up %s", self.A_up.shape)
+        # Residual data -> data
+        self.residual = instantiate(model_config.model.residual)
```
JPXKQX:
```
instantiate(
    model_config.model.residual,
    ...
    sub_graph_down=self._graph_data[(self._graph_name_data, "to", self._graph_name_truncation)],
    sub_graph_up=self._graph_data[(self._graph_name_truncation, "to", self._graph_name_data)],
)
````

Comment (JPXKQX) [reply models/src/anemoi/models/models/encoder_processor_decoder.py]: `_graph_name_truncation` is a new key in the config

File: models/tests/layers/test_residual.py
Comment (mchantry) [inline models/tests/layers/test_residual.py:1]: Is this supposed to be fully commented out @matschreiner ?

File: models/tests/models/test_encprocdec.py
Comment (mchantry) [inline models/tests/models/test_encprocdec.py:1]: @matschreiner same for this one?

Comment (matschreiner) [reply models/tests/models/test_encprocdec.py:1]: I was talking with @JPXKQX and he said that we would probably need to remove the tests because of how I was relying on pickled files, but we can run the tests in the other file actually :)  @mchantry.

Comment (mchantry) [reply models/tests/models/test_encprocdec.py:1]: ah yes, I see now

Comment (mchantry) [reply models/tests/models/test_encprocdec.py:1]: If we could manage the pickles do you feel strongly about having this test?

**Comment by havardhhaugen** [comment]
Nice work @matschreiner.  I'm wondering if we can generalize the way weighting is done to make it more configurable and easier to add new methods in the future? And allow scaling by source node attributes? (Specifically for stretched grid I would like to weight by the area weight of input nodes to account for the non-uniform spatial resolution).

File: models/tests/models/test_encprocdec.py
Comment (matschreiner) [reply models/tests/models/test_encprocdec.py:1]: Nono, I'm completely fine with not keeping this test. :) But I think it would be valuable in general to have unit tests for the models. I would like to develop against test cases instead of running the pipeline when developing modules.

File: models/src/anemoi/models/models/encoder_processor_decoder.py [hunk 2]
```diff
@@ -57,31 +54,31 @@ def __init__(
...
-            self.A_down = self._make_truncation_matrix(self._truncation_data["down"])
-            LOGGER.info("Truncation: A_down %s", self.A_down.shape)
-        if "up" in self._truncation_data:
-            self.A_up = self._make_truncation_matrix(self._truncation_data["up"])
-            LOGGER.info("Truncation: A_up %s", self.A_up.shape)
+        # Residual data -> data
+        self.residual = instantiate(
+            model_config.model.residual,
+            num_data_nodes=graph_data[self._graph_name_data].num_nodes,
+            num_truncation_nodes=graph_data[self._graph_name_hidden].num_nodes,
```
havardhhaugen: Should this be `self._graph_name_truncation`? Or is it assumed that you always truncate via the hidden mesh nodes?

matschreiner: Yes, nice catch!
This is supposed to be the truncation graph - which of course can be the hidden graph but doesn't have to.
However looking at it again now, I think it needs to be changed a little bit. Since this modules can be a skip connections or other modules that don't depend on the graph, I think I need to find another way of passing these arguments..

mchantry:
Great work @matschreiner could there be some documentation added please?
The design choice that you've gone for means that the default configs are unchanged, but this makes documentation even more important so that users know they can change the residual blocks away from the skip connection.

models/src/anemoi/models/schemas/models.py
```diff
@@ -188,7 +191,10 @@ class BaseModelSchema(PydanticBaseModel):
         ...,
         discriminator="target_",
     )
-    "GNN decoder schema."
+    residual: Union[SkipConnectionSchema, TruncationMapperSchema, NoConnectionSchema] = Field(
```
mchantry: The use of a default factory is nice for preventing a change to configs, what do you think about adding a commented out line in the configuration file for models `#residual: SkipConnectionSchema` so that people can see where to change the type of residual connection?

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+        self.project_up = SparseProjector(
+            edge_index=sub_graph_up.edge_index,
+            weights=sub_graph_up.edge_length.squeeze(),
+            src_size=num_truncation_nodes,
+            dst_size=num_data_nodes,
+        )
+
+    def forward(self, x, grid_shard_shapes=None, model_comm_group=None, *args, **kwargs):
+        self.to(x.device)
```
ssmmnn11: please add comment why we move it to device in forward please (as we had before)

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,154 @@
...
+    weights,
+    num_target_nodes,
+):
+    total = torch.zeros(num_target_nodes, device=weights.device)
+    norm = total.scatter_add_(0, edge_index[1].long(), weights)
+    norm = norm[edge_index[1]]
+    return weights / (norm + 1e-8)
+
+
+def _gaussian_distance_weighting(weights: torch.Tensor, sigma: float = 1.0) -> torch.Tensor:
```
ssmmnn11: sigma needs to be configurable

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,154 @@
...
+            edge_index,
+            weights,
+            (src_size, dst_size),
+            device=edge_index.device,
+        ).coalesce()
+
+    def forward(self, x, *args, **kwargs):
+        out = []
+        for i in range(x.shape[0]):
+            out.append(torch.sparse.mm(self.projection_matrix.T, x[i, ...]))
```
ssmmnn11: you transpose the matrix in every application, please either store transposed matrix or do it once in forward.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,154 @@
...
+        return torch.stack(out)
```
ssmmnn11: in the original code we deactivated amp (mixed precision). this now runs with mixed precision? is it casting the skipped connection to e.g. float16? I would like control over that. could you make it configurable?

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -131,59 +125,13 @@ def __init__(
...
-                out.append(self._multiply_sparse(x[i, ...], A))
-        return torch.stack(out)
-
     def _get_shard_shapes(self, x, dim=0, shard_shapes_dim=None, model_comm_group=None):
         if shard_shapes_dim is None:
             return get_shard_shapes(x, dim, model_comm_group)
         else:
             return apply_shard_shapes(x, dim, shard_shapes_dim)
 
-    def _apply_truncation(self, x, grid_shard_shapes=None, model_comm_group=None):
```
ssmmnn11, line 164: Do you plan to adapt AnemoiEnsModelEncProcDec as well?

ssmmnn11 (dismissed):
some comments :-)

matschreiner:
> Nice work @matschreiner. I'm wondering if we can generalize the way weighting is done to make it more configurable and easier to add new methods in the future? And allow scaling by source node attributes? (Specifically for stretched grid I would like to weight by the area weight of input nodes to account for the non-uniform spatial resolution).

Thanks :) Syure, good idea, now you can specify which edge_attribute you want to weigh by in the config file :)

models/src/anemoi/models/layers/residual.py
ssmmnn11: I forgot, can we add something like:

def _row_normalize_weights(edge_index, weights, num_target_nodes, power=1.0):
    powered = weights ** power
    total = torch.zeros(num_target_nodes, device=weights.device)
    norm = total.scatter_add_(0, edge_index[1].long(), powered)
    norm = norm[edge_index[1]]
    return weights / (norm + 1e-8)

matschreiner: In fact, I was wondering whether this should be deleted altogether? 
I changed the code now such that it is configurable which edge attribute to use for the weighting, and since the gaussian distance weight is implemented here https://anemoi.readthedocs.io/projects/graphs/en/latest/modules/edge_attributes.html#anemoi.graphs.edges.attributes.GaussianDistanceWeights maybe it would be better to simply create a graph with that edge attribute?

ssmmnn11: good point. that is a much better option. leave everything to the graph creation, where we have all the options already.

models/src/anemoi/models/layers/residual.py
matschreiner: Done, I also moved the call to the SparseProjector, where the sparse matrix actually lives

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+
+    def forward(self, x, *args, **kwargs):
+        x = x[:, -1, ...]  # pick current date
+        return torch.zeros_like(x, device=x.device, dtype=x.dtype)
+
+
+class TruncationMapper(nn.Module):
```
JPXKQX: I would probably rename this to `TruncatedConnection` or something similar without the "mapper"

models/src/anemoi/models/layers/residual.py
matschreiner: I made it work as the original code now, where you have a flag to enable/disable the the amp context :) Do you want even more control so you can choose the exact type?
I'm not so experienced with these kinds of low level details, so if possible could you provide a small snippet of what you want?

This is what I have now:
```
with torch.amp.autocast(device_type="cuda", enabled=self.autocast):
    for i in range(x.shape[0]):
        out.append(torch.sparse.mm(self.projection_matrix, x[i, ...]))

```

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,180 @@
...
+        return x
+
+    def _get_shard_shapes(self, x, dim=0, shard_shapes_dim=None, model_comm_group=None):
+        if shard_shapes_dim is None:
+            return get_shard_shapes(x, dim, model_comm_group)
+        else:
+            return apply_shard_shapes(x, dim, shard_shapes_dim)
+
+
+class SparseProjector(nn.Module):
```
JPXKQX: I think the SparseProjector is something very similar to a Mapper. The only difference is that it is not trainable but it can still be also used outside of a residual connection (e.g. in an encoder/decoder). So, I would move it outside of the residual.py. I think we have several options here: a new file? in the mappers.py, perhaps?

JPXKQX: For context, once this PR has been merged, we will probably want to use the `SparseProjector` to define an encoder that performs linear interpolation

models/src/anemoi/models/layers/residual.py
ssmmnn11: maybe similar to this:

        if torch.cuda.is_available():
            with torch.amp.autocast(device_type="cuda", enabled=self.autocast):
                out1 = torch.sparse.mm(A, x)
        else:
            with torch.amp.autocast(device_type="cpu", enabled=self.autocast):
                out1 = torch.sparse.mm(A, x)

I think the check if cuda is available etc. is still required, but maybe there is a better way to do it.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class NoConnection(nn.Module):
+    """No-op connection that returns a zero tensor with the same shape as the last timestep."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+
+    def forward(self, x, *args, **kwargs):
+        x = x[:, -1, ...]  # pick current date
```
ssmmnn11, line 34: maybe "take latest step"? these don't have to be dates.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,180 @@
...
+        # as they can't be broadcast correctly when using DDP.
+        self._to_device(x)
+
+        out = []
+        with torch.amp.autocast(device_type="cuda", enabled=self.autocast):
+            for i in range(x.shape[0]):
+                out.append(torch.sparse.mm(self.projection_matrix, x[i, ...]))
+        return torch.stack(out)
+
+    def _to_device(self, x):
```
ssmmnn11: you can always just do .to(device), if it is on device already, it is a nop - so you don't need this extra logic here.

matschreiner: Oh right. For bigger models you have to be a bit careful because it will loop through all parameters and check each if they are on the right device.. But of course here we only have the matrix, so it's fine :)

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,180 @@
...
+
+def _row_normalize_weights(
+    edge_index,
+    weights,
+    num_target_nodes,
+):
+    total = torch.zeros(num_target_nodes, device=weights.device)
+    norm = total.scatter_add_(0, edge_index[1].long(), weights)
+    norm = norm[edge_index[1]]
+    return weights / (norm + 1e-8)
```
havardhhaugen: Can we allow scaling by (source) node attributes as well? I guess this could be done in the graph creation by creating a new edge attribute which also contains the source node scaling, but I'm not sure how easy this is since node and edge attributes are separated in graph creation. Something like:
```
def _row_normalize_weights(
    edge_index: torch.Tensor,
    edge_attr: torch.Tensor,
    num_target_nodes: int,
    src_node_attr: torch.Tensor = None,
) -> torch.Tensor:
    if src_node_attr is not None:
        weights = edge_attr*src_node_attr[edge_index[0]]
    else:
        weights = edge_attr
    total = torch.zeros(num_target_nodes, device=weights.device)
    norm = total.scatter_add_(0, edge_index[1].long(), weights)
    norm = norm[edge_index[1]]
    return weights / (norm + 1e-8)
```

models/src/anemoi/models/models/encoder_processor_decoder.py
matschreiner: I could, didn't think about it. This should maybe be in a different PR?

models/src/anemoi/models/schemas/models.py
matschreiner: Yep, I have include the residual in all the example config files so it's clear to people that it can be configured, but we don't throw an error if someone does not specify it. 
I also added the documentation of the residual modules in docs under models/layers and I added it to the example config in the documentation. LMK if you want more documentation and also point me to where I should write it. 

Other than that I have adressed all comments in the PR, so I think we are ready to go :)

models/src/anemoi/models/models/encoder_processor_decoder.py
mchantry: If it's not done here, please can you open an issue so we have awareness that it needs doing.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def _to_channel_shards(self, x, grid_shard_shapes=None, model_comm_group=None):
+        return self._reshard(x, shard_channels, grid_shard_shapes, model_comm_group)
+
+    def _to_grid_shards(self, x, grid_shard_shapes=None, model_comm_group=None):
+        return self._reshard(x, gather_channels, grid_shard_shapes, model_comm_group)
+
+    def _reshard(self, x, fn, grid_shard_shapes=None, model_comm_group=None):
+        if grid_shard_shapes is not None:
+            shard_shapes = self._get_shard_shapes(x, 0, grid_shard_shapes, model_comm_group)
```
japols: This breaks with num_gpus_per_model > 1 unfortunately.
You want the shard shapes of the original tensor for both cases, i.e. compute once in fwd via `shard_shapes = self._get_shard_shapes()` and pass them to `self._to_channel/grid_shards` respectively.

japols (commented):
Nice work, @matschreiner!
It would be nice to have an example config with a TruncationMapper residual connection. Also possibly adding the choice to save/reuse truncation matrices using hardware.files/paths.truncation similar to how we can reuse graphs via graph.overwrite.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,127 @@
...
+        up_weight = torch.ones(sub_graph_up.edge_index.shape[1], device=sub_graph_up.edge_index.device)
+        down_weight = torch.ones(sub_graph_down.edge_index.shape[1], device=sub_graph_down.edge_index.device)
+
+        if edge_weight_attribute:
+            up_weight = sub_graph_up[edge_weight_attribute].squeeze() * up_weight
+            down_weight = sub_graph_down[edge_weight_attribute].squeeze() * down_weight
+
+        if src_node_weight_attribute:
+            up_weight = sub_graph_up[src_node_weight_attribute].squeeze() * up_weight
+            down_weight = sub_graph_down[src_node_weight_attribute].squeeze() * down_weight
```
havardhhaugen: Does this work? I thought the node attributes were stored in `graph[node_name][src_node_weight_attribute]`. And the node attribute has to be indexed to get the correct shape and match the order of edges in the weight tensors.

matschreiner: Oh, yeah, right, this was made pretty late in the day, I'm writing tests for it now :)

JPXKQX: In this case, the selection is an edge attribute. There are several options here: 1) Use the `AttributeFromSourceNodes` function implemented in anemoi-graphs to define the edge attribute in the graph, then use this directly as the weight. 2) Stick with a node attribute and select it from the full graph instead of the subgraph.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,127 @@
...
+            autocast=autocast,
+        )
+
+    def forward(self, x, grid_shard_shapes=None, model_comm_group=None, *args, **kwargs):
+        batch_size = x.shape[0]
+        x = x[:, -1, ...]  # pick latest step
+        shard_shapes = self._get_shard_shapes(x, 0, grid_shard_shapes, model_comm_group)
+
+        x = einops.rearrange(x, "batch ensemble grid features -> (batch ensemble) grid features")
+        x = self._to_channel_shards(x, shard_shapes, model_comm_group)
```
japols, line 103: only switch to channel_shards `if grid_shard_shapes is not None` (same for going back to grid shards)

matschreiner: This is actually how it already works - _reshard only reshards if shard_shapes is not None :)

models/src/anemoi/models/layers/residual.py
```diff
@@ -76,8 +76,8 @@ def __init__(
             down_weight = sub_graph_down[edge_weight_attribute].squeeze() * down_weight
 
         if src_node_weight_attribute:
-            up_weight = sub_graph_up[src_node_weight_attribute].squeeze() * up_weight
-            down_weight = sub_graph_down[src_node_weight_attribute].squeeze() * down_weight
+            down_weight = graph[data_nodes][src_node_weight_attribute].squeeze() * down_weight
+            up_weight = graph[truncation_nodes][src_node_weight_attribute].squeeze() * up_weight
```
havardhhaugen: This still looks like it will fail due to shape error (in general len(nodes) != len(edges), probably this should also be the case for the unit test). 
```
down_weight = graph[data_nodes][src_node_weight_attribute].squeeze()[sub_graph_down.edge_index[0]]
up_weight = graph[truncation_nodes][src_node_weight_attribute].squeeze()[sub_graph_up.edge_index[0]]
```

models/src/anemoi/models/layers/residual.py
japols: don't think it does since `self._get_shard_shapes()` always builds them, it's a bit confusing... sry

you  can remove _get_shard_shapes() entirely and do this instead:

```
shard_shapes = apply_shard_shapes(x, 0, grid_shard_shapes) if grid_shard_shapes is not None else None
```

matschreiner: Ah right, ok, I'll use your code bit, thanks :)

models/src/anemoi/models/layers/residual.py
matschreiner: Oh, yeah, of course.. Thanks for catching that! 
Sorry for having to correct it twice :)

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -58,43 +54,37 @@ def __init__(
             Graph definition
         """
         super().__init__()
-        self._graph_data = graph_data
```
mchantry, line 61: @matschreiner why remove this?

models/src/anemoi/models/schemas/models.py
```diff
@@ -222,7 +225,10 @@ class BaseModelSchema(PydanticBaseModel):
         ...,
         discriminator="target_",
     )
-    "GNN decoder schema."
```
mchantry, line 225: Why remove this please @matschreiner ?

training/src/anemoi/training/config/layers/residual_connection.yaml
mchantry, line 1: @matschreiner I am being slow sorry, what's the purpose of this file?

models/src/anemoi/models/models/encoder_processor_decoder.py
matschreiner: The graph is only used for instantiation of the model, so I removed it to reduce the statefulness of the model. :)

models/src/anemoi/models/schemas/models.py
matschreiner: Oh, thanks, no point in removing it, I overwrote it by mistake.

models/src/anemoi/models/models/encoder_processor_decoder.py
mchantry: Yes, I think this was already true before this PR, right?
Thanks for the clarification.

JPXKQX: I think the graph is set as part of the state so it is stored with the weights when a checkpoint is created (not sure who can confirm this)

training/src/anemoi/training/config/layers/residual_connection.yaml
matschreiner: The idea was that the residual configuration should be defined in this file so that users can see how to configure the different residual classes when they open it. The general configuration would then reference this. I just forgot to link the two together. I don't know if this is a good idea though.

mchantry: Shouldn't this be in documentation? And then users can add it to a model config yaml?

models/tests/models/test_encprocdec.py
mchantry: @matschreiner can you delete this file please?

models/src/anemoi/models/schemas/residual.py
```diff
@@ -0,0 +1,36 @@
...
+from anemoi.utils.schemas import BaseModel
+
+
+class SkipConnectionSchema(BaseModel):
+    target_: Literal["anemoi.models.layers.residual.SkipConnection"] = Field(
+        default="anemoi.models.layers.residual.SkipConnection", alias="_target_"
+    )
+
+
+class NoConnectionSchema(BaseModel):
```
theissenhelen, line 14: If these two schemas are exactly the same and only differ in the target value, it may be beneficial to have one schema covering bother options instead.

Issue #446, opened 2025-07-31
Pytorch geometric doesn't '|' type hints


anaprietonem:
### What happened?

When dropping 3.9, we updated type hints across the packages and found out that pytorch-geomtric introduced an inspector class in the newer versions that crashes when '|' is used. 

But even str | None - geometric complaints AttributeError: 'types.UnionType' object has no attribute '__qualname__'.  The pipe creates a types.UnionType and that doesn't have __qualname__ so the inspector crashes

see https://github.com/pyg-team/pytorch_geometric/issues/10138

### What are the steps to reproduce the bug?

```

from torch_geometric.nn import MessagePassing

class Tmp(MessagePassing):
    def __init__(self, in_channels: int, out_channels: int) -> None:
        super().__init__(aggr='add')
    def forward(self, x: Tensor, edge_index: Tensor) -> None:
        ...
        return None        

    def message(self, x_j: Tensor, edge_attr: Tensor | None = None) -> None:
        return None

tmp = Tmp(123, 123)

```

### Version

Latest version

### Platform (OS and architecture)

ATOS

### Relevant log output

```shell

```

PR #461, opened 2025-08-05
Feature/postprocessor for nans in diagnostic fields

sahahner:
## Description
Place NaNs in the output of diagnostic variables at locations of NaNs of another variable.

The new logic of the imputer takes the NaN locations from the input and places them in the output. This is not possible for diagnostic variables as they are not present in the input.

### DIFF
--- a/models/src/anemoi/models/preprocessing/postprocessor.py
+++ b/models/src/anemoi/models/preprocessing/postprocessor.py
@@ -202,23 +202,11 @@ class NormalizedReluPostprocessor(Postprocessor):
         return postprocessor_function
 
 
-class ConditionalZeroPostprocessor(Postprocessor):
-    """Sets values to specified value where another variable is zero.
+class ConditionalPostprocessor(Postprocessor):
+    """Base class for postprocessors that conditionally apply a transformation based on another variable.
 
-    Expects the config to have keys corresponding to customizable values and lists of variables to postprocess and a variable to use for postprocessing.:
-
-    ```
-    default: "none"
-    remap: "x"
-    0:
-        - y
-    5.0:
-        - x
-    3.14:
-        - q
-    ```
-
-    If "x" is zero, "y" will be postprocessed with 0, "x" with 5.0 and "q" with 3.14.
+    This class is intended to be subclassed for specific implementations.
+    It expects the config to have keys corresponding to customizable values and lists of variables to postprocess.
     """
 
     def __init__(
@@ -243,24 +231,9 @@ class ConditionalZeroPostprocessor(Postprocessor):
             self.masking_variable, None
         )
 
-    def _get_postprocessor_function(self, method: float, name: str):
-        """For ConditionalZeroPostprocessor, the 'method' is the constant value to fill
-        when the masking variable is zero. This function simply returns the value.
-        """
-        LOGGER.info(
-            f"ConditionalZeroPostprocessor: replacing valus in {name} with value {method} if {self.masking_variable} is zero."
-        )
-        return method
-
     def _expand_subset_mask(self, x: torch.Tensor, mask: torch.tensor) -> torch.Tensor:
         """Expand the subset of the mask to the correct shape."""
         return mask.expand(*x.shape[:-2], -1)
-
-    def get_zeros(self, x: torch.Tensor) -> torch.Tensor:
-        """Get zero mask from data"""
-        # The mask is only saved for the last dimension (grid)
-        idx = [slice(0, 1)] * (x.ndim - 2) + [slice(None), slice(None)]
-        return (x[idx] == 0).squeeze()
 
     def fill_with_value(self, x: torch.Tensor, index: list[int], fill_mask: torch.tensor):
         for idx_dst, value in zip(index, self.postprocessorfunctions):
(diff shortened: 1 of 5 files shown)


models/src/anemoi/models/interface/__init__.py
```diff
@@ -140,6 +139,8 @@ def predict_step(
                 grid_shard_shapes = [shape[-2] for shape in shard_shapes]
                 x = shard_tensor(x, -2, shard_shapes, model_comm_group)
 
+            x = self.pre_processors(x, in_place=False)
```
mchantry: Are there any significant memory impacts from this?

mchantry: Why has this moved?

sahahner: This has moved already in the base branch. Both pre- and postprocessors need to be sharded or not sharded

sahahner:
Without the postprocessor there are no NaNs in diagnostic fields (for example, the model predicts a diagnostic sea ice variable over land)
gnn_pred_val_sample_rstep00_batch0000_rank0_epoch000(1)

With the postprocessors one can copy NaN values from another variable to the diagnostic one
gnn_pred_val_sample_rstep00_batch0000_rank0_epoch001

sahahner:
> @sahahner great work, but should there be a little documentation on how to use?

I added a paragraph to the documentation on how to use the new ConditionalNaNPostprocessor in combination with an imputer for diagnostic variables. How it needs to be configured in the config file is stated in the docstring of the class itself. 

I would like to leave the class docstring general, as the ConditionalNaNPostprocessor can also be used without an imputer.

mchantry (approved):
LGTM, thanks.

models/docs/modules/preprocessing.rst
```diff
@@ -50,10 +50,14 @@ mask is updated for every batch during training.
...
-batch, also for input timesteps. Therefore, any NaNs in the target data
-are imputed to enable proper loss computation.
+result, no NaNs are reintroduced into the diagnostic output fields by
+the imputer. In contrast, during training, diagnostic variables are
+included in the batch, also for input timesteps. Therefore, any NaNs in
+the target data are weighted by zero to enable proper loss computation.
+For introducing NaNs to diagnostic variables, the
+``anemoi.models.preprocessing.postprocessor.ConditionalNaNPostprocessor``
+class can be used. It replaces values in the diagnostic variables with
+NaN if a specified masking variable is NaN.
```
NRaoult: I know I reviewed this a bit already, but reading through now, I think it's better to start with training and move on to inference, instead of the training being in the middle. Also, the last sentence could be clearer. And do we need to say here that the chosen masking variable must be a prognostic?

Maybe something like:
"During training, diagnostic variables are included in each batch, and therefore at the input timesteps. Any NaNs in the target data are weighted by zero to enable proper loss computation. During inference, however, NaN locations for diagnostic variables are not available—those fields aren’t part of the model input—so the imputer never reintroduces NaNs into the diagnostic output. To insert NaNs into diagnostic variables, the "anemoi.models.preprocessing.postprocessor.ConditionalNaNPostprocessor" class used be used. This masks diagnostic variable entries by setting them to NaN wherever the chosen masking variable is NaN.

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -286,7 +259,73 @@ def inverse_transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Ten
                 f"({self.num_training_output_vars}) or inference shape ({self.num_inference_output_vars})",
             )
 
-        zero_mask = self.get_zeros(x[..., masking_variable])
+        postprocessor_mask = self.get_locations(x[..., masking_variable])
```
HCookie, line 272: `self.get_locations` should be mocked an abstractmethod

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -286,7 +259,73 @@ def inverse_transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Ten
...
+
+    The module sets "y" NaN, at NaN locations of "x".
+    """
+
+    def _get_postprocessor_function(self, method: float, name: str):
+        """For ConditionalNaNPostprocessor, the 'method' is a NaN to fill
+        when the masking variable is NaN. This function simply returns a NaN.
+        """
+        LOGGER.info(
+            f"ConditionalNaNPostprocessor: replacing valus in {name} with value NaN if {self.masking_variable} is NaN."
```
HCookie:
```suggestion
            f"ConditionalNaNPostprocessor: replacing values in {name} with value NaN if {self.masking_variable} is NaN."
```

models/src/anemoi/models/schemas/data_processor.py
```diff
@@ -42,8 +42,9 @@ class NormalizerSchema(BaseModel):
 class ImputerSchema(BaseModel):
     default: str = Field(literals=["none", "mean", "stdev"])
     "Imputer default method to apply."
-    maximum: Union[list[str], None]
-    minimum: Union[list[str], None]
+    maximum: Union[list[str], None] = Field(default_factory=list)
+    minimum: Union[list[str], None] = Field(default_factory=list)
+    mean: Union[list[str], None] = Field(default_factory=list)
```
HCookie, line 47: 3.10+ syntax can be used here

models/src/anemoi/models/preprocessing/postprocessor.py
sahahner: oh yes, you are right! good that you spotted that.

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -286,7 +267,73 @@ def inverse_transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Ten
...
+        """
+        LOGGER.info(
+            f"ConditionalNaNPostprocessor: replacing values in {name} with value NaN if {self.masking_variable} is NaN."
+        )
+        return torch.nan
+
+    def get_locations(self, x: torch.Tensor) -> torch.Tensor:
+        """Get NaN mask from data"""
+        # The mask is only saved for the last dimension (grid)
+        idx = [slice(0, 1)] * (x.ndim - 2) + [slice(None), slice(None)]
```
pinnstorm: Just to make sure I understand the indexing here, you are basically selecting the 0 index for all dimensions other than the last two which will be (grid, variable)?

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -286,7 +267,73 @@ def inverse_transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Ten
...
+        return (x[idx] == 0).squeeze()
+
+
+class ConditionalNaNPostprocessor(ConditionalPostprocessor):
+    """Sets values to NaNs where another variable is NaN.
+
+    Expects the config to have list of variables to postprocess and a variable to use for postprocessing.:
+
+    ```
+    default: "none"
```
pinnstorm, line 320: Is it worth adding a very brief section in the training/docs/user-guide/training.rst docs to describe the Postprocessor? Potentially not necessary right now

pinnstorm (approved):
LGTM! I just left a couple of questions but seems to work well for land from @NRaoult results too 🙌

NRaoult (approved):

NRaoult (dismissed):
Tested with some land variables (snow depth prognostic, snow cover diagnostic). All works well!!!

Top **with** postprocesser, bottom **without** postprocesser
Screenshot 2025-08-07 at 08 29 45
Screenshot 2025-08-07 at 08 31 01
All LGTM 👍

models/docs/modules/preprocessing.rst
sahahner: I have also added that the chosen variable should be prognostic. Nevertheless, I do not want to add an assert in the code, as some day we might have another way to add NaNs to output fields of diagnostic variables, that could then be copied.

models/src/anemoi/models/schemas/data_processor.py
anaprietonem: torch-geometric has issues with the | notation - https://github.com/ecmwf/anemoi-core/issues/446. In this file there shouldn't be an issue (since the schemas for pydantic and test passed [okey](https://github.com/ecmwf/anemoi-core/actions/runs/16806647310)) but I think until we figure out further/understand the issue better, we should keep type hints as they are in models. That's what we had to do when dropping 3.9

pinnstorm (approved):
LGTM 🚀

models/src/anemoi/models/preprocessing/postprocessor.py
sahahner: Good questions. I noticed that it would be better to cover all dimensions (but the variable one) by this mask to allow here for changing NaN locations over time as well.

models/src/anemoi/models/preprocessing/postprocessor.py
```diff
@@ -286,7 +269,73 @@ def inverse_transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Ten
...
+        when the masking variable is NaN. This function simply returns a NaN.
+        """
+        LOGGER.info(
+            f"ConditionalNaNPostprocessor: replacing values in {name} with value NaN if {self.masking_variable} is NaN."
+        )
+        return torch.nan
+
+    def get_locations(self, x: torch.Tensor) -> torch.Tensor:
+        """Get NaN mask from data"""
+        # reference/masking variable is already selected. Mask covers all remaining dimensions.
```
pinnstorm, line 340: Nice!

pinnstorm (approved):

PR #422, opened 2025-07-15
fix: pydantic model validator not working in transformer schema

theissenhelen:
## Description
The model validator in the transformer schema checking the allowed extra config entries was not working as expected. 

## What problem does this change solve?
The model validator now correctly checks the type of the allowed extra config entries

##  Additional notes ##

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

Issue #410, opened 2025-07-09
Model Transformation Layer - Post-loading modifications (freezing, transfer learning, adapters)


JesperDramsch:
## Overview
During my work towards #248 I have converged on a solution that requires a refactor of the model initialisation.

**UPDATE**: This issue is now part of the 5-PR checkpoint architecture refactor. It implements the **Model Transformation Layer** that applies post-loading modifications to models.

## Parent Issue
Part of #248 - Checkpoint System Refactor

## Related Architecture Components
- Foundation: #493 (Pipeline Infrastructure)
- Works with: #458 (Checkpoint Acquisition), #494 (Loading Orchestration)
- Integration: #495 (System Integration)

## Original Problem
Specifically in https://github.com/ecmwf/anemoi-core/blob/d51219f0ef598a149dc2c7901b055947785a26b2/training/src/anemoi/training/train/train.py#L217

We have a lot of branching that is getting increasingly difficult to extend and maintain.

## New Architecture Role
This issue implements the **Model Transformation Layer** - the third layer in our pipeline architecture:

```
┌─────────────────────────────────────────────────┐
│    Model Transformation Layer (THIS ISSUE)      │
│         (Post-loading modifications)            │
├─────────────────────────────────────────────────┤
│         Loading Orchestration Layer             │
│    (Strategies for applying checkpoints)        │
├─────────────────────────────────────────────────┤
│        Checkpoint Acquisition Layer             │
│      (Obtaining checkpoint from sources)        │
└─────────────────────────────────────────────────┘
```

### 1. ModelModifier Base Class (`training/src/anemoi/training/train/modify.py`)
```python
class ModelModifier(ABC):
    """Base class for all model modifiers"""
    @abstractmethod
    def apply(self, model: nn.Module) -> nn.Module:
        """Apply modification to model"""
        pass
```

### 2. Implemented Modifiers
- `FreezingModelModifier`: Selective parameter freezing by module name
- `TransferLearningModelModifier`: Transfer learning with fallback for missing checkpoint system
- Future: `LoRAModifier`, `QuantizationModifier`, etc.

### 3. ModelModifierApplier
Orchestrates application of multiple modifiers in sequence:
```python
class ModelModifierApplier:
    def process(self, base_model: nn.Module, config: DictConfig) -> nn.Module
```

## Configuration
```yaml
training:
  model_modifier:
    modifiers:
      - _target_: "anemoi.training.train.modify.FreezingModelModifier"
[...]

Issue #248, opened 2025-04-10
Anemoi [Fine-tuning, Transfer Learning, Model Freezing] Roadmap


JesperDramsch:
During 2025, I plan to work on the fine-tuning of Anemoi models. This issue should work as a roadmap and info-dump, as well as a discussion ground for different users.

## Current State

The current state of fine-tuning in Anemoi revolves around re-training the entire model.

While this makes sense in pre-training the dynamics with single-step rollout and then finalizing the training with multi-step rollout, it is disadvantageous when we want to fine-tune on data with different distributions. This is the `naive fine-tuning` approach:

1. Train model $`M`$ with weights $`W`$ on dataset (or task) $`A`$
2. Train model $`M'`$ modifying the weights $`W`$ to $`W'`$ with a lower learning rate (and other tricks to make it work)

The disadvantages here are that the extensive pre-training of the model often captures nuances that may be "forgotten" by retraining the model. This has been observed in the training process and often leads to a lot of "finicky" training set-ups.

### Implementation details

This is currently implemented as warm-starts and forking of the training, is implemented in multiple config entries in `training`:

```yaml
# resume or fork a training from a checkpoint last.ckpt or specified in hardware.files.warm_start
run_id: null
fork_run_id: null
transfer_learning: False # activate to perform transfer learning
load_weights_only: False # only load model weights, do not restore optimiser states etc.
```
cf: https://github.com/ecmwf/anemoi-core/blob/d02e0bb2a9bf1700de90c8d862f80a510f03eceb/training/src/anemoi/training/config/training/default.yaml#L3-L7

This has implications for traceability and the automation of training pipelines. But it also has certain drawbacks: the forking of a run has to be specified by run_id that is expected to be on the same system, which discourages collaboration and sharing of checkpoints. These were design choices made early in the project before we anticipated the breadth of adoption.

## Proposed Solution

To address these limitations, I propose implementing a comprehensive fine-tuning capability in Anemoi that includes:

### 1. Enhanced Model Freezing

Building on PR #61, we need to extend the submodule freezing functionality to enable more granular control over which parts of the model are trainable during fine-tuning. This will allow:

- Freezing arbitrary submodules at multiple levels of the model hierarchy
- Partial freezing of specific parameter groups
- Possibly even differential learning rates for different model components

### 2. Integration with PEFT Library

[...]

PR #442, opened 2025-07-29
feat: Model Transformation Layer — Native PipelineStage modifiers (#410)

JesperDramsch:
**Part of the checkpoint system refactor (#248).** Implements the Model Transformation Layer as native `PipelineStage` classes built on Phase 1 infrastructure (PR #501).

> ⚠️ **Architecture change from original PR.** The original `train/modify.py` (sync `ModelModifier.apply(model)` interface) has been **deleted and replaced** with native async `PipelineStage` implementations in `checkpoint/modifiers/`. See design decisions below.

### What this PR does

- **`FreezingModifierStage`** — native `PipelineStage` that freezes specified submodules. Full feature port from the original `FreezingModelModifier`: strict mode, gradient validation, parameter counting, dot-notation submodule lookup.
- **Deletes `train/modify.py`** — `ModelModifierApplier`, `FreezingModelModifier`, and `TransferLearningModelModifier` are all removed. Transfer learning now lives in the loading layer (#998). Pipeline handles orchestration.
- **Config templates** for all 4 loading strategies (`weights_only.yaml`, `transfer_learning.yaml`, `warm_start.yaml`, `cold_start.yaml`)
- **Package exports** and catalog discovery for the modifier layer

### Architecture

```
CheckpointContext enters with model (weights already loaded by loading strategy)
        ↓
  FreezingModifierStage freezes specified submodules
        ↓
CheckpointContext exits with model modified + metadata["modifiers_applied"] updated
```

Composes naturally with sources and loaders:
```python
pipeline = CheckpointPipeline(stages=[
    LocalSource(),
    TransferLearningLoader(skip_mismatched=True),
    FreezingModifierStage(submodules_to_freeze=["encoder"]),
])
```

### Why `train/modify.py` was deleted (not wrapped)

The original plan was thin adapter wrappers around the existing sync classes. That turned out to be wrong:

- PR #410's code was prototype, not production-ready
- Wrapping sync `apply(model)` inside async `process(context)` creates dual code paths
- `ModelModifierApplier` duplicated pipeline orchestration that Phase 1 already handles
- `TransferLearningModelModifier` mixed loading + modification concerns — these are now separate layers

Building native `PipelineStage` classes is cleaner: no indirection, single code path, clear separation of concerns.

### Current status

🟢 **Ready for review.** `main` is merged in, so the diff is now scoped to the modifier delta.

- ✅ g1: FreezingModifierStage native port + train/modify.py deleted
- ✅ g2: Config templates
[...]

PR #464, opened 2025-08-06
feat: Checkpoint Acquisition Layer — Local, HTTP, S3 sources (#458)

### LINKED ISSUE #458: Refactor restart of training [warm start, forked runs, restarts]
As part of #248 I mentioned a possible refactor of how we resume runs. On #442 @JPXKQX mentioned the following:

> Thanks Jesper, I think this refactoring makes a lot of sense. Would it make sense to have another modifier "ResumeRun..." to which we could bring all the logic from  `run_id`, `fork_run_id`, `load_only_weights` and `warm_start`?

_Originally posted by @JPXKQX in https://github.com/ecmwf/anemoi-core/issues/442#issuecomment-3140014167_
            

I believe this could be a good opportunity to make the code more intuitive and also reduce the direct dependency on MLFlow run_ids. (Still think it's an easy way to resume, but we can add flexibility to add checkpoints from anywhere.)

JesperDramsch:
**Part of the checkpoint system refactor (#248).** Implements the Checkpoint Acquisition Layer as native `PipelineStage` classes built on the Phase 1 infrastructure (PR #501).

> ⚠️ **Complete reimplementation.** The original PR #464 code (sync `CheckpointSource.fetch()` interface) has been replaced with a fresh implementation using Phase 1's async `PipelineStage.process(context)` pattern. The old review findings are tracked

### What this PR does

Adds three checkpoint source implementations that fetch checkpoints from different backends and populate `CheckpointContext.checkpoint_data` for downstream loading strategies:

| Source | What it does | Key dependency |
|--------|-------------|----------------|
| `LocalSource` | Loads from local filesystem | `torch.load()` + Phase 1 `detect_checkpoint_format()` |
| `HTTPSource` | Async download with retry | Phase 1 `download_with_retry()` (aiohttp, exponential backoff) |
| `S3Source` | Download from S3-compatible storage | `boto3` (lazy import, credential delegation) |

All three extend `CheckpointSource(PipelineStage)` and implement `async process(context) -> context`.

### Architecture

```
CheckpointContext enters with checkpoint_path or config["url"]
        ↓
  Source stage loads/downloads checkpoint
        ↓
CheckpointContext exits with checkpoint_data populated + checkpoint_format detected
```

Data flows through `CheckpointContext`, not return values. Each source is a standalone pipeline stage — no shared dispatcher, no registry routing.

### Design decisions

- **Reimplementation, not adaptation**: PR #458's download logic is 5–15 lines per source. The registries and class hierarchy from #458 are replaced by Phase 1's `ComponentCatalog` and `CheckpointPipeline`. Wrapping 600-line classes to reuse 15 lines would add needless indirection.
- **Local + HTTP + S3 only**: GCS and Azure deferred per review feedback (Ana: remove cloud provider refs). Trivial to add later — ~10 lines each, same pattern.
- **Phase 1 error types everywhere**: `CheckpointNotFoundError`, `CheckpointSourceError` — never stdlib exceptions.
- **S3 credential delegation**: Entirely boto3's default chain. Works across AWS, EWC, MinIO, Ceph without code changes.
- **S3 error mapping**: Every boto3 error code mapped to actionable messages with Phase 1 exception types.
- **Lazy imports**: `boto3` imported inside method, not at module level. `ImportError` includes pip install command.

### What's NOT in this PR

| Deferred | Why | Tracked in |
|----------|-----|-----------|
| GCS source | Ana review feedback — remove cloud refs | Follow-up PR |
| Azure source | Same | Follow-up PR |
| MLFlow source | Phase 3 backwards compat | #495 |
| Checkpoint caching | Architecture supports it, not needed yet | Follow-up PR |
| Progress reporting | Phase 1 has hooks for it | Follow-up PR |
| safetensors format | Phase 1 enum doesn't include it | Follow-up issue |

### Test coverage

151+ tests passing, 5 skipped (network-dependent, gated by `ANEMOI_TEST_NETWORK`):

```
training/tests/checkpoint/sources/
├── test_base.py      — 3 tests (inheritance, abstract, interface)
├── test_local.py     — ~30 tests (load, format detect, errors, edge cases)
├── test_http.py      — ~30 tests (download_with_retry usage, error wrapping)
└── test_s3.py        — ~30 tests (URL parsing, error mapping, lazy import, cleanup)
```

Test philosophy: real-behavior tests for LocalSource (zero mocks, actual torch.save/load), mock-only for network sources (boto3, aiohttp).

### How to review

```bash
# Run the full source test suite
pytest training/tests/checkpoint/sources/ -v --tb=short

# Check no Phase 1 files were modified
git diff main -- training/src/anemoi/training/checkpoint/base.py \
  training/src/anemoi/training/checkpoint/pipeline.py \
  training/src/anemoi/training/checkpoint/exceptions.py
```

### Try it yourself

```bash
git checkout feature/checkpoint-sources
cd training
pip install -e .

# Load a local checkpoint through the pipeline
python -c "
import asyncio, torch
from pathlib import Path
from anemoi.training.checkpoint.base import CheckpointContext
from anemoi.training.checkpoint.sources.local import LocalSource

# Create a dummy checkpoint
torch.save({'state_dict': {'layer.weight': torch.randn(10, 5)}, 'epoch': 3}, '/tmp/test.ckpt')

# Run it through the source stage
source = LocalSource()
ctx = CheckpointContext(checkpoint_path=Path('/tmp/test.ckpt'))
result = asyncio.run(source.process(ctx))

print(f'Format:  {result.checkpoint_format}')
print(f'Keys:    {list(result.checkpoint_data.keys())}')
print(f'Weights: {result.checkpoint_data[\"state_dict\"][\"layer.weight\"].shape}')
"
```

For S3 (if you have credentials):
```bash
python -c "
import asyncio
from anemoi.training.checkpoint.base import CheckpointContext
from anemoi.training.checkpoint.sources.s3 import S3Source

source = S3Source()
ctx = CheckpointContext(config={'url': 's3://your-bucket/your-checkpoint.ckpt'})
result = asyncio.run(source.process(ctx))
print(f'Loaded {len(result.checkpoint_data[\"state_dict\"])} parameters from S3')
"
```

### Phase 1 note

6 Phase 1 files have bugfixes on this branch (283 insertions, 137 deletions). Key dependency: `formats.py:detect_format_from_data()` is called by `sources/base.py`. These fixes are also cherry-picked onto the loading-orchestration branch. Pragmatic approach: include fixes in whichever PR merges first, the other gets a clean rebase.

**Closes #458**
**Part of #248**

---

<details>
<summary>Original PR description (archived 2026-03-20)</summary>

## Description

This PR introduces a comprehensive checkpoint loading system that supports loading model checkpoints from various sources including local files, S3, HTTP/HTTPS, Google Cloud Storage, and Azure Blob Storage. The implementation provides a modular, extensible architecture that separates checkpoint retrieval from model loading strategies and includes robust error handling and validation.

This change addresses several key problems in the current checkpoint loading workflow:

**New Feature**: Adds multi-source checkpoint loading capabilities with support for:
- **Local filesystem**: Standard file path loading
- **HTTP/HTTPS**: Download from web URLs  
- **Amazon S3**: S3 bucket support with boto3
- **Google Cloud Storage**: GCS support with google-cloud-storage
- **Azure Blob Storage**: Azure support with azure-storage-blob

**Enhanced Model Loading Strategies**:
- **StandardModelLoader**: Standard PyTorch Lightning checkpoint loading
- **TransferLearningModelLoader**: Handle size mismatches and partial weight loading
- **WeightsOnlyModelLoader**: Load only model weights, skip optimiser states

**Architecture Improvements**:
- **Separation of concerns**: Checkpoint retrieval vs model loading logic
- **Extensibility**: Easy to add new sources or loading strategies via registry pattern
- **Testability**: Mock-friendly design with clear interfaces
- **Error handling**: Informative error messages and graceful failures
- **Performance**: Efficient temporary file handling and cleanup

**Closes #458**

Should be updated and merged after #410 / #422 

### Implementation Details

The PR introduces two main components:

**CheckpointLoaders (`checkpoint_loaders.py`)**:
- Abstract base class `CheckpointLoader` with pluggable implementations
- `LocalCheckpointLoader` for filesystem access
- `RemoteCheckpointLoader` with multi-cloud support
- Automatic source detection and loader selection
- Comprehensive error handling with clear messages

**ModelLoading (`model_loading.py`)**:
- Abstract base class `ModelLoader` for different loading strategies
- Standard Lightning checkpoint compatibility
- Transfer learning with size mismatch handling
- Metadata preservation and data_indices validation
- Registry pattern for extensibility

### Usage Examples

```python
# Load from any supported source
checkpoint = load_checkpoint_from_source("s3://bucket/model.ckpt")
checkpoint = load_checkpoint_from_source("https://example.com/model.ckpt")
checkpoint = load_checkpoint_from_source("/local/path/model.ckpt")

# Load model with transfer learning
model = load_model_from_checkpoint(
    model=my_model,
    checkpoint_source="s3://bucket/pretrained.ckpt",
    loader_type="transfer_learning",
    skip_mismatched=True
)

# Register custom loaders
registry.register("custom_source", CustomCheckpointLoader())
registry.register("custom_strategy", CustomModelLoader())
```

### Testing and Quality Assurance

- **100% line coverage** across all loaders and strategies
- **Integration tests** for multi-source scenarios
- **Error handling tests** for edge cases and failures
- **Mock-based testing** to avoid external dependencies

```bash
# Run all checkpoint loading tests
pytest training/tests/utils/test_checkpoint_loaders.py -v
pytest training/tests/utils/test_model_loading.py -v

# Test with different sources (requires cloud credentials)
pytest training/tests/utils/ -k "remote" --cloud-tests
```

### Compatibility and Dependencies

- **Dependencies**: Optional cloud dependencies (boto3, google-cloud-storage, azure-storage-blob)
- **Type safety**: Full type annotations and strict typing
- **Code quality**: Passes all linting rules (ruff, black, isort)

### Future Extensions

This architecture makes it easy to add:
- Additional cloud providers (OCI, DigitalOcean Spaces, etc.)
- Custom authentication mechanisms
- Caching and optimisation strategies
- Checkpoint validation and integrity checks
- Progress tracking for large downloads

</details>

----
📚 Documentation preview 📚: https://anemoi-training--464.org.readthedocs.build/en/464/

----
📚 Documentation preview 📚: https://anemoi-graphs--464.org.readthedocs.build/en/464/

----
📚 Documentation preview 📚: https://anemoi-models--464.org.readthedocs.build/en/464/


JesperDramsch:
Yes exactly, so it's much easier to say "hey I have this checkpoint in an s3 bucket" or similar.

mchantry:
@JesperDramsch thanks for the contribution. Is there an active use-case for the GCS loading? There is an active user-base for azure, so I think implementing the azure one would be a nice addition, but wonder if you could save yourself some work by leaving GCS for when a use-case arises?

training/pyproject.toml
```diff
@@ -91,13 +91,16 @@ optional-dependencies.profile = [
 optional-dependencies.remote = [
   "aiohttp>=3.8",
 ]
+optional-dependencies.s3 = [
+  "boto3>=1.28",
```
anaprietonem: In anemoi-utils, we rely on obstore do access s3 objects. This is then used by datasets. https://github.com/ecmwf/anemoi-utils/blob/2baa4812a4b3a186c7f39d6bb04ed0010ae15717/pyproject.toml#L63. I think it could be good to harmonise that

training/tests/unit/checkpoint/sources/test_s3.py
```diff
@@ -0,0 +1,83 @@
...
+        importlib.reload(importlib.import_module("anemoi.training.checkpoint.sources.s3"))
+    finally:
+        if saved:
+            sys.modules["boto3"] = saved
+
+
+@pytest.mark.asyncio
+async def test_s3_source_parses_s3_url() -> None:
+    """s3://bucket/path/to/model.ckpt -> bucket='bucket', key='path/to/model.ckpt'."""
+    source = S3Source(url="s3://my-bucket/checkpoints/model.ckpt")
```
anaprietonem, line 53: could we consider using the s3 bucket where we have other test data for the integration tests to check this or we prefer any s3 reading to be outside of the unit-tests?

training/tests/checkpoint/sources/conftest.py
```diff
@@ -0,0 +1,15 @@
+"""Source-specific test fixtures."""
+
+from pathlib import Path
+
+import pytest
+import torch
+
+
+@pytest.fixture
+def sample_checkpoint(tmp_path: Path) -> Path:
```
anaprietonem: this could be useful for other tests, so could be  just added it as part of the already existing conftest.py

training/src/anemoi/training/checkpoint/formats.py
```diff
@@ -518,17 +544,32 @@ def convert_lightning_to_pytorch(
     return pytorch_checkpoint
 
 
-def is_format_available(checkpoint_format: Literal["lightning", "pytorch", "state_dict"]) -> bool:  # noqa: ARG001
+def is_format_available(checkpoint_format: Literal["lightning", "pytorch", "safetensors", "state_dict"]) -> bool:
```
anaprietonem: It's nice to see safetenors could be easily considered here. But since that package is not yet considered or used in anemoi, I suggest any references to it should be removed from this PR.

training/src/anemoi/training/checkpoint/pipeline.py
```diff
@@ -39,7 +39,7 @@
 ...     'stages': [
 ...         {'_target_': 'anemoi.training.checkpoint.sources.LocalSource',
 ...          'path': '/tmp/checkpoint.pt'},
-...         {'_target_': 'anemoi.training.checkpoint.loaders.WeightsOnlyLoader',
+...         {'_target_': 'anemoi.training.checkpoint.loading.LoadingStrategy',
```
anaprietonem: I assume this is coming from the changes in the loaders, but this won't work on this PR?

training/src/anemoi/training/checkpoint/formats.py
```diff
@@ -26,12 +26,68 @@
...
+    Returns
+    -------
+    str
+        Format of the checkpoint: "lightning", "pytorch", or "state_dict"
+    """
+    if not isinstance(checkpoint_data, dict):
+        return "pytorch"
+
+    # Check for Lightning-specific keys (exclude generic training state keys)
+    lightning_specific_keys = {
```
anaprietonem: how robust is this to changes in PTL or torch?

training/src/anemoi/training/checkpoint/catalog.py
```diff
@@ -285,7 +285,7 @@ def _get_sources(cls) -> dict[str, str]:
     def _get_loaders(cls) -> dict[str, str]:
         """Get the registry of loaders, discovering if needed."""
         if cls._loaders is None:
-            cls._loaders = cls._discover_components("anemoi.training.checkpoint.loaders", "LoadingStrategy")
+            cls._loaders = cls._discover_components("anemoi.training.checkpoint.loading", "LoadingStrategy")
```
anaprietonem, line 296: similar comment as https://github.com/ecmwf/anemoi-core/pull/464/changes#r3008039354.

training/src/anemoi/training/checkpoint/__init__.py
```diff
@@ -46,8 +46,17 @@
 try:
     from .validation import validate_pipeline_health
```
anaprietonem: is this working? and what's the `validate_pipeline_health` checking?

training/src/anemoi/training/checkpoint/sources/s3.py
```diff
@@ -0,0 +1,271 @@
...
+
+from anemoi.training.checkpoint.sources.base import CheckpointSource
+
+if TYPE_CHECKING:
+    from anemoi.training.checkpoint.base import CheckpointContext
+
+LOGGER = logging.getLogger(__name__)
+
+
+class S3Source(CheckpointSource):
```
anaprietonem, line 46: Similar to the comment about the pyproject. A lot of this functionality doesn't seem very related or dependant of training such as settings up s3 client with boto3. We already have some of this with obstore. And potentially these sources could be considered at inference time to load the checkpoints from a bucket. So I think, first would be good to understand if obstore could be used here and then how much of these functionality could be moved or imported from utils.

training/src/anemoi/training/checkpoint/sources/base.py
```diff
@@ -0,0 +1,129 @@
...
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Abstract base class for checkpoint sources.
+
+Checkpoint sources are responsible for acquiring checkpoint data from
+various locations (local filesystem, S3, HTTP, etc.) and populating
```
anaprietonem, line 13: how are those sources identified? will the user needs to define/instantiate the required in the config or are we inferring it from the checkpoint file path used? I would imagine the latter being easier

anaprietonem (changes requested):
Nice work Jesper, very nice to start having this functionality! left some comments for changes and some questions to clarify

training/tests/unit/checkpoint/sources/test_s3.py
JesperDramsch: That would be great actually.

training/pyproject.toml
JesperDramsch: I will check. Best to re-use dependencies in my opinion.

training/src/anemoi/training/checkpoint/formats.py
JesperDramsch: Oh yeah valid, this slipped in when I copied my old prototype, sorry.

training/src/anemoi/training/checkpoint/sources/base.py
JesperDramsch: It will have to be user-supplied.

training/pyproject.toml
JesperDramsch: I have now completely switched the source to use the implementation in `anemoi-utils` thank you.

JesperDramsch: Fixed in 1c4761611 (and the underlying boto3→obstore swap in fa4095953). `[s3]` extra now declares `anemoi-utils[s3]>=0.4.37`, which pulls obstore. End-to-end verified against the public NOAA S3 bucket.

training/tests/unit/checkpoint/sources/test_s3.py
JesperDramsch: Added as `test_s3_source_downloads_from_real_public_bucket` in 1c4761611 — a `@pytest.mark.network`-gated test that downloads from the public NOAA GHCN bucket via the real `anemoi.utils.remote.s3.download_file` path. Skips cleanly when `obstore` or `anemoi-utils[s3]` are not installed. If there is an internal team bucket you would prefer over the public NOAA one, happy to swap it in.

training/tests/checkpoint/sources/conftest.py
JesperDramsch: Done in 1c4761611. `sample_checkpoint` now lives in `training/tests/checkpoint/conftest.py` so other layers can reuse it. The single-fixture `sources/conftest.py` was removed.

training/src/anemoi/training/checkpoint/formats.py
JesperDramsch: Done in 1c4761611. All safetensors references removed from `formats.py` (and the `Literal[...]` type alias no longer includes it). Validation utilities also no longer probe for the safetensors module.

training/src/anemoi/training/checkpoint/pipeline.py
JesperDramsch: Done in 1c4761611. The `from_config` docstring example no longer references `LoadingStrategy` from the `loading` package (which is not present on this PR's branch).

training/src/anemoi/training/checkpoint/formats.py
JesperDramsch: Hardened in 1c4761611. The `pytorch-lightning_version` key is now the primary marker, and the auxiliary key set falls back to a 2+ match so detection survives a single Lightning rename. Added a docstring note about the tested PTL/torch range. Existing format tests still pass.

training/src/anemoi/training/checkpoint/catalog.py
JesperDramsch: No code change, as the design is intentional. `_get_loaders` calls `_discover_components("anemoi.training.checkpoint.loading", "LoadingStrategy")`, which returns an empty dict + warning if the `loading` package is missing (see `catalog.py:280-298`). 

So on this branch the catalog reports zero loaders and zero modifiers without failing and once PR #998 (loading) and PR #442 (modifiers) merge into main, the same call discovers them automatically. 

The pipeline/`from_config` docstring example at `pipeline.py:42` was the only place this assumption leaked into user-visible code; that has been fixed in 1c4761611.

training/src/anemoi/training/checkpoint/__init__.py
JesperDramsch: Working now. The previous code was a `try/except ImportError` sentinel because `checkpoint/validation.py` had never been written. Cherry-picked the implementation from PR #998 in 1e0f008ba. `validate_pipeline_health(context)` now does three things:
- Checks every `stage_{i}_{name}` metadata key is a completion marker (no `"failed:"` prefix).
- If any source stage ran, asserts `context.model.weights_initialized is True` so a swallowed loader error does not silently produce a random-weight model.
- Verifies structural invariants on the context (no optimizer without model, no `pl_module` with a non-Lightning format, etc.) and surfaces any error reported by the pre-execution environment/config validation.

API mirrors `validate_checkpoint` in `utils.py` — returns `True` or raises `CheckpointValidationError(.validation_errors=[...])`. Tests at `training/tests/checkpoint/test_validation.py` (21 tests, all green).

training/src/anemoi/training/checkpoint/sources/s3.py
JesperDramsch: Done in fa4095953. `S3Source` now calls `anemoi.utils.remote.s3.download_file(url, target, overwrite, resume, verbosity)` directly (obstore-backed). No boto3 client construction in this PR. The `pyproject.toml` `[s3]` extra was switched to `anemoi-utils[s3]>=0.4.37`, which pulls obstore transitively. boto3 and botocore are no longer dependencies.

training/src/anemoi/training/checkpoint/sources/base.py
JesperDramsch: Documented in 1c4761611. Added a paragraph at the top of `sources/base.py` clarifying that source selection is explicit via Hydra `_target_`. 

There is no scheme-based dispatcher that automatically routes `s3://` to `S3Source`. Each source has a `supports(...)` classmethod but it is opt-in validation, not automatic dispatch. The motivation is that someone could legitimately want `LocalSource` against a FUSE-mounted S3 path without the pipeline second-guessing the scheme.

JesperDramsch:
@mchantry agreed — dropping GCS from this PR. Filed #1139 for `AzureSource` as a follow-up since there is an active user base for it. We can pick that up after this PR lands.

JesperDramsch:
Addressed every comment in @anaprietonem's review, and resynced with main so CI can run. Commits in this round:

  - fa4095953 — switch S3Source to anemoi.utils.remote.s3.download_file (obstore-backed); drop boto3/botocore entirely
  - 1e0f008ba — implement validate_pipeline_health + CheckpointPipelineValidator (was a raising ImportError sentinel before)
  - 1c4761611 — batch review-feedback commit (safetensors, docstring, Lightning detection, conftest move, real-S3 test, sources/base.py docs)
  - 1b01bf7a4 — merge origin/main (resolved conflicts in pyproject.toml and test_utils.py; moved new tests into the renamed tests/unit/checkpoint/ layout)
  - 3b7c78cc2 — split anemoi-utils config into settings.toml + settings.secrets.toml per the CI error

(Still trying to fix the CI, but that should be a few iterations at most.)

##  Per-comment status

  - pyproject obstore harmonisation. [s3] extra is now anemoi-utils[s3]>=0.4.37 (fa4095953).
  - real S3 bucket integration test. @pytest.mark.network test downloads from the public NOAA GHCN bucket (1c4761611 + 3b7c78cc2 for the config-file split).
  - move fixture to parent conftest. sample_checkpoint lives in tests/unit/checkpoint/conftest.py; sources/conftest.py removed (1c4761611).
  - remove safetensors references. All six references in formats.py plus the Literal[...] type alias are gone; validation module no longer probes for it either (1c4761611).
  - pipeline.py docstring example. from_config example no longer references the loading package that doesn't exist on this PR's branch (1c4761611).
  - formats.py Lightning detection robustness. pytorch-lightning_version is now the primary marker, with a 2+ auxiliary-key fallback so a single Lightning rename doesn't silently break
  detection. Docstring notes the tested PTL/torch range (1c4761611).
  - catalog.py cross-PR reference. Reply only — design is intentional. _discover_components returns an empty dict + warning if the package is missing, so the catalog correctly reports zero
  loaders/modifiers on this branch and discovers them once #998 and #442 land.
  - validate_pipeline_health. Sentinel replaced with a real implementation that checks stage completion, the source-loaded-weights invariant, structural consistency, and the pre-execution
  validation status. API mirrors validate_checkpoint. 21 tests at tests/unit/checkpoint/test_validation.py (1e0f008ba).
  - s3.py use obstore. S3Source calls anemoi.utils.remote.s3.download_file directly. boto3 and botocore are no longer dependencies anywhere (fa4095953).
  - sources/base.py: how are sources identified. Docstring paragraph documents that source selection is explicit via Hydra _target_, not URL-scheme inference (1c4761611).

> [!NOTE]
> The review asked for changes in pipeline.py:42 (docstring) and formats.py:51 (Lightning detection). Flagged that decision in 1c4761611's commit message.

training/src/anemoi/training/checkpoint/catalog.py
anaprietonem: thanks for the clarification Jesper. Could you please add a comment in the code explaining that's the expected behaviour and that this functionality will be extended with PRs 998 and 442? So if users look at this they understand that's work in progress.

training/src/anemoi/training/checkpoint/__init__.py
```diff
@@ -46,8 +46,17 @@
...
-    # Validation module not available
-    validate_pipeline_health = None
+    from typing import NoReturn
+
+    def validate_pipeline_health(*_args: object, **_kwargs: object) -> NoReturn:  # type: ignore[misc]
+        """Sentinel: raise when the validation module is unavailable."""
+        msg = (
+            "validate_pipeline_health is not available because the "
+            "validation module could not be imported. Ensure that "
+            "anemoi.training.checkpoint.validation is installed."
```
anaprietonem: why this error message? (as in this should be installed whenever someone does `pip install anemoi-training` right) (there is no separate checkpoint package(?))

training/src/anemoi/training/checkpoint/sources/base.py
anaprietonem: Thanks Jesper, for that particular example what would be the error raise if someone tries to use a s3:// path while instantiating a LocalSource?

training/src/anemoi/training/checkpoint/catalog.py
JesperDramsch: Done, added a docstring to both `_get_loaders` and `_get_modifiers` explaining that the packages are populated by PRs #998 and #442, and that an empty registry with a discovery warning is the expected behaviour until those land.

training/src/anemoi/training/checkpoint/__init__.py
JesperDramsch: Good catch, that fallback was misleading. It's gone now. The validation module ships with anemoi-training so there's no reason for it to be missing at import time. Silly leftover from the gating logic before.

training/src/anemoi/training/checkpoint/sources/base.py
JesperDramsch: If someone configures `LocalSource` with `s3://bucket/key`, the path gets interpreted as a literal local path and `Path("s3://bucket/key").resolve()` won't exist on disk, so you get a `CheckpointNotFoundError`. Added that to the module docstring with a note that callers wanting scheme-based routing should dispatch on `urlparse(uri).scheme` themselves before picking a `_target_`.

MeraX:
Do you plan to add a page to the documentation for this PR?

anaprietonem:
> Do you plan to add a page to the documentation for this PR?

Had a chat with @JesperDramsch about this yesterday. There is already some docs available https://anemoi.readthedocs.io/projects/training/en/latest/checkpoint_pipeline_configuration.html. Before merging we'd update those to reflect that the sources are merged (so update from planned)  and Jesper already updated https://github.com/ecmwf/anemoi-core/pull/464/commits/3006aeb868f749baa36494bcb332949698c8b6fa.

JesperDramsch:
Integration tests pass: https://github.com/ecmwf/anemoi-core/actions/runs/26514398223

anaprietonem (approved):
Thanks Jesper for addressing the comments and implementing this feature! nice work!

Issue #257, opened 2025-04-11
Consolidate testing set-ups (integration tests in training, with testing utilities in utils, and tests in datasets)


VeraChristina:
### What maintenance does this project need?

The testing set-ups for anemoi-datasets (using testing utilities from anemoi-utils) and integration tests in anemoi-training evolved in parallel. We should consolidate the set-ups prior to moving the set-up for integration tests to other anemoi packages. This includes:

- potential refactor of downloading utilities for test data to accomodate needs in training
- agreeing on flags/markers and using them consistently
- reviewing the set-ups more generally to align them

VeraChristina:
Related issue on utils: https://github.com/ecmwf/anemoi-utils/issues/147

PR #468, opened 2025-08-07
test: replace longtest markers with slow

VeraChristina:
## Description
This PR replaces the longtest markers and --longtests cli option for pytest with slow markers and --slow.

Make markers and cli for slow tests consistent across the anemoi packages.

https://github.com/ecmwf/anemoi-core/issues/257

### DIFF
--- a/.github/workflows/integration-tests-hpc.yml
+++ b/.github/workflows/integration-tests-hpc.yml
@@ -41,7 +41,7 @@ jobs:
             source .venv/bin/activate
             pip install --upgrade pip
             pip install -e ./training[all,tests] -e ./models[all,tests] -e ./graphs[all,tests]
-            python3 -m pytest -v training/tests/integration --longtests
+            python3 -m pytest -v training/tests/integration --slow
             deactivate
             rm -rf $REPO_NAME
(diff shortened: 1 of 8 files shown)


aaron-hopkinson (approved):
LGTM

MeraX:
Hi Vera and Arron, @ecmwf/anemoisecurity 

I would like to raise security concerns regarding the additional dependency introduced by this PR. `pytest-skip-slow` is a very small package, maintained by an honourable single developer. As it it such a small package, it is probably reviewed by much less people than other dependencies like pytorch. This means, we cannot rely on the "community" when it comes to updates of `pytest-skip-slow`. Thus one of us would have to review updates of `pytest-skip-slow` to make sure that they are save, work well with our code and don't add any 3rd-party dependencies, that we don't need. 

For me personally, it would be much easier to keep track of the code for `--slow` when if if were part of any of the anemoi packages. Thus in my opinion, the potential risks and dependency-maintenance-costs of using an external dependency for --slow are much higher than the benefit we get from not maintaining 20 lines of the code on our own.

Therefore I would recommend implementing the `--slow` tag in anemoi-utils.

PR #469, opened 2025-08-07
feat: Logging to AzureML with ML Flow

timothyas:
## Description

I have been logging metrics to Azure ML, which requires Azure's special blend of mlflow. This PR adds those features, which I think might be useful to a broader audience of people who either run on Azure, Azure ML, or simply log and view metrics on AML. I'm currently running on an on-prem HPC machine but logging to AML with this code, however I think it will be necessary for those actually running on AML as well.

cc @mchantry 

Using mlflow to log to Azure ML requires some azure-specific code and packages.

No open issue, please let me know if you'd like me to open one up to describe the situation, I'd be happy to do so.

##  Additional notes ##

I got this working *just enough* to my liking, but there are still some issues unfortunately.

* tests: I'm not sure how to make tests with this, since it would require logging to AML... If anyone has any ideas I'm all ears
* TokenAuth: is not necessary with AML, see necessary helper PR at ecmwf/anemoi-utils#200
* terminal logs: with regular mlflow, terminal logs get updated by simply overwriting the terminal_log.txt. Overwriting files is not allowed with azure-mlflow. Moreover this appears to happen on each process (?) that's running, so many times over and over. I put in some try/except logic to try writing these files, but just pass whenever a failure happens. This means the terminal_log.txt stops updating pretty early on in the training process. I'm not really sure what to recommend here.
* diagnostic plots: these also seem to get written by each process (worker?). However, since each unique plot has a unique name, putting the try/except logic here doesn't really matter - each process tries to log the plot and if it succeeds, that means it was the first one and the plot gets stored. If the process fails, then it just moves on passively since a different one beat it to the job. no problem but maybe not the cleanest solution.
* model checkpoints: I simply could not get model checkpoints to log to AML, and I don't know what is going on there. For my purposes, I don't really care though, since they get stored locally on the machine where I will use the checkpoints anyway. Also, @mariahpope at EPIC/NOAA has been kicking off some anemoi training runs on AML, and noticed the checkpoints are stored. So, this may not be a problem.
* hyperparameter logging: in the second half of the method `log_hyperparams_in_mlflow`, the logger tries to log "expanded_params" after truncating them. I have no idea why but I could not get this to work for AML, despite lots of attempts with ChatGPT and Microsoft Copilot. This is the reason for overriding this method in the Azure subclass.
* I think there are a lot of different ways to establish a connection to an AML workspace, I did it here with the following code block:

```python
        sp_auth = ServicePrincipalAuthentication(
            tenant_id=os.environ["AZURE_TENANT_ID"],
            service_principal_id=os.environ["AZURE_CLIENT_ID"],
            service_principal_password=os.environ["AZURE_CLIENT_SECRET"],
        )
        ws = Workspace(
            subscription_id=os.environ["AZURE_SUBSCRIPTION_ID"],
            resource_group=aml_resource_group,
            workspace_name=aml_workspace_name,
            auth=sp_auth,
        )
```

Note there's a check before those lines to make sure the necessary environment variables are defined for the user. I would be very open to making this more flexible based on what true AML users think, or we could let this evolve as more users start to flex this feature.

Let me know if these limitations are acceptable, and I'd be happy to put some warning flags for the user so it's clear what will and will not be logged to AML. If anyone wants these to be worked out in the PR, I'd love to hear some suggestions - I'm pretty new to AML and leaned heavily on some LLMs to make this work, but "we" all got stuck on these issues :)

---

### DIFF
--- a/training/pyproject.toml
+++ b/training/pyproject.toml
@@ -64,7 +64,7 @@ dependencies = [
   "zarr>=2.14.2",
 ]
 
-optional-dependencies.all = [  ]
+optional-dependencies.all = [ "azure" ]
 optional-dependencies.azure = [ "azure-ai-ml", "azure-identity", "azureml-mlflow", "strenum" ]
 optional-dependencies.dev = [
   "anemoi-training[all,docs,tests]",


phinate:
Just to chime in that I've independently noticed many of the same issues as you @timothyas as an AzureML user. Happy to help take this forward if you'd like to divide up the work; will test the current state of your branch on our infrastructure as a first step! Thanks for this. :)

phinate:
I'm working on the SDK v2 migration on a fork-of-your-fork @timothyas, and just wanted to check: how are you (and colleague(s)) running things on AzureML? For me, I'm running a Command Job via `az ml job create`, and am working with that as the use-case. Want to make sure I don't end up accidentally preventing you from running things!

phinate:
> Hi! I have also been running Command Jobs (either do it via CLI or within a notebook using python sdk), so that is great.

Excellent! I'm actually having different results based on how I'm running at the moment; trying to avoid the secrets-based approach with a `ServicePrincipal`, and instead use a `ManagedIdentity` or a `UserIdentity`. Not sure what you've been getting to work, but for sure I'd love to know what your current workflow is + if you could test things once they're stable on our side :)

timothyas:
Hey @phinate thanks so much for all of this!! I'm really not sure why or how I'm using an older version of azure mlflow... I definitely appreciate your efforts for an upgrade and I'll upgrade my setup accordingly. I'm curious if that would fix some of the issues I ran into, like those related to logging hyper parameters. 

I would be happy to iterate with you on updates to this PR. My workflow is kind of strange actually. I typically run on an HPC machine, but since we happen to have an azure workspace set up, I just use azure mlflow to log to that as a server. However, from the NOAA side we may be using azure resources and AML directly soon, and @mariahpope has been getting that going. 

Just a heads up and explanation for a slow response, I'm on leave the rest of this week but I'll jump back on this on Monday.

timothyas:
Hi @phinate I pushed a couple of rapid updates since the main HPC machine I'm using is out for maintenance tomorrow. I hope that's OK. Basically:

* Turns out the default identity works just fine for me, azure figures it all out under the hood. So no complicated and overspecific ServicePrincipal, ClientSecret authentication etc. What you implemented is much cleaner and more general so big thanks for that!
* I only needed to modify `trk_uri` -> `tracking_uri`, since the latter gets passed to the MLFlowLogger instantiation. If that is passed as None, no logging ends up happening ... ([cb8fd24](https://github.com/ecmwf/anemoi-core/pull/469/commits/cb8fd2405c197f31e420c621061e720008e0a219))
* I merged in upstream/main to resolve some conflicts, really minor stuff.

Since I just got this working I haven't yet really been able to think about your more substantive issues yet, but here are some answers to your questions
* I have not gotten checkpoint logging to work with mlflow... For me this is fine since they get dumped locally onto the HPC file system, which is where I want to use them anyway, but this definitely needs to be addressed. I'm hoping with updated versioning (?) this can get sorted without too much headache
* Metrics have always been logging for me, here's a screenshot example of what I see on AML: 
   Screenshot 2025-08-19 at 4 33 49 PM
   ... and much more. All the plots (histograms, predictions, spectra, trainable features...) show up in the "images" tab, and the metrics tab is full of a ton of stuff. I'm wondering if cb8fd24](https://github.com/ecmwf/anemoi-core/pull/469/commits/cb8fd2405c197f31e420c621061e720008e0a219) solves your problem? I hope so...
* Also regarding the mlflow name vs the AML name, I'm not totally married to this code. I'll think about it a bit more to make sure it's not critical, but I'm leaning toward dropping it anyway. Right now I get two ~16 character hashes, one created by mlflow and one created by AML, and I always end up going onto AML and renaming it anyway to something that is sensible.

phinate:
You're very welcome for the changes! Thanks for tidying up some things for minimal diff to main. :)

> * I have not gotten checkpoint logging to work with mlflow... For me this is fine since they get dumped locally onto the HPC file system, which is where I want to use them anyway, but this definitely needs to be addressed. I'm hoping with updated versioning (?) this can get sorted without too much headache

Could you help me understand what you mean by updated versioning here? I'm identifying strongly with your question mark 😅

In past AML work, it's just been a case of linking the right directory to the `Outputs` tab via the job YAML and using the SDK. I can take a look at this if you're not around to!

> * Metrics have always been logging for me, here's a screenshot example of what I see on AML:

Ah, very cool! I've just seen that these metrics do exist for me, but log to a totally different job called `anemoi-debug` that is launched alongside my existing job automatically without my configuring of it. That's pretty confusing to me -- have you encountered this behaviour before? (I might not understand why this is useful in a more general MLFlow context...)

timothyas:
@phinate Sorry that was confusing. By version updates, I meant from the Azure SDK v1 -> v2, which as I understand is what you incorporated?

That is really strange. I have not seen that behavior, regarding the logging, since I only have 1 "AML Job" which is what gets sent from my HPC machine to AML. @mariahpope, next time you're running on AML would you be able to try it with this feature branch to see if you get this "double job" issue that @phinate is seeing? (PS I'm happy to provide a debug config with 1 degree GFS data so you don't have to update all your yamls if you don't want to)

training/src/anemoi/training/diagnostics/mlflow/azureml.py
mchantry, line 1: Please could the standard header be added to this file

# (C) Copyright 2025 Anemoi contributors.
#
# This software is licensed under the terms of the Apache Licence Version 2.0
# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
#
# In applying this licence, ECMWF does not waive the privileges and immunities
# granted to it by virtue of its status as an intergovernmental organisation

timothyas: Good catch @mchantry! Just added in [2e1351e](https://github.com/ecmwf/anemoi-core/pull/469/commits/2e1351e0dccdaa27f64c5dcf0110941b569f5e0f)

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -115,7 +115,13 @@ def _output_figure(
                 logger.experiment.log({exp_log_tag: wandb.Image(fig)})
             if self.config.diagnostics.log.mlflow.enabled:
                 run_id = logger.run_id
-                logger.experiment.log_artifact(run_id, str(save_path))
+                try:
+                    logger.experiment.log_artifact(run_id, str(save_path))
+                except Exception as e:
+                    if "Resource Conflict" in str(e):
```
gmertes: I suppose this case doesn't throw a unique exception, and a string comparison is the only way? If so - absolutely fine.

training/src/anemoi/training/diagnostics/logger.py
```diff
@@ -70,11 +71,20 @@ def get_mlflow_logger(config: BaseSchema) -> None:
...
+    kw = {}
+    if not config.diagnostics.log.mlflow.use_azure:
+        logger_class = AnemoiMLflowLogger
+    else:
+        logger_class = AnemoiAzureMLflowLogger
+        kw["aml_resource_group"] = config.diagnostics.log.mlflow.aml_resource_group
+        kw["aml_workspace_name"] = config.diagnostics.log.mlflow.aml_workspace_name
+        kw["aml_subscription_id"] = config.diagnostics.log.mlflow.aml_subscription_id
+        kw["aml_identity"] = config.diagnostics.log.mlflow.aml_identity
+
```
gmertes: We should avoid an if driven by a boolean, this doesn't scale. We should go for one of these:
- A lookup table with an enum in the schema, like logger = "default" or "azure" , that then translates to the logger class.
- Use hydra instantiate with the class target in the config. This way, the kwargs could also be automatically passed and don't need to be hardcoded here (but they should still be part of a schema).

@anaprietonem do you have thoughts on this?

training/src/anemoi/training/diagnostics/mlflow/azureml.py
```diff
@@ -0,0 +1,371 @@
...
+            self,
+            experiment_name=experiment_name,
+            run_name=run_name,
+            tracking_uri=tracking_uri,
+            tags=tags,
+            save_dir=save_dir,
+            log_model=log_model,
+            prefix=prefix,
+            run_id=run_id,
+        )
```
gmertes: If this whole block is identical to the base class, then we should avoid copy-paste. Instead, in the base class, refactor this block into a private function that you can then just call from the derived class.

We should also avoid inheriting from concrete classes. I suggest refactoring both loggers to have a shared based class. This doesn't have to be a complete ABC, just a place for shared code. The base class can be what is currently `AnemoiMLflowLogger` , and the new concrete `AnemoiMLflowLogger` can just be an empty class inheriting from the base. For future-proofing.

training/src/anemoi/training/diagnostics/mlflow/azureml.py
```diff
@@ -0,0 +1,371 @@
...
+            Whether to run offline or not, by default False
+        authentication : bool | None, optional
+            Whether to authenticate with server or not, by default None
+        log_hyperparams : bool | None, optional
+            Whether to log hyperparameters, by default True
+        on_resume_create_child: bool | None, optional
+            Whether to create a child run when resuming a run, by default False
+        max_params_length: int | None, optional
+            Maximum number of params to be logged to Mlflow
+        """
```
gmertes, line 166: Since python inherits docstrings from the base class, I think in the case of really long ones like this they can be omitted. I think a docstring should only have a single source of truth unless they are different. But this personal preference, happy for others to chime in.

training/src/anemoi/training/diagnostics/mlflow/logger.py
```diff
@@ -237,7 +237,14 @@ def _remove_csi(line: bytes) -> bytes:
...
             logfile.flush()
-        self.experiment.log_artifact(self.run_id, str(self.file_save_path))
+
+        # This try/except is for AML logging
+        # typically terminal_output.txt gets updated by overwriting, but this is not allowed on AML
+        # since this happens tons of times, just silently pass if it fails
+        try:
+            self.experiment.log_artifact(self.run_id, str(self.file_save_path))
+        except:
+            pass
```
gmertes: Failing silently here could hide real problems when used with the default logger. I would do an optional flag like `fail_silently` that is only enabled by the Azure logger.

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -283,6 +283,16 @@ class MlflowSchema(BaseModel):
...
+    use_azure: bool = False
+    "If true, log to azure ml workspace server"
+    aml_identity: str | None = None
+    "Type of identity to use for logging in with Azure ML."
+    aml_resource_group: str | None = None
+    "If using AML to log with MLFlow, name of the resource group"
+    aml_workspace_name: str | None = None
+    "If using AML to log with MLFlow, name of the workspace"
+    aml_subscription_id: str | None = None
+    "If using AML to log with MLFlow, subscription ID"
```
gmertes: I think this needs to be refactored into its own schema. This goes together with not using a bool `use_azure`. Instead, the `MlflowSchema` could have a key like `server` that then takes a server-specific schema with extra options. In the case of the default server, it could be empty. 

I think @anaprietonem will have better insight and suggestions.

training/src/anemoi/training/diagnostics/mlflow/logger.py
phinate: Nothing intelligent to add here, just wanted to say that my linter flagged that this pattern is preferred to `try`/`except`/`pass`:

```python
from contextlib import suppress


with suppress(Exception):
    self.experiment.log_artifact(self.run_id, str(self.file_save_path))
```

I do think we could think of something better here though; maybe logging a new file per-epoch? Or at least shipping this feature with no terminal logging for now, since we have the logs already on AzureML?

phinate:
@timothyas: I'm particularly curious about the way your workflow treats the environment variable-side of things. For example, we've talked a bit about the workspace authentication, but really my runs have actually now stopped using that with the introduction of the environment variable-based access to MLFlow. If you're working in an environment where, indeed, you _don't_ have those variables to hand on the machine that's running Anemoi, we may need to explicitly cover that case just like we did with the workspace-based fallback.

Please correct me if this isn't necessary!

timothyas:
Hi everyone!

First, thanks @anaprietonem and @gmertes for the review, that is super helpful. I definitely agree about the refactoring comments, and I'll be happy to iterate on that. However, I think it would be best to first iterate with @phinate to get this to a place where it's working for us. Seems like the main issues here are
* the "double runs" issue (which sounds like is fixed thanks to @anaprietonem and @phinate, I just need to check how this works in my setup)
* logging checkpoints (right now this isn't happening at all... TBD why)
* logging hyperparameters would be nice and should be possible

After that I'll be happy to refactor this and clean up some of the dirty try/except blocks that are not really ideal (+ other comments).

@phinate to your question, let me see if I get access to these environment variables, and if that code block works for me.

timothyas:
@phinate I don't have this environment variable, but adding that line doesn't really matter for my case since it will just pass through that line same as before. As I understand it, previously you were getting two "jobs" on Azure, with that was doing the actual computation, and the other one created by Anemoi (via the PTL logger) which had all the metrics. In my case, I only see the latter one, and in my view it doesn't really matter what it's labelled as on AML since it has the run_name which I can correspond to my HPC job.

Let me know if that makes sense... But TL; DR I'll put that line in since it solves your problem and doesn't break my setup.

I'll try to chug away at checkpoint logging after the weekend.

timothyas:
Awesome that's great to see @phinate! Are you getting checkpoints too? 

Also it's great to hear all your thoughts related to restarting, etc. I'll look forward to the issue, and also to the PR to this PR whenever you get a chance. Thanks!

training/src/anemoi/training/schemas/diagnostics.py
anaprietonem: Yes I'd suggest to have a different schema for the AzureLogger that inherits from the MlflowSchema and add the extra entries!

training/src/anemoi/training/diagnostics/mlflow/azureml.py
```diff
@@ -0,0 +1,382 @@
...
+        import mlflow
```
anaprietonem: any thoughts on simplifying this by just calling the super for the __init__ of the AnemoiMLflowLogger?

training/src/anemoi/training/diagnostics/mlflow/azureml.py
```diff
@@ -0,0 +1,382 @@
...
+        self._forked = forked
+        self._flag_log_hparams = log_hyperparams
+        self._max_params_length = max_params_length
+
+        self._fork_run_server2server = None
+        self._parent_run_server2server = None
+        self._parent_dry_run = False
+
+        # <-- Azure specific stuff -->
+        # we don't need authenticate, this just lets us easily subclass the logger
```
anaprietonem: Regarding the authentication:
- above there is a mention to the `aml_identity` and that's used for azure authentication?
seems a bit confusing to have a entry in this logger with `authentication:None`. We could look at using the authentication flag already existing in the logger to carry out the azure one? (rather than SSO the AnemoiMlogger uses).

training/src/anemoi/training/diagnostics/mlflow/azureml.py
```diff
@@ -0,0 +1,382 @@
...
+        self._resumed = resumed
+        self._forked = forked
+        self._flag_log_hparams = log_hyperparams
+        self._max_params_length = max_params_length
+
+        self._fork_run_server2server = None
+        self._parent_run_server2server = None
+        self._parent_dry_run = False
+
+        # <-- Azure specific stuff -->
```
anaprietonem: related to the comment of reusing the init - the azure specific could also be moved to a private method specific of this logger and simply called then called it first in the init?

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: If the plots are just saved by rank0 and include the batch and epoch, when are you seeing error of duplicated artifacts? if this just specific to the AzureLogger?

timothyas:
Hi everyone, thanks for all of the input! Sorry for the slow progress on this one, I have a bunch of September deadlines that I had to tend to this week so I didn't get time to work on this PR. I'm hoping to dive into it next week.

@phinate thanks for all of the thoughts here. I think I would like to prioritize cleaning up this PR and getting it in before getting more functionality related to the forking/resuming issues. For example I don't want you to keep having to work off this branch to do any dev work. Thanks for your patience already.

@anaprietonem great comments, will dive deeper as soon as I can.

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -283,6 +287,8 @@ class MlflowSchema(BaseModel):
     "Specifies the maximum number of retries for MLflow HTTP requests, default 35."
     max_params_length: int = MAX_PARAMS_LENGTH
     "Maximum number of hpParams to be logged with mlflow"
+    save_dir: str | None = None
```
anaprietonem, line 360: For this we'd need an entry to the config linked to hardware.paths.logs.mlflow as at the moment this is hardcoded and not configurable

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -291,6 +297,55 @@ def clean_entity(cls: type["MlflowSchema"], values: dict[str, Any]) -> dict[str,
         return values
 
 
+class AzureMlflowSchema(BaseModel):
```
anaprietonem: why not inheriting from `MlflowSchema` and reusing the keys? we could also have a BaseMlflowSchema for common keys and then MlflowSchema and AzureMlflowSchema if ieasier

training/src/anemoi/training/diagnostics/logger.py
```diff
@@ -27,79 +27,46 @@ def get_mlflow_logger(config: BaseSchema) -> None:
...
+        if logger_config["save_dir"] is None:
+            logger_config["save_dir"] = config.hardware.paths.logs.mlflow
 
-    offline = config.diagnostics.log.mlflow.offline
-    if not offline:
-        tracking_uri = config.diagnostics.log.mlflow.tracking_uri
-        LOGGER.info("AnemoiMLFlow logging to %s", tracking_uri)
-    else:
-        tracking_uri = None
+        # only create save_dir if it's specified in schema (e.g. it's not for Azure)
```
anaprietonem: we could move this to the AnemoiMlflowLogger, cause in that case the save_dir is not configurable but directly controlled with the hardware config

training/src/anemoi/training/diagnostics/logger.py
```diff
@@ -27,79 +27,46 @@ def get_mlflow_logger(config: BaseSchema) -> None:
...
-    os.environ["_MLFLOW_HTTP_REQUEST_MAX_RETRIES_LIMIT"] = str(http_max_retries + 1)
-    # these are the default values, but set them explicitly in case they change
-    os.environ["MLFLOW_HTTP_REQUEST_BACKOFF_FACTOR"] = "2"
-    os.environ["MLFLOW_HTTP_REQUEST_BACKOFF_JITTER"] = "1"
-
-    from anemoi.training.diagnostics.mlflow import LOG_MODEL
-    from anemoi.training.diagnostics.mlflow import MAX_PARAMS_LENGTH
-    from anemoi.training.diagnostics.mlflow.logger import AnemoiMLflowLogger
+    # TODO: this seems a bit dirty
+    log_system = logger_config.pop("system")
```
anaprietonem: I think the docdict Config that is used below is much cleaner

anaprietonem:
> Hi everyone, I was able to do some work on this PR and focused on creating a separate `AnemoiAzureMLflowLogger` schema, and created a base class from which that class and the `AnemoiMLflowLogger` classes could be inherited from. I liked @gmertes suggestion to add the `_target_` option to the existing schemas, this seemed like a really clean way to streamline the options, but still have generality. I'd be curious what folks (esp. @gmertes @anaprietonem) think of the implementation.

I like the idea of the  BaseAnemoiMLflowLogger, regarding the log_model the problem there is that if that's not passed to the get_mlflow_logger how does one control or activate the logging of model checkpoints to mlflow?

training/src/anemoi/training/diagnostics/logger.py
```diff
@@ -27,79 +27,46 @@ def get_mlflow_logger(config: BaseSchema) -> None:
...
-    log_model = getattr(config.diagnostics.log.mlflow, "log_model", LOG_MODEL)
-
-    logger = AnemoiMLflowLogger(
-        experiment_name=config.diagnostics.log.mlflow.experiment_name,
-        project_name=config.diagnostics.log.mlflow.project_name,
-        tracking_uri=tracking_uri,
-        save_dir=save_dir,
-        run_name=config.diagnostics.log.mlflow.run_name,
+    logger = instantiate(
+        logger_config,
```
anaprietonem, line 33: are you proposing the pass the logger_config to the init and then handle all the settings there? Otherwise I see a mix at the moment and it's not clear how some options like 'offline' that are removed from this function are later handled in the init

training/src/anemoi/training/schemas/diagnostics.py
timothyas: Agreed. I just committed that, but I'm happy to undo that if this ends up not being the format we settle on.

training/src/anemoi/training/schemas/diagnostics.py
timothyas: We could totally do that. I made a separate schema because the options: offline, authentication, and save_dir don't make sense for azureml (in our first draft we just had an assertion error raise if offline was set to True). But we could bring back the assertion error and inherit, let me know what you think.

Just to give us something concrete, I'll make a BaseMLflowSchema for common options. But I'm happy to update that as needed.

training/src/anemoi/training/diagnostics/logger.py
timothyas: Yeah I think that would be cleaner, but this implementation was just a start. I think it would be a bit more straightforward to remove as much of this "preamble" stuff and handle everything in the logger class. I'm a bit new to omegaconf and hydra though so I think my implementation is a bit rough around the edges.

training/src/anemoi/training/diagnostics/logger.py
timothyas: I'm not sure I totally understand. It was my impression that save_dir was basically the only option that needed to be handled before getting passed to the logger class, since it's in the "hardware" scope of the config, whereas I'm only passing the diagnostics.log.mlflow options to the logger class.

training/src/anemoi/training/diagnostics/logger.py
timothyas: Sorry I don't totally understand what you're suggesting. Could you elaborate?

For context, I wasn't sure how to handle these options (log_system, log_terminal, expand_hyperparams). I guess these lines could also be moved inside of the logger constructor ?

anaprietonem: Sure! my point was mostly about mixing the .pop and dotDict `config.diagnostics.log.mlflow.system`. Since now we have a mix of both with settings like `forked = config.training.fork_run_id is not None` and `log_system = logger_config.pop("system")`. I think it would be more consistent to avoid using the pop to access the config key and rather use the DoctDict interface

training/src/anemoi/training/diagnostics/logger.py
```diff
@@ -27,79 +26,33 @@ def get_mlflow_logger(config: BaseSchema) -> None:
...
+
+    # TODO: this seems a bit dirty
+    log_system = logger_config.pop("system")
+    log_terminal = logger_config.pop("terminal")
+    expand_hyperparams = logger_config.pop("expand_hyperparams")
+
+    # Defaults that exist outside the scope of config.diagnostics.log.mlflow
+    if "save_dir" in logger_config:
+        if logger_config["save_dir"] is None:
+            logger_config["save_dir"] = config.hardware.paths.logs.mlflow
```
timothyas: @anaprietonem I have this line in here since there was a version of it in the original code. But since `config.hardware.paths.logs.mlflow` is now in the default config yaml, do you think we can just remove this line (and nested if statements)?

timothyas:
> I like the idea of the BaseAnemoiMLflowLogger, regarding the log_model the problem there is that if that's not passed to the get_mlflow_logger how does one control or activate the logging of model checkpoints to mlflow?

@anaprietonem In both the current and original implementation this is happening through the config.diagnostics.logs.mlflow.log_model option, but I think this current implementation just cleans up the code a bit by moving the code blocks from `get_mlflow_logger` to the `BaseAnemoiMLflowLogger.__init__` method. Note that the new default to the constructor for log_model is the value set in `anemoi.training.diagnostics.mlflow.__init__.py`.

training/src/anemoi/training/schemas/diagnostics.py
timothyas: One downside of doing it this way, rather than having the azure version inherit all options, is that the yaml situation gets a bit more complicated for users. For me to use the azure _target_, I can no longer modify all options from a top level config.yaml file, and instead I have to go into the diagnostics/evaluation.yaml file and remove the offline, authentication, and save_dir options. 

So, maybe it is better to just inherit the same schema, even though those options would get ignored. I'm not sure if you have a recommendation there.

training/src/anemoi/training/diagnostics/logger.py
timothyas: Gotcha, thank you. I switched this to using `.get` and `del`, which Microsoft Copilot tells me is more consistent for the DotDict interface, but let me know if you have something else in mind.

training/src/anemoi/training/schemas/diagnostics.py
anaprietonem: do you mean that you can't do  in your top level config:
```
diagnostics:
    mlflow:
      _target_: AzureMLllfowLogger
      tracking
      offline
      etc?
      
 ```
 what's the error you get if you try that?

Comment (timothyas) [reply training/src/anemoi/training/schemas/diagnostics.py]: Yeah, the `_target_: AzureMLflowLogger` works, but the issue is the options `offline`, `authentication`, and `save_dir`. I removed these from the azuremlflowlogger class, and so they can't be passed in the yaml. For example:

```yaml
diagnostics:
  log:
   mlflow:                                                
      _target_: "anemoi.training.diagnostics.mlflow.azureml.AnemoiAzureMLflowLogger"                                          
      enabled: True      
      offline: False          
      authentication: False   
      log_model: True
      # etc
```

running config validate throws the error

```
$ anemoi-training config validate --config-name debug.yaml

💣 3 validation errors for BaseSchema
diagnostics.log.mlflow.MlflowSchema._target_
  Input should be 'anemoi.training.diagnostics.mlflow.logger.AnemoiMLflowLogger' [type=literal_error, input_value='anemoi.training.diagnost...AnemoiAzureMLflowLogger', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/literal_error                                            
diagnostics.log.mlflow.AzureMlflowSchema.offline           
  Extra inputs are not permitted [type=extra_forbidden, input_value=False, input_type=bool]                                   
    For further information visit https://errors.pydantic.dev/2.11/v/extra_forbidden
diagnostics.log.mlflow.AzureMlflowSchema.authentication
  Extra inputs are not permitted [type=extra_forbidden, input_value=False, input_type=bool]
    For further information visit https://errors.pydantic.dev/2.11/v/extra_forbidden
2025-09-10 08:52:38 ERROR 💣 Exiting
```

but after removing the offline and authentication options, it's all fine. This is the downside to how I've implemented it, so I'm leaning toward including these in the schema and putting assertion errors in the logger classes to keep the user from running an azure logger offline

anaprietonem: Okey I will try to think about this a bit more. I have checkout some of the changes in this branch and tested and indeed can reproduce the above, so now it's easier for me to see if we can propose a solution. The one that comes to mind would be to to something like the scalars config https://github.com/ecmwf/anemoi-core/tree/main/training/src/anemoi/training/config/training where we can have the default that would use the 'BaseSchema' with the common keys and then specific config for the mlflow block? But that might also not be ideal from a user perspective

training/src/anemoi/training/diagnostics/logger.py
anaprietonem: I don't think I yet understand the need for doing:
```
    logger_config = OmegaConf.to_container(convert_to_omegaconf(config.diagnostics.log.mlflow))
    log_system = logger_config.pop("system")

```
 could one you just do?:
```
    logger_config = config.diagnostics.log.mlflow
    log_system = logger_config.log_system

```

training/src/anemoi/training/schemas/diagnostics.py
anaprietonem: A work around could be:
```
from pydantic import BaseModel as PydanticBaseModel

class MlflowBaseSchema(PydanticBaseModel):
    enabled: bool
    "Use MLflow logger."
    log_model: bool | Literal["all"] | None = None
    "Log checkpoints created by ModelCheckpoint as MLFlow artifacts. \
            If True, checkpoints are logged at the end of training. If 'all', checkpoints are logged during training."
    tracking_uri: str | None = None
    "Address of local or remote tracking server."
    experiment_name: str
    "Name of experiment."
    project_name: str
    "Name of project."
    system: bool
    "Activate system metrics."
    terminal: bool
    "Log terminal logs to MLflow."
    run_name: str | None
    "Name of run."
    on_resume_create_child: bool
    "Whether to create a child run when resuming a run."
    expand_hyperparams: list[str] = Field(default_factory=lambda: ["config"])
    "Keys to expand within params. Any key being expanded will have lists converted according to `expand_iterables`."
    http_max_retries: PositiveInt = Field(example=35)
    "Specifies the maximum number of retries for MLflow HTTP requests, default 35."
    max_params_length: int = MAX_PARAMS_LENGTH
    "Maximum number of hpParams to be logged with mlflow"
    save_dir: str | None = None
    "Directory to save logs to when offline=True, default=config.hardware.paths.logs.mlflow"

    @root_validator(pre=True)
    def clean_entity(cls: type["MlflowSchema"], values: dict[str, Any]) -> dict[str, Any]:  # noqa: N805
        if values["enabled"] is False:
            values["tracking_uri"] = None
        return values


class MlflowSchema(MlflowBaseSchema):
    target_: Literal["anemoi.training.diagnostics.mlflow.logger.AnemoiMLflowLogger"] = Field(
        alias="_target_",
    )
    offline: bool
    "Run MLflow offline. Necessary if no internet access available."
    authentication: bool
    "Whether to authenticate with server or not"

class AzureMlflowSchema(MlflowBaseSchema):
    target_: Literal["anemoi.training.diagnostics.mlflow.azureml.AnemoiAzureMLflowLogger"] = Field(
        alias="_target_",
    )
    identity: str | None = None
    "Type of identity to use for logging in with Azure ML."
    resource_group: str | None = None
    "Name of the AzureML resource group"
    workspace_name: str | None = None
    "Name of the AzureML workspace"
    subscription_id: str | None = None
    "AzureML subscription ID"
    azure_log_level: str = "WARNING"
    "Log level for all azure packages (azure-identity, azure-core, etc)"


and then 

class LoggingSchema(BaseModel):
    wandb: WandbSchema
    "W&B logging schema."
    tensorboard: TensorboardSchema
    "TensorBorad logging schema."
    mlflow: Annotated[MlflowSchema,AzureMlflowSchema,Field(discriminator="target_")]
    "MLflow logging schema."
    interval: PositiveInt
    "Logging frequency in batches."

```
the only downside is the PydanticBaseModel allow for extra keys so what's happening is the azure one will get the defaults for offline and authentication. We have those cases in a few other examples in the code and it works okey.

timothyas: I like that suggestion, thanks @anaprietonem. I'll go ahead and implement that.

training/src/anemoi/training/diagnostics/logger.py
timothyas: Oh I see, the reason is because logger_config is passed to the creation of the logger class, which does not take the argument "system". So I'm removing it from the config here, but maintaining the value to handle the system logging later. However, we could change the code to move these lines into the logger constructor:

```
    config_params = OmegaConf.to_container(convert_to_omegaconf(config), resolve=True)
    logger.log_hyperparams(
        config_params,
        expand_keys=expand_hyperparams,
    )

    if log_terminal:
        logger.log_terminal_output(artifact_save_dir=config.hardware.paths.plots)
    if log_system:
        logger.log_system_metrics()
```

That would remove the need for all this argument "popping" etc. What do you think about that?

anaprietonem: Thanks for clarifying that Tim, I like that suggestion!

training/src/anemoi/training/schemas/diagnostics.py
timothyas: So I tried this out and it did not work... Sorry for all the back and forth on this, but after thinking about this, I think what I would lean toward is what I've now implemented in 820b0e7. In that commit I went toward having the Azure schema inherit the standard mlflow schema. Even though it accepts extra options that technically do not apply to the azure, it seems like the simplest situation coding wise, and for the user. By enforcing the required choice for offline, I think it keeps the user from wasting time with an inconsistent config while being relatively straightforward in implementation. I'm curious what you think...

training/src/anemoi/training/diagnostics/logger.py
timothyas: Done in 1da387a

training/src/anemoi/training/diagnostics/logger.py
timothyas: Cleaning up some conversations, this has been implemented by creating a separate schema.

training/src/anemoi/training/diagnostics/mlflow/azureml.py
timothyas: This has been cleaned up with a bit better class inheritance. Resolving to de-clutter the conversations.

training/src/anemoi/training/diagnostics/mlflow/azureml.py
timothyas: I agree with this and implemented this. The Azure logger class only has the new arguments.

training/src/anemoi/training/diagnostics/logger.py
timothyas: @anaprietonem I implemented that suggestion in 7f83552, so now the only config option that gets removed is the "enabled" option, since if we're in this code block, mlflow logging is certainly enabled, and it doesn't really make sense to pass to the logger class constructor.

I kept the "log_hyperparams, log_terminal, log_system" calls at the bottom, since some of these use config options that are out of the scope of logging. However, all the option handling has been removed, so it's a bit cleaner.

Let me know what you think!

training/src/anemoi/training/diagnostics/logger.py
timothyas: I think this is much cleaner now, I'll resolve this since we're discussing more or less the same issue in a separate conversation.

phinate:
Hey @anaprietonem, thanks for checking in :) 

From my side, I've been waiting for the dust to settle on the refactor-type changes you and @timothyas have been working on. I have a little bit of bandwidth to help move things along if either of you would like me to look things over or write some tests like I alluded to earlier!

training/src/anemoi/training/diagnostics/mlflow/logger.py
phinate: Did we have any better ideas here @timothyas?

training/src/anemoi/training/diagnostics/mlflow/azureml.py
```diff
@@ -0,0 +1,289 @@
...
+            "`diagnostics.mlflow.resource_group`, `diagnostics.mlflow.workspace_name`."
+        )
+        raise ValueError(msg)
+    LOGGER.info("Attempting to get Workspace object...")
+    ws = ml_client.workspaces.get(ml_client.workspace_name)
+    LOGGER.info("Succeeded getting the current workspace!")
+    return ws
+
+
+class AnemoiAzureMLflowLogger(AnemoiMLflowLogger):
```
phinate: I'm not up-to-date with the most recent changes, but I was wondering why an abstract base class was introduced (`BaseAnemoiMLflowLogger`) if we don't use it for Azure here?

timothyas: Nice catch @phinate, I maybe got a little mixed up here. As it is, the `AnemoiAzureMLflowLogger` should inherit the base class. I guess right now the base class is not 100% necessary, but maybe it's useful for any potential other versions later down the line.

timothyas: Fixed in [261947c](https://github.com/ecmwf/anemoi-core/pull/469/commits/261947c99ac75b49746ea787e455193257cbd4da)

training/src/anemoi/training/diagnostics/logger.py
anaprietonem: Yep looks much better, thank you!

training/src/anemoi/training/schemas/diagnostics.py
anaprietonem: Sorry I missed this one. What I find a bit conflicting at the moment is that schemas do not follow the structure of the code classes (we have AnemoiMLflowLogger, BaseAnemoiMLflowLogger and AnemoiAzureMLflowLogger) but while in the schemas that's not the same pattern. I was looking at the code snipped I posted above and realised that the loggingSchema should have been 

```
class LoggingSchema(BaseModel):
    wandb: WandbSchema
    "W&B logging schema."
    tensorboard: TensorboardSchema
    "TensorBorad logging schema."
    mlflow: Annotated[MlflowSchema | AzureMlflowSchema,Field(discriminator="target_")]
    "MLflow logging schema."
    interval: PositiveInt
    "Logging frequency in batches."

```
(note there was a missing | in the mlflow section). Using that it runs okey for me, could you test it? otherwise you can copy paste the error trace and I can help debugging.

anaprietonem:
Changes LGTM, I think if we could fix the schemas and @timothyas managed to find a way to improve the try/expect for the artifacts it should be ready. Regarding testing @phinate if you are happy to give that a go, it would be great. I am aware unit tests for the existing functionality on the 'AnemoiMlflowLogger' are not very extensive, so if you'd like some help let me know!

training/src/anemoi/training/diagnostics/mlflow/logger.py
timothyas: I don't really have a better solution here. The terminal output is just a continually updated stream. We could put time stamps on it, but we'd end up with a text file every 30s which is probably not ideal.

Maybe a cleaner option is to enforce that the Azure logger simply does not log terminal output. I'll add that in since it seems cleaner than passing options all the way through the LogsMonitor class, but if anyone has a better idea I'd be open to it

training/src/anemoi/training/schemas/diagnostics.py
timothyas: Thanks @anaprietonem! I added the `Annotated[ ]` around this, and made sure it works for me ([8a017bf](https://github.com/ecmwf/anemoi-core/pull/469/commits/8a017bf7f0289176d44779164422465e7d934d7a)).

That's a fair point regarding schema/class structure. Would you prefer it if I make the class structure consistent with the Schema structure? i.e. get rid of the BaseAnemoiMLflowLogger?

phinate:
Following up from a conversation with @anaprietonem: the test I think would be nice to have is just to check that we can instantiate the `AzureMLFlowLogger` class with the arguments, and maybe see if we can get some kind of log after setting the right environment variables that the class requires.

I think if we can do that, then I'd be happy to defer other tests to a focused PR that looks at the whole logging suite.

training/src/anemoi/training/diagnostics/logger.py
```diff
@@ -27,79 +26,23 @@ def get_mlflow_logger(config: BaseSchema) -> None:
...
-        log_hyperparams=log_hyperparams,
-        authentication=config.diagnostics.log.mlflow.authentication,
-        on_resume_create_child=config.diagnostics.log.mlflow.on_resume_create_child,
-        max_params_length=max_params_length,
     )
     config_params = OmegaConf.to_container(convert_to_omegaconf(config), resolve=True)
     logger.log_hyperparams(
         config_params,
-        expand_keys=config.diagnostics.log.mlflow.expand_hyperparams,
+        expand_keys=logger.expand_hyperparams,
```
timothyas: @anaprietonem these lines were deleted in #327, and are now creating a conflict for this PR. I'm curious what your preference is for resolving that conflict, should I remove these lines or keep them?

anaprietonem: Hey @timothyas, for context yes those changes were introduced here https://github.com/ecmwf/anemoi-core/pull/327/files. So now the log_hyperparams function is simply called as part of the trainer set up so than we can also log a few other things related to the loss scaling. Hope that's clear and you can resolve the conflict but let me know if you need more help!

timothyas: Ah perfect, that's great to know, thank you @anaprietonem! I will ~Merge upstream/main into this feature branch~ <- Actually I'll wait to do that so I don't mess with @phinate's PR. I'll just remove these lines from logger.py

After merging with upstream/main, I think it would make sense to get rid of the `expand_hyperparams` init argument and attribute to the mlflow loggers. These were added by this PR, but since the `expand_keys` is handled by the schema, and this is passed as an argument to the expand hyperparams method, there's no reason for duplication (and possibly conflicting default values) here.

timothyas: Ok so actually I had to do something different here in order to satisfy Ruff... Ruff was calling out a bunch of "unused argument" instances. One of those instances is related to the "expand_keys" argument here, which is not used by the AML logger, since parameters are not individually logged due to the strict 200 parameter limit (we just do it as a single json file and that's it). So, I decided to rely on the class attribute `expand_hyperparameters` rather than as an extra argument to the method.

I hope that makes sense, let me know if you'd like more explanation.

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -115,7 +116,12 @@ def _output_figure(
                 logger.experiment.log({exp_log_tag: wandb.Image(fig)})
             if self.config.diagnostics.log.mlflow.enabled:
                 run_id = logger.run_id
-                logger.experiment.log_artifact(run_id, str(save_path))
+                try:
```
anaprietonem: @timothyas, I like you added the check for raising an exception just if the AzureMlflow logger! could be worth adding a one-linear comment with some context about this? (so people are aware is just a problem related to Azure that might be resolved in the future but for now we can't remove?)

As an alternative - you could consider overwriting the log_artifact method of the experiment in `AnemoiAzureMLflowLogger` and perform the try except there?

timothyas: Thanks for this suggestion @anaprietonem, I actually like your second idea to overwrite the logger.experiment.log_artifact. This keeps the goofy try/except logic isolated. I implemented that now just testing if it works.

timothyas: Appears to be working via updates in 3019aa8. This is much cleaner, now there are no modifications to diagnostics/callbacks/plot.py - thanks again!

anaprietonem:
> Thanks @anaprietonem, that sounds great! Hopefully no big issues with the parameters being logged...

 I looked a bit more in detail, the difference I was seeing it's mostly since now the config entries are being logged slightly different so that we don't see 'config.data.processor...' but 'data.processor...' . Since we have these entries on the clean_params: (list of prefixes to remove)
 
 ```
        "hardware",
        "data",
        "dataloader",
        "model",
        "training",
        "diagnostics",
        "graph",
 ````
 
 we'd need to remove then to keep the same behaviour as before. @timothyas could you that change as part of that PR please?

timothyas:
@anaprietonem I just removed the prefixes that you mentioned in c9eeef0, although I have to be honest I don't really understand why that would have changed in this PR. Let me know if this is what you were looking for. Thanks!

timothyas:
Woohoo! Thanks @anaprietonem! Let me know if you'd like any help with the other PR.

PR #475, opened 2025-08-11
test: add multigpu option to pytest cli

VeraChristina:
## Description
Add multigpu option to pytest cli to by default skip tests marked as requiring multiple GPUs


cathalobrien:
LGTM!!1!

Issue #477, opened 2025-08-12
IndexError in training/losses/utils.py


efare04:
### What happened?

I'm trying to run anemoi-training for the first time, with GraphForecaster and using a GNN. The run has succesfully passed config validation, graph creation, and diagnostics setup, and then I think the issue has occurred somewhere in model initialisation. The run crashes with an IndexError when attempting to log the final variable scalings.

The final few lines of the error traceback are:
  File "/data/users/sprai/.conda/envs/sprai_test/lib/python3.12/site-packages/anemoi/training/losses/utils.py", line 33, in print_variable_scaling
    log_text += f"{name}: {variable_scaling[idx]:.4g}, "
                           ~~~~~~~~~~~~~~~~^^^^^
IndexError: invalid index of a 0-dim tensor. Use `tensor.item()` in Python or `tensor.item<T>()` in C++ to convert a 0-dim tensor to a number

I think the solution would just be to change line 33 of [utils.py](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/losses/utils.py) to `log_text += f"{name}: {variable_scaling[idx].item():.4g}, "`.

### What are the steps to reproduce the bug?

I'm using a dataset of ocean variables (we're trying to forecast SST) from GloSea6 which I created using [dataset_recipe.txt](https://github.com/user-attachments/files/21735294/dataset_recipe.txt). The results of `anemoi-datasets inspect` for this dataset are:
```shell
📦 Path          : ./glosea-o25-19990801-19990831-24h-v1.zarr
🔢 Format version: 0.30.0

📅 Start      : 1999-08-01 12:00
📅 End        : 1999-08-31 12:00
⏰ Frequency  : 1d
🚫 Missing    : 0
🌎 Resolution : None
🌎 Field shape: [1207, 1442]

📐 Shape      : 31 × 10 × 1 × 1,740,494 (2 GiB)
💽 Size       : 576.4 MiB (576.4 MiB)
📁 Files      : 105

   Index │ Variable │     Min │     Max │         Mean │       Stdev
   ──────┼──────────┼─────────┼─────────┼──────────────┼────────────
       0 │ siconc   │       0 │   0.997 │     0.104715 │      0.2897
       1 │ sivolu   │       0 │ 9.74737 │     0.126527 │    0.407793
       2 │ sohefldo │   -1329 │     769 │     -12.1304 │     79.4818
[...]

PR #478, opened 2025-08-13
Fix/one variable model

lzampier:
## Description

The current version of anemoi fails when models with just one prognostic variable are trained because of some .squeeze statements that reduce arrays to scalars. I use .reshape(-1) instead. This seems to fix the issue in my tests and still works with more complex models with multiple variables.

Enables models with one prognostic variable to be trained in anemoi-train. This is arguably an edge case, which should be enabled nevertheless.

This change addresses issue #477. 

### DIFF
--- a/training/src/anemoi/training/diagnostics/plots.py
+++ b/training/src/anemoi/training/diagnostics/plots.py
@@ -229,15 +229,15 @@ def plot_power_spectrum(
     grid_pc_lon, grid_pc_lat = np.meshgrid(regular_pc_lon, regular_pc_lat)
 
     for plot_idx, (variable_idx, (variable_name, output_only)) in enumerate(parameters.items()):
-        yt = y_true[..., variable_idx].squeeze()
-        yp = y_pred[..., variable_idx].squeeze()
+        yt = (y_true if y_true.ndim == 1 else y_true[..., variable_idx]).reshape(-1)
+        yp = (y_pred if y_pred.ndim == 1 else y_pred[..., variable_idx]).reshape(-1)
 
         # check for any nan in yt
         nan_flag = np.isnan(yt).any()
 
         method = "linear" if nan_flag else "cubic"
         if output_only:
-            xt = x[..., variable_idx].squeeze()
+            xt = (x if x.ndim == 1 else x[..., variable_idx]).reshape(-1)
             yt_i = griddata((pc_lon, pc_lat), (yt - xt), (grid_pc_lon, grid_pc_lat), method=method, fill_value=0.0)
             yp_i = griddata((pc_lon, pc_lat), (yp - xt), (grid_pc_lon, grid_pc_lat), method=method, fill_value=0.0)
         else:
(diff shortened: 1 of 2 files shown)


training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -229,15 +229,15 @@ def plot_power_spectrum(
     grid_pc_lon, grid_pc_lat = np.meshgrid(regular_pc_lon, regular_pc_lat)
 
     for plot_idx, (variable_idx, (variable_name, output_only)) in enumerate(parameters.items()):
-        yt = y_true[..., variable_idx].squeeze()
-        yp = y_pred[..., variable_idx].squeeze()
+        yt = (y_true if y_true.ndim == 1 else y_true[..., variable_idx]).reshape(-1)
```
anaprietonem, line 232: @lzampier - why do you need the if here - could just do y_true[..., variable_idx]).reshape(-1)? and same for x

lzampier: Probably yes. I was overcautious. Let me test later.

anaprietonem: Great, apart from that I think it could be good if we could have a test added here for the `print_variable_scaling` where we can see if it works fine for one and more than one variable https://github.com/ecmwf/anemoi-core/blob/5a311574b9fb3ab6d610e8e80fc3d91eab061a82/training/src/anemoi/training/losses/utils.py

could you give that a go? I can review later or if you have doubts just drop me a message!

lzampier: I checked, and yes we need the `(y_true if y_true.ndim == 1 else` part, otherwise the plotting fails. Regarding the tests, I have done an attempt that you can see in the newest commit. Let me know if you think it is appropriate. The behavior is a bit strange to test, so maybe you have a better idea.

training/tests/unit/train/test_print_variable_scaling.py
```diff
@@ -0,0 +1,129 @@
+# (C) Copyright 2024 Anemoi contributors.
```
anaprietonem: header should be 2025

training/tests/unit/train/test_print_variable_scaling.py
```diff
@@ -0,0 +1,129 @@
...
+    names = ["swh", "mwp"]
+    indices = _fake_indices(names)
+    shapes = [(2,), (1, 2), (2, 1), (1, 1, 2)]  # all contain exactly 2 elements
+
+    for shape in shapes:
+        base = np.array([1.0, 2.5]).reshape(shape)
+        loss = _FakeLoss(base)
+
+        handler = _ListHandler()
+        old_level = LOGGER.level
```
anaprietonem: why do we need this for the LOGGER?

training/tests/unit/train/test_print_variable_scaling.py
```diff
@@ -0,0 +1,129 @@
...
+    payload = message.removeprefix(prefix)
+    # The function ends with a trailing ", " — filter empty tail
+    parts = [p for p in payload.split(", ") if p]
+    return len(parts)
+
+
+# ------------------------------ Tests ---------------------------------------
+
+
+def test_print_variable_scaling_single_var_flattens_but_not_scalar() -> None:
```
anaprietonem: my proposal for the test would be to reuse the tests/unit/train/test_loss_scaling.py -test_variable_loss_scaling_vals 

There we can get the data_indices and loss and then call `print_variable_scaling(loss, indices)` - checking that the  test passes with one and many variables

anaprietonem (approved):
Nice fix, and tests have been included! GTM

PR #479, opened 2025-08-13
feat: flash attention v3

cathalobrien:
## Description
Adds a flash attention v3 option to models/layers/attention.py

flash attention v3 is optimised for hoppers and newer GPUs, with up to 2x faster attention kernels. For more info see [here](https://github.com/Dao-AILab/flash-attention?tab=readme-ov-file#flashattention-3-beta-release)

you can use flash-attn v3 with the following config entry
```yaml
model.processor.attention_implementation=flash_attention_v3
```

I compared the loss (seed=42) for SDPA vs flash-attn v3 and it matched to the 7th decimal place
Screenshot 2025-08-13 at 16 51 34

I also changed the error message reporting for the flash attention v2 wrapper, to actually print the error. This improves the user experience in the case that flash attn is installed but you need a newer gcc. Now anemoi will print the GLIBC error, rather then telling you to install flash attn when it's already installed

faster attention


mchantry:
> TODO will merge both flash_attention wrappers into one wrapper to minimise config complexity

Would be good to get output saying which version was chosen, for posterity.

cathalobrien:
> > TODO will merge both flash_attention wrappers into one wrapper to minimise config complexity
> 
> Would be good to get output saying which version was chosen, for posterity.

done, good catch

models/src/anemoi/models/layers/attention.py
mchantry, line 1: @cathalobrien does this now need updating?

models/src/anemoi/models/layers/attention.py
```diff
@@ -266,23 +293,53 @@ def forward(
...
+            if version.parse(flash_attn.__version__):
+                raise RuntimeError("Error: Flash-attn version is too low. Update to 2.6.0 or higher.")
+            else:
+                from flash_attn.layers.rotary import RotaryEmbedding
 
-            self.attention = flash_attn.flash_attn_func
+        self.attention = flash_attn_func
 
-        self.use_rotary_embeddings = use_rotary_embeddings
+        self.use_rotary_embeddings = (
```
mchantry: This is a second step of security, correct? Code should already exit if using rotary with v3?
(not saying this is a bad idea to have two steps)

models/src/anemoi/models/layers/attention.py
```diff
@@ -315,16 +372,28 @@ def forward(
...
+        else:
+            out = self.attention(
+                query,
+                key,
+                value,
+                causal=False,
+                window_size=(window_size, window_size) if window_size is not None else (-1, -1),
+                dropout_p=dropout_p,
+                softcap=softcap,
+                alibi_slopes=alibi_slopes,
```
mchantry, line 404: Do we need something to ensure users don't try to use alibi_slopers with v3?

models/src/anemoi/models/layers/attention.py
cathalobrien: yeah good point I can remove that

models/src/anemoi/models/layers/attention.py
cathalobrien: done

models/src/anemoi/models/layers/attention.py
```diff
@@ -126,6 +127,16 @@ def __init__(
         else:
             self.alibi_slopes = None
 
+        if self.attention_implementation == "flex_attention":
```
mchantry: @cathalobrien does this have a direct connection to flashv3?

mchantry: If it's useful it can still go in, just understanding.

cathalobrien: no oops, i was just playing with flex attention to see how it compares (performance seems noticeably worse) and forgot to remove this line. It's gone now

models/src/anemoi/models/layers/attention.py
```diff
@@ -244,15 +261,18 @@ def forward(
...
         sequence_len = query.shape[-2]
 
         if window_size is not None and (self.mask is None or tuple(self.mask.shape) != (sequence_len, sequence_len)):
             self.update_mask(sequence_len, window_size=window_size, device=query.device)
 
-        with torch.nn.attention.sdpa_kernel(backends=[torch.nn.attention.SDPBackend.MATH]):
+        from torch.nn.attention import SDPBackend
+        from torch.nn.attention import sdpa_kernel
+
+        with sdpa_kernel(backends=[SDPBackend.FLASH_ATTENTION, SDPBackend.EFFICIENT_ATTENTION, SDPBackend.MATH]):
```
mchantry: @cathalobrien why is this changing? Could you please add a comment to explain what this line does.

cathalobrien: before SDPA was limited to the MATH backend. I think that's too restrictive. With this change, SDPA will try use faster backends if they're available.

mchantry (approved):
Great contribution, thanks.

PR #486, opened 2025-08-17
docs: update diffusion documentation

ssmmnn11:
## Description
Fixes for diffusion documentation

----
📚 Documentation preview 📚: https://anemoi-training--486.org.readthedocs.build/en/486/

----
📚 Documentation preview 📚: https://anemoi-graphs--486.org.readthedocs.build/en/486/

----
📚 Documentation preview 📚: https://anemoi-models--486.org.readthedocs.build/en/486/

### DIFF
--- a/training/docs/user-guide/diffusion-set-up.rst
+++ b/training/docs/user-guide/diffusion-set-up.rst
@@ -78,7 +78,7 @@ The diffusion configuration includes:
 -  `rho`: Controls the noise schedule distribution
 -  `noise_embedder`: Sinusoidal embeddings for noise conditioning
 -  `inference_defaults`: Default parameters for noise scheduler and
-   sampler
+   sampler, these are not used during training.
 
 .. code:: yaml
 


training/docs/user-guide/diffusion-set-up.rst
```diff
@@ -75,6 +94,33 @@ The diffusion model uses conditional layer normalization to condition
 the latent space on the noise level, enabling the model to denoise
 appropriately at different noise scales.
 
+*************************
+ Inference configuration
+*************************
+
+The `inference_defaults` block specifies default parameters for
+sampling:
```
anaprietonem, line 102: Might be clear, but we could be more explicit and say - none of these parameters are considered during training so no need to modify the defaults when implementing the model?

ssmmnn11: it is mentioned now in the config description above.

PR #492, opened 2025-08-20
ci: Add checkpoint migrations workflow

gmertes:
## Description
The upcoming checkpoint migrations rely on the migration files having a reference to the next anemoi-models version that the migration is targeting. This is done with a placeholder `%NEXT_ANEMOI_MODELS_VERSION%` that is replaced on release time with the version being released.

This workflow does that. It runs on the release-please (DeployDuck) PRs, and it updates the placeholder in all the migration files and commits it to the release-please branch. If there are no migrations to update, it does nothing. Since release-please regularly force-pushes its branch to the latest state of main, any new migrations made since the opening of the PR will be automatically updated.

When the release PR is merged, the migrations go into main with the correct versions.

I made it a general workflow where we can put all future things to do at release time.

#386

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -1,0 +1,43 @@
+# This workflow runs on release PRs created by release-please (DeployDuck).
+# Here we can do extra things required before releasing.
+name: "[PR] Release"
+
+on:
+  pull_request:
+    branches:
+      - main
+      - hotfix/*
+    paths:
+      - .release-please-manifest.json
+    types: [opened, synchronize, reopened]
+
+jobs:
+  migrations:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


bdvllrs (approved):
Thanks a lot for this! I don't see any abvious issues with it and it looks good.

If I understand correctly, this runs whenever a PR targetting main is updating `.release-please-manifest.json`, i.e., the release-please PR.

bdvllrs:
Now that I think about it, would it also work if the version changes (first commit is for a minor version, but a breaking change is pushed and it's updated to a major version)? Because the second time, all of the `%NEXT_ANEMOI_MODELS_VERSION%` would have already been replaced and would keep the first prediction of the version.

bdvllrs:
Perfect then!

anaprietonem (approved):
LGTM!

Issue #484, opened 2025-08-15
More comprehensive integration tests in training


VeraChristina:
### Is your feature request related to a problem? Please describe.

We have added infrastructure for integration tests in training and have added tests to cover most uses cases, and a couple of additional tests (restart, restart from existing checkpoint, use existing graph, etc.). However, some aspects of training are not tested or only tested for a couple of use cases and some problems are missed because we use datasets with rless parameters for these tests.

### Describe the solution you'd like

A couple of things we could think about adding:
- [ ] more comprehensive tests for checkpoints / checkpoint migrations -- currently only testing gnn global
- [ ] include test for transformer - likely easier once Triton kernel is done
- [ ] tests for rollout -- currently not tested
- [x] add multi-gpu tests to test sharding for different models (probably partially covered in benchmark tests)
- [ ] review datasets used for testing, can we keep them small and still catch more of the potential problems?
- [ ] tests for forking runs (potentially better placed in system-level tests)
- [ ] test for transfer-learning and load_weights only
- [x] plotting callbacks

Depending on how comprehensively we want to test these, it might be enough to add a few tests, or it might be better to revisit the existing structure of fixtures to make them more reusable.

mchantry:
Small addition. Switch to minimal default output from integration tests, with an option to run debugging logger level. Currently identifying failures in CI is harder than needed.

cathalobrien:
FYI about transformer tests
```
include test for transformer - likely easier once Triton kernel is done
```
triton kernel has some numerics issues and its not so high priority. we can do transformer tests with SDPA tho since [this](https://github.com/ecmwf/anemoi-core/commit/2bf1842d7fc24c0eb1bb4e1d29093542f93d17f5) commit was merged. They won't be fast, so I wouldn't add benchmark tests, but they won't OOM and should give the right answer

PR #497, opened 2025-08-21
feat: Extend restart checkpoint to consider graphtransformer

anaprietonem:
## Description
Related to first point of https://github.com/ecmwf/anemoi-core/issues/484

Extend restart checkpoint test with graphtransformer as it was just considering the gnn. 

[
](https://github.com/ecmwf/anemoi-core/issues/484)

##  Additional notes ##

### DIFF
--- a/training/tests/integration/conftest.py
+++ b/training/tests/integration/conftest.py
@@ -236,23 +236,28 @@ def gnn_config(testing_modifications_with_temp_dir: DictConfig, get_tmp_paths: G
 
 @pytest.fixture
 def architecture_config_with_checkpoint(
+    request: pytest.FixtureRequest,
     architecture_config: tuple[DictConfig, str],
     get_test_data: GetTestData,
 ) -> tuple[DictConfig, str]:
     cfg, dataset_url = architecture_config
 
+    model_type = request._pyfuncitem.callspec.params["architecture_config"][0]
+
     # Pick checkpoint path based on the model
-    if "gnn" in cfg.model.architecture:
-        ckpt_name = ("anemoi-integration-tests/training/checkpoints/testing-checkpoint-gnn-global-2025-07-31.ckpt",)
-    elif "graphtransformer" in cfg.model.architecture:
-        ckpt_name = (
+    if "gnn" in model_type:
+        existing_ckpt = get_test_data(
+            "anemoi-integration-tests/training/checkpoints/testing-checkpoint-gnn-global-2025-07-31.ckpt",
+        )
+    elif "graphtransformer" in model_type:
+        existing_ckpt = get_test_data(
             "anemoi-integration-tests/training/checkpoints/testing-checkpoint-graphtransformer-global-2025-07-31.ckpt",
         )
+
     else:
         msg_error = f"Unknown architecture in config: {cfg.model.architecture}"
         raise ValueError(msg_error)
 
-    existing_ckpt = get_test_data(ckpt_name)
     checkpoint_dir = Path(cfg.hardware.paths.output + "checkpoint/dummy_id")
     checkpoint_dir.mkdir(parents=True, exist_ok=True)
     shutil.copy(existing_ckpt, checkpoint_dir / "last.ckpt")
--- a/training/tests/integration/conftest.py
+++ b/training/tests/integration/conftest.py
@@ -8,6 +8,7 @@
 # nor does it submit to any jurisdiction.
 
 
+import logging
 import os
 import shutil
 from pathlib import Path
@@ -20,6 +21,8 @@ from omegaconf import OmegaConf
 
 from anemoi.utils.testing import GetTestData
 from anemoi.utils.testing import TemporaryDirectoryForTestData
+
+LOGGER = logging.getLogger(__name__)
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


training/tests/integration/conftest.py
```diff
@@ -232,14 +235,29 @@ def gnn_config(testing_modifications_with_temp_dir: DictConfig, get_tmp_paths: G
...
+    architecture_config: tuple[DictConfig, str],
     get_test_data: GetTestData,
 ) -> tuple[DictConfig, str]:
-    cfg, dataset_url = gnn_config
-    existing_ckpt = get_test_data(
-        "anemoi-integration-tests/training/checkpoints/testing-checkpoint-global-2025-07-31.ckpt",
-    )
+    cfg, dataset_url = architecture_config
+
+    model_type = request._pyfuncitem.callspec.params["architecture_config"][0]
```
VeraChristina: Instead of accessing this here, how about returning "gnn" or "graphtransformer" in the architecture_config fixture?

anaprietonem: Yes I was wondering the best way to do this. That's a good idea and probably makes this test cleaner. That extra parameter won't be useful in tests like `test_training_cycle_architecture_configs` or `test_config_validation_architecture_configs` but if that's fine, then happy to make that change!

training/tests/integration/conftest.py
```diff
@@ -232,14 +235,29 @@ def gnn_config(testing_modifications_with_temp_dir: DictConfig, get_tmp_paths: G
...
+
+    # Pick checkpoint path based on the model
+    if "gnn" in model_type:
+        existing_ckpt = get_test_data(
+            "anemoi-integration-tests/training/checkpoints/testing-checkpoint-gnn-global-2025-07-31.ckpt",
```
anaprietonem, line 255: Also note that for now, this is a copy of the existing ckpt in the bucket with the name updated to account for the different architectures and still not break the current tests. Once we are happy with the changes here I can take care of cleaning the checkpoints bucket folder

training/tests/integration/conftest.py
VeraChristina: Yes, I'd prefer to pass it explicitly if we need it here, and I think it's fine to just ignore the extra parameter in the other tests. There are probably other ways but this seems fine

anaprietonem: I changed this!

anaprietonem:
@VeraChristina @bdvllrs, this branch is now updated with the latest main (that includes the checkpoint migration functionality). The two ckpt in the bucket has the migrations key (gnn and graphtransformer), and the tests passed okey https://github.com/ecmwf/anemoi-core/actions/runs/17154626521

training/tests/integration/conftest.py
```diff
@@ -84,6 +87,7 @@ def architecture_config(
     get_tmp_paths: GetTmpPaths,
 ) -> tuple[DictConfig, str]:
```
bdvllrs: Update the type annotation here to `tuple[DictConfig, str, str]`

training/tests/integration/conftest.py
```diff
@@ -238,15 +242,25 @@ def migrator() -> Migrator:
 
 
 @pytest.fixture
-def gnn_config_with_checkpoint(
+def architecture_config_with_checkpoint(
     migrator: Migrator,
-    gnn_config: tuple[DictConfig, str],
+    architecture_config: tuple[DictConfig, str],
```
bdvllrs: Same here, the type annotation here should be `tuple[DictConfig, str, str]`

training/tests/integration/test_training_cycle.py
```diff
@@ -31,13 +31,13 @@ def test_training_cycle_architecture_configs(
     architecture_config: tuple[DictConfig, str],
     get_test_archive: GetTestArchive,
 ) -> None:
-    cfg, url = architecture_config
+    cfg, url, _ = architecture_config
     get_test_archive(url)
     AnemoiTrainer(cfg).train()
 
 
 def test_config_validation_architecture_configs(architecture_config: tuple[DictConfig, str]) -> None:
```
bdvllrs: Update type hint

training/tests/integration/test_training_cycle.py
```diff
@@ -31,13 +31,13 @@ def test_training_cycle_architecture_configs(
     architecture_config: tuple[DictConfig, str],
```
bdvllrs: Update type hint

training/tests/integration/test_training_cycle.py
```diff
@@ -156,8 +156,8 @@ def test_restart_training(gnn_config: tuple[DictConfig, str], get_test_archive:
 
 
 @skip_if_offline
-def test_loading_checkpoint(gnn_config_with_checkpoint: DictConfig, get_test_archive: callable) -> None:
-    cfg, url = gnn_config_with_checkpoint
+def test_loading_checkpoint(architecture_config_with_checkpoint: DictConfig, get_test_archive: callable) -> None:
```
bdvllrs: Change type hint for `architecture_config_with_checkpoint` to `tuple[DictConfig, str]`

bdvllrs (changes requested):
Hi, except from the type hints mismatch, this looks good to me!

anaprietonem:
thanks @bdvllrs for the catch! i had completely missed updating the type hints, should be good now

bdvllrs (approved):
Thanks! All looks good now :slightly_smiling_face:

PR #508, opened 2025-08-26
feat: Allow custom MLFLOW_SYSTEM_METRICS_SAMPLING_INTERVAL

halvarsu:
Can be set either through config or via environment variable

## Description
The MLFlow system metrics logging interval was reduced to 100s in #333, but it is hard coded and overwrites the environment variable. This PR allows changing the value either through the config or via setting the environment variable MLFLOW_SYSTEM_METRICS_SAMPLING_INTERVAL

new feature I guess?

No issue

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/config/diagnostics/evaluation.yaml
+++ b/training/src/anemoi/training/config/diagnostics/evaluation.yaml
@@ -50,6 +50,7 @@ log:
     experiment_name: 'anemoi-debug'
     project_name: 'Anemoi'
     system: False
+    system_metrics_sampling_interval: 100 # in seconds
     terminal: True
     run_name: null # If set to null, the run name will be the a random UUID
     on_resume_create_child: True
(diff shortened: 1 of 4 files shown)


anaprietonem:
hey @halvarsu what's your use case or motivation to need to increase the sampling for the system metrics? We realised that a higher sampling means runs are very data heavy and saturates our server and database. That's why we prefer to simply reduce the default sampling frequency and not make this a parameter that can be controlled by users

halvarsu:
I am trying to debug some OOM issues and would like to control the sampling frequency to see the trend in memory usage. I don't mind removing the config option to make it harder for users to set the variable, it works just as well for my use case to set the option with an environment variable. However, this is not possible at thee moment as it is hard coded to overwrite the environment variable.

I am working with short jobs (~10 min) on a local server, so the servers and databases should not be affected.

PR #512, opened 2025-08-29
chore(training): Add explicit encoder/decoder shard_strategy in all model configs

japols:
## Description
Consistently add shard_strategy to all model configs.

##  Additional notes ##


anaprietonem:
> Did you consider making this setting mandatory on the Pydantic side?

Hey Marek,  regarding your second question so when Jan introduced this feature https://github.com/ecmwf/anemoi-core/pull/366 the design was to make a new config entry 'shard_strategy' but so that it was not a breaking-change. (ie - if users have that specified in their configs then that's used otherwise the default strategy ('edges') defined in the code is used. ). If we were to make this a mandatory setting in Pydantic that would mean this key would be a config breaking change and hence for those using 'older configs' where there is no 'shard strategy' the code will fail and request the pass this explicitly.  So far I believe we support two types of strategies (Jan feel free to chime in here pls) and 'edges' is the most performant one so having it predefined in the code and as a not-breaking change seems to me a good compromise for users to take benefit or it and reduce the possibilities they choose the wrong value for this field.

mchantry (approved):

PR #515, opened 2025-09-01
Fix dry run check to use current run object

phinate:
A quick fix to something found when setting

```yaml
diagnostics:
  log:
    mflow:
      on_resume_create_child: False
```

I don't think this code path was previously explored by tests (it's an undefined variable error) — probably a bigger issue to review how the MLFlow logging is being tested.


anaprietonem (approved):
LGTM! thanks for spotting the bug Nathan. And indeed this part of the mlflow logger don't have great test coverage as you can see from https://github.com/ecmwf/anemoi-core/blob/main/training/tests/unit/diagnostics/mlflow/test_mlflow_logger.py. If you have any thoughts on how to extend this it would be good to discuss them. We should probably add tests for the AzureMLflow logger. 

I am glad to hear that setting the flag 'on_resume_create_child' to False helps with the issue you were seeing! If in the future you want to have nested runs for resuming in Azure we might need to rethink this logic.

PR #517, opened 2025-09-02
test(training): Add option for caching downloaded integration test data

matschreiner:
## Description
This PR adds an option for caching data downloaded for integration tests. When enabled, datasets are downloaded and extracted into a which can be specified as an environment variable `$ANEMOI_TEST_CACHE` or otherwise will default to `$home/.anemoi_test_cache` , so subsequent test runs reuse the same data instead of re-downloading/re-extracting. 
This greatly improves the ability to run the tests locally as otherwise downloading the data takes ~10 minutes on a cabled connection. 

To run the tests with caching run: 
`pytest training/tests/integration/test_training_cycle.py::test_training_cycle_lam --cache`

----
📚 Documentation preview 📚: https://anemoi-training--517.org.readthedocs.build/en/517/

----
📚 Documentation preview 📚: https://anemoi-graphs--517.org.readthedocs.build/en/517/

----
📚 Documentation preview 📚: https://anemoi-models--517.org.readthedocs.build/en/517/


training/tests/integration/conftest.py
```diff
@@ -26,6 +27,16 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def pytest_addoption(parser: Parser) -> None:
+    group = parser.getgroup("anemoi")
+    group.addoption("--cache", action="store_true", default=False, help="Enable caching of test datasets")
+
```
VeraChristina: Could you move adding the cli option to the top level conftest.py, i.e. in `tests/`?

training/tests/integration/conftest.py
```diff
@@ -18,6 +18,7 @@
 from hydra import initialize
 from omegaconf import DictConfig
 from omegaconf import OmegaConf
+from pytest import Parser
```
VeraChristina: Oh, and just importing pytest and then using pytest.Parser should make the pre-commit happy :)

training/tests/integration/conftest.py
```diff
@@ -45,8 +56,10 @@ def testing_modifications_with_temp_dir(tmp_path: Path) -> DictConfig:
 
 
 class GetTmpPaths:
-    def __init__(self, temporary_directory_for_test_data: TemporaryDirectoryForTestData) -> None:
+    def __init__(self, temporary_directory_for_test_data: TemporaryDirectoryForTestData, cache_path: str = "") -> None:
         self.temporary_directory_for_test_data = temporary_directory_for_test_data
+        if cache_path:
+            self.temporary_directory_for_test_data.base_dir = Path(cache_path)
```
VeraChristina, line 61: Potentially worth adding a comment here, as this deviates from the intended use of the temporary_directory_for_test_data fixture

matschreiner: Right, I'm happy to add a comment, but I'm not sure what it should say :)

VeraChristina:
Just adding here for context and visibility of our earlier discussion about more explicit handling of temp versus cached paths.

Setting the self.temporay_directory_for_test_data.base_dir in GetTmpPaths works to cache the downloaded testing data. It would be more clean to more explicitly handle temp versus cached paths, and to not just overwrite the base dir in the fixtures that handle temp dirs.

However, that might require changes in utils. Will think about other options and get back to you.

training/tests/integration/conftest.py
```diff
@@ -45,8 +55,10 @@ def testing_modifications_with_temp_dir(tmp_path: Path) -> DictConfig:
 
 
 class GetTmpPaths:
```
VeraChristina, line 57: Can we rename this to make it more explicit that these are not necessarily tmp paths? Something like GetTestDataPaths

training/tests/integration/conftest.py
```diff
@@ -71,8 +83,12 @@ def __call__(self, config: DictConfig, list_datasets: list[str]) -> tuple[str, l
 
 
 @pytest.fixture
-def get_tmp_paths(temporary_directory_for_test_data: TemporaryDirectoryForTestData) -> GetTmpPaths:
-    return GetTmpPaths(temporary_directory_for_test_data)
+def get_tmp_paths(temporary_directory_for_test_data: TemporaryDirectoryForTestData, cache_enabled: bool) -> GetTmpPaths:
```
VeraChristina, line 86: Can we rename this function as well for clarity (see above)?

training/tests/integration/conftest.py
```diff
@@ -45,8 +55,10 @@ def testing_modifications_with_temp_dir(tmp_path: Path) -> DictConfig:
 
 
 class GetTmpPaths:
-    def __init__(self, temporary_directory_for_test_data: TemporaryDirectoryForTestData) -> None:
+    def __init__(self, temporary_directory_for_test_data: TemporaryDirectoryForTestData, cache_path: str = "") -> None:
```
VeraChristina, line 58: Small nitpic: could we default cache_path to `None` instead of ""? Makes the intent a bit clearer

training/tests/integration/conftest.py
```diff
@@ -71,8 +83,12 @@ def __call__(self, config: DictConfig, list_datasets: list[str]) -> tuple[str, l
 
 
 @pytest.fixture
-def get_tmp_paths(temporary_directory_for_test_data: TemporaryDirectoryForTestData) -> GetTmpPaths:
-    return GetTmpPaths(temporary_directory_for_test_data)
+def get_tmp_paths(temporary_directory_for_test_data: TemporaryDirectoryForTestData, cache_enabled: bool) -> GetTmpPaths:
+    cache_path = ""
```
VeraChristina, line 87: as above, could we default cache_path to `None` instead of ""?

training/tests/integration/conftest.py
VeraChristina: Something like this maybe?
```
# NOTE: Overwriting base_dir here to enable caching.
# This is a workaround; see issue #XYZ in utils for a more robust approach.
```

VeraChristina (commented):
Thanks for adding the docs!

I have added one small nitpic in the comments below (on defaulting cache_path to `None`).

On the bigger picture: since temporary_directory_for_test_data.base_dir wasn’t really meant to be overwritten, this approach is a bit of a workaround. But since changing this properly in utils may take more coordination, how about we go with this approach, make the workaround a little more explicit, and add an issue in utils to add a caching option to the testing data utilities?

To make it clearer, I’d suggest (1) renaming the GetTmpPaths class to something like GetTestDataPaths to not disguise the not-actually-tmp paths, and (2) adding a short comment above the base_dir assignment noting that it’s a temporary workaround. It’d also be great to (3) open an issue in utils to track adding a proper caching mechanism there longer term.

PR #532, opened 2025-09-05
feat: add step scalers

bdvllrs:
## Description
This allows setting different scalers for different levels.

This can be useful:
1. to set to 0 coef just part of the levels of a variable
2. to define different scalers for different ranges of levels
3. or at the extreme case, to set any function by setting a constant value for each level

I originally made this for 1. and debugging potentially out-of-distribution levels of a variable.

Checklist:
- [x] implement the feature
- [ ] add tests
- [ ] add docs

##  Additional notes ##


training/src/anemoi/training/losses/scalers/variable_level.py
```diff
@@ -81,62 +76,155 @@ def get_scaling_values(self, **_kwargs) -> torch.Tensor:
...
     """Polynomial scaling, (slope * variable_level)^2, yaxis shift by self.y_intercept."""
 
-    def get_level_scaling(self, variable_level: float) -> torch.Tensor:
+    def get_level_scaling(self, variable_level: float) -> float:
         return (self.slope * variable_level) ** 2 + self.y_intercept
 
 
-class NoVariableLevelScaler(BaseVariableLevelScaler):
-    """Constant scaling by 1.0."""
+class ConstantLevelScaler(BaseVariableLevelScaler):
```
mc4117, line 156: I'm not sure the constant level scaler is needed? Because you could just use the novariablelevel scaler and then control the value of that scaler in the normal variable scalings?

training/src/anemoi/training/losses/scalers/variable_level.py
```diff
@@ -81,62 +76,155 @@ def get_scaling_values(self, **_kwargs) -> torch.Tensor:
...
             # Apply variable level scaling
             assert variable_level is not None, f"Variable {variable_name} has no level to scale."
-            variable_level_scaling[idx] = self.get_level_scaling(float(variable_level))
+            variable_level_scaling[idx] = self.get_level_scaling(variable_level)
 
         return variable_level_scaling
 
 
-class LinearVariableLevelScaler(BaseVariableLevelScaler):
+class BaseLinearVariableLevelScaler(BaseVariableLevelScaler):
```
mc4117, line 94: I think there are some things missing in this new class that were there previously like the logging message?

mc4117:
I like the idea of doing a step level scaler! However I'm wondering if it could not already be implemented using the relu level scaler with gradient 0 and setting the minimum value? But with this new functiontionality it will also probably be cleaner

training/src/anemoi/training/losses/scalers/variable_level.py
bdvllrs: Do you mean the logging of the interceptions/slope?

In that case, it's still in the LinearVariableLevelScaler class. I don't think it makes sense to keep it to the base class that deosn't necessarily have a slope/intercept.

bdvllrs:
Hi! Thanks for the comments. However I don't understand how you can get the same behavior with the ReLU scaler. With the step scaler you can have for many different groups of levels have different scalers (with potentially different slopes, intercepts). 

Can you give me an example how you could have the same behavion with the relu scaler?

bdvllrs:
Hi @mc4117, thanks for your answer, this indeed works with a caveat: the schemas currently doesn't allow for ints or bools (https://github.com/ecmwf/anemoi-core/blob/e049e839e3a1a1abf293ae5b7d8b2e9d7036efae/training/src/anemoi/training/schemas/training.py#L348).
I'm opening a new PR #571 to fix this.

In my opinion, this makes this PR irrelevant: I like the current way of doing it more, and I think it's better to not have multiple ways of achieving the same thing, so I'll be closing this PR.

PR #537, opened 2025-09-10
fix(training): nan mask weights gather for plots

japols:
## Description
Avoid gathering unresolved `nan_mask_weight` scaler with `grid_dim=1`.

Fixes an bug where PlotLoss attempted to gather a nan_mask_weight scalar that wasn’t actually sharded but simply constant across the dimension (shape 1).

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/diagnostics/callbacks/plot.py
+++ b/training/src/anemoi/training/diagnostics/callbacks/plot.py
@@ -918,7 +918,7 @@ class PlotLoss(BasePerBatchPlotCallback):
 
             self.loss = copy.deepcopy(pl_module.loss)
 
-            if hasattr(self.loss.scaler, "nan_mask_weights"):
+            if hasattr(self.loss.scaler, "nan_mask_weights") and self.loss.scaler.nan_mask_weights.shape[2] != 1:
                 self.loss.scaler.nan_mask_weights = pl_module.allgather_batch(self.loss.scaler.nan_mask_weights)
 
             super().on_validation_batch_end(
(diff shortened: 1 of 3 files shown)


training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -918,7 +918,8 @@ def on_validation_batch_end(
 
             self.loss = copy.deepcopy(pl_module.loss)
 
-            if hasattr(self.loss.scaler, "nan_mask_weights"):
+            # gather nan_mask_weights along grid dim if necessary, don't gather if grid dim is broadcastable
+            if hasattr(self.loss.scaler, "nan_mask_weights") and self.loss.scaler.nan_mask_weights.shape[-2] != 1:
```
anaprietonem: @japols what do you mean here with broadcastable? Is this related to having sharding activated, probably could be good to clarify for users

sahahner:
That is indeed a good point. We can remove the scaler for the lam and stretched grid scalers. 
Nevertheless, the general idea is to also define scalers that are not necessarily used (for example, the tendency scalers). Since the nan mask scaler is well tested for the global model, and I know users (not only myself :) ) who already use it, I would like to leave it in [global.yaml](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/config/training/scalers/global.yaml).

mchantry:
> Does it make sense to also comment out the `nan_mask_weights` in the scalers config if it is no longer being used?

@havardhhaugen there is a style at present to create the dictionary irrespective of use, and then just choose which ones to use in the training config. What you propose is reasonable but would break this style.

anaprietonem:
PR missing another fix since output mask in on cpu when using plots as reported by Fernando (AEMET) - Jan planning to fix this in this PR and then we can finish review and merge

anaprietonem:
@icedoom888  this is related to the issue you reported https://github.com/ecmwf/anemoi-core/issues/603 - could you please test this branch??

I run a simple LAM test with sharding and plots https://mlflow.ecmwf.int/#/experiments/264/runs/94dde82af22b4dc3a5ab5c4b7ac3d589

PR #544, opened 2025-09-14
feat: 402 time interpolation mass conserving interpolation for accumulated fields

### LINKED ISSUE #402: Time Interpolation: Mass-Conserving Interpolation for Accumulated Fields

<h1><strong>Mass-Conserving Interpolation for Accumulated Fields</strong></h1>

<p>When down-scaling <strong>accumulated</strong> variables—such as total precipitation (<code inline="">tp</code>) and convective precipitation (<code inline="">cp</code>)—from a 6-hourly forecast to hourly resolution, naïve interpolation yields hourly values that <strong>do not sum to the original 6-hour total</strong>.<br>
This breaks water-balance conservation and can distort any downstream verification or post-processing steps that integrate precipitation over time.</p>

### Proposed Algorithm

1. **Predict per-hour logits**
   Generate one raw score $z_{t+k}$ for each hourly sub-interval, $k = 1,\dots,6$.

2. **Softmax to obtain allocation weights**

   $$w_{t+k} = \frac{\exp(z_{t+k})}{\sum_{j=1}^{6}\exp(z_{t+j})}, \quad k = 1,\dots,6.$$

3. **Scale by the 6-hour accumulation**

   $$\Delta A_{t+k} = w_{t+k}\, A_{t+6}, \quad k = 1,\dots,6.$$

### Properties

* **Mass-conserving**:

  $$\sum_{k=1}^{6}\Delta A_{t+k} = A_{t+6}$$

  (since softmax ensures $\sum_{k=1}^{6} w_{t+k} = 1$).

* **Non-negative**: Each hourly increment is guaranteed non-negative, $\Delta A_{t+k}\ge 0$.

* **Differentiable**: The entire interpolation procedure is end-to-end differentiable, allowing gradient-based optimization.

Any alternative implementations for the 1st and 2nd point?

### Additional context

_No response_

### Organisation

_No response_

roxas6path:
closes #402

## Description
Context and Discussion on PR -->: #402

- Non Breaking change only affects Time Interpolator

- Implements the mass conservation for temporal interpolation of accumulated variables
- Implements a processor to ensure input accumulated data has the correct format
- Updates the interpolator config

## Summary of Changes
- Mass conservation logic
- pydantic checks to ensure that mass_conservation is setup correctly, with helpful error statements
- ZeroOverwriter to set LHS boundary to 0 for specific variables
- Adds Tests for ZeroOverwiter postprocessor, Mass Conservation logic
- Adds Docs: Getting Stated with interpolator
- Adds Docs: ZeroOverwriter Postprocessor
- Ratify config.data.timestep purpose

#402
##  Additional notes ##

@Magnus-SI @radiradev @frazane @sahahner 

----
📚 Documentation preview 📚: https://anemoi-training--544.org.readthedocs.build/en/544/

----
📚 Documentation preview 📚: https://anemoi-graphs--544.org.readthedocs.build/en/544/

----
📚 Documentation preview 📚: https://anemoi-models--544.org.readthedocs.build/en/544/

### DIFF
--- a/training/src/anemoi/training/config/interpolator.yaml
+++ b/training/src/anemoi/training/config/interpolator.yaml
@@ -5,7 +5,7 @@ defaults:
 - diagnostics: evaluation
 - hardware: example
 - graph: multi_scale
-- model: transformer
+- model: graphtransformer
 - training: interpolator
 - _self_
 


training/src/anemoi/training/config/diagnostics/evaluation.yaml
```diff
@@ -36,7 +36,7 @@ log:
     offline: False
     log_model: False
     project: 'Anemoi'
-    entity: ???
+    entity: null
```
mchantry: @Rilwan-Adewoyin could you move this change into a tiny separate PR please. It's the right thing to do, but not appropriate here.

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -5,17 +5,91 @@ defaults:
...
-- model: graphtransformer
+- model: transformer
 - training: interpolator
 - _self_
 
 config_validation: True
 
 data:
   frequency: 1h
+  timestep: 1h
```
radiradev: Isn't timestep ignored by the interpolator? Should set this to null or None?

models/src/anemoi/models/models/interpolator.py
radiradev, line 1: In this PR and as well as the original implementation assumes that the forcing indices are on the boundary (which makes sense). Should we add a check for this?

e.g x1, x2, x3, x4, x5 = model(x0, x6) works

but something like  x0, x3, x4 = model(x1, x2) would likely not work

models/src/anemoi/models/models/interpolator.py
```diff
@@ -180,3 +186,193 @@ def forward(
...
+        # NOTE: When interpolating between 0 and 6, this makes outputs for 1, 2,3 ,4, 5, 6
+        input_constraint_indxs = self.map_accum_indices["constraint_idxs"]
+        target_indices = self.map_accum_indices["target_idxs"]
+
+        # (B, T, …, V_acc)
+        logits = y_preds[..., target_indices]  # (B,T,E,G,V_acc)
+        zeros = torch.zeros_like(logits[:, 0:1])
+        weights = F.softmax(torch.cat([logits, zeros], dim=1), dim=1)  # shape (B, T=interp_steps+1, E, G, V_acc)
+
+        if not include_right_boundary:
```
radiradev, line 344: When would the right boundary not be included? I don't see how that ensures that the quantity is conserved

training/src/anemoi/training/config/test_benchmark.py
mc4117, line 1: I think this was probably uploaded by mistake?

models/src/anemoi/models/models/interpolator.py
```diff
@@ -180,3 +186,193 @@ def forward(
...
+
+            y_preds = torch.cat([y_preds, torch.zeros_like(y_preds[:, 0:1])], dim=1)
+
+            # Add the energy conserved variables to the right boundary
+
+            constraints = x_input[:, -1:, ..., input_constraint_indxs]  # shape (B, 1, E, G, V_acc)
+
+            y_preds_accum = weights * constraints  # shape (B, T=interp_steps+1, E, G, V_acc)
+
+            y_preds[:, -1:, ..., y_index_ex_target_indices] = x_input[
```
mc4117, line 390: Am I correct in understanding that this corrects the last step so the precipitation adds up to 6hrs?

models/src/anemoi/models/models/test_benchmark.py
mc4117, line 1: wrongly uploaded?

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -5,17 +5,91 @@ defaults:
...
 - hardware: example
 - graph: multi_scale
-- model: graphtransformer
+- model: transformer
 - training: interpolator
 - _self_
 
 config_validation: True
 
 data:
```
mc4117: I would maybe put this into its own timeinterpolator config under data as it's a bit difficult to understand

training/src/anemoi/training/config/test_benchmark.py
roxas6path: Thanks, so yeah I'm not really sure what this file is, I thought it appeared while I was rebasing on to main. In any case thanks I'll delete it . Done in commit d97d5c2c52bd38180994765f50004610403a0326

models/src/anemoi/models/models/interpolator.py
roxas6path: kinda yes - and more generally it holds for any of the accumulated variables.

In more detail (using 6hr -> 1hr eg.) this "include_right_boundary" is only used during inference where we need to predictions for the intermediate interpolation steps 1,2,3,4,5 and the boundary window 6. For instaneous values, we can just copy the values from the boundary window. For accumulated variables we need to do the correction you've mentioned

roxas6path: I'll clean the comments and annotate the steps better, also I need to update the 'energy conservation' to 'mass conservation'

training/src/anemoi/training/config/interpolator.yaml
roxas6path: So have a model dependent config in the data folder?

models/src/anemoi/models/models/test_benchmark.py
roxas6path: Yep, thanks deleted it

training/src/anemoi/training/config/diagnostics/evaluation.yaml
roxas6path: Yep updated this

training/src/anemoi/training/config/interpolator.yaml
mc4117: I see your point. My feeling here would be that it's justified but happy to be contradicted

training/src/anemoi/training/config/interpolator.yaml
roxas6path: So I believe this was incorrect previously - timestep should be accounted for. There have been changes in the inference code to factor this in as well

radiradev: Could you point me to where this is used in the code? And why is this 1h instead of 6h?

models/src/anemoi/models/models/interpolator.py
roxas6path: Okay - I can add some assertions during interpolator init which checks that the min(input_times)<min(target_times) and  max(input_times) > max(target_times) and input_time[0] == 0

I think this should enforce behaviour to normal interpolation, i.e. calculating intermediate values given a sequence of data

training/src/anemoi/training/config/interpolator.yaml
roxas6path: anemoi-core (datasets) : https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/data/dataset/singledataset.py

anemoi-core (datamodules) : https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/data/datamodule/singledatamodule.py

anemoi-inference: I can check on Monday, but I think theres logic in here which also forms Mars request in a way along the lines of steps=target_time[:]*timestep

My understanding is that across the anemoi-core and anemoi-inference I think  data.frequency represents the temporal increment of the underlying data and data.timestep represents what what time increment we want our model to operate on. When training a 6 hourly --> 1 hourly model, we use an underlying dataset with freq 1hour and we want our model to operate on hourly data. 

This is reflected further in our config, when we use the target_time = [0,6] and input_time=[1,2,3,4,5] the units for these times are (or should be) the timestep parameter. 
E.g. We want the `0th*1hr` and `6th*1hr` as inputs, and then we want to target the `1st*1hr`, `2nd*1hr` , ... etc

models/src/anemoi/models/models/interpolator.py
roxas6path: Hey just saw this now - so:
- Generally during training the right_boundary isn't needed as output since we calc losses only on the intermediary interpolated steps

- Prior to the conditional branching, based on include_right_boundary, to ensure the quantity is conserved correctly, the line `weights = F.softmax(....)` is performed. At this stage weights are calculated for steps 1 to 6, ensuring they all add to 1 e.g. weights w_1, w_2, w_3, w_4, w_5, w_6 all add to 1 and  -->&-->  interpolated output accumulation values = target_i  = \sum(w_i * right_boundary).

- The right boundary is include for inference when we need to output interpolation steps 1,2,3,4,5 and the final step 6. At final step 6 only the accumulated vars will changed as they get converted from 6hr accums to 1hr accums.

training/src/anemoi/training/config/interpolator.yaml
radiradev: For datasets and datamodule the argument is passed around but I don't see it being used in any logic. For `anemoi-inference` the code has been refactored to not include `timestep` logic because I thought it was an argument that could be ambiguous.

models/src/anemoi/models/preprocessing/overwriter.py
```diff
@@ -0,0 +1,96 @@
...
+        self.num_inference_input_vars = len(self.data_indices.model.input.name_to_index)
+        self._train_time_to_idxs: dict[int, list[int]] = {}
+        self._infer_time_to_idxs: dict[int, list[int]] = {}
+
+        train_map = self.data_indices.data.input.name_to_index
+        infer_map = self.data_indices.model.input.name_to_index
+
+        # Accept config.groups as:
+        # - list[{"vars": [...], "time_index": [int]}]
+        if not isinstance(groups, list):
```
sahahner: Do the pydantic schemas already catch these list/dict missmatches?

Would it be possible to move the asserts into a config/indices validate function as we did it for the [postprocessors](https://github.com/ecmwf/anemoi-core/blob/c6930d9acf5d5574961946c6516c1f11de2efefc/models/src/anemoi/models/preprocessing/postprocessor.py#L55), for example?

models/src/anemoi/models/preprocessing/overwriter.py
```diff
@@ -0,0 +1,96 @@
...
+
+            missing_train = [v for v in vars_list if v not in train_map]
+            missing_infer = [v for v in vars_list if v not in infer_map]
+            if missing_train or missing_infer:
+                LOGGER.debug("SetToZero: variables not found (train=%s, infer=%s)", missing_train, missing_infer)
+
+            self._train_time_to_idxs.setdefault(time_index, [])
+            self._infer_time_to_idxs.setdefault(time_index, [])
+
+            for idx in train_idxs:
```
sahahner, line 80: Does the list of data indices to be set to zero look the same for every time index?

models/src/anemoi/models/preprocessing/overwriter.py
```diff
@@ -0,0 +1,96 @@
...
+
+from torch import Tensor
+
+from anemoi.models.data_indices.collection import IndexCollection
+from anemoi.models.preprocessing import BasePreprocessor
+
+LOGGER = logging.getLogger(__name__)
+
+
+class SetToZero(BasePreprocessor):
```
sahahner: Could we go for ZeroOverwriter here? For the other preprocessors, we repeat the file name in the class name. It would be nice to keep this symmetric, I think.

models/src/anemoi/models/models/interpolator.py
```diff
@@ -180,3 +186,245 @@ def forward(
...
+
+        Returns
+        -------
+        Tensor
+            Model output (after post-processing)
+        """
+        with torch.no_grad():
+
+            assert (
+                len(batch.shape) == 5
```
sahahner, line 234: asserting for a 5-dimensional shape here, but the error message asks for 4. Is that correct?

training/src/anemoi/training/config/interpolator.yaml
roxas6path: commit 4875b2d5402696490f7ff33fab2ce57ab3589ef0 moves the data related config settings from the base config to a config under data, with a setup assuming only two accumulated fields exist "tp" and "cp".

Also further explanation is added to help the user understand how to set it up.

lmk if this looks good to you

training/src/anemoi/training/config/interpolator.yaml
Rilwan-Adewoyin: Hmm I do think I remember us discussing this a while ago and us not being sure on what timestep should be set to.

Expanding on my previous message timestep is used in the following ways:

Training and dataloading (anemoi-core)

- Schema definition

    - training/src/anemoi/training/schemas/data.py:

        - DataSchema.timestep declared as a string field; user-provided config value is made available to runtime code.
        
- Validation and conversion (seconds), ratio vs frequency, relative time indices
    - training/src/anemoi/training/data/datamodule/singledatamodule.py:
        - timeincrement cached property converts config.data.timestep and config.data.frequency to seconds, asserts timestep is a multiple of frequency, logs, and returns the ratio used to step through dates.
relative_date_indices() builds the time indices list using self.timeincrement (which is based on timestep/frequency) and training.multistep_input/rollout settings.
_get_dataset(...) passes timestep=self.config.data.timestep to the dataset constructor.
training/src/anemoi/training/data/datamodule/ensdatamodule.py:
Passes through timestep=self.config.data.timestep to the ensemble dataset.
Dataset consumption
training/src/anemoi/training/data/dataset/singledataset.py:
Constructor stores self.timestep = timestep passed in from the datamodule.
statistics_tendencies() tries to call self.data.statistics_tendencies(self.timestep) which uses the configured timestep to fetch tendency statistics.
training/src/anemoi/training/data/dataset/ensdataset.py:
Inherits and forwards timestep as above.
Note: The training integration configs reference timestep (e.g.,

models/src/anemoi/models/models/interpolator.py
roxas6path: Thanks for catching this, the assertion statement should read "5-dimensional tensor", this has been addressed in commit 28322eeada4d10e8a0f91639bf37dafa8b34e552.
lmk what you think

training/src/anemoi/training/config/interpolator.yaml
roxas6path: Hi @radiradev, i think it'd be good to discuss this over a video call next week?

I agree that in parts of the code timestep's utility has been removed - it'd be good to discuss if there is an implementation which allows config.data.timestep and config.data.frequency to have the same meaning/functionality in time interpolator as in other models and then whether or not this implementation is worth even following compared to what is currently in place

There are still areas where having config.data.timestep operate differently for time interpolator meant extra patches had to be made [when trying to utilise anemoi-inference / prepML to pre-load, to file, all MARS data required for an inference run] so it'd be good to discuss all these factors together.

models/src/anemoi/models/models/interpolator.py
roxas6path: Hey @radiradev so I've implemented this via pydantic checks - you can check out the implementation in commit dedb040403530f55f6769aa70a5afc7b76abcb69

models/src/anemoi/models/preprocessing/overwriter.py
roxas6path: Yep good idea, I've implemented this in commit 42802c15d241d98e9a41bc3eaf37e14d37adce44 .

models/src/anemoi/models/preprocessing/overwriter.py
roxas6path: Hey so the way SetToZero is implemented the objects  `self._train_time_to_idxs` and `self._infer_time_to_idxs` collate all combinations of time_index and variable_index which must be set to 0 e.g. in the code block https://github.com/ecmwf/anemoi-core/blob/42802c15d241d98e9a41bc3eaf37e14d37adce44/models/src/anemoi/models/preprocessing/overwriter.py#L70C1-L75C69

you can see that `self._train_time_to_idxs` and `self._infer_time_to_idxs` hold information for all groups, by essentially each storing a dictionary dict[key, list[int]] where the key is the time_index and the list[int] represents all variables that must be set to 0 at that time index.

models/src/anemoi/models/preprocessing/overwriter.py
roxas6path: So this change has been implemented in commit b3711da57904151fa8c52b9554b5aefee767c5bd

Rilwan-Adewoyin:
> Hi Rilwan, realising the Zero-Overwriter in the preprocessor logic is very fitting. Would you mind including a few tests so that there are some example configurations one can have a look at? Eventually, this should also be covered in a few lines of documentation.

1) So on the tests I have added these in commit bba4129478cbd8622156ca87b75a5660b1d933bd

2) I think most of the additional config complexity comes from the additional mass-conservation stuff. As far as an exempler config the following existing config files cover everything except for the particular dataset chosen:

base config: https://github.com/ecmwf/anemoi-core/blob/feat/402-time-interpolation-mass-conserving-interpolation-for-accumulated-fields/training/src/anemoi/training/config/interpolator.yaml

data config: https://github.com/ecmwf/anemoi-core/blob/feat/402-time-interpolation-mass-conserving-interpolation-for-accumulated-fields/training/src/anemoi/training/config/data/zarr_interpolator.yaml

training config: https://github.com/ecmwf/anemoi-core/blob/feat/402-time-interpolation-mass-conserving-interpolation-for-accumulated-fields/training/src/anemoi/training/config/training/interpolator.yaml

These configs provide an example of how to train a model where you only want to perform mass conservation on tp and cp. 

I'll also add some extra pydantic logic to ensure that if someone chooses to use mass conservation that helpful error messages can are emitted if the config is created incorrectly

frazane:
Maybe I am missing it but is there a check somewhere to verify that the disaggregated precipitation indeed sums up to the right bound?

Rilwan-Adewoyin:
> Maybe I am missing it but is there a check somewhere to verify that the disaggregated precipitation indeed sums up to the right bound?

Good idea @frazane , I've just added that in the following commit 63d6f0dcde27dd0623d976dc4f857df2d0bc3d6b , I'll do a check over it on Monday. It's current implementation focuses on testing the AnemoiModelEncProcDecInterpolator.resolve_mass_conservations function, checking the output values sum to the input value.

training/src/anemoi/training/schemas/base_schema.py
```diff
@@ -108,6 +108,94 @@ def check_bounding_not_used_with_data_extractor_zero(self) -> Self:
...
+
+        remap_mismatch = [t for (t, c) in pairs if remap.get(t) != c]
+        if remap_mismatch:
+            error_title = "normalizer_remap_mismatch"
+            error_msg = f"""data.processors.normalizer.config.remap must map each target to its corresponding constraint
+             variable. {remap_mismatch}. By default every pairing in config.model.mass_conserving_accumulations must be
+             included in config.data.processors.normalizer.config.remap"""
+
+            raise PydanticCustomError(error_title, error_msg)
+
```
roxas6path, line 120: @theissenhelen  - for reference, the addition to the Base Schema

context: When enforcing mass_conserving_accumulations for interpolation (via the config.model) the following config elements must also be set in config.data

models/src/anemoi/models/preprocessing/overwriter.py
```diff
@@ -0,0 +1,98 @@
...
+
+from torch import Tensor
+
+from anemoi.models.data_indices.collection import IndexCollection
+from anemoi.models.preprocessing import BasePreprocessor
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ZeroOverwriter(BasePreprocessor):
```
anaprietonem, line 24: can we have a dosctring describing what's the purpose of this class?

models/src/anemoi/models/schemas/data_processor.py
```diff
@@ -252,25 +287,12 @@ class PreprocessorTarget(str, Enum):
...
 
 
 class PreprocessorSchema(BaseModel, validate_assignment=False):
     target_: PreprocessorTarget = Field(..., alias="_target_")
     "Processor object from anemoi.models.preprocessing.[normalizer|imputer|remapper]."
-    config: Union[dict, NormalizerSchema, ImputerSchema, PostprocessorSchema, RemapperSchema]
+    config: Union[dict, NormalizerSchema, ImputerSchema, PostprocessorSchema, RemapperSchema, ZeroOverwriterSchema]
     "Target schema containing processor methods."
-
-    @model_validator(mode="after")
```
anaprietonem, line 264: why has this been deleted?

models/src/anemoi/models/models/interpolator.py
```diff
@@ -180,3 +186,246 @@ def forward(
...
+
+            # Gather output if needed
+            if gather_out and model_comm_group is not None:
+                y_preds = gather_tensor(
+                    y_preds, -2, apply_shard_shapes(y_preds, -2, grid_shard_shapes), model_comm_group
+                )
+
+        return y_preds
+
+    def resolve_mass_conservations(self, y_preds, x_input, include_right_boundary=False) -> torch.Tensor:
```
anaprietonem, line 300: from what I see the mass conservation logic is just applicable/available for the interpolator right? - could this be reused for other models  or is the logic dependent/linked to the interpolator?
Proposal could be to abstract this to another into a class in models like physical_constrains so then we can build extend those in the future for enforcing other type of conservations

anaprietonem: Can you update the docs with a brief explanation on how's this implemeted/can be used please?

training/src/anemoi/training/schemas/base_schema.py
anaprietonem: Rilwan is the mass_conserving_accumulations is specified just as part of the interpolator model why is this defined at the level of the BaseSchema and not part of the `InterpolatorModelSchema` ?

roxas6path: So yep, the mass_conserving_accumulations is specified in the InterpolatorModelSchema as you mentioned ->(https://github.com/ecmwf/anemoi-core/blob/4d5c2067a3fc9a6b96ba4197c0f2860523cc1fa6/models/src/anemoi/models/schemas/models.py#L266C1-L273C6)

However, the consistency check needs to happen at a level which can access both the model and data configs, which under the current code structure leaves the BaseSchema as the only place it can go ceteris paribus

models/src/anemoi/models/models/interpolator.py
roxas6path: > from what I see the mass conservation logic is just applicable/available for the interpolator right? - could this be reused for other models or is the logic dependent/linked to the interpolator? Proposal could be to abstract this to another into a class in models like physical_constrains so then we can build extend those in the future for enforcing other type of conservations

So, in a similar vein, earlier @sahahner and I discussed a more ideal structure for the resolve_mass_conservations which in which it was a postprocessor - In this sense it could be used flexibly like the "bounding" physical constraints. However, with the current stage of the code my understanding is that this would require a level of change appropriate for a follow up PR.

Context:
The issue was that given a sample with multiple timesteps of data, postprocessors currently operate on a single timestep one at a time, the logic for resolve_mass_conservation requires  an operation to be performed after all timesteps have been produced e.g. after the full rollout has completed.

models/src/anemoi/models/preprocessing/overwriter.py
roxas6path: implemented in bb8bbd79faf4e1ef303db94ce7ea5d4721bec01d

models/src/anemoi/models/models/interpolator.py
roxas6path: Docs updated in 8dbdd97981fc5fc6267f9c3d264bf5e96c5d6fd0
- Getting started section for interpolator in anemoi training docs
- ZeroOverwriter section in anemoi models docs
@anaprietonem @Magnus-SI @sahahner 

Time Interpolator Training: https://anemoi--544.org.readthedocs.build/projects/training/en/544/user-guide/interpolator-set-up.html

ZeroOverwriter: https://anemoi--544.org.readthedocs.build/projects/models/en/544/modules/preprocessing.html

models/tests/preprocessing/test_preprocessor_overwriter.py
```diff
@@ -0,0 +1,115 @@
...
+                "zero_overwriter": {
+                    "groups": [
+                        {
+                            "vars": ["tp_accum", "cp_accum"],
+                            "time_index": [0],
+                        }
+                    ]
+                },
+                # training data categories
+                "forcing": ["tp_accum", "cp_accum", "z", "q"],
```
sahahner: To test the indices, it would be nice to have a diagnostic variable before the forcing variables in the training input tensor.

models/tests/preprocessing/test_preprocessor_overwriter.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    return base
+
+
+def test_zero_overwriter_not_inplace_training(zero_overwriter, training_input_data) -> None:
+    """ZeroOverwriter should not modify input when in_place=False (training shape)."""
+    x = training_input_data.clone()
+    x_old = x.clone()
+    _ = zero_overwriter(x, in_place=False)
+    assert torch.allclose(x, x_old, equal_nan=True)
```
sahahner: Would it be possible to output a reasonable error message?

models/src/anemoi/models/preprocessing/overwriter.py
```diff
@@ -0,0 +1,115 @@
...
+    groups: list[dict]
+        - vars: list[str]
+            Variable names to overwrite with zero.
+        - time_index: list[int]
+            Timesteps (relative to the input window) to set to zero, e.g., [0].
+
+    Notes
+    -----
+    - Operates on inputs; indices are built for both training and inference inputs.
+    - If `groups` is missing or malformed, no variables are overwritten.
```
sahahner: I wonder whether it would be safer to have an assert instead of a warning in the case the config file is wrong?

models/src/anemoi/models/schemas/data_processor.py
roxas6path: Thanks for catching this - I should have reverted this - it was by accident 

- reverted this change in commit db40988581a399a026f15e4fa504c68095ef5643

models/src/anemoi/models/preprocessing/overwriter.py
roxas6path: Sorry, this note and the related code should have been removed.

In an earlier comment you suggested moving this log.warning to be a check in the pydantic implementation - I implemented your suggestion in an earlier commit.

I've removed this note and the related code in commit e086f8958578f1aa9a0857cc63075987a1063526

Thanks for catching this

models/tests/preprocessing/test_preprocessor_overwriter.py
roxas6path: Updated this in commit 574e18ae10199d6dba425b8cd05e3c215d2d202e

models/tests/preprocessing/test_preprocessor_overwriter.py
roxas6path: Yep updated this in 663b886b350aac3e45563708e7907b8e8010a30c

training/docs/user-guide/interpolator-set-up.rst
```diff
@@ -0,0 +1,238 @@
...
+
+   data:
+     frequency: 1h
+     timestep: 1h
+
+****************************
+ Changes in training config
+****************************
+
+Use the deterministic forecaster task (no ensemble specialization
```
anaprietonem, line 33: does the interpolator works with the ensemble  models? If not we mention it explicitly that for now that's not supported

training/docs/user-guide/interpolator-set-up.rst
```diff
@@ -0,0 +1,238 @@
...
+       - "cp_accum"
+       # other forcing fields (topography, solar, etc.)
+
+     diagnostic:
+       - "tp"
+       - "cp"
+
+     normalizer:
+       default: "mean-std"
+       remap:
```
anaprietonem, line 131: does the output results from the interpolator then contain tp_accum and cp_accum or are those renamed back to the original variables names?

training/docs/user-guide/interpolator-set-up.rst
```diff
@@ -0,0 +1,238 @@
+############################
+ Time Interpolator training
```
anaprietonem, line 2: @Rilwan-Adewoyin thanks for nice guide!

training/src/anemoi/training/data/datamodule/singledatamodule.py
```diff
@@ -80,7 +80,8 @@ def data_indices(self) -> IndexCollection:
     def relative_date_indices(self, val_rollout: int = 1) -> list:
         """Determine a list of relative time indices to load for each batch."""
         if hasattr(self.config.training, "explicit_times"):
-            return sorted(set(self.config.training.explicit_times.input + self.config.training.explicit_times.target))
+            indices = set(self.config.training.explicit_times.input + self.config.training.explicit_times.target)
+            return sorted(self.timeincrement * idx for idx in indices)
```
anaprietonem, line 84: why is this needed here?

training/src/anemoi/training/schemas/base_schema.py
anaprietonem: I have been thinking a bit about this. @Rilwan-Adewoyin  could try doing something like:

```
class BaseSchema(BaseModel):
    data: DataSchema
    model: InterpolatorModelSchema  # could also be a Union if you have multiple model types
    config_validation: bool = True

    @model_validator(mode="after")
    def run_cross_section_validations(self) -> "BaseSchema":
        """Run any sub-schema cross-validations."""
        if not self.config_validation:
            return self

        for name, value in self.__dict__.items():
            if hasattr(value, "validate_cross_config"):
                value.validate_cross_config(self)

        return self
        
  and then 
  
  class InterpolatorModelSchema(BaseModel):
    mass_conserving_accumulations: Optional[Dict[str, str]] = None

    def validate_cross_config(self, full_config: "BaseSchema") -> None:
        """Perform mass-conserving accumulation validation."""
        mca = self.mass_conserving_accumulations
        if not mca:
            return  # nothing to check
       .....
 ```

File: models/src/anemoi/models/preprocessing/overwriter.py [hunk 21]
```diff
@@ -0,0 +1,108 @@
...
+        self._infer_time_to_idxs: dict[int, list[int]] = {}
+
+        train_map = self.data_indices.data.input.name_to_index
+        infer_map = self.data_indices.model.input.name_to_index
+
+        # Accept config.groups as:
+        # - list[{"vars": [...], "time_index": [int]}]
+        for entry in groups:
+            if not isinstance(entry, MutableMapping):
+                LOGGER.warning("ZeroOverwriter group must be a dict, got %r; skipping.", type(entry).__name__)
```
sahahner: I wonder wether some of these warnings should be an assert.

models/src/anemoi/models/preprocessing/overwriter.py
```diff
@@ -0,0 +1,108 @@
...
+        self.num_training_input_vars = len(self.data_indices.data.input.name_to_index)
+        self.num_inference_input_vars = len(self.data_indices.model.input.name_to_index)
+        self._train_time_to_idxs: dict[int, list[int]] = {}
+        self._infer_time_to_idxs: dict[int, list[int]] = {}
+
+        train_map = self.data_indices.data.input.name_to_index
+        infer_map = self.data_indices.model.input.name_to_index
+
+        # Accept config.groups as:
+        # - list[{"vars": [...], "time_index": [int]}]
```
sahahner: Could the time_index be the key of the dictionary, similar to how the constant imputer is defined, please? Then we could follow the same design. I am happy to help regarding this.

sahahner: I would propose this config file entry for overwriting the timestep zero of the variables tp_accum and cp_accum with zero.
```
               zero_overwriter
                    0:
                            - tp_accum
                            - cp_accum
```

sahahner: I had a look at the docmentation and the "time_index" can in fact be a list of indices (very useful), so the solution from the constant imputer is not feasible.

Then, I think we should at least name the entry "time_indices" to avoid confusion.

training/src/anemoi/training/schemas/base_schema.py
roxas6path: Yep looks good - will implement this

training/docs/user-guide/interpolator-set-up.rst
roxas6path: The output results from the interpolator would contain tp and cp - only the input variables are renamed to tp_accum and cp_accum. The output variables remain with the correct standard name

training/src/anemoi/training/schemas/base_schema.py
roxas6path: @anaprietonem implemented in 26e8d94a4f0bdc4ef25b27b76b191b5bd2740db1 and 3041ced5a67f9fbde0f89d0c803dc4fa47d8e17e

training/docs/user-guide/interpolator-set-up.rst
roxas6path: Do you mean 1) can it train on an ensemble or 2) Can it be used to interpolate an ensemble?

In the first case - I haven't tested this, @Magnus-SI  have you tested using it for ensemble?

For 2) I think this should be possible

training/src/anemoi/training/data/datamodule/singledatamodule.py
roxas6path: @anaprietonem 

The outcome is exemplified in the docs here:
https://anemoi--544.org.readthedocs.build/projects/training/en/544/user-guide/interpolator-set-up.html#:~:text=%2C%205%5D-,Examples%3A,training.explicit_times%3A%20input%3A%20%5B0%2C%204%5D%2C%20target%3A%20%5B1%2C%202%2C%203%5D,-Mass%2Dconserving%20time

It just standardises the way the interpolator workflow reads the data to the way the other workflows e.g. forecastor and diffusion read the data. Whereby frequency refers to data temporal spacing and timestep is the 'temporal spacing' the model operates on and and what is extracted from the data.

It means that if you want to train a 24hour -> 6hour model, on 2 datasets - where 1 is hourly and the other is 6 hourly then the explicit_times will stay constant, but the config.data.frequency will just change to reflect the variation in source dataset.

I believe in a previous iteration of the interpolator code timestep still had this purpose

models/src/anemoi/models/preprocessing/overwriter.py
roxas6path: updated in [37bd2fb](https://github.com/ecmwf/anemoi-core/pull/544/commits/37bd2fb7f476a81e788a8b6dcf8f092dca471687)

models/src/anemoi/models/preprocessing/overwriter.py
roxas6path: updated in [185a03a](https://github.com/ecmwf/anemoi-core/pull/544/commits/185a03a762a75bb538f0ca2041f91326b22e0563)

models/src/anemoi/models/preprocessing/overwriter.py
sahahner: thanks. Now it is clearer.

anaprietonem (approved):
Thanks Rilwan for addressing the comments and the nice work. Before merging it would be good if you can update the branch with main and check tests still pass https://github.com/ecmwf/anemoi-core/actions/runs/18645927148 and https://github.com/ecmwf/anemoi-core/actions/runs/18645957033. 


@Rilwan-Adewoyin currently there seems to be some errors with flash_attention - the integration tests are trying to picked that up and failing. Please check that before merging

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -1,22 +1,16 @@
 defaults:
-- data: zarr
+- data: zarr_interpolator
 - dataloader: native_grid
 - datamodule: single
 - diagnostics: evaluation
 - hardware: example
 - graph: multi_scale
-- model: graphtransformer
```
anaprietonem, line 8: the integration tests errors are likely coming from this. We don't test yet the transformer as part of our integration tests, as flash attention can't be installed and we had issues when compiling flex attention. I'd change this back to graph-transformer for now

training/src/anemoi/training/train/tasks/interpolator.py
```diff
@@ -88,7 +88,7 @@ def _step(
         self,
         batch: torch.Tensor,
         validation_mode: bool = False,
-    ) -> tuple[torch.Tensor, Mapping[str, torch.Tensor]]:
+    ) -> tuple[torch.Tensor, Mapping[str, torch.Tensor], list[torch.Tensor]]:
```
frazane, line 91: This now doesn't match the signature of the abstract method on the parent class.

models/docs/modules/preprocessing.rst
```diff
@@ -69,3 +69,44 @@ The module contains the following classes:
...
+****************
+ ZeroOverwriter
+****************
+
+Overwrite selected timesteps of specified input variables with zero.
+
+This preprocessor operates on inputs before the model and is
+model-independent. It is useful whenever a variable should be reset to
+zero at certain timesteps within each input window (for example, for
+accumulated or windowed variables).
```
frazane, line 82: Do I understand correctly that this simply sets the specified index in the input window to zero? This will not work when cumulated variables are accumulated over the forecast itself. 

Concretely: when we predict precipitation up to say 120h, it is common to accumulate the precipitation over the entire length of the forecast. So at 66h you would have 66h accumulation, at 90h you have 90h, and so on up to 120h accumulation. The `ZeroOverwriter` would not suffice in this case because the right-bound would still be the wrong value. If we want to predict from 6h to 12h, then we should not only set the 6h value to zero but also subtract the 6h value from the 12h value. 

We need either an additional preprocessor to handle that case, or to implement a more generic preprocessor that disaggregates previously accumulated variables over an input window (and sets the first value to zero).

EDIT: or a pre-processor in anemoi-inference @gmertes @HCookie ?

roxas6path: Hi @frazane - This merged [PR](https://github.com/ecmwf/anemoi-transform/pull/168) in anemoi-inference highlighted the behaviour you are discussing and also implements a functional solution for it. The related [github issue](https://github.com/ecmwf/anemoi-transform/issues/167#event-20188002169) provides info on the implemented solution.

frazane: I missed it. Awesome!

training/src/anemoi/training/train/tasks/interpolator.py
roxas6path: Thanks for highlighting this - It turns out the abstract method had the wrong signature - it seems that when step is called even the base class expects three outputs in training_step, validation_step as shown below:

`train_loss, _, _ = self._step(batch)`

I updated the base class _step signature in [99923bc](https://github.com/ecmwf/anemoi-core/pull/544/commits/99923bc0c363a1951489196d1ce0a7ba43221453)

training/src/anemoi/training/config/interpolator.yaml
roxas6path: noted, updated in [b16e868](https://github.com/ecmwf/anemoi-core/pull/544/commits/b16e8684686eecfe2d4952e828345f12e49e3efa) - I'll check if tests are now passing soon

PR #571, opened 2025-09-30
fix: allow int values in variable group schema

bdvllrs:
## Description
Previously the variable groups schema did not allow int values for complex descriptions like:
```yaml
variable_groups:
     l_50:
        param: [z]
        level: [50]  # not allowed
```

Related to and supersedes PR #532

##  Additional notes ##

### DIFF
--- a/training/docs/modules/losses.rst
+++ b/training/docs/modules/losses.rst
@@ -206,6 +206,12 @@ are available.
         is_model_level: True
         param: 'z'
 
+The list of available metadata attributes is:
+
+.. autoclass:: anemoi.transforms.variable.Variable
+   :members:
+   :no-undoc-members:
+
 If metadata is not available, complex variable groups cannot be defined,
 and an error will be raised.
 
--- a/training/src/anemoi/training/utils/variables_metadata.py
+++ b/training/src/anemoi/training/utils/variables_metadata.py
@@ -66,7 +66,6 @@ class ExtractVariableGroupAndLevel:
         variable_groups: dict[str, GROUP_SPEC | dict[str, GROUP_SPEC]],
         metadata_variables: dict[str, dict | Variable] | None = None,
     ) -> None:
-
         if isinstance(variable_groups, DictConfig):
             variable_groups = OmegaConf.to_container(variable_groups, resolve=True)
 
--- a/training/src/anemoi/training/schemas/training.py
+++ b/training/src/anemoi/training/schemas/training.py
@@ -345,7 +345,7 @@ class BaseTrainingSchema(BaseModel):
     "Scalers to use in the computation of the loss and validation scores."
     validation_metrics: dict[str, LossSchemas]
     "List of validation metrics configurations."
-    variable_groups: dict[str, str | list[str] | dict[str, str | bool | list[str]]]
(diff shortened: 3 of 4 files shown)


training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -66,7 +66,6 @@ def __init__(
         variable_groups: dict[str, GROUP_SPEC | dict[str, GROUP_SPEC]],
         metadata_variables: dict[str, dict | Variable] | None = None,
     ) -> None:
-
```
mc4117, line 69: Could we keep this extra line?

training/docs/modules/losses.rst
```diff
@@ -206,6 +206,12 @@ are available.
         is_model_level: True
         param: 'z'
 
+The list of available metadata attributes is:
+
+.. autoclass:: anemoi.transforms.variable.Variable
+   :members:
+   :no-undoc-members:
+
```
mc4117, line 227: It could be nice to add an example in the documentation with your step scalers?

training/src/anemoi/training/utils/variables_metadata.py
bdvllrs: Woops, sorry for this. This comes from some debugging...

training/docs/modules/losses.rst
bdvllrs: As I expressed in the previous PR, I don't think step scalers are relevant anymore as it can be done with different variable_groups. Do you think that #532 is worth continuing?

bdvllrs:
I've tried addind unit tests for this use case, but this seems a bit more tricky than expected. I'll take another crack at it another time.

training/docs/modules/losses.rst
mc4117: I just meant an example of how you can do scalers on individual variables with the current set-up. But also happy for it not to be included if you don't think it's relevant

bdvllrs:
Thanks a lot! I think I fixed it? Maybe you can review the test I added, in particular the hard-coded metadata that I added. What I missed was that the `VariableFromDict` needs mars metadata to work.

training/docs/modules/losses.rst
bdvllrs: Done!

training/docs/modules/losses.rst
```diff
@@ -206,6 +206,40 @@ are available.
...
+   variable_groups:
+     default: sfc
+     pl:
+        is_pressure_level: True
+     l_50:  # this needs to come first to take priority
+        param: ["z"]
+        level: [50]
+     l:
+        param: ["z"]
```
bdvllrs, line 241: The priority based on the order here is not very well documented. Also the chosen dict-like structure (as opposed to a list) doesn't suggest that the order is important.

Do you think it would be worth implementing a priority system where more restrictive groups (that gather less variables) have priority over more generic groups? (Probably in another PR)

mc4117: I like the idea but I think it might be a bit difficult to implement?

bdvllrs: Ok, I'll take a look in another branch. In any case, this is ready for review!

Issue #552, opened 2025-09-17
Refactor mappers to allow dynamic behaviour


JPXKQX:
### Is your feature request related to a problem? Please describe.

Currently, mappers in anemoi-models are static: they receive the subgraph (edge indices & edge attributes) at initialisation, storing it as a mapper attribute. This design limits flexibility when applying the model to novel use cases (observations, multi-domain training, ...). Redesigning them to be dynamic, so that the subgraph is passed during forward(), would make mappers more extensible. 

Should all mappers be strictly dynamic? Not necessarily. If benchmarks show a significant performance degradation when always passing the subgraph at forward(), we plan to introduce a caching mechanism. This would allow subgraphs to be stored and reused internally through specialised classes. 

### Current Status 

``` 
self.encoder = instantiate(  
  model_config.model.encoder, 
  sub_graph=graph[(self._graph_name_data, "to", self._graph_name_hidden)], 
  **encoder_kwargs 
  ...
) 
x_data_latent, x_latent = self.encoder( 
        (x_data_latent, x_hidden_latent), 
        **kwargs 
        ...
) 
``` 

###  Proposed Solution 

``` 
self.encoder = instantiate(  
     model_config.model.encoder, 
     **encoder_kwargs,
     ...
) 

x_data_latent, x_latent = self.encoder( 
        (x_data_latent, x_hidden_latent), 
        edge_index=graph[(self._graph_name_data, "to", self._graph_name_hidden)].edge_index, 
        **kwargs,
       ...
) 
```

ssmmnn11:
the original mappers always received nodes, edges and edge attributes. There will not be a performance impact. I don' think we need caching in any form. This will just depend on how we set up the graph data.

PR #574, opened 2025-09-30
feat(models): mapper refactor

ssmmnn11:
## Description
As described in #552 refactor mappers to enable both flexible graph behaviour, static, dynamic etc. 
- simplify mapper structure (Mixins), every mapper has its own pre_process and post_proscess implementation
- introduce graph providers to make mappers more flexible and allow for future dynamic graphs

Updated description:

Mapper/Processor Refactor - Graph Providers
Refactors of how mappers and processors handle graph connectivity. Graph edge management is now separated from the neural network layers into dedicated “Graph Providers”.

 Changes:
 - New GraphProvider abstraction (graph_providers.py) with multiple implementations:
   - StaticGraphProvider - for fixed pre-computed graphs
   - NoOpGraphProvider - for potential edge-less architectures
   - DynamicGraphProvider - placeholder for on-the-fly graph construction
   - ProjectionGraphProvider - for sparse matrix operations (interpolation etc.)
 - Mappers/Processors no longer own edge state - edges are passed in during forward()
 - Models create and manage graph providers, calling get_edges() and passing results to layers
 - Simplified SparseProjector (now stateless)
 
 Why:
 - Multi-dataset support: Different datasets can use different graphs without changing the mapper/processor code
 - Dynamic graphs for observations: Enables future work on observation datasets where connectivity changes at runtime (e.g., k-NN graphs computed on-the-fly)
 - Composability: Graph management is now orthogonal to the neural network architecture - swap graph providers without touching mappers
 - Cleaner separation of concerns: Each component has a single responsibility, net result is more modular code with less coupling.
 
 Migration script included for existing checkpoints.
 

----
📚 Documentation preview 📚: https://anemoi-training--574.org.readthedocs.build/en/574/

----
📚 Documentation preview 📚: https://anemoi-models--574.org.readthedocs.build/en/574/


models/src/anemoi/models/layers/mapper.py
```diff
@@ -269,9 +260,11 @@ def __init__(
 
         Linear = self.layer_factory.Linear
 
-        self._register_edges(sub_graph, sub_graph_edge_attributes, src_grid_size, dst_grid_size, trainable_size)
+        # Initialize graph mode (static or dynamic)
+        self._init_graph_mode(sub_graph, sub_graph_edge_attributes, src_grid_size, dst_grid_size, trainable_size)
 
-        self.trainable = TrainableTensor(trainable_size=trainable_size, tensor_size=self.edge_attr.shape[0])
+        if self.graph_mode == "static":
```
JPXKQX: Would it make sense to move the `TrainableTensor` instantiation inside the `_init_graph_mode()`?

models/src/anemoi/models/layers/mapper.py
```diff
@@ -208,10 +199,10 @@ def __init__(
...
         num_heads: int,
         mlp_hidden_ratio: int,
-        sub_graph: HeteroData,
-        sub_graph_edge_attributes: list[str],
-        src_grid_size: int,
-        dst_grid_size: int,
+        sub_graph: Optional[HeteroData] = None,
+        sub_graph_edge_attributes: Optional[list[str]] = None,
+        src_grid_size: Optional[int] = None,
+        dst_grid_size: Optional[int] = None,
```
JPXKQX: I think that we need the `edge_dim` here. In fact, we could probably include this in the `GraphEdgeMixin`, right?

models/src/anemoi/models/layers/mapper.py
```diff
@@ -193,8 +123,69 @@ def _expand_edges(self, edge_index: Adj, edge_inc: Tensor, batch_size: int) -> A
...
+        if cpu_offload:
+            self.proc = nn.ModuleList([offload_wrapper(x) for x in self.proc])
+
+    @abstractmethod
+    def pre_process(self, x, shard_shapes, model_comm_group=None, x_src_is_sharded=False, x_dst_is_sharded=False):
+        pass
+
+    @abstractmethod
+    def post_process(self, x_dst, shapes_dst=None, model_comm_group=None, keep_x_dst_sharded=False, **kwargs):
+        pass
```
JPXKQX, line 90: We could add `forward()` to the BaseMapper class

JPXKQX, line 90:
```suggestion
        pass
 
    @abstractmethod
    def forward(
        self,
        x: PairTensor,
        batch_size: int,
        shard_shapes: tuple[tuple[int], tuple[int]],
        model_comm_group: Optional[ProcessGroup] = None,
        x_src_is_sharded: bool = False,
        x_dst_is_sharded: bool = False,
        keep_x_dst_sharded: bool = False,
        **kwards
    ) -> PairTensor:
        pass
```

models/src/anemoi/models/layers/mapper.py
```diff
@@ -47,99 +48,28 @@
 NUM_CHUNKS_INFERENCE_MAPPER = int(os.environ.get("ANEMOI_INFERENCE_NUM_CHUNKS_MAPPER", NUM_CHUNKS_INFERENCE))
 
 
-class BaseMapper(nn.Module, ABC):
-    """Base Mapper from souce dimension to destination dimension."""
-
-    def __init__(
+class GraphEdgeMixin:
```
matschreiner: If we want to handle graphs inside the mappers, I think inheriting from a mixin that covers both fixed and dynamic cases is probably not the best pattern. What happens if we want to support a third way of supplying graphs? Do we extend the mixin with another condition for another mode?
Instead, I’d suggest isolating the graph handling into its own class. A BaseGraphProvider interface could define how graphs are supplied, with implementations like FixedGraphProvider and DynamicGraphProvider. Then the mapper would simply depend on this provider:
```
class GraphTransformerBaseMapper(BaseMapper):
    def __init__(self, *, graph_provider: BaseGraphProvider, ...):
        ...
```
This way, adding new strategies for graphs later is just a matter of subclassing BaseGraphProvider, and the mapper is kept isolated from the graph logic.

models/src/anemoi/models/layers/mapper.py
matschreiner, line 1: I think handling graphs inside the mappers is too late in the data pipeline. This couples graph logic to the mapper and reduces modularity - I think mappers should receive a graph, not be responsible for how to get it.

matschreiner: This is because the graph depends both on the model architecture and the data domain, so including the graph in the model couples it to data which is not ideal

ssmmnn11: This is a step to make it easier to disentangle that in the future. Mappers should recieve edges and edge attributes and not have the graph baked in. However, this change should keep backward compatibility. In addition it simplifies the inheritance structure. Once we have a solid use case for the dynamic mappers we can iterate. 

We also currently use the graph for meta data, where we should use the dataset directly. The plan is to improve upon that in the future.

models/src/anemoi/models/layers/mapper.py
ssmmnn11: @matschreiner have a look at

https://github.com/ecmwf/anemoi-core/compare/feat/mapper-refactor-graph-provider?expand=1

models/src/anemoi/models/layers/mapper.py
ssmmnn11: added edge_dim

models/src/anemoi/models/layers/graph_providers.py
```diff
@@ -0,0 +1,268 @@
...
+    Graph providers encapsulate the logic for supplying edge indices and attributes
+    to mapper and processor layers. This allows for different strategies (static, dynamic, etc.)
+    to be implemented and swapped without modifying the mapper classes.
+    """
+
+    @abstractmethod
+    def get_edges(
+        self,
+        batch_size: int,
+        edge_index: Optional[Adj] = None,
```
matschreiner: I think that the goal of this is to abstract all graph provision logic into a single baseclass. 
By including the graph in this signature we defeat the purpose of the class because it means that graph has already been provided.

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -0,0 +1,268 @@
...
+        -------
+        tuple[Tensor, Adj]
+            Edge attributes and expanded edge index
+        """
+        edge_attr = self.trainable(self.edge_attr, batch_size)
+        edge_index = self._expand_edges(self.edge_index_base, self.edge_inc, batch_size)
+        return edge_attr, edge_index
+
+
+class DynamicGraphProvider(BaseGraphProvider):
```
matschreiner, line 304: The reason for taking `edge_index` as arguments is so that the DynamicGraphProvider can act as a noop on dynamic graphs. Then maybe the name DynamicGraphProvider is a little misleading, since it is only passing the graph along without doing any work to get it.

Would it be possible to somehow hook the DynamicGraphProvider up to the graph stream so that it is actually providing the graph to the mapper?

ssmmnn11: I now have only a example implementation of the dynamic graph provider. We will implement one that will create graphs on the fly ; other options are also to build one to subsample a big graph etc.

models/src/anemoi/models/layers/mapper.py
ssmmnn11: I merged the new structure now.

models/src/anemoi/models/layers/graph_provider.py
matschreiner: Great, thanks, I think it looks good!

models/src/anemoi/models/layers/mapper.py
ssmmnn11: moved to the graph provider now.

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -1,154 +1,40 @@
...
-def _row_normalize_weights(edge_index: torch.Tensor, weights: torch.Tensor, num_target_nodes: int) -> torch.Tensor:
-    total = torch.zeros(num_target_nodes, device=weights.device)
-    norm = total.scatter_add_(0, edge_index[1].long(), weights)
-    norm = norm[edge_index[1]]
-    return weights / (norm + 1e-8)
 
 
 class SparseProjector(torch.nn.Module):
-    """Constructs and applies a sparse projection matrix for mapping features between grids.
+    """Applies sparse projection matrix to input tensors.
```
theissenhelen, line 5: If the logic gets moved to the ProjectionGraphProvider, this class could be removed as it essentially just performs a matmul.

models/src/anemoi/models/layers/graph_providers.py
```diff
@@ -0,0 +1,631 @@
...
+from torch_geometric.typing import Adj
+
+from anemoi.models.distributed.graph import shard_tensor
+from anemoi.models.distributed.khop_edges import sort_edges_1hop_sharding
+from anemoi.models.layers.graph import TrainableTensor
+
+LOGGER = logging.getLogger(__name__)
+
+
+def _shard_edges_1hop(
```
JPXKQX: I see that this is only used in this file, but should we move it to models/distributed ?? (perhaps in a new graph_provider.py file)

ssmmnn11: very similar code is in the mappers and in the GNN processor, good idea.

ssmmnn11: done

dietervdb-meteo:
I successfully trained a (sharded) stretched grid model that loads a prebuilt graph from disk with this branch, which confirms the `lam_with_fixed_graph` integration test.

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -698,15 +697,36 @@ def get_node_trainable_tensors(self, node_attributes: NamedNodesAttributes) -> d
...
-        }
-
-        if isinstance(model.processor, GraphEdgeMixin):
-            trainable_modules[model._graph_name_hidden, model._graph_name_hidden] = model.processor
-
-        return {name: module for name, module in trainable_modules.items() if module.trainable.trainable is not None}
+        trainable_modules = {}
+
+        # Check encoder
+        if (
```
JPXKQX, line 703: are these being tested in the unit/integration tests? `model.encoder.graph_provider` should be `model.encoder_graph_provider`

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -0,0 +1,594 @@
+# (C) Copyright 2025 Anemoi contributors.
```
JPXKQX, line 1: Should we rename this to graph_provider.py for consistency with other files like mapper.py, processor.py?

PR #588, opened 2025-10-06
feat(training): Refactor optimizer creation to support custom and torch optimizers

gabrieloks:
## Description
Refactoring of the optimizer creation to support custom and torch optimizers.
This PR enables users to select different PyTorch optimizers and also allows the use of custom ones, such as AdEMAMix.
##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--588.org.readthedocs.build/en/588/

----
📚 Documentation preview 📚: https://anemoi-graphs--588.org.readthedocs.build/en/588/

----
📚 Documentation preview 📚: https://anemoi-models--588.org.readthedocs.build/en/588/

### DIFF
--- a/training/src/hydra_plugins/anemoi_searchpath/anemoi_searchpath_plugin.py
+++ b/training/src/hydra_plugins/anemoi_searchpath/anemoi_searchpath_plugin.py
@@ -68,4 +68,14 @@ class AnemoiSearchPathPlugin(SearchPathPlugin):
                 )
                 LOGGER.info("Prepending current user directory (%s) to the search path.", cwd_path)
                 LOGGER.debug("Search path is now: %s", search_path)
+
+        # Add package config path as lowest priority fallback (issue #570)
+        # This enables discovery of default configs like 'training/default'
+        # Appended (not prepended) so user configs maintain higher priority
+        search_path.append(
+            provider="anemoi-package-searchpath-plugin",
+            path="pkg://anemoi.training/config",
+        )
+        LOGGER.debug("Appended package config path (pkg://anemoi.training/config) to search path.")
+
         LOGGER.info("Search path is now: %s", search_path)
--- a/training/tests/unit/commands/test_config.py
+++ b/training/tests/unit/commands/test_config.py
@@ -38,3 +38,41 @@ def test_dump_config(config_generator: ConfigGenerator) -> None:
             mock_copy_files.assert_called_once_with(config_path, mock.ANY)
             assert output_path.exists()
             assert OmegaConf.load(output_path) == {"test": "value"}
+
+
+def test_validate_config_uses_package_path(config_generator: ConfigGenerator) -> None:
+    """Test that validate_config works correctly with package configs.
+
+    This test verifies the fix for issue #570: the AnemoiSearchPathPlugin
+    adds 'pkg://anemoi.training/config' to the search path, enabling
+    discovery of package configs like 'training/default'.
+
+    Note: The package path is added by the plugin, not by the initialize() call,
+    so we just verify the method executes successfully.
+    """
+    with mock.patch("anemoi.training.commands.config.initialize"), mock.patch(
+        "anemoi.training.commands.config.compose",
+        return_value=OmegaConf.create({"test": "value"}),
+    ), mock.patch("anemoi.training.commands.config.BaseSchema"):
+        # Call validate_config - it should succeed
+        # The AnemoiSearchPathPlugin will add the package path automatically
+        config_generator.validate_config("test-config", mask_env_vars=False)
+
+        # If we get here without exception, the validation worked
+
+
+def test_validate_config_with_mask_env_vars(config_generator: ConfigGenerator) -> None:
+    """Test that validate_config works with mask_env_vars option."""
+    with mock.patch("anemoi.training.commands.config.initialize"), mock.patch(
+        "anemoi.training.commands.config.compose",
+        return_value=OmegaConf.create({"test": "value"}),
+    ), mock.patch("anemoi.training.commands.config.BaseSchema"), mock.patch.object(
+        config_generator,
+        "_mask_slurm_env_variables",
+        return_value=OmegaConf.create({"test": "masked"}),
+    ) as mock_mask:
+        # Call validate_config with mask_env_vars=True
+        config_generator.validate_config("test-config", mask_env_vars=True)
+
+        # Verify that _mask_slurm_env_variables was called
+        mock_mask.assert_called_once()


training/src/anemoi/training/train/tasks/base.py
```diff
@@ -734,53 +731,86 @@ def validation_step(self, batch: torch.Tensor, batch_idx: int) -> None:
...
-                convert_to_omegaconf(self.config),
-                "diagnostics.log.mlflow.expand_hyperparams",
-                default=["config"],
-            )
-            # Log hyperparameters
-            self.logger.log_hyperparams(
-                hyper_params,
-                expand_keys=expand_keys,
-            )
+        return {"scheduler": scheduler, "interval": "step"}
```
HCookie, line 761: Are you looking to generalise the LR Scheduler in this PR too?

training/pyproject.toml
```diff
@@ -56,7 +56,7 @@ dependencies = [
   "pyshtools>=4.13",
   "pytorch-lightning>=2.1",
   "timm>=0.9.2",
-  "torch>=2.2",
```
HCookie, line 59: Why has this been changed?

training/src/anemoi/training/train/tasks/base.py
gabrieloks: We could, potentially

training/pyproject.toml
gabrieloks: Sorry, just flash-attn stuff. This should not go in the final version.

training/src/anemoi/training/train/tasks/base.py
anaprietonem: could we be great to check if we can use https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.CosineAnnealingLR.html rather than the one from Timm

training/src/anemoi/training/config/training/default.yaml
```diff
@@ -37,11 +37,36 @@ swa:
   enabled: False
   lr: 1.e-4
 
-# Optimizer settings
+# =====================================================================
+# Optimizer configuration
+# =====================================================================
 optimizer:
```
radiradev, line 43: We can just pass in arbitrary objects as:
```
optimizer:
  _target_: torch.optim.Adam
  _partial_: true
  lr: 0.001
  weight_decay: 0.0
```

training/src/anemoi/training/train/tasks/base.py
radiradev, line 1: If we allow arbitrary objects, then there would be no need for the extra logic here.

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -734,37 +735,64 @@ def validation_step(self, batch: torch.Tensor, batch_idx: int) -> None:
...
+                optimizer_class=optimizer_cls,
+                **kwargs,
             )
         else:
-            optimizer = torch.optim.AdamW(
-                self.trainer.model.parameters(),
-                lr=self.lr,
-                **self.optimizer_settings.kwargs,
-            )
+            optimizer = optimizer_cls(params, lr=self.lr, **kwargs)
```
radiradev: I guess you'd need `optimizer=hydra.utils.instantiate(cfg.optimizer)` here

training/src/anemoi/training/train/tasks/base.py
gabrieloks: We would still have to take care of zero but I think you are right. Instantiating directly from the config seems much simpler

training/src/anemoi/training/optimizers/AdEMAMix.py
```diff
@@ -0,0 +1,190 @@
+"""-----------------------------------------------------------------------------
+Copyright (c) 2024 Apple Inc.
+
+This source file is part of the ADEMAMix project:
+https://github.com/apple/ml-ademamix
+
+Licensed under the MIT License.
+See the LICENSE file in the root directory of this source tree for details.
```
floriankrb: Is the LICENCE file in the root directory of **this** source tree the right file to point to?

gabrieloks: Good catch!

training/src/anemoi/training/optimizers/AdEMAMix.py
```diff
@@ -0,0 +1,190 @@
...
+    def f_inv(t):
+        return math.pow(0.5, 1 / (t + 1))
+
+    if step < warmup:
+        a = step / float(warmup)
+        return f_inv((1.0 - a) * f(beta_start) + a * f(beta_end))
+    return beta_end
+
+
+class AdEMAMix(Optimizer):
```
mc4117, line 46: would this now be the default optimiser in anemoi? And do we expect to have more optimisers? I was wondering if we should have a base.py or default.py in this folder?

radiradev: @mc4117 I would rather not have this as the default optimizer. We can keep AdamW as the default one, initialised in the config?

mchantry: Agreed, AdamW should be the default. There's just a question about whether it's in the configs explicitly, or pydantic fills in the default so we don't break configs. What do you think @radiradev ?

mc4117: sorry I didn't mean that we should make this the default. I more meant the opposite, that we should have a default.py where the AdamOptimiser would be there as a class. But maybe it's not necessary

radiradev: @mc4117 I don't think we should have AdamW in the code, as we can initialize it directly through the config with some default like:
```
optimizer:
  _target_: torch.optim.AdamW
  _partial_: true
  lr: 0.001
  weight_decay: 0.0
```

All future optimizers would inherit from [`torch.optim.Optimizer`](https://github.com/pytorch/pytorch/blob/0fabc3ba44823f257e70ce397d989c8de5e362c1/torch/optim/optimizer.py#L339) anyway so don't think we'd need a base one, unless we do something that is not supported by it. 

@mchantry I guess for backwards compatibility with configs we could have pydantic instantiate AdamW in case the old config syntax is used - but maybe with a warning that it is going to get deprecated.

training/docs/modules/optimizers.rst
```diff
@@ -0,0 +1,114 @@
...
+evolution of the learning rate during training. By default,
+``anemoi-training`` uses a ``CosineLRScheduler`` from ``timm.scheduler``
+with optional warm-up steps and minimum learning rate.
+
+The scheduler is created by ``BaseGraphModule._create_scheduler`` and
+returned to the trainer together with the optimizer in
+``configure_optimizers``. The default scheduler configuration follows a
+cosine decay with per-step updates:
+
+.. code:: yaml
```
mc4117: in the documentation, it looks like you can specify the scheduler from the config but I can't see how this works in the code because it looks hard coded to me?

training/src/anemoi/training/train/tasks/base.py
floriankrb: Something surprising when reading this code: 
_ There is a concept of what a scheduler is, with a type Scheduler here (`CosineLRScheduler` is a kind of Scheduler)
_  and something called `_create_scheduler` which does not return an object of type Scheduler but a dict.

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -734,37 +735,47 @@ def validation_step(self, batch: torch.Tensor, batch_idx: int) -> None:
 
         return val_loss, y_preds
 
-    def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[dict]]:
-        """Configure the optimizers and learning rate scheduler.
+    def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[dict[str, Any]]]:
+        """Create optimizer and LR scheduler based on Hydra config."""
+        optimizer = self._create_optimizer_from_config(self.config.training.optimizer)
+        scheduler = self._create_scheduler(optimizer)
+        return [optimizer], [scheduler]
```
floriankrb, line 741: From the constructor of `CosineLRScheduler`, I would think that a scheduler contains an optimiser.

In such a case, I would expect `return [scheduler.optimizer], [scheduler]` here.
Or just `return [scheduler]`
Or even `return Schedulers(scheduler)`

This would implies using an object instead of a dictionary, and you may not want to adress this in this PR.

gabrieloks: Yeah, for this PR I would keep it like this and when there is a clear idea of what we want to do with the schedulers, we might want to address this :)

training/src/anemoi/training/train/tasks/base.py
gabrieloks: This was done so that the optimizer and scheduler could be created in separate functions. For now, it’s not very clear to me how we should refactor the schedulers, so I would leave that for another PR.

training/docs/modules/optimizers.rst
```diff
@@ -0,0 +1,103 @@
...
+
+***********************
+ Configuration in YAML
+***********************
+
+An example configuration for using ``AdEMAMix`` is shown below:
+
+.. code:: yaml
+
+   optimizer:
```
JPXKQX, line 87: I would add an `__init__.py` file to the 'optimisers/' folder so that we can specify the AdeMAMix class directly

training/src/anemoi/training/optimizers/AdEMAMix.py
```diff
@@ -0,0 +1,203 @@
...
+            "betas": betas,
+            "eps": eps,
+            "alpha": alpha,
+            "beta3_warmup": beta3_warmup,
+            "alpha_warmup": alpha_warmup,
+            "weight_decay": weight_decay,
+        }
+        super().__init__(params, defaults)
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
```
JPXKQX: You can remove this, right?

training/docs/modules/optimizers.rst
gabrieloks: I would do this if we ever planned to add additional custom optimizers, but I don’t think that’s likely. This approach also makes it very clear to the user that they’re using an optimizer provided by Anemoi (anemoi.training.optimizers...) rather than one from PyTorch (torch.optim...).

training/src/anemoi/training/optimizers/AdEMAMix.py
gabrieloks: I will remove this 👍

training/src/anemoi/training/optimizers/AdEMAMix.py
```diff
@@ -0,0 +1,200 @@
+"""AdEMAMix optimizer implementation.
+
+-----------------------------------------------------------------------------
+Copyright (c) 2024 Apple Inc.
```
VeraChristina, line 4: Related to licenses, we'll need to add this to a `NOTICE` file as an external contribution. See [this explanation](https://github.com/ecmwf/codex/blob/87d2f57e9ba77a0cf544aeb4afdd23829b43f2cd/Legal/copyright_and_licensing.md?plain=1#L22)

gabrieloks: Created this NOTICE file and included the MIT license and ACKNOWLEDGEMENTS from the Apple AdEMAMix repository.

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -734,37 +734,31 @@ def validation_step(self, batch: torch.Tensor, batch_idx: int) -> None:
...
-                lr=self.lr,
-                **self.optimizer_settings.kwargs,
-            )
+    def _create_optimizer_from_config(self, opt_cfg: Any) -> torch.optim.Optimizer:
+        """Instantiate optimizer directly via Hydra config (_target_ style)."""
+        params = filter(lambda p: p.requires_grad, self.parameters())
+
+        # Convert schema to dict if needed
+        if hasattr(opt_cfg, "model_dump"):
+            opt_cfg = opt_cfg.model_dump(by_alias=True)
```
VeraChristina, line 749: Are we removing the option for a zero redundancy optimizer, or will this just be instantiated via the config as other optimizers?

gabrieloks: The zero redundancy optimizer can be instantiated via the config as other optimizers. I left an example in the default config: [here](https://github.com/ecmwf/anemoi-core/pull/588/files#diff-6dadf1a4c514d12f180a11cf35b3b85d80cc63693c423eb459588037ca67cff3R68:~:text=%23%20Optional%3A%20configuration%20for%20ZeroRedundancyOptimizer)

PR #594, opened 2025-10-08
Feature/multi dataset integration

mchantry:
## Description
Supports multiple time-aligned datasets as inputs and outputs for training.
e.g. 
ds_t -> ds_{t+1}

##  Additional notes ##


training/src/anemoi/training/config/datamodule/multi.yaml
JPXKQX, line 1: Can we remove this one?

training/src/anemoi/training/config/data/multi_debug.yaml
```diff
@@ -0,0 +1,81 @@
...
+    forcing: ${data.default.forcing}
+    diagnostic: ${data.default.diagnostic}
+    normalizer: ${data.default.normalizer}
+    processors: ${data.default.processors}
+
+  era5_copy:
+    # ERA5_copy uses same configuration as ERA5 for debugging
+    forcing: ${data.default.forcing}
+    diagnostic: ${data.default.diagnostic}
+    normalizer: ${data.default.normalizer}
```
JPXKQX: Is the `normalizer` key needed?

training/src/anemoi/training/train/train.py
```diff
@@ -137,30 +141,42 @@ def initial_seed(self) -> int:
...
             graph_filename = None
 
+        # Create new graph
         from anemoi.graphs.create import GraphCreator
 
         graph_config = convert_to_omegaconf(self.config).graph
+
+        # ALWAYS override dataset from dataloader config (ignore dummy in graph config)
+        if hasattr(graph_config.nodes.data.node_builder, "dataset"):
+            graph_config.nodes.data.node_builder.dataset = dataset_path
```
MeraX, line 169: Would a similar mechanism be required for ICON Node builders? see also https://github.com/ecmwf/anemoi-core/pull/627

training/src/anemoi/training/config/debug_multi_eracerra.yaml
```diff
@@ -0,0 +1,64 @@
...
+  files:
+    truncation: null
+    truncation_inv: null
+    graph: graph_anemoi_new_${data.resolution}.pt
+
+    # Primary dataset for ERA5
+    dataset: aifs-ea-an-oper-0001-mars-${data.resolution}-1979-2023-6h-v8.zarr
+
+    # Secondary dataset for ERA5_copy (using same file for debugging)
+    dataset_b: cerra-rr-an-oper-0001-mars-5p5km-1984-2022-6h-v3-hmsi.zarr
```
radiradev: Is this working currently? I've been trying to run with two datasets with non-identical datetimes and getting errors

radiradev: Here I mean, one is a subset of datetimes of the other as it's the case in the example above.

training/src/anemoi/training/data/multidataset.py
```diff
@@ -0,0 +1,237 @@
...
+                shuffle=shuffle,  # Will be overridden in __iter__
+                label=f"{label}_{name}",
+            )
+
+        # Use the first dataset as the primary for shared properties
+        self.primary_dataset = next(iter(self.datasets.values()))
+
+        # Verify all datasets have the same number of valid indices
+        primary_count = len(self.primary_dataset.valid_date_indices)
+        for name, dataset in self.datasets.items():
```
radiradev: A better check here would be to compare the dates directly - this check wouldn't catch two datasets with the same length but different dates.

training/src/anemoi/training/data/dataset/multidataset.py
```diff
@@ -0,0 +1,237 @@
...
+    @cached_property
+    def statistics(self) -> dict:
+        """Return combined statistics from all datasets."""
+        combined_stats = {}
+        for name, dataset in self.datasets.items():
+            combined_stats[name] = dataset.statistics
+        return combined_stats
+
+    @cached_property
+    def statistics_tendencies(self) -> dict:
```
radiradev: This returns tendencies for datasets that are present but nothing for the datasets that don't have them. Code later assumes that this should be instead set to `None` for datasets that do not have tendencies.

training/src/anemoi/training/config/debug_multi_eracerra.yaml
mchantry: What do you mean about non-identical datetimes radi?

The first implementation of multiple datasets aim to use datasets that are fully time aligned, and feature the time datetimes

training/src/anemoi/training/data/dataset/multidataset.py
JPXKQX: Fixed, thank you!

training/src/anemoi/training/data/multidataset.py
JPXKQX: Done!

models/src/anemoi/models/interface/__init__.py
```diff
@@ -75,28 +75,77 @@ def __init__(
...
+            Data indices for the dataset
+        statistics_tendencies : dict, optional
+            Tendencies statistics for the dataset
+
+        Returns
+        -------
+        tuple
+            (pre_processors, post_processors, pre_processors_tendencies, post_processors_tendencies)
+        """
+        from anemoi.training.utils.config_utils import get_dataset_data_config
```
anaprietonem: this is not good, it's making anemoi.training a dependency in anemoi-models

VeraChristina: agree! The inference task in the system-level tests fails on a similar import in [models/src/anemoi/models/models/interpolator.py](https://github.com/ecmwf/anemoi-core/pull/594/files#diff-dafd9781f62a6e1a5f87f69826a694a511b9d4eea91e97de82cddd40efee85cc)

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
```diff
@@ -76,13 +86,11 @@ def _assemble_input(self, x, fcstep, batch_ens_size, grid_shard_shapes=None, mod
             (
                 einops.rearrange(x, "batch time ensemble grid vars -> (batch ensemble grid) (time vars)"),
                 node_attributes_data,
+                torch.ones(batch_ens_size * x.shape[3], device=x.device).unsqueeze(-1) * fcstep,
```
anaprietonem, line 89: where is this coming from ?

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
```diff
@@ -143,42 +158,85 @@ def forward(
...
             Output tensor
         """
-        batch_size, ensemble_size = x.shape[0], x.shape[2]
+        dataset_names = list(x.keys())
+
+        # Extract and validate batch sizes across datasets
+        batch_sizes = [x[dataset_name].shape[0] for dataset_name in dataset_names]
+        ensemble_sizes = [x[dataset_name].shape[2] for dataset_name in dataset_names]
+
+        # Assert all datasets have the same batch and ensemble sizes
```
anaprietonem: If we need all datasets to have same batch size and ensemble sizes would it be easier to not allow dictionarificaiton of those fields or do these checks when reading the config?

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
```diff
@@ -98,18 +106,25 @@ def _assemble_input(self, x, fcstep, batch_ens_size, grid_shard_shapes=None, mod
...
+        x_out: torch.Tensor,
+        x_skip: torch.Tensor,
+        batch_size: int,
+        batch_ens_size: int,
+        dtype: torch.dtype,
+        dataset_name: str = None,
+    ):
+        ensemble_size = batch_ens_size // batch_size
+        x_out = (
+            einops.rearrange(x_out, "(bs e n) f -> bs e n f", bs=batch_size, e=ensemble_size).to(dtype=dtype).clone()
```
anaprietonem, line 120: Why now the ensemble dimension is not being flattened with the batch size as before?

models/src/anemoi/models/models/base.py
```diff
@@ -202,32 +275,48 @@ def predict_step(
...
-                y_hat_shard_shapes = apply_shard_shapes(y_hat, -2, grid_shard_shapes)
-                y_hat = gather_tensor(y_hat, -2, y_hat_shard_shapes, model_comm_group)
+                for dataset_name in dataset_names:
+                    y_hat_shard_shapes = apply_shard_shapes(y_hat[dataset_name], -2, grid_shard_shapes[dataset_name])
+                    y_hat[dataset_name] = gather_tensor(y_hat[dataset_name], -2, y_hat_shard_shapes, model_comm_group)
 
         return y_hat
+
+    @abstractmethod
+    def fill_metadata(self, md_dict) -> None:
```
anaprietonem, line 328: This is just used for the diffusion model ?

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
```diff
@@ -203,25 +261,36 @@ def forward(
...
+                x_out,
+                x_skip_dict[dataset_name],
+                batch_size,
+                batch_ens_size,
+                dtype=x[dataset_name].dtype,
+                dataset_name=dataset_name,
+            )
 
-        return x_out
+        return x_out_dict
```
anaprietonem, line 286: There is no fill_metadata for this class or I am missing it? it's it using the one from the deterministic there the ensemble size is hardcoded to 1

models/src/anemoi/models/samplers/diffusion_samplers.py
```diff
@@ -268,24 +285,29 @@ def sample(
 class DPMpp2MSampler(DiffusionSampler):
     """DPM++ 2M sampler (DPM-Solver++ with 2nd order multistep)."""
 
-    def __init__(self, **kwargs):
+    def __init__(self, dtype: torch.dtype = torch.float64, **kwargs):
```
anaprietonem, line 288: Could this be an issue when training with half precision?

training/src/anemoi/training/config/data/multi.yaml
```diff
@@ -0,0 +1,87 @@
+# Multi-dataset data configuration for debugging with era5 and era5_copy
+format: zarr
+# Time frequency requested from dataset
+frequency: 6h
+# Time step of model (must be multiple of frequency)
+timestep: 6h
+
+# Dataset-specific configurations
+datasets:
+  era5:
```
anaprietonem, line 10: Do we have a way to check that the spelling of the dataset entries passed across all sections is consistent? (say if I say 'era5', 'cerra' for the data section that then I don't by accident do 'ERA5" or 'cerrra' somewhere else ?

training/src/anemoi/training/config/diagnostics/plot/detailed_cerra.yaml
```diff
@@ -0,0 +1,73 @@
+asynchronous: True # Whether to plot asynchronously
```
anaprietonem: Why is this called detailed_cerra? ( as in we don't have anything like detailed_era5 so that's I got confused)

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
VeraChristina: Yes, sorry, this is confusing. It is using the one from the deterministic because it would be identical: The `shapes` refer to the shape of the input tensor, where ensemble size is always 1 because we don't provide different ensemble members as input to the model. So we don't actually use the ensemble dimension in the input tensor

training/src/anemoi/training/data/datamodule.py
```diff
@@ -33,99 +34,105 @@ class AnemoiDatasetsDataModule(pl.LightningDataModule):
...
     def metadata(self) -> dict:
+        """Return metadata from all training datasets."""
         return self.ds_train.metadata
 
     @cached_property
     def supporting_arrays(self) -> dict:
-        return self.ds_train.supporting_arrays | self.grid_indices.supporting_arrays
+        """Return supporting arrays from all training datasets."""
+        # Each dataset has its own supporting arrays, no assumptions about sharing
+        return self.ds_train.supporting_arrays
```
anaprietonem: Why no longer `self.grid_indices.supporting_arrays`?

training/src/anemoi/training/data/datamodule.py
```diff
@@ -33,99 +34,105 @@ class AnemoiDatasetsDataModule(pl.LightningDataModule):
...
-                "Falling back rollout to: %s",
-                rollout_value,
-            )
+            LOGGER.warning("Falling back rollout to: %s", rollout_value)
 
         rollout = max(rollout_value, val_rollout)
-
         multi_step = self.config.training.multistep_input
-        return [self.timeincrement * mstep for mstep in range(multi_step + rollout)]
+        return list(range(multi_step + rollout))
```
anaprietonem, line 123: why is the `timeincrement` no longer considered here?

training/src/anemoi/training/data/dataset.py
```diff
@@ -8,36 +8,29 @@
...
-from anemoi.models.distributed.balanced_partition import get_balanced_partition_range
 from anemoi.training.data.grid_indices import BaseGridIndices
-from anemoi.training.utils.seeding import get_base_seed
-from anemoi.training.utils.usable_indices import get_usable_indices
 
 LOGGER = logging.getLogger(__name__)
 
 
-class NativeGridDataset(IterableDataset):
+class NativeGridDataset:
```
anaprietonem, line 26: Why is this no longer inheriting from `IterableDataset`? I understand the interface changed and no longer has an __iter__ method but I think then the docstrings could be updated to reflect that

training/src/anemoi/training/diagnostics/callbacks/evaluation.py
```diff
@@ -154,10 +154,16 @@ def _eval(self, pl_module: pl.LightningModule, batch: torch.Tensor) -> None:
         batch: torch.Tensor
             Batch tensor (bs, input_steps + forecast_steps, latlon, nvar)
         """
-        loss = torch.zeros(pl_module.loss.num_scales, dtype=batch.dtype, device=pl_module.device, requires_grad=False)
+        loss = torch.zeros(
+            1,
```
anaprietonem, line 164: is this not going to conflict with the multiscale loss? ( as in you're hardcoding the num_scales to be always 1 (maybe @JPXKQX  this is what you were talking about the other day, so fine if you're still working on it)

training/src/anemoi/training/losses/loss.py
anaprietonem, line 119: to match what we had before shouldn't this be `config.training.metrics or defaultdict(list)`?

training/src/anemoi/training/schemas/schema_utils.py
```diff
@@ -0,0 +1,15 @@
...
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from typing import Literal
+from typing import TypeAlias
+from typing import TypeVar
+
+T = TypeVar("T")
+DatasetDict: TypeAlias = dict[Literal["datasets"], dict[str, T]]
```
anaprietonem, line 18: I think this could use a comment to explain it's just to identify which dictionaries or entries of the schemas are indexed with dataset names

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -159,20 +170,34 @@ def __init__(
...
+
+        # Create output_mask dictionary for each dataset
+        self.output_mask = {}
+        for name in self.dataset_names:
+            self.output_mask[name] = instantiate(config.model.output_mask, graph_data=graph_data[name])
+
+        # Handle supporting_arrays merge for multi-dataset
+        # Multi-dataset: merge supporting arrays from all output masks
+        combined_supporting_arrays = supporting_arrays.copy()
+        for mask in self.output_mask.values():
```
anaprietonem: If there is a mismatch in the index names between supporting arrays and output_mask this could go mixed up.

models/src/anemoi/models/interface/__init__.py
JPXKQX: Added here: https://github.com/ecmwf/anemoi-utils/pull/256

training/src/anemoi/training/train/train.py
```diff
@@ -188,7 +209,7 @@ def model(self) -> pl.LightningModule:
         "Please use a different activation function."
 
         kwargs = {
-            "config": self.config,
+            "config": convert_to_omegaconf(self.config),
```
anaprietonem, line 212: Why is this changing?

training/src/anemoi/training/train/tasks/interpolator.py
```diff
@@ -64,60 +65,90 @@ def __init__(
...
+                    self.target_forcing_indices[dataset_name] = [self.target_forcing_indices[dataset_name]]
+            else:
+                self.target_forcing_indices[dataset_name] = []
+
+            self.use_time_fraction[dataset_name] = target_forcing_config[dataset_name].time_fraction
 
-        self.use_time_fraction = config.training.target_forcing.time_fraction
+        self.num_tfi = {name: len(idxs) for name, idxs in self.target_forcing_indices.items()}
 
-        self.multi_step = 1
```
anaprietonem, line 78: I believe we had to add this when were looking at the callbacks for the temporal interpolator? @VeraChristina  @mc4117

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -290,8 +323,65 @@ def __init__(
...
+        self.grid_shard_shapes = dict.fromkeys(self.dataset_names, None)
+        self.grid_shard_slice = dict.fromkeys(self.dataset_names, None)
+
+    def _get_loss_name(self) -> str:
+        """Get the loss name for multi-dataset cases."""
+        # For multi-dataset, use a generic name or combine dataset names
+        return "multi_dataset"
+
+    def _get_task_type_from_config(self, config: dict) -> str:
+        task_class_name = str(config.training.model_task).split(".")[-1]
```
anaprietonem: Could this not be implemented with the pydantic schema or during the config validation loading ?

training/src/anemoi/training/train/tasks/interpolator.py
mc4117: Yes and also for inference. Otherwise inference goes looking for the t-6 step in the inputs

training/src/anemoi/training/train/tasks/interpolator.py
```diff
@@ -64,60 +65,90 @@ def __init__(
...
         self.boundary_times = config.training.explicit_times.input
         self.interp_times = config.training.explicit_times.target
         sorted_indices = sorted(set(self.boundary_times + self.interp_times))
         self.imap = {data_index: batch_index for batch_index, data_index in enumerate(sorted_indices)}
 
-        self.rollout = 1
```
mc4117, line 84: Also shouldn't be removed

models/src/anemoi/models/models/interpolator.py
```diff
@@ -174,26 +234,50 @@ def forward(
         if self.latent_skip:
             x_latent_proc = x_latent_proc + x_latent
```
mc4117, line 225: Maybe silly question but this skip connection still work given that they are now lists?

training/src/anemoi/training/train/tasks/base.py
VeraChristina: Ideally, we wouldn't get this from the config but fill it in via a subclass method. (But we need this for the metadata passed on to the model in the init of the task, so the task class isn't initialised yet.) 

If we do it as part of the schema / config validation, how do we handle configs that disable config validation?

training/src/anemoi/training/data/dataset.py
JPXKQX: `MultDataset` is now implementing the `__iter__()` method. We could rename `MultDataset` -> `NativeGridDataset`, and `NativeGridDataset` to something else. This also relates to PR #756

training/src/anemoi/training/diagnostics/callbacks/evaluation.py
JPXKQX: TBD, We will compute the loss for different datasets, different rollout steps and different scales. We need to decide the order of operations to compute the final value. Currently, we add the different datasets (which implies that all datasets have the same number of scales), then the different rollout steps, and finally, we add together all the scales. The main reason for adding the scales at the end is to support its logging in the progress bar and mlflow (I think)

JPXKQX: @ssmmnn11 is already looking into this

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
JPXKQX: It is coming from the lines below. Instead of having 2 `torch.cat`, we have only 1 now. 

The only change we needed was to change the `x_data_latent.shape[:-1]` to `batch_ens_size * x.shape[3]`. In line 87, we can see that x_data_latent.shape[:-1] is the product of (batch, ensemble, grid).

The order of data in the latent space is still the same, so this won't affect the checkpoints migration.

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
JPXKQX: I think it's the same, we are just merging both `einops.rearrange` into one

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
JPXKQX: I think that's the case, we compute 1 `batch_size` and `ensemble_size`. We could just get the batch_size from the first key, but this is a way to check it's consistent

training/src/anemoi/training/config/diagnostics/plot/detailed_cerra.yaml
JPXKQX: I think this file wasn't intended to be committed

models/src/anemoi/models/models/interpolator.py
VeraChristina: At this point `x_latent` and `x_latent_proc` should be tensors. In line 223, the processor takes a tensor and outputs a tensor `x_latent_proc`. In line 210, we combine the latent tensors from multiple datasets into one tensor `x_latent`. So the skip connection doesn't change as it skips the processor (which as before is only one and doesn't handle dicts/lists of tensors).

models/src/anemoi/models/samplers/diffusion_samplers.py
JPXKQX: It shouldn't. We're just copying it from the other diffusion sampler.

training/src/anemoi/training/config/data/multi.yaml
JPXKQX: no, I hope it fails at some point

training/src/anemoi/training/schemas/schema_utils.py
JPXKQX: Done!

training/src/anemoi/training/data/datamodule.py
JPXKQX: This relates to the issue reported by Dieter. Now, the `relative_time_indices` are multiplied by the `timeincrement` inside the dataset. So this `timeincrement` is specific to each dataset, fixing the issue.

training/src/anemoi/training/train/tasks/interpolator.py
JPXKQX: why? is it used somewhere?

models/src/anemoi/models/interface/__init__.py
anaprietonem: thanks both!

training/src/anemoi/training/config/data/multi.yaml
anaprietonem: How difficult would be to do this with the config validation or users would be useful to define this in the top level config and then pass that entry to each section?

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
anaprietonem: All good, yes thanks for the clarification Vera!

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
anaprietonem: Thanks Mario, got it  now!

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
anaprietonem: I meant more the asserts below? as in those are also in deterministic model, diffusion etc

training/src/anemoi/training/train/tasks/base.py
anaprietonem: so currently we have https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/schemas/base_schema.py the `SchemaCommonMixin` so for example `expand_paths` gets applied regardless. I was wondering if something like this could be considered here.

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
VeraChristina: I don't think we dictionarify these, this is just sanity checking

training/src/anemoi/training/data/dataset.py
anaprietonem: I think MultiDataset is fine, maybe `NativeGridDataset` is what I found it a bit less intuitive now as the doscstring still mentions it's an iterative dataset. But fine if we want to address this later.

class NativeGridDataset:
    """Iterable dataset for AnemoI data on the arbitrary grids."""

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
anaprietonem: Yes sorry maybe I was not very clear from my comment, but I mostly was thinking that the assert for batch size and ensemble size is repeated across all models, so we could just make that a method in the base model that extract the ensemble and batch size and check across datasets.

training/docs/adrs/adr-001.md
```diff
@@ -2,56 +2,54 @@
...
-- The checkpoint metadata contains everything needed to describe a variable (without having to reopen the dataset).
-- The model inputs and outputs and the metadata is load bearing, not the dataset.
+- The configuration now supports a dictionary-based approach ("dict-ification") for datasets and related components.
+- The main entry point is a `datasets:` field, which is a dictionary where each key is a dataset name and the value is its configuration.
+- For each dataset, you can specify different `forcing`, `diagnostic`, and `target` variables, enabling flexible use cases such as emulation, downscaling, and more.
+- This approach allows users to define per-dataset settings for data processing, normalisation, variable selection and others.
+- The single-dataset configuration is deprecated in favor of this new structure, but backward compatibility is temporarily maintained with a deprecation warning (without config validation).
+- Not all components are yet configurable per-dataset. For example, `graphs` and `model` remain global for now. Per-dataset configuration for these will be considered in follow-up PRs.
```
anaprietonem: I'd clarify here that for graphs - providing pre-built graphs for each dataset would overcome the problem for now, and that for models - encoder and processors do get instantiated per dataset but that at a config level, it's yet not possible to configure those mappers differently.

training/docs/adrs/adr-001.md
```diff
@@ -2,56 +2,54 @@
...
 ## Consequences
 
-<!--Discuss the impact of this decision, including benefits, trade-offs, and potential technical debt.-->
+**Benefits:**
+- Enables advanced workflows (downscaling, emulation, etc.) by allowing different datasets and variable roles in a single configuration.
+- Lays the foundation for future per-dataset configuration in other components (e.g., graphs, model).
+
+**Trade-offs:**
+- Some configuration complexity increases due to the dictionary-based structure.
+- Not all components are yet fully per-dataset configurable. Further work is needed.
```
anaprietonem, line 45: I would mention here for now this is for reanalysis dataset, further work in the future will include support for observation datasets.

anaprietonem: And also I would mention that for now the dataset also need to have temporal alignment,  and this will be kept as a requirement for now.

JPXKQX: yes, that's important

training/docs/adrs/adr-001.md
JPXKQX: Done!

training/src/anemoi/training/config/data/multi.yaml
JPXKQX: Not sure. I imagine checking 2 fields with the same keys would be feasible, but it wouldn't be easy with so many fields. I also think this would break compatibility with old configurations, as I would already need a dataset name. It is probably worth discussing this later if it turns out to be a recurrent issue.

training/src/anemoi/training/data/dataset.py
JPXKQX: True, docstring wasn't updated. It should be updated now

training/src/anemoi/training/diagnostics/callbacks/evaluation.py
JPXKQX: It's fixed now

training/src/anemoi/training/data/datamodule.py
JPXKQX: They are added back now

training/src/anemoi/training/train/tasks/base.py
JPXKQX: Well spotted! This was wrong. It should be fixed now

training/src/anemoi/training/losses/loss.py
JPXKQX: I think this function is still applied per-dataset, so zconfig.training.metrics` is no longer a good default

training/src/anemoi/training/config/data/multi.yaml
MeraX: > How difficult would be to do this with the config validation or users would be useful to define this in the top level config and then pass that entry to each section?

Would validation be somewhat easier if dataset names were not keys but values like in the graphs section? e.g.
```
datasets:
  - name: era5
    forcing: ...
  - name: cera
    forcing: ...
```

training/docs/user-guide/training.rst
```diff
@@ -400,12 +414,13 @@ sum equals 0.25 of the sum of all node weights with the following config
 setup
 
 .. code:: yaml
-
-   node_loss_weights:
-      _target_: anemoi.training.losses.nodeweights.ReweightedGraphNodeAttribute
-      target_nodes: data
-      scaled_attribute: cutout
-      weight_frac_of_total: 0.25
```
VeraChristina, line 408: Is this still up to date? Or do we do this via scalers now? Noticed it when doing the dictionaryfication indentation

JPXKQX: This is now done via scalers, so it is out of date. I would delete this paragraph and refer to the Loss section instead.

training/src/anemoi/training/train/tasks/base.py
VeraChristina: As discussed offline, we now get the task type from the task instead of the config

training/docs/user-guide/training.rst
VeraChristina: done!

JPXKQX: thanks!

training/src/anemoi/training/train/tasks/interpolator.py
mc4117: Yes in the callbacks and in inference too

JPXKQX: We can keep it then, but we should start thinking about why we need the rollout here

JPXKQX: Done!

training/src/anemoi/training/config/data/multi.yaml
JPXKQX: Hi Marek, I think this decision was made to ensure consistency with other parts of the code, such as the graph, scalers, metrics and losses. However, I agree that this is a valid point, so I would recommend raising it as a separate piece of work

Issue #513, opened 2025-09-01
Move paths and files out of hardware in the configs


matschreiner:
Paths for inputs/outputs, datasets, and log locations are currently configured under `hardware`. This is unintuitive as I think hardware should describe compute resources like GPUs, devices, and nodes, not files/paths. 
I propose introducing a new section in the config called `filesystem` or `io` and deprecating path and files keys under hardware.

anaprietonem:
Hey @matschreiner  thanks for opening this issue. What I find  useful from the current 'hardware' config setup is that a single file captures all the keys related to transferring and setting up anemoi on a given system. So for example when moving from ATOS to Leonardo I know the hardware config is the one I need to change as details about file paths and hardware are different. Separating this further could then make this process more cumbersome. What are your thought about this?

@floriankrb  Also suggested that the part related to files/paths is not ideally split as probably all things related to reading files could be controlled by the settings.toml and just configure the output paths. There is a PR https://github.com/ecmwf/anemoi-core/pull/238 where we are trying to simplify this point. 

We can adapt that PR to address your point and rethink how the hardware config is set up!

matschreiner:
@floriankrb @anaprietonem 
Thanks both, ok, I see the benefit of having a single place to touch when moving between HPCs. But I think there's still small itch of having files and paths under hardware - what would you think of having both hardware and fs stuff under a system config? So the hardware folder would be called system and the example config would instead look like 

``` 
.../config/system/example.yaml 
defaults:
-  paths: ...
-  files: ...
-  hardware: ...
``` 

Is the paths exclusively read only? How about if you want to train on your own dataset?

anaprietonem:
Naming that system sounds good and I think probably is less confusing to users, so I like that suggestion. Paths could just be about defining output paths for logs etc. And so then in files we could make it so if people want to pass an absolute path to their dataset (as the case you suggest @matschreiner then they can do so) but if just passed a dataset name then the path is constructed based on ~/.config/anemoi/settings.toml (as @floriankrb  said with some priority policy. It might then be worth renaming those categories to:
```
.../config/system/example.yaml 
defaults:
-  filesystem:
         - inputs
         - outputs
-  hardware: ...
```

PR #598, opened 2025-10-10
fix(training)!: Refactor configuration by introducing system schema with hardware, paths, and files subschemas

matschreiner:
# Description
This PR reorganizes the configuration structure by introducing a new top-level schema called system, which groups the subschemas hardware, storage, and files (see [issue #513](https://github.com/ecmwf/anemoi-core/issues/513)).

Previously, paths and files were defined under hardware, which was confusing since they describe filesystem layout and not than compute resources.
It also wasn’t clear whether paths refer to directories or files. It has now been renamed to storage, clarifying its role as the definition of directory structure for inputs, outputs, and logs.

```
system/
├── hardware.yaml
├── input.yaml
└── output.yaml
```

The PR also isolates the concatenation logic for paths in the pydantic scheme so we don't need to write out the full paths for all outputs/logs/etc in each field. This is very brittle and used to happen in both code throughout the framework and inside the nested configuration files. This is now isolated to happen in only one place. 

----
📚 Documentation preview 📚: https://anemoi-training--598.org.readthedocs.build/en/598/

----
📚 Documentation preview 📚: https://anemoi-graphs--598.org.readthedocs.build/en/598/

----
📚 Documentation preview 📚: https://anemoi-models--598.org.readthedocs.build/en/598/

### DIFF
--- a/training/tests/integration/config/test_stretched.yaml
+++ b/training/tests/integration/config/test_stretched.yaml
@@ -86,7 +86,7 @@ diagnostics:
         every_n_batches: ${diagnostics.plot.frequency.batch}
         #Defining the accumulation levels for precipitation related fields and the colormap
         accumulation_levels_plot: [0, 0.05, 0.1, 0.25, 0.5, 1, 1.5, 2, 3, 4, 5, 6, 7, 100] # in mm
-        precip_and_related_fields: ${diagnostics.plot.precip_and_related_fields}
+        precip_and_related_fields: ["tp"]
         colormaps: ${diagnostics.plot.colormaps}
       # - _target_: anemoi.training.diagnostics.callbacks.plot.PlotSpectrum
       #   # every_n_batches: 100 # Override for batch frequency
@@ -98,6 +98,6 @@ diagnostics:
       - _target_: anemoi.training.diagnostics.callbacks.plot.PlotHistogram
         sample_idx: ${diagnostics.plot.sample_idx}
         every_n_batches: ${diagnostics.plot.frequency.batch}
-        precip_and_related_fields: ${diagnostics.plot.precip_and_related_fields}
+        precip_and_related_fields: ["tp"]
         parameters:
         - tp
--- a/training/tests/integration/config/test_lam.yaml
+++ b/training/tests/integration/config/test_lam.yaml
@@ -21,10 +21,27 @@ system:
     truncation: null
     truncation_inv: null
 
-
 # need this modification since some variables set in the default zarr config are not part of the dataset
 data:
   forcing:
+  - "cos_latitude"
+  - "cos_longitude"
+  - "sin_latitude"
+  - "sin_longitude"
+  - "cos_julian_day"
+  - "cos_local_time"
+  - "sin_julian_day"
+  - "sin_local_time"
+  diagnostic:
+  - "tp"
+
+  normalizer:
+    default: "mean-std"
+    std:
+    - "tp"
+    min-max:
+    max:
+    none:
     - "cos_latitude"
     - "cos_longitude"
     - "sin_latitude"
@@ -56,7 +73,7 @@ diagnostics:
         every_n_batches: ${diagnostics.plot.frequency.batch}
         #Defining the accumulation levels for precipitation related fields and the colormap
         accumulation_levels_plot: [0, 0.05, 0.1, 0.25, 0.5, 1, 1.5, 2, 3, 4, 5, 6, 7, 100] # in mm
-        precip_and_related_fields: ${diagnostics.plot.precip_and_related_fields}
+        precip_and_related_fields: [tp]
         colormaps: ${diagnostics.plot.colormaps}
       - _target_: anemoi.training.diagnostics.callbacks.plot.PlotSpectrum
         # every_n_batches: 100 # Override for batch frequency
@@ -68,6 +85,6 @@ diagnostics:
       - _target_: anemoi.training.diagnostics.callbacks.plot.PlotHistogram
         sample_idx: ${diagnostics.plot.sample_idx}
         every_n_batches: ${diagnostics.plot.frequency.batch}
-        precip_and_related_fields: ${diagnostics.plot.precip_and_related_fields}
+        precip_and_related_fields: [tp]
         parameters:
         - tp


training/src/anemoi/training/config/system/example.yaml
```diff
@@ -1,10 +1,5 @@
...
 defaults:
-- paths: example
-- files: example
-
-# number of GPUs per node and number of nodes (for DDP)
-accelerator: auto
-num_gpus_per_node: 1
-num_nodes: 1
-num_gpus_per_model: 1
+  - storage: example
```
anaprietonem: If I am following correctly, shouldn't this be system/example.yaml? (I think that's why now the unit tests are failing too)

training/src/anemoi/training/config/training/lam.yaml
```diff
@@ -113,13 +113,13 @@ validation_metrics:
 variable_groups:
   default: sfc
   pl:
-     param: [q, t, u, v, w, z]
+    param: [q, t, u, v, w, z]
```
anaprietonem, line 115: just curious, is this because of a different formatter being used in the yamls?

training/tests/integration/config/test_stretched.yaml
```diff
@@ -2,9 +2,9 @@
 dataloader:
   dataset:
     cutout:
-      - dataset: ${hardware.paths.data}/${hardware.files.dataset}
+      - dataset: ${system.storage.dataset}/${system.files.dataset}
         thinning: 25 # ???-value in stretched config
-      - dataset: ${hardware.paths.data}/${hardware.files.forcing_dataset}
+      - dataset: ${system.storage.dataset}/${system.files.forcing_dataset}
```
anaprietonem: I can't see a key called dataset in storage, should this be `data`?

training/src/anemoi/training/schemas/base_schema.py
```diff
@@ -122,7 +122,7 @@ class UnvalidatedBaseSchema(PydanticBaseModel):
     """Datamodule configuration."""
     diagnostics: Any
     """Diagnostics configuration such as logging, plots and metrics."""
-    hardware: Any
+    system: Any
     """Hardware configuration."""
```
anaprietonem, line 148: should this say system configuration including hardware and I/O settings?

training/src/anemoi/training/config/training/lam.yaml
matschreiner: I think so! 
Maybe we could add formatting specifications in the pyproject file for consistency?

training/src/anemoi/training/config/system/input/example.yaml
```diff
@@ -0,0 +1,6 @@
+dataset: ???
+graph: ???
```
mchantry, line 2: In the previous version graph paht had a defined default. Could this be considered here?

matschreiner: sure! Lets see if this works: @copilot add a default path for the graph.

anaprietonem:
There is now a migration script so checkpoints won't break. Code runs with pydantic on or off but behaviour is different (in terms how output folders are build), so we should address this.  (mlflow [runs](https://mlflow.ecmwf.int/#/compare-runs?runs=[%22a99ea6ca877c4a1f9c6eb0aa1ec62810%22,%222f955be2eab9472db35ba2bf47e88c1e%22]&experiments=[%22264%22]))
Screenshot 2025-10-31 at 10 58 52

potential solution: have the expand paths functionality available to for the unvalidated schema

anaprietonem:
@dietervdb-meteo @matschreiner  - I did some changes so that behaviour should be now equivalent from a 'paths' perspective with and without config validation. Would you have some time to test it or review this? There is also now a migration script so that hopefully helps to not break the checkpoints

dietervdb-meteo (changes requested):
Ok, this seems to work and makes sense. Two comments though:
1. why is `root` a required entry for `output`? I have a setup where I like certain outputs (logs) to go one way and other ouptuts (checkpoints) a quite different way. I can set this up by specifying `root: ""` but `root: ` is not accepted, nor is just leaving out the `root` entry, both of which one could naturally expect to coincide with that.
2. Previously checkpoints where stored as `<ckpt_folder_from_config>/<run_id>/last.ckpt` now they seem to be stored as `<ckpt_folder_from_config>/last.ckpt`. Is this intended? I think having the `<run_id>` subfolder added automatically is a good way to protect against overwriting checkpoints by accident. (It might also be needed for run continuation?)

anaprietonem:
> Ok, this seems to work and makes sense. Two comments though:
> 
> 1. why is `root` a required entry for `output`? I have a setup where I like certain outputs (logs) to go one way and other ouptuts (checkpoints) a quite different way. I can set this up by specifying `root: ""` but `root: ` is not accepted, nor is just leaving out the `root` entry, both of which one could naturally expect to coincide with that.
> 2. Previously checkpoints where stored as `<ckpt_folder_from_config>/<run_id>/last.ckpt` now they seem to be stored as `<ckpt_folder_from_config>/last.ckpt`. Is this intended? I think having the `<run_id>` subfolder added automatically is a good way to protect against overwriting checkpoints by accident. (It might also be needed for run continuation?)

Hey Dieter, thanks for having a look.

2. I will have a look at this, it should be adding the run_id otherwise this will be an issue when running multiple runs at the same time.
1. @matschreiner  maybe you can comment on this one? I kept the original design. But Dieter just to check I follow - your point is you'd like to decouple the building of the paths so the root defined for checkpoints and plots or logs, is not dependent on a high-level system root?

Screenshot 2025-11-10 at 10 07 20
Screenshot 2025-11-10 at 10 07 52

dietervdb-meteo:
I would simply suggest having `""` as a default for that root path, and possibly interpreting `None` i.e. `root: ` also in that way.
 
I.e. have a setup where the following three are equivalent: 
a) `root` not specified in output config.
b) `root: ` in the output config
c) `root: ""` in the output config

Currently only c) is supported.

dietervdb-meteo (approved):
As previously, LGTM

dietervdb-meteo:
Very good point... I overlooked that when reviewing, sorry. So used to go and look for config changes in the default configs inside the code rather than the docs...

PR #600, opened 2025-10-13
fix(models): replace placeholder of migration script anemoi-models version in main

bdvllrs:
## Description
Before a new version is published, the anemoi-models version stored in migration scripts is set as a placeholder `%NEXT_ANEMOI_MODELS_VERSION%`. If the migrations are run on the main branch with script still having the placeholder, the checkpoint will incorrect information of the version of the scripts.

This replaces the placeholder with the current `__version__` of anemoi-models (as the future version is impossible to predict).

This will allow a better feedback by `anemoi-models migration inspect`. 

##  Additional notes ##

### DIFF
--- a/.github/workflows/integration-tests-hpc.yml
+++ b/.github/workflows/integration-tests-hpc.yml
@@ -51,7 +51,7 @@ jobs:
             #SBATCH --qos=ng
             #SBATCH --gpus=1
             #SBATCH --gres=gpu:1
-            #SBATCH --mem=16G
+            #SBATCH --mem=30G
           troika_user: ${{ secrets.HPC_CI_INTEGRATION_USER }}
   benchmark-tests:
     runs-on: hpc


anaprietonem:
sorry what do you mean with `This is also only for models trained on main and not on a specific version, so (hopefully) not for production models.`? and does `0.3.dev1` have an impact of the way compatibility groups are defined or when running the sync?

bdvllrs:
Sorry, my mistake. It will not affect trained models but I meant models migrated to the main branch, or to a specific PR. 

This does not affect compatibility group. However as the script checksum signature will change (as the version of anemoi-models will change from the placeholder to the actual version), the migrator script will fail to migrate further migrations.  But this is already the case without this fix.

anaprietonem:
> Here are some answers to your questions:
> 
> * the checkpoint in between are tested correctly by CI because we restart all migrations from a testing checkpoint that should not contain migration scripts with placeholders. The fact that the tested script contain placeholder does not apply the quality of testing.
> * Gert's pipeline will still update the version in the scripts because my fix does not change the scripts, it only changes the perceived version when reading the migrations.

Thanks Benjamin, I think then I understand better the scope of the fix, and it's fine to merge!

anaprietonem (approved):

PR #609, opened 2025-10-14
feat: new interactive graph visualization

frazane:
## Description
Adds a new lightweight way to visualize graphs interactively. No added dependencies (the javascript code runs entirely in the broswer and sources libraries via CDN). The way this is implemented is quite simple: the data is embedded directly in the HTML in json format using a jinja template. Then javascript code does the rest (yes, definitely with the help of coding assistants).

The largest graph on which this was tested was a cutout graph with n320 grid globally and a 2km grid over central Europe (the one we currently have at MeteoSwiss). Loaded in about 6s and then still runs relatively smoothly. Still, we might be reaching a limit with >1'500'000 nodes with lots of connections - we might need to find some optimizations to the code at some point as I am sure there are some left.

Nothing changes in terms of interface - this will simply add another output to the command:

```shell
anemoi-graphs inspect graph.pt output_dir/
```

## Features
- fast GPU-based rendering
- full 3D scene navigation
- interactive controls for nodes and edges (toggles, radius scaling, colors, etc.)
- nodes selection and highlighting
- globe map
- only for nodes coordinates and edges (no attributes values, for now) 

## Graphs tested
- [x] n320 global (+ hidden)
- [x] CERRA LAM graph from RMI (+ hidden)
- [x] COSMO SGM graph from MeteoSwiss (+hidden)
- [x] n320 global hierarchical (+ multiple hidden levels) 

----
image

----
📚 Documentation preview 📚: https://anemoi-training--609.org.readthedocs.build/en/609/

----
📚 Documentation preview 📚: https://anemoi-graphs--609.org.readthedocs.build/en/609/

----
📚 Documentation preview 📚: https://anemoi-models--609.org.readthedocs.build/en/609/

### DIFF
--- a/graphs/src/anemoi/graphs/plotting/interactive_3d_html.py
+++ b/graphs/src/anemoi/graphs/plotting/interactive_3d_html.py
@@ -61,7 +61,7 @@ def extract_nodes_edges(
         edges_keys = graph.edge_types
 
     out_nodes = {n: graph[n].x for n in nodes}
-    out_edges = {"_".join(e) : graph[e].edge_index for e in edges_keys}
+    out_edges = {"_".join(e): graph[e].edge_index for e in edges_keys}
 
     return out_nodes, out_edges
 
@@ -84,6 +84,7 @@ def to_nodes_json(lats, lons, prefix="P"):
 def to_edges_json(names1, names2, pairs):
     edges = [[names1[i], names2[j]] for i, j in pairs]
     return edges
+
 
 def plot_interactive_graph_3d(
     graph: HeteroData,


graphs/src/anemoi/graphs/plotting/interactive_html_light.py
```diff
@@ -0,0 +1,907 @@
+import numpy as np
+import torch
+from jinja2 import Template
+
+HTML_TEMPLATE = """
```
HCookie: Can you move the html template into a dedicated jinja2 file?

dietervdb-meteo:
Hi @frazane , is this always plotting things on a globe? LAM graphs for example cover only a limited area, would they get rendered (on part of the globe?) in the current implementation? (Don't have time to test now, can send you a LAM graph in case you would have time, in any case this comment should not stop you merging/developping this PR)

JPXKQX:
Hi Francesco, sorry for the delayed review. I just have one minor comment about the location: could we put it in `interactive_3d_html.py` and rename the previous one to `interactive_2d_html.py`, and use the current CLI from inspect.py? This would allow us to reuse some GraphInspector functionality, such as loading the graph or subsetting it with the area argument.

Also, I would set the default edges to 'None' and iterate over 'graph.edge_types', which will return all the edges already defined in the graphs.

Thanks for the contribution, really cool! Looking forward to have it merged

frazane:
@ecmwf/anemoisecurity please ignore the review request, It was created by mistake during a merge with main. Note to self: "never use VSCode git UI ever again" :)

matschreiner:
Hi @frazane — this looks awesome, thanks a lot for implementing it!
I was on vacation last week so I didn’t get a chance to review earlier.
Would it make sense for me to open an issue with a few comments?

1) When zooming in, navigation becomes a tricky because the rotation sensitivity doesn’t seem aligned with the zoom level. Even a small drag rotates the globe quite a lot when zoomed in.

2) Could node scaling depending on connection be optional? In some views - for example on stretched grid the global nodes become extremely small compared to the cutout nodes e.g., in the screenshot below there are actually four global nodes, but they’re too tiny to see.

3) Why are the nodes rendered as triangles instead of circles/dots?

screenshot_1763034576

PR #629, opened 2025-10-23
fix(models): processor chunking

japols:
## Description
Remove the separate ProcessorChunk class and flatten all layers directly into the BaseProcessor.
Chunking is now handled dynamically at runtime by grouping layers into checkpointed segments.

Previously, the Processor class held a list of ProcessorChunks which held its own ModuleList of layers, meaning that checkpointed layer groupings were tied to the chunking configuration saved in the model checkpoint.
When resuming training with a different num_chunks, the restored module structure no longer matched the saved one, causing checkpoint mismatches. Now we only have one flat list of all layers (Blocks) in the Processor Class and chunking is handled dynamically.

##  Additional notes ##
Tested with all models, i.e. GT, Transformer, GNN, PointWiseMLP

----
📚 Documentation preview 📚: https://anemoi-training--629.org.readthedocs.build/en/629/

----
📚 Documentation preview 📚: https://anemoi-graphs--629.org.readthedocs.build/en/629/

----
📚 Documentation preview 📚: https://anemoi-models--629.org.readthedocs.build/en/629/

### DIFF
--- a/models/src/anemoi/models/distributed/transformer.py
+++ b/models/src/anemoi/models/distributed/transformer.py
@@ -18,6 +18,42 @@ from torch.distributed.distributed_c10d import ProcessGroup
 from anemoi.models.distributed.utils import get_memory_format
 
 
+def _alltoallwrapper(output_list: list, input_list: list, group: ProcessGroup):
+    """Wrapper function for all_to_all across NCCL, MPI and Gloo backends.
+    There is no all_to_all primitive for the Gloo backend. In that case each
+    process broadcasts its tensor asynchronously.
+
+    Retuns nothing but modifies output_list in-place
+
+    """
+    comm_size = dist.get_world_size(group=group)
+
+    if dist.get_backend(group) == "gloo":
+
+        # Need to check torch version here bc the syntax for dist.send/recv changed in torch v2.6
+        torch_version = torch.__version__.split(".")
+        torch_major_version = int(torch_version[0])
+        torch_minor_version = int(torch_version[1])
+        if torch_major_version <= 2 and torch_minor_version < 6:
+            raise NotImplementedError("Gloo all_to_all not implemented for torch < v2.6")
+
+        reqs = []
+        rank = dist.get_rank(group=group)
+        # Here we implement the linear shift algorithm from Hofmann and Ruenger, 2013
+        for i in range(0, comm_size):
+            j = (i - rank + comm_size) % comm_size
+            if j != rank:
+                # exchange data with rank j
+                reqs.append(dist.isend(input_list[j], group_dst=j, group=group))
+                reqs.append(dist.irecv(output_list[j], group_src=j, group=group))
+            else:
+                output_list[rank] = input_list[rank]
+        for req in reqs:
+            req.wait()
+    else:
+        dist.all_to_all(output_list, input_list, group=group)
+
+
 def _headsalltoall(input_: Tensor, shapes: list, group: Optional[ProcessGroup] = None) -> Tensor:
     """Apply all_to_all along the head dimension.
 
@@ -52,7 +88,7 @@ def _headsalltoall(input_: Tensor, shapes: list, group: Optional[ProcessGroup] =
         for rank in range(comm_size)
     ]
 
-    dist.all_to_all(output_list, input_list, group=group)
+    _alltoallwrapper(output_list, input_list, group=group)
 
     # Note: torch.cat already creates a contiguous tensor.
     return torch.cat(output_list, dim=-2).contiguous(memory_format=input_format)
@@ -79,7 +115,7 @@ def _seqalltoall(input_: Tensor, shapes: list, group: Optional[ProcessGroup] = N
 
     output_list = [torch.empty_like(input_list[comm_rank]) for _ in range(comm_size)]
 
-    dist.all_to_all(output_list, input_list, group=group)
+    _alltoallwrapper(output_list, input_list, group=group)
 
     # Note: torch.cat already creates a contiguous tensor.
     return torch.cat(output_list, dim=-3).contiguous(memory_format=input_format)


training/src/anemoi/training/config/model/point_wise.yaml
```diff
@@ -33,16 +33,28 @@ processor:
 
 encoder:
```
japols, line 34: @NRaoult I had to change these defaults to run the point wise model, what do you think?

NRaoult: Looks good to me, but maybe also check with @JPXKQX ?

JPXKQX: Perfect!

models/src/anemoi/models/layers/processor.py
```diff
@@ -111,11 +118,10 @@ def __init__(
         )
 
         self.build_layers(
-            PointWiseMLPProcessorChunk,
+            PointWiseMLPProcessorBlock,
             num_channels=num_channels,
-            num_layers=self.chunk_size,
+            hidden_dim=(mlp_hidden_ratio * num_channels),
```
matschreiner, line 149: Should this be hidden_channels?

models/src/anemoi/models/layers/processor.py
```diff
@@ -213,14 +219,13 @@ def __init__(
...
             num_channels=num_channels,
-            num_layers=self.chunk_size,
-            layer_kernels=self.layer_factory,
-            mlp_hidden_ratio=mlp_hidden_ratio,
+            hidden_dim=(mlp_hidden_ratio * num_channels),
             num_heads=num_heads,
-            window_size=window_size,
             qk_norm=qk_norm,
+            window_size=window_size,
+            layer_kernels=self.layer_factory,
```
matschreiner, line 249: why does this take both the block TransformerProcessorBlock and a layer_factory?

models/src/anemoi/models/layers/processor.py
japols: I aggree, that's indeed a better name. Not sure if renaming block components should be part of this PR though as it's mainly about restructuring the chunking by moving code around.

models/src/anemoi/models/layers/processor.py
japols: layer_factory provides linear and layernorm kernels to the block

matschreiner: Okay, so the layer factory is not the layers built by the build_layers function. I think this function is hard to read. Maye it could be made clearer if the signature was something like 

```
build_layers(self, layer_class, num_layers, layer_kwargs):
    ...
```

models/src/anemoi/models/layers/block.py
```diff
@@ -242,6 +242,7 @@ def __init__(
         mlp_extra_layers: int = 0,
         update_src_nodes: bool = True,
         layer_kernels: DotDict,
+        edge_dim: Optional[int] = None,
```
anaprietonem, line 245: is this change related to the chunking fix?

models/src/anemoi/models/layers/block.py
anaprietonem: shouldn't this also be tuple[Tensor,Tensor]?

models/src/anemoi/models/layers/processor.py
```diff
@@ -420,14 +436,15 @@ def __init__(
...
-            num_layers=self.chunk_size,
-            layer_kernels=self.layer_factory,
+            GraphTransformerProcessorBlock,
+            in_channels=num_channels,
+            hidden_dim=(mlp_hidden_ratio * num_channels),
+            out_channels=num_channels,
             num_heads=num_heads,
-            mlp_hidden_ratio=mlp_hidden_ratio,
-            qk_norm=qk_norm,
+            num_chunks=1,
```
anaprietonem: why are we are setting the num_chunks=1 here? and how does this relate to then having in the configs `  num_chunks: 4` for example for the GraphTransformerProcessor?

models/src/anemoi/models/layers/processor.py
```diff
@@ -48,10 +48,10 @@ def __init__(
         """Initialize BaseProcessor."""
```
jakob-schloer:
```suggestion
        """Initialize BaseProcessor.

        Parameters
        ----------
        num_layers : int
            Number of processor layers.
        num_channels : int
            Number of channels, i.e. feature dimension of the processor state.
        num_chunks: int
            Number of chunks of the processor. The num_chunks and num_layers, defines how many layers are grouped together for checkpointing, i.e. chunk_size = num_layers/ num_chunks. 
        cpu_offload : bool
            Whether to offload processing to CPU, by default False
        layer_kernels : DotDict
            A dict of layer implementations e.g. layer_kernels.Linear = "torch.nn.Linear"
            Defined in config/models/<model>.yaml
        **kwargs : dict
            Additional keyword arguments
        """
```

models/src/anemoi/models/layers/processor.py
```diff
@@ -63,22 +63,29 @@ def offload_layers(self, cpu_offload):
...
             ],
         )
 
+    def run_layer_chunk(self, chunk_start: int, data: tuple, *args, **kwargs) -> tuple:
+        for layer_id in range(chunk_start, chunk_start + self.chunk_size):
+            data = self.proc[layer_id](*data, *args, **kwargs)
+
+        return data
+
     def run_layers(self, data: tuple, *args, **kwargs) -> Tensor:
```
jakob-schloer:
```suggestion
    def run_layers(self, data: tuple, *args, **kwargs) -> tuple:
```

jakob-schloer: I wonder why we return tuples because we never really make use of them. Even in the GNN and GraphTransformer block the returned `edge_attr` are not used. However, this is a general design question and goes beyond the PR. Anyway, at the moment every return is a tuple :)

models/src/anemoi/models/layers/processor.py
jakob-schloer: Just a suggestion to update the docstrings to make it clearer what chunks actually are.

models/src/anemoi/models/layers/block.py
japols: yes it is, I had to move that logic from chunk.py to block.py

models/src/anemoi/models/layers/processor.py
japols: it's the "inner" chunking that only works in inference via env variable (should be replaced with something better soon)... just moved that from chunk.py to here

models/src/anemoi/models/layers/block.py
japols: true :)

models/src/anemoi/models/layers/processor.py
japols: I think the GNNForwardMapper also updates src nodes which is probably the reason for tuples here @ssmmnn11 ?

models/src/anemoi/models/layers/processor.py
japols: removed num_chunks from block initialisation, now it's only configurable via env variable during inference and set to 1 otherwise (as it was before)

training/src/anemoi/training/utils/checkpoint.py
```diff
@@ -80,6 +83,9 @@ def transfer_learning_loading(model: torch.nn.Module, ckpt_path: Path | str) ->
     # Load the checkpoint
     checkpoint = torch.load(ckpt_path, weights_only=False, map_location=model.device)
 
+    # apply chunking migration (fails silently otherwise leading to hard to debug issues)
```
anaprietonem, line 86: can we add a comment that this 'silent failing' is due the strict=False and that in the future we will look at more robust way to implement this to avoid this hack?

japols, line 86:
```suggestion
    # apply chunking migration (fails silently otherwise leading to hard to debug issues)
    # this is due to loading with strict=False, planning to make this more robust in the future
```

PR #209, opened 2025-03-21
feat: Timeweights

OpheliaMiralles:
## Description

Originally designed to be able to provide lead time decay weights for nowcasting, but can be useful for any time of weighting scheme in the time dimension.

## Type of Change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Issue Number

Partially solves #63 

## Code Compatibility

-   [x] I have performed a self-review of my code

### Code Performance and Testing

-   [ ] I have added tests that prove my fix is effective or that my feature works
-   [ ] I ran the [complete Pytest test](https://anemoi.readthedocs.io/projects/training/en/latest/dev/testing.html) suite locally, and they pass
-   [ ] I have tested the changes on a single GPU
-   [ ] I have tested the changes on multiple GPUs / multi-node setups
-   [ ] I have run the Benchmark Profiler against the old version of the code
-   [x] If the new feature introduces modifications at the config level, I have made sure to update Pydantic Schemas and default configs accordingly

### Dependencies

-   [ ] I have ensured that the code is still pip-installable after the changes and runs
-   [ ] I have tested that new dependencies themselves are pip-installable.
-   [x] I have not introduced new dependencies in the inference portion of the pipeline

### Documentation

-   [x] My code follows the style guidelines of this project
-   [x] I have updated the documentation and docstrings to reflect the changes
-   [x] I have added comments to my code, particularly in hard-to-understand areas

## Additional Notes

The relative_date_indices might overlap with #153 for some parts, given that it was originally inspired by the time interpolation branch published on anemoi-training before the refactoring.

PR #636, opened 2025-10-31
feat: multi output steps

dietervdb-meteo:
## Description
This PR provides flexibility to have the model predict multiple time steps at once, i.e. apart from mutliple input steps one can now also specify multiple output steps. It supports, for example, a forecaster that takes say three inputs, [-6h, -3h, 0h] and forecasts two output times  [+3h, +6h]. This is all in a single model step, as before one can forecast further in the future through rollout. A rollout step will, in the current implementation, bridge all output steps:
<p align="center">
illustration
</p>
Experimentation with these multiple output step models has shown they are a promising approach to increase time resolution of forecasters while keeping the error growth through rollout under control. If you want to experiment yourself with such models it is recommended to use the frozen branch `multi-out-tmp`, rather than the one this PR is based on, since this one can still undergo drastic changes until merger.

  
## Features added
* Support for all forecasters (deterministtic, ens-crps, ens-(tend)diffusion) to predict multiple future lead times in a single model forward, and propagate these throug rollout.
* Support for the `Interpolator`  to predict all target times at once
* Loss scalers allowing separate weights for the different output steps

## Notes on the technical implementation/design:
1. This PR introduces an additional dimension (time) to the model output. `x_out` now has the dimensions (batch, time, ensemble, grid, variables), for all currently implemented models. 
2. It has been a design choice to create each output step as a difference to the last input step. One could alternatively consider to create them as a cumulative difference wrt the last input step (suggested by @frazane ) and other options could be possible. It seems reasonable to stick with this implementation for now and refer (configurable?) alternatives to the future.
4. It has been a design choice to have a single rollout step cover all output steps. E.g. if we have a model with output steps [3h, 6h] then  rollout 1 is [3h, 6h], rollout 2 is [9h, 12h] etc. One could imagine a rollout step that covers only a first few output steps. It is suggested to leave such (configurable?) option to possible future extensions. See some comments below for alternatives to consider.
5. We have currently chosen to provide a default value of 1 for `config.training.multistep_output`, so that one is not forced to specify this variable explicitly in the config. This makes this PR non-breaking. Making the specification of time steps (both in and output) more flexibly configurable is something being considered for the future, which would break configs. We thus opt to break the config at such a later stage once, rather than have a temporary breaking change by this first implementation.

----
📚 Documentation preview 📚: https://anemoi-training--636.org.readthedocs.build/en/636/

----
📚 Documentation preview 📚: https://anemoi-graphs--636.org.readthedocs.build/en/636/

----
📚 Documentation preview 📚: https://anemoi-models--636.org.readthedocs.build/en/636/

### DIFF
--- a/graphs/src/anemoi/graphs/nodes/builders/from_file.py
+++ b/graphs/src/anemoi/graphs/nodes/builders/from_file.py
@@ -174,6 +174,7 @@ class LimitedAreaNPZFileNodes(NPZFileNodes):
         margin_radius_km: float = 100.0,
     ) -> None:
         super().__init__(npz_file, name, lat_key, lon_key)
+
         self.area_mask_builder = KNNAreaMaskBuilder(reference_node_name, margin_radius_km, mask_attr_name)
 
     def register_nodes(self, graph: HeteroData) -> None:
--- a/training/src/anemoi/training/diagnostics/callbacks/evaluation.py
+++ b/training/src/anemoi/training/diagnostics/callbacks/evaluation.py
@@ -54,7 +54,7 @@ class RolloutEval(Callback):
     def _eval(
         self,
         pl_module: pl.LightningModule,
-        batch: torch.Tensor | dict[str, torch.Tensor],
+        batch: dict[str, torch.Tensor],
     ) -> None:
         batch_tensor = batch
         if isinstance(batch, dict):


models/src/anemoi/models/models/base.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         self._graph_name_data = model_config.graph.data
         self._graph_name_hidden = model_config.graph.hidden
         self.multi_step = model_config.training.multistep_input
+        self.multi_out = model_config.training.multistep_output
```
OpheliaMiralles: Would be nice to have more explicit names for input/output time steps, e.g. `self.input_timesteps` and `self.output_timesteps` would be clearer to me than `multi_step` and `multi_out`

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -128,15 +128,19 @@ def _assemble_output(self, x_out, x_skip, batch_size, ensemble_size, dtype):
...
             )
             .to(dtype=dtype)
             .clone()
         )
 
-        # residual connection (just for the prognostic variables)
+        # residual connection
+        # add time dimension and extend along output steps
+        x_skip = x_skip.unsqueeze(1).expand(-1, self.multi_out, -1, -1, -1)
+        # unly use for the prognostic variables)
```
OpheliaMiralles: Why do we have to do this, couldn't we keep the time dimension in the input (it is already there because we have `multi_step`, then it will be in `x_skip` too, instead of re-creating one for outputs?

training/src/anemoi/training/losses/scalers/output_steps.py
```diff
@@ -0,0 +1,43 @@
...
+
+import torch
+
+from anemoi.training.losses.scalers.base_scaler import BaseScaler
+from anemoi.training.utils.enums import TensorDim
+
+LOGGER = logging.getLogger(__name__)
+
+
+class OutputStepScaler(BaseScaler):
```
OpheliaMiralles: Would suggest TimeStepScaler for extra clarity (also we don't need "Output", since it's a loss scaler it will be applied to outputs anyway), also I can add my custom timeweights (#209) here when it's merged

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -105,29 +108,34 @@ def advance_input(
...
+            # )
+
+            # get new "constants" needed for time-varying fields
+            x[:, -(i + 1), :, :, self.data_indices.model.input.forcing] = batch[
+                :,
+                self.multi_step + (rollout_step + 1) * self.multi_out - (i + 1),
+                :,
+                :,
+                self.data_indices.data.input.forcing,
+            ]
```
OpheliaMiralles: Something similar is used in interpolators (see e.g. #635 https://github.com/ecmwf/anemoi-core/pull/635/files#diff-6a9db7ac519f329605972e899a971ee694ff77b7a3b861ec1374d4b1083c183e) but handling of indices is managed int the `__init__` of the task and somewhat differently, we should see if we need to coordinate.

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -162,17 +170,20 @@ def rollout_step(
             ...,
             self.data_indices.data.input.full,
         ]  # (bs, multi_step, latlon, nvar)
+
+        required_time_steps = rollout * self.multi_out + self.multi_step
```
OpheliaMiralles: Same here in interpolator we use `explicit_times`, maybe for consistency we could adopt the same naming scheme.

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -162,17 +170,20 @@ def rollout_step(
...
         )
-        assert batch.shape[1] >= rollout + self.multi_step, msg
+        assert batch.shape[1] >= required_time_steps, msg
 
         for rollout_step in range(rollout or self.rollout):
             # prediction at rollout step rollout_step, shape = (bs, latlon, nvar)
             y_pred = self(x)
-
-            y = batch[:, self.multi_step + rollout_step, ..., self.data_indices.data.output.full]
+            fc_times = [self.multi_step + rollout_step * self.multi_out + i for i in range(self.multi_out)]
```
OpheliaMiralles: `self.multi_step + rollout_step * self.multi_out` is `required_time_steps`, maybe can be re-used

OpheliaMiralles (commented):
I think this would be great to apply those changes to the interpolator, because there is a direct usecase of an output time dimension here. I would be very glad to have this merged to be able to add the `LeadTimeDecay` weights introduced less properly in #209 (@VeraChristina). Thanks a lot for this @dietervdb-meteo. (tests don't pass right now)

OpheliaMiralles:
> > I think this would be great to apply those changes to the interpolator, because there is a direct usecase of an output time dimension here. I would be very glad to have this merged to be able to add the `LeadTimeDecay` weights introduced less properly in #209 (@VeraChristina). Thanks a lot for this @dietervdb-meteo. (tests don't pass right now)
> 
> I agree that there is a lot of overlap between the two. There are in a way two differences: this is a "forecasting" model and also predicts all targets at the same forward pass. Because it predicts the targets in the same forward pass, I think it simplifies a lot of the dataloading and training code that was there for the interpolator.
> 
> Might make sense sometime in the future to compare an interpolator that predicts states jointly v the current one that predicts the individually.

Don't you think it would be nice to actually add to this PR some changes to the interpolator so that it also returns all predictions in the same forward pass?

models/src/anemoi/models/models/base.py
dietervdb-meteo: I agree, but not sure people want to change names in the config. So maybe we can simply go with `self.multistep_input` and `self.multistep_output`? Or do you think it is worth changing the keys in the config as well? Other thing to do is to converge on how timesteps are defined in the interpolator vs the forecaster. So maybe we can leave this until then as well...

models/src/anemoi/models/models/encoder_processor_decoder.py
dietervdb-meteo: It would only remove the `unsqueeze` from this line and would lead to an additional code change higher up (indexing with a size one list rather than an integer). But maybe that is indeed better long term.

training/src/anemoi/training/losses/scalers/output_steps.py
dietervdb-meteo: Thanks, good suggestion. Implemented

training/src/anemoi/training/train/tasks/forecaster.py
dietervdb-meteo: They are not exactly the same. In `required_time_steps` the variable `rollout` is the **maximum** number of rollout steps. In `fc_times` the variable`rollout_step` is the **current** rollout step (which changes throughout the for loop).

models/src/anemoi/models/models/encoder_processor_decoder.py
OpheliaMiralles: No indeed that's a bigger change, it would mainly mean we never stack time with variables and always keep the time dimension but maybe it's worth asking at the ATS whether it is desirable or not

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -162,17 +167,20 @@ def rollout_step(
...
         ]  # (bs, multi_step, latlon, nvar)
+
+        required_time_steps = rollout * self.multi_out + self.multi_step
         msg = (
             "Batch length not sufficient for requested multi_step length!"
-            f", {batch.shape[1]} !>= {rollout + self.multi_step}"
+            f", {batch.shape[1]} !>= {required_time_steps}"
         )
-        assert batch.shape[1] >= rollout + self.multi_step, msg
+        assert batch.shape[1] >= required_time_steps, msg
```
mc4117: I'm slightly confused as to why you don't need to change the dataloader to make sure that the batches have the required timesteps?

dietervdb-meteo: Indeed I did make such changes, see https://github.com/ecmwf/anemoi-core/pull/636/files#diff-0d21f109bcee370bd2de2200010ea018e90b014947e8cb8e009b457d5ad8e4df

mc4117: oh sorry indeed I missed this, it's such a small change to the code! Very neat PR

training/src/anemoi/training/train/tasks/forecaster.py
dietervdb-meteo: see https://github.com/ecmwf/anemoi-core/issues/648

training/src/anemoi/training/train/tasks/forecaster.py
dietervdb-meteo: see #648

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -105,29 +105,34 @@ def advance_input(
...
-            self.data_indices.model.output.prognostic,
-        ]
-
-        x[:, -1] = self.output_mask.rollout_boundary(
-            x[:, -1],
-            batch[:, self.multi_step + rollout_step],
-            self.data_indices,
-            grid_shard_slice=self.grid_shard_slice,
-        )
+        x = x.roll(-self.multi_out, dims=1)
```
mc4117: Am I correct in thinking this means all the output steps are now an input in to the model? This essentially then makes the model an lstm?

mc4117: It would also maybe make the model different in the first step and subsequent rollout steps?

dietervdb-meteo: It is pretty much identical to the implementation we have without this PR, no? I.e. taking all output as input (see also point 4. in the PR description above).

To make sure we are on the same page:

a) Current default.  2 steps in, 1 step out.  Say -6, 0 in and 6 out. We then advance (`roll`) by 1 to 0, 6 in and 12 out. I.e. all output is used as input, but the input still contains a bit more than purely the output of the previous step.

b) Current option. 1 step in, 1 step out. Say 0 in and 6 out. We then advance (`roll`) by 1 to 6 in and 12 out. I.e. all output is used as input, and this is now the only input. I assume this is what you refer to as "lstm" like. So this is already an option in the current version of the code.

c) New option I explored using this PR. 3 steps in, 2 steps out. Say -6, 3, 0 in and 3, 6 out. We then advance (`roll`) by 2 to 0, 3, 6 in and 9,12 out. I.e. all output is used as input, but the input still contains a bit more than purely the output of the previous step, just like the current default.

d) Additional options with this PR are quite limitless, as long as `multistep_input >= multistep_output` (I should probably insert an assert statement checking this in the code at some point)

But maybe I misunderstand what you are trying to tell me?

dietervdb-meteo: After some offline discussions with @mc4117 I now understand better what she has in mind: come to a set-up where it is possible to say have two input steps, 0, -6, and then two output steps +3, +6 and then use +3 only as a diagnostic (i.e. not feed it back as input under rollout). In the current implementation this is not possible, but it could be interesting to have such flexibility.

training/src/anemoi/training/losses/scalers/time_step.py
```diff
@@ -0,0 +1,114 @@
...
+        super().__init__(norm=norm)
+        del kwargs
+        self.weights = weights
+
+    def get_scaling_values(self) -> torch.Tensor:
+        return torch.tensor(self.weights, dtype=torch.float32)
+
+
+class LeadTimeDecayScaler(TimeStepScaler):
+    """Class to decrease the weight as a function of lead time."""
```
dietervdb-meteo:
```suggestion
    """Class to decrease (default) or increase (optional) the weight as a function of lead time."""
```

training/src/anemoi/training/train/tasks/forecaster.py
OpheliaMiralles: That's what is done in the interpolator, if you have no boundary times but only predict +1, ..., +6 with multi step inputs -6, ..., 0. The interpolator can be also be used as a forecaster, it's just not autoregressive. Once this is merged, I think we need to refactor the interpolator to support any timesteps as input and output too (not necessarily boundary and target times). @Magnus-SI

training/src/anemoi/training/losses/scalers/time_step.py
```diff
@@ -0,0 +1,114 @@
...
+        decay_type: str = "linear",
+        inverse: bool = False,
+        norm: str | None = None,
+        **kwargs,
+    ) -> None:
+        """Initialise Scaler.
+
+        Parameters
+        ----------
+        decay_factor : float
```
dietervdb-meteo, line 94:
```suggestion
        output_lead_times : list
        The lead times for which output is generated by a single model forward.
        decay_factor : float
```

training/src/anemoi/training/schemas/training.py
```diff
@@ -182,6 +182,14 @@ class GraphNodeAttributeScalerSchema(BaseModel):
     "Normalisation method applied to the node attribute."
 
 
+class TimeStepScalerSchema(BaseModel):
```
dietervdb-meteo, line 187: Hi @OpheliaMiralles , can you add a schema for `LeadTimeDecayScaler` that way I can run a test with it.

OpheliaMiralles: Yep will do

dietervdb-meteo:
Just ran a toy training with the `LeadTimeDecayScaler` and at least technically seems to work neatly!

models/src/anemoi/models/models/base.py
```diff
@@ -61,6 +61,7 @@ def __init__(
...
```
OpheliaMiralles: We could set a default of 1 here to not break everyone's code (model_config.training.training.multistep_output if hasattr(...) else 1)

dietervdb-meteo: I thought of that, but typically the consensus is not to make too many 'hidden' defaults even at the cost of having to add a few extra lines to configs every time anemoi updates.

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -111,6 +111,8 @@ def _assemble_output(self, x_out, x_skip, batch_size, ensemble_size, dtype):
         for bounding in self.boundings:
             # bounding performed in the order specified in the config file
             x_out = bounding(x_out)
+            
+        x_out = x_out.contiguous() # necessary after expand()
```
MicheleCattaneo: This addition seemed necessary on the MeteoSwiss HPC system, which uses Nvidia GPUs.
Without it, the following error is thrown:
```
File "<PATH TO VENV>/.venv/lib/python3.12/site-packages/torch/distributed/distributed_c10d.py", line 3341, in all_gather
    work = group.allgather([tensor_list], [tensor])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Tensors must be contiguous
```
Interestingly, this does not seem to happen in other settings, such as on LUMI with AMD GPUs. Nonetheless, I think this behavior is to be expected, since `torch.Tensor.expand` creates a [view](https://docs.pytorch.org/docs/stable/generated/torch.Tensor.expand.html#torch-tensor-expand) of the data. 

Does anybody have an idea why this only happens in some settings, or whether there is a better way to solve it? A possible alternative might be to use `x_skip = x_skip.unsqueeze(1).repeat(1, self.multi_out, 1, 1, 1)` instead of `expand`.

models/src/anemoi/models/models/base.py
OpheliaMiralles: Ah, I just pushed a change with a default... I think it is taking long to merge because it might break most people's configs, but I guess it is fine  to assume that if we don't explicitly set a multiple step then the output is only one time step, no?

dietervdb-meteo: Hi @OpheliaMiralles , the reason I have held off on updating this PR a bit was the decision that it will only be merged after the multi-dataset changes (https://github.com/ecmwf/anemoi-core/pull/594) are merged, and this will require some refactor of the code. I thought it a better use of my (and others?) time to do a big refactor once, rather than trying to incrementally keep up with the changes in core (currently we can also not merge automatically anymore). The plan would be to sync with people involved early in the new year, and draw up a plan something like this: refactor -> compatibility with interpolator -> compatibility with other parts of code (crps, diffusion, plotting callbacks etc) -> merge.

models/src/anemoi/models/models/base.py
```diff
@@ -61,7 +61,9 @@ def __init__(
         self._graph_name_data = model_config.graph.data
         self._graph_name_hidden = model_config.graph.hidden
         self.multi_step = model_config.training.multistep_input
-        self.multi_out = model_config.training.multistep_output
+        self.multi_out = (
+            model_config.training.multistep_output if hasattr(model_config.training, "multistep_output") else 1
```
dietervdb-meteo: If we want to set a default I think it is better to do it via the pydantic schema of the config, which allows for defaults.

models/src/anemoi/models/models/base.py
OpheliaMiralles: OK if you prefer. But the multi-datasets work is not really related to predicting several time steps at once, so we could also have done it in parallel if people had time/were engaged.

dietervdb-meteo: It's not about time/engagement, it was essentially an ATS decision that this should wait till after multi-datasets. If you want we can set up a PR where we merge this branch into the mutli-datasets branch, but not sure that is worth it, as multi-datasets is supposed to be merged 'soon' (it waits on an inference implementation). Maybe @mchantry or @JPXKQX can give us an estimate.

dietervdb-meteo: @OpheliaMiralles, if you want to speed up things we can have a meeting to see how we can plan things (see my message on slack)

OpheliaMiralles: Haven't seen the message on slack! But sure we can discuss there

training/src/anemoi/training/train/tasks/diffusionforecaster.py
ssmmnn11: with multiple output steps, will the tendencies for each step be properly normalised? or are they now all using the same normalisation, e.g. for 12h tendencies?

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -59,18 +165,19 @@ def _rollout_step(
...
             y_pred = self(x)
-
-            y = batch[:, self.multi_step + rollout_step, ..., self.data_indices.data.output.full]
-            LOGGER.debug("SHAPE: y.shape = %s", list(y.shape))
+            fc_times = [self.multi_step + rollout_step * self.multi_out + i for i in range(self.multi_out)]
+            y = batch[:, fc_times, ...]
+            y = y[..., self.data_indices.data.output.full]
             # y includes the auxiliary variables, so we must leave those out when computing the loss
 
             loss, metrics_next, y_pred = checkpoint(
```
ssmmnn11, line 83: will the multiple timestep output work with the tendency scalers (account for relative change of variables)? or do you scale all outputs with the same tendency weight?

training/src/anemoi/training/train/tasks/rollout.py
```diff
@@ -89,10 +89,19 @@ def _advance_input(
...
         rollout_step: int,
     ) -> torch.Tensor:
+        """Default implementation used by simple rollout tasks.
+
+        Supports model outputs shaped like:
+        - (B, E, G, V)
+        - (B, T, E, G, V)  (uses the last time step)
+        """
+        # If the prediction carries a time dimension, use the last step.
+        y_pred_step = y_pred[:, -1] if y_pred.ndim == 5 else y_pred
```
ssmmnn11: why don't we changes it to carry always a time dimension, one output step only would then just be 1? Would that be an option? This is already the case with ensemble size in a similar way.

training/src/anemoi/training/train/tasks/obsinterpolator.py
```diff
@@ -0,0 +1,150 @@
...
+from torch import Tensor
+from torch_geometric.data import HeteroData
+
+from anemoi.models.data_indices.collection import IndexCollection
+from anemoi.training.train.tasks.base import BaseGraphModule
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ObsGraphInterpolator(BaseGraphModule):
```
ssmmnn11: Is this part of the multi output step refactor, or should this go into another pull request?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -926,6 +930,20 @@ def _plot(
                 ...,
                 pl_module.data_indices.data.output.full,
             ]
+
+            # Some tasks/models return predictions/targets without an explicit
+            # time dimension.
+            #
+            # - deterministic: (B, G, V)
```
ssmmnn11: why do we not have the time dimension now for all models?

training/src/anemoi/training/losses/scalers/time_step.py
```diff
@@ -0,0 +1,116 @@
+# (C) Copyright 2024 Anemoi contributors.
```
ssmmnn11, line 1: 2026? :-)

training/src/anemoi/training/losses/scalers/time_step.py
```diff
@@ -0,0 +1,116 @@
...
+
+import torch
+
+from anemoi.training.losses.scalers.base_scaler import BaseScaler
+from anemoi.training.utils.enums import TensorDim
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TimeStepScaler(BaseScaler):
```
ssmmnn11, line 20: Is the plan to always have a time step scaler that defaults to one, and then different ones that can do different scalings?

ssmmnn11 (dismissed):
will be great to have multi output step capabilities :-)

training/src/anemoi/training/diagnostics/callbacks/plot.py
dietervdb-meteo: Actually we do, for all models now `x_out` has dimensions `(batch, time, ensemble, grid, variables)`. Note that for some models this pr will probably only support a time dimension of size 1. Which ones is still to be decided.  The modifications filtering on tensor shape by @OpheliaMiralles are put into place to get a first working (i.e. integration test passing) version of the code. We hope to clean this up further soon.

training/src/anemoi/training/losses/scalers/time_step.py
dietervdb-meteo: One does not need to specify a time step scaler, then loss is simply averaged over the time dimension. So no time step scaler is equivalent to a time step scaler with equal weights. The basic `TimeStepScaler` allows to put an arbitrary weight for each separate time step. There are some derived classes (see lower in the same file) that allow to specify the weights using some functions.

training/src/anemoi/training/train/tasks/rollout.py
OpheliaMiralles: Yes I think @dietervdb-meteo wants to fuse the _advance_input from rollout.py and forecaster.py so it will be the case when it is done

training/src/anemoi/training/train/tasks/obsinterpolator.py
OpheliaMiralles: Interpolator models are much faster to run with multi output steps so both deterministic interpolators are updated to avoid the expensive loop, and it is part of this PR

training/src/anemoi/training/train/tasks/rollout.py
dietervdb-meteo: This has now been done.

training/src/anemoi/training/train/tasks/diffusionforecaster.py
dietervdb-meteo: In this PR we do not plan to support multi-out for diffusion yet. This can hopefully be added as a feature later, so the model output will already carry a time dimension, but there will only be one entry (timestep).

training/src/anemoi/training/train/tasks/forecaster.py
dietervdb-meteo: Hi @ssmmnn11 , thanks for raising this point. In the current implementation the tendency scalers work with more than one output step (at least technically, did not do a detailed check on the result). All output steps are, separatly per variable, scaled by the same per variable tendency statistic, which is based on the frequency of the dataset. So as long as the frequency of the dataset matches the frequency of the output steps this should be correct. At least that is as far as I understand the code. If you or someone with experience with the tendency scaler can have a second look that could be helpful.

training/src/anemoi/training/losses/scalers/time_step.py
dietervdb-meteo: Strictly speaking that piece of code was written in 2025, but I'm not sure if we need to bother. If someone wants to earn an easy commit, go for it.

training/src/anemoi/training/train/tasks/obsinterpolator.py
VeraChristina: Thanks for the context on the interpolator updates. That makes sense. However, it doesn’t really address whether the _new_ interpolator task should be included here. I think adding the new task would be better done in a separate PR to keep the scope focused.

training/tests/integration/conftest.py
```diff
@@ -316,6 +316,36 @@ def ensemble_config(
...
+
+    use_case_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/test_ensemble_crps.yaml")
+    assert isinstance(use_case_modifications, DictConfig)
+
+    tmp_dir, rel_paths, dataset_urls = get_tmp_paths(use_case_modifications, ["dataset"])
+    use_case_modifications.system.input.dataset = str(Path(tmp_dir, rel_paths[0]))
+
+    OmegaConf.set_struct(template.training, False)  # allow adding multistep_output
+    cfg = OmegaConf.merge(template, testing_modifications_callbacks_on_with_temp_dir, use_case_modifications)
+    OmegaConf.resolve(cfg)
```
VeraChristina: would it be possible to re-use the ensemble config fixture and only add the two multistep modifications on top? similar to how it's done in the `mlflow_dry_run_config` fixture

training/tests/integration/conftest.py
```diff
@@ -353,6 +383,27 @@ def gnn_config(testing_modifications_with_temp_dir: DictConfig, get_tmp_paths: G
...
+    get_tmp_paths: GetTmpPaths,
+) -> tuple[DictConfig, str]:
+    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="test_multi_out"):
+        template = compose(config_name="config")
+
+    use_case_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/test_multi_out.yaml")
+    assert isinstance(use_case_modifications, DictConfig)
+
+    tmp_dir, rel_paths, dataset_urls = get_tmp_paths(use_case_modifications, ["dataset"])
+    use_case_modifications.system.input.dataset = str(Path(tmp_dir, rel_paths[0]))
```
VeraChristina, line 384: As above, would it be possible to reuse the `gnn_config` fixture, and add the multi_step modifications on top? Potentially loading them from a file as done for the `imputer_modifications.yaml` to include the scaler

training/tests/integration/conftest.py
```diff
@@ -531,3 +582,37 @@ def mlflow_dry_run_config(gnn_config: tuple[DictConfig, str], mlflow_server: str
...
+) -> tuple[OmegaConf, str]:
+    overrides = request.param
+
+    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="test_diffusion"):
+        template = compose(config_name="diffusion", overrides=overrides)
+
+    use_case_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/test_diffusion.yaml")
+    tmp_dir, rel_paths, dataset_urls = get_tmp_paths(use_case_modifications, ["dataset"])
+    use_case_modifications.system.input.dataset = str(Path(tmp_dir, rel_paths[0]))
+
```
VeraChristina, line 585: as above, would it be possible to reuse the diffusion config fixture?

training/tests/integration/conftest.py
```diff
@@ -531,3 +582,37 @@ def mlflow_dry_run_config(gnn_config: tuple[DictConfig, str], mlflow_server: str
...
+        [
+            "model=graphtransformer_diffusiontend",
+            "training.model_task=anemoi.training.train.tasks.GraphDiffusionTendForecaster",
+        ],
+    ],
+    ids=["diffusion", "diffusiontend"],
+)
+def multi_out_diffusion_config(
+    request: pytest.FixtureRequest,
+    testing_modifications_callbacks_on_with_temp_dir: OmegaConf,
```
VeraChristina: Do we need to include the callbacks here? Or are they covered by the tests for the multistep ensemble?

training/tests/integration/config/test_multi_out.yaml
```diff
@@ -0,0 +1,27 @@
...
+dataloader:
+  training:
+    datasets:
+      data:
+        end: 1979-01-08 12:00:00
+  validation:
+    datasets:
+      data:
+        start: 1979-01-08 18:00:00
+
```
VeraChristina: would it be possible to reuse either the `test_config.yaml` or the `gnn_config` fixture directly, and only add the multistep_modifications on top, as done with the imputer?

If not, can we please use the dataset used in the other tests (except the global, for which this will be changed as well)?

models/src/anemoi/models/models/interpolator.py
```diff
@@ -127,7 +127,9 @@ def _assemble_output(self, x_out, x_skip, batch_size, ensemble_size, dtype, data
         # residual connection (just for the prognostic variables)
         if x_skip is not None:
             # residual connection (just for the prognostic variables)
-            x_out[..., self._internal_output_idx[dataset_name]] += x_skip[..., self._internal_input_idx[dataset_name]]
+            x_out[..., self._internal_output_idx[dataset_name]] += (
+                x_skip[..., self._internal_input_idx[dataset_name]].unsqueeze(1).expand(-1, self.multi_out, -1, -1, -1)
+            )
```
mc4117: Do we not need `x_out = x_out.contiguous()` here as well as we do for the other models?

training/src/anemoi/training/losses/kcrps.py
```diff
@@ -47,14 +47,14 @@ def _kernel_crps(self, preds: torch.Tensor, targets: torch.Tensor) -> torch.Tens
...
+            Predicted ensemble, shape (batch_size, n_out_steps, n_vars, latlon, ens_size)
         targets : torch.Tensor
-            Ground truth, shape (batch_size, n_vars, latlon)
+            Ground truth, shape (batch_size, n_out_steps, n_vars, latlon)
 
         Returns
         -------
         kCRPS : torch.Tensor
-            The point-wise kernel CRPS, shape (batch_size, 1, latlon).
+            The point-wise kernel CRPS, shape (batch_size, n_out_steps, n_vars, latlon).
```
mc4117, line 57: I don't think there is a variable dimension here? I think it's actually one dimensional?

models/src/anemoi/models/models/interpolator.py
MicheleCattaneo: We probably do. I included `x_out = x_out.contiguous()` to make it work on the meteoswiss HPC (nvidia gpus). What's odd is that Dieter had no issues running the code on lumi before this change (amd gpus). However, `expand` is for sure creating a view and not a copy of the tensor, which is not contiguous:
```
>>> x = torch.randn(16,1,16,16)
>>> x.is_contiguous()
True
>>> x_view = x.expand(-1,6,-1,-1)
>>> x_view.is_contiguous()
False
```
For later `gatherall` operations, all tensors are required to be contiguous (at least it seems for nvidia hardware?)
Alternatively, we can think of using something else rather and `expand`

MicheleCattaneo: Actually, since we already unsqueeze the second dimension, I guess `expand` is redundant, since we can rely on broadcasting. And the result should be contiguous

OpheliaMiralles: @MicheleCattaneo I'm not actually sure why we needed it in the first place

training/src/anemoi/training/losses/kcrps.py
OpheliaMiralles, line 57:
```suggestion
            The point-wise kernel CRPS.
```

OpheliaMiralles: Indeed

models/src/anemoi/models/models/interpolator.py
MicheleCattaneo: Do you mean the unsqueeze? The skip connection for all multi-outs is done w.r.t time 0, which misses the time dimension. So far you add a time dimension with unsqueeze and then use expand to "repeat" the values as many times as output steps there are, so that you can add the two tensors.
But with a singleton dimension for time it should be already enough thanks to broadcasting

MicheleCattaneo: Actually, I stepped in with the debugger again, and the tensor becomes non-contiguous before the `expand`. What makes it non contiguous is actually `einops.rearrange` a few lines above. 

From the [docs](https://einops.rocks/api/rearrange/):

"Output: tensor of the same type as input. If possible, a view to the original tensor is returned. "

As far as I know, when you have a view of a tensor, you can't be sure that it will be contiguous. What do you think? isn't it odd that this issue never popped up before? rearrange was there before, just used differently.

training/src/anemoi/training/train/tasks/interpolator.py
```diff
@@ -107,18 +125,18 @@ def get_target_forcing(self, batch: dict[str, torch.Tensor], interp_step: int) -
...
+                    itemgetter(*self.interp_times)(self.imap),
+                    ...,
+                ][
+                    ...,
                     self.target_forcing_indices[dataset_name],
                 ]
 
             if self.use_time_fraction[dataset_name]:
-                target_forcing[dataset_name][..., -1] = (interp_step - self.boundary_times[-2]) / (
+                target_forcing[dataset_name][..., -1] = (self.interp_times - self.boundary_times[-2]) / (
```
radiradev: Surely this is not needed if the model is always predicting all the steps

training/src/anemoi/training/train/tasks/interpolator.py
```diff
@@ -82,15 +94,21 @@ def __init__(
             self.use_time_fraction[dataset_name] = target_forcing_config[dataset_name].time_fraction
 
         self.num_tfi = {name: len(idxs) for name, idxs in self.target_forcing_indices.items()}
-
         self.boundary_times = config.training.explicit_times.input
         self.interp_times = config.training.explicit_times.target
```
radiradev, line 101: Wouldn't this always be the steps in between? Shouldn't this change try to simplify the interpolator

training/src/anemoi/training/train/tasks/interpolator.py
```diff
@@ -129,46 +147,30 @@ def _step(
...
-            metrics.update(metrics_next)
-            y_preds.append(y_pred)
-
-        loss *= 1.0 / len(self.interp_times)
-        return loss, metrics, y_preds
+        target_forcing = self.get_target_forcing(batch)
+        y_pred = self(x_bound, target_forcing)
+        y = {}
+        for dataset_name, dataset_batch in batch.items():
+            y[dataset_name] = dataset_batch[:, itemgetter(*self.interp_times)(self.imap), ...][
```
radiradev: Again, should't we just pick all the steps in between say (0,6) rather than explicitly mapping indices

OpheliaMiralles: How do you know what all the steps mean? I had a config with boundary times 0 36 (freq 10min) and the steps I wanted to predict were e.g. 1,2,3,6,9,12,18. I don't see how I can do that without mapping indices...

training/src/anemoi/training/train/tasks/interpolator.py
OpheliaMiralles: It gives info about where in the interval the data is, I don't see why it would be useless? It is an additional information

training/src/anemoi/training/train/tasks/interpolator.py
OpheliaMiralles: It's hard to see what you mean because you commented a deleted line

training/src/anemoi/training/train/tasks/interpolator.py
radiradev: the model always predicts all the target indices, so I don't see why you need to add this extra conditioning? Before it was needed as the model did a separate forward for each step

training/src/anemoi/training/train/tasks/interpolator.py
radiradev: self.interp_times = config.training.explicit_times.target I assumed this would always be contiguous but fair enough that there could be another use case

training/src/anemoi/training/train/tasks/interpolator.py
radiradev: I assumed contiguous indices. But even in your case (e.g. 1,2,3,6,9,12,18) - shouldn't it be the job of the of the dataloader? Wouldn't the batch contain data that doesn't get used?

radiradev: Maybe it's something for a later refactor rather than now doing this as part of this PR

models/src/anemoi/models/models/interpolator.py
MicheleCattaneo: @OpheliaMiralles why did you set this as resolved? Did somebody find the reason?

graphs/src/anemoi/graphs/nodes/builders/from_file.py
```diff
@@ -173,10 +173,10 @@ def __init__(
         mask_attr_name: str | None = None,
         margin_radius_km: float = 100.0,
     ) -> None:
-        super().__init__(npz_file, name, lat_key, lon_key)
```
JPXKQX, line 176: We need to drop this change, Issue #819

models/src/anemoi/models/models/interpolator.py
mc4117: I have added back in x.contiguous

OpheliaMiralles: Yes I set resolved because contiguous was added back but actually if einops rearrange is the source we should seriously wonder why the problem was not there before

models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
```diff
@@ -745,7 +752,12 @@ def _assemble_input(
         node_attributes_data = self.node_attributes[dataset_name](self._graph_name_data, batch_size=bse)
         grid_shard_shapes = grid_shard_shapes[dataset_name] if grid_shard_shapes is not None else None
 
-        x_skip = self.residual[dataset_name](x, grid_shard_shapes, model_comm_group)
+        x_skip = self.residual[dataset_name](x, grid_shard_shapes, model_comm_group)[
+            ..., self._internal_input_idx[dataset_name]
+        ]
+        assert x_skip.ndim == 4, "Residual must be (batch, ensemble, grid, vars)."
+        x_skip = x_skip.unsqueeze(1).expand(-1, self.multi_out, -1, -1, -1)
```
JPXKQX: I would move this inside the residual so that it always returns a time dimension.

models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
```diff
@@ -745,7 +752,12 @@ def _assemble_input(
...
+        x_skip = einops.rearrange(x_skip, "batch time ensemble grid vars -> (batch ensemble) grid (time vars)")
```
JPXKQX, line 839: I understand that this change is due to the fact that we now have an extra dimension. However, I believe that we could remove this `einops.rearrange` by editing the next `einops.rearrange` used in the `if self.condition_in_residual` (in this same method)

ssmmnn11: this should be one oper and inside of the residual so we can checkpoint if needed later.

graphs/src/anemoi/graphs/nodes/builders/from_file.py
dietervdb-meteo: Done in [c6fbf73](https://github.com/ecmwf/anemoi-core/pull/636/commits/c6fbf73ddd4675c72a2334b6584e533d91d74ccc)

OpheliaMiralles:
I would have appreciated a discussion before reverting previous commits I made.

First, regarding the multi-dataset merge: only a limited number of integration tests were failing, and it would have been useful to understand and address those issues together rather than reverting the whole change, which results in duplicated work.

Second, concerning the interpolator: I think it would be helpful to have a broader discussion rather than reverting changes and proceeding with a plan that currently seems to be driven by a single perspective. This is a complex topic, and I believe we would benefit from sharing views and agreeing on a common approach.

More generally, I would welcome more collective discussion before major reversions, especially on topics where multiple reasonable approaches exist.

mc4117:
This PR has the single output interpolator that exists in main because my understanding is that various people are still using this implementation in their work. I implemented the multiple output interpolator as is done on this branch that @Magnus-SI  and I have been training from on for a while https://github.com/ecmwf/anemoi-core/tree/feature/ens_interp. We certainly have to support this version but if there is a different use case, this is an open PR so we can make changes. Happy to take this discussion offline or discuss in the meeting tomorrow

models/src/anemoi/models/layers/residual.py
```diff
@@ -31,13 +31,25 @@ def __init__(self, graph: HeteroData | None = None) -> None:
...
 
         Should be overridden by subclasses.
         """
         pass
 
+    @staticmethod
+    def _expand_time(x: torch.Tensor, multi_out: int | None) -> torch.Tensor:
+        if multi_out is None:
+            return x
+        return x.unsqueeze(1).expand(-1, multi_out, -1, -1, -1).contiguous()
```
MicheleCattaneo: I would think that `expand` + `contiguous` is the same as just using `repeat`. Is there a difference?

training/docs/user-guide/configuring.rst
```diff
@@ -78,6 +78,42 @@ You can also change the GPU count to whatever you have available:
 This matches the interface of the underlying defaults in Anemoi
 training.
 
+*********************************
+ Multistep Input and Output
+*********************************
+
+Anemoi uses ``multistep_input`` and ``multistep_output`` to control how many time
+steps the model consumes and predicts in a single forward pass.
```
mc4117:
```suggestion
steps the model injests as input and predicts in a single forward pass.
```

training/docs/user-guide/configuring.rst
```diff
@@ -78,6 +78,42 @@ You can also change the GPU count to whatever you have available:
...
+
+-  ``multistep_input``: number of past timesteps provided as model input.
```
mc4117:
```suggestion
-  ``multistep_input``: number of past timesteps provided as model input. When set to 1, only `t_{0}` is used.
```

models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
dietervdb-meteo: Done

models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
dietervdb-meteo: done

models/src/anemoi/models/layers/residual.py
dietervdb-meteo: done (in alternative fashion)

training/src/anemoi/training/diagnostics/callbacks/evaluation.py
```diff
@@ -54,12 +54,15 @@ def __init__(self, config: OmegaConf, rollout: int, every_n_batches: int) -> Non
     def _eval(
         self,
         pl_module: pl.LightningModule,
-        batch: torch.Tensor,
+        batch: torch.Tensor | dict[str, torch.Tensor],
```
JPXKQX: I think this will always be a dictionary

training/src/anemoi/training/data/dataset.py
```diff
@@ -49,6 +49,15 @@ def statistics(self) -> dict:
...
     def statistics_tendencies(self, timestep: datetime.timedelta | None = None) -> dict | None:
         """Return dataset tendency statistics."""
+        if timestep is None:
+            timestep = getattr(self, "timestep", None)
+        if timestep is None:
+            msg = "timestep must be provided to compute tendency statistics."
+            raise ValueError(msg)
+        return self.statistics_tendencies_for_timestep(timestep)
+
+    def statistics_tendencies_for_timestep(self, timestep: str) -> dict | None:
```
JPXKQX: Do we need two functions here?

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -291,6 +291,8 @@ class PlotSchema(PydanticBaseModel):
     "Handle plotting tasks without blocking the model training."
     datashader: bool
     "Use Datashader to plot."
+    output_steps: PositiveInt | None = Field(default=None, example=1)
+    "Max number of output steps to plot per rollout for multi-step outputs (forecast mode)."
```
VeraChristina: I'd rather make these a required key of each plotting callback configuration, in line with how we handle plotting configs more generally

graphs/src/anemoi/graphs/nodes/builders/from_file.py
```diff
@@ -174,7 +174,6 @@ def __init__(
         margin_radius_km: float = 100.0,
     ) -> None:
         super().__init__(npz_file, name, lat_key, lon_key)
-
```
JPXKQX, line 177: I think this line change would trigger release-please to create a new version of anemoi-graphs, which may not be needed

training/src/anemoi/training/diagnostics/callbacks/evaluation.py
dietervdb-meteo: Ok, cleaned up.

graphs/src/anemoi/graphs/nodes/builders/from_file.py
dietervdb-meteo: Put the line back and removed the graphs label.

PR #645, opened 2025-11-05
feat(models,training): Ensemble Interpolator

Magnus-SI:
## Description
Allows for training a probabilistic interpolator/temporal downscaler by using the same method as the ensemble forecaster.

##  Additional notes ##
This branch and draft PR is for testing purposes. Cleaner code and proper tests will be added once we can confirm that this method works well. But feel free to comment on the method itself.

## TODO ##
- [ ] Implement target parallelism for faster training.
- [x] Complete a full training run and verify.
- [x] Consider adding the option to output all targets simultaneously.

### DIFF
--- a/models/src/anemoi/models/models/ensinterp.py
+++ b/models/src/anemoi/models/models/ensinterp.py
@@ -1,0 +1,195 @@
+# (C) Copyright 2024 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+import logging
+from typing import Optional
+
+import einops
+import torch
+from hydra.utils import instantiate
+from torch.distributed.distributed_c10d import ProcessGroup
+from torch_geometric.data import HeteroData
+
+from anemoi.models.distributed.graph import shard_tensor
+from anemoi.models.distributed.shapes import get_shard_shapes
+from anemoi.models.models import AnemoiModelEncProcDec
+from anemoi.utils.config import DotDict
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AnemoiModelEncProcDecEnsInterp(AnemoiModelEncProcDec):
+    """Message passing graph neural network with ensemble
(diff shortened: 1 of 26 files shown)


models/src/anemoi/models/models/ensinterp.py
```diff
@@ -0,0 +1,196 @@
...
+
+from anemoi.models.distributed.graph import shard_tensor
+from anemoi.models.distributed.shapes import get_shard_shapes
+from anemoi.models.models import AnemoiModelEncProcDec
+from anemoi.utils.config import DotDict
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AnemoiModelEncProcDecEnsInterp(AnemoiModelEncProcDec):
```
mc4117, line 28: would it not make more sense to subclass from the interpolator directly?

training/src/anemoi/training/train/tasks/ensinterp.py
```diff
@@ -0,0 +1,395 @@
...
+if TYPE_CHECKING:
+
+    from omegaconf import DictConfig
+    from torch.distributed.distributed_c10d import ProcessGroup
+    from torch_geometric.data import HeteroData
+
+LOGGER = logging.getLogger(__name__)
+
+
+class GraphEnsInterp(BaseGraphModule):
```
mc4117, line 33: same here? It would remove a lot of need for duplicate code and make it easier to see what are the real changes needed for EnsInterpolators?

models/src/anemoi/models/models/ensinterp.py
Magnus-SI: The problem is that we need to inherit functionality both from the ensemble and interpolator classes, and this applies to almost all of the subfunctions. As far as I can see, _calculate_input_dim is the only function that is fully inheritable without making ensemble specific changes. But I agree that it's unfortunate to duplicate this much code.

training/src/anemoi/training/train/tasks/ensinterp.py
Magnus-SI: Agreed, though I was planning to inherit from the GraphEnsForecaster class here. The GraphInterpolator class only modifies __init__ _step and forward, and we need to change __init__ and _step anyway. forward would be the only function inheritable as is. While from the GraphEnsForecaster, everything related to ensemble parallel is fully inheritable.

mc4117:
This has been implemented instead by PR #896 and PR #1069

PR #663, opened 2025-11-12
fix(training): provide more informative error when user specifies inexistent node attribute 

frazane:
Small fix to make the error message more informative. Instead of just telling a user that the specified attribute does not exist in the graph data, it also provides a list of the ones that are available. So for instance, if one mistakenly specified "cutout_mask" instead of "cutout", they will know right away what the problem is.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,6 +1,6 @@
 ## How to Contribute
 
-Please see the [read the docs](https://anemoi.readthedocs.io/en/latest/dev/contributing.html).
+Please see the [read the docs](https://anemoi.readthedocs.io/en/latest/contributing/contributing.html).
 
 
 ## Contributors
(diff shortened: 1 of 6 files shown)


training/src/anemoi/training/losses/scalers/node_attributes.py
```diff
@@ -95,7 +95,9 @@ def __init__(
             **kwargs,
         )
         if self.scaling_mask_attribute_name not in self.nodes:
```
JPXKQX:
```suggestion
        if self.scaling_mask_attribute_name not in self.nodes.node_attrs():
```

JPXKQX (changes requested):
Hi Francesco, thank you for raising this issue. I added one suggestion that is not related to your proposed changes, but which I think would also improve the clarity of the code. Another minor suggestion would be to include the name of the class in the error message to make it easier to track errors. What do you think?

PR #669, opened 2025-11-13
feat: activate minmium plotting for integration tests

anaprietonem:
## Description
Testing plots it's not part our current testing suite. Since most of them are implemented via callbacks a simple way would be to activate a minimal plotting set up for the integration test to ensure we don't break them in the future. 

This PR proposes those changes:
- Spectrum plot it's quite slow so for now it's deactivated for the SG model
- Since the test dataset for LAM and SG is a small one we just test plotting tp
- The plots for the interpolator are deactivated - while the callbacks run the output produced is not scientifically correct. More work will be done in the future to address this (@mc4117 - notified about this in slack channel)
- In the hierarchical model - current plotting of edge features is not support - this plot is for now deactivated. (Discussed with @icedoom888 )

cc: @VeraChristina  - having those tests would also be useful in the context of multiple datasets to ensure there is no feature regression for plotting. 

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
+++ b/training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
@@ -286,7 +286,7 @@ class PlotLoss(_PlotLoss):
             trainer,
             pl_module,
             outputs,
-            batch[0][:, :, 0, :, :],
+            batch[:, :, 0, :, :],
             batch_idx,
         )
 


training/tests/integration/config/testing_modifications.yaml
```diff
@@ -12,7 +12,7 @@ dataloader:
 
 diagnostics:
   plot:
-    callbacks: []
```
VeraChristina, line 15: I am wondering if we want to activate the plotting callbacks in all tests (including restart training tests, starting from checkpoints, training with graph for lam, etc.) or just in the basic training cycle tests for the different use cases -- just thinking of test time, otherwise I think it'd be fine

mc4117: That makes sense to me. I agree it's probably not necessary to run the callbacks for all the tests, as long as it's tested once for each model type (time-interpolator, ensemble etc)

mc4117 (approved):
LGTM! Thanks for fixing the issue with the old eda callback!

PR #670, opened 2025-11-13
feat(models): add configurable residual connections in enc-proc-dec

JPXKQX:
## Description

This is a follow-up to PR #451 . The branch has been moved to anemoi-core (so we can now run integration tests), and rebased with main.

- It abstracts the residual connection around a new class, `BaseResidualConnection`. It currently supports 3 options: `SkipConnection`, `TruncatedConnection`, and `NoConnection`.
- It also creates the `SparseProjector` class which handles the projection logic. This class could be used in the future for the multi-scale loss or other use cases.

Encapsulation of residual connection functionality.

----
📚 Documentation preview 📚: https://anemoi-training--670.org.readthedocs.build/en/670/

----
📚 Documentation preview 📚: https://anemoi-graphs--670.org.readthedocs.build/en/670/

----
📚 Documentation preview 📚: https://anemoi-models--670.org.readthedocs.build/en/670/

### DIFF
--- a/graphs/docs/usage/create_sparse_matrices.rst
+++ b/graphs/docs/usage/create_sparse_matrices.rst
@@ -74,7 +74,7 @@ For example, to export only the ``gauss_weight`` attribute for the
 
    % anemoi-graphs export_to_sparse graph_recipe.yaml output_dir/ \
        --edges-attributes-name gauss_weight \
-       --edges-name data->down
+       --edges-name data down
 
 You can specify multiple attributes or subgraphs by repeating the
 arguments:
@@ -84,8 +84,8 @@ arguments:
    % anemoi-graphs export_to_sparse graph_recipe.yaml output_dir/ \
        --edges-attributes-name gauss_weight \
        --edges-attributes-name another_weight \
-       --edges-name data->down \
-       --edges-name down->data
+       --edges-name data down \
+       --edges-name down data
 
 This flexibility allows you to generate only the sparse matrices you
 need for your application, reducing storage and processing time.
--- a/graphs/src/anemoi/graphs/commands/export_to_sparse.py
+++ b/graphs/src/anemoi/graphs/commands/export_to_sparse.py
@@ -18,7 +18,7 @@ class ExportToSparse(Command):
 
     Example usage specifying an edge attribute:
     ```
-    anemoi-graphs export-to-sparse graph.pt output_path --edge_attribute_name edge_attr
+    anemoi-graphs export-to-sparse graph.pt output_path --edge-attribute-name edge_attr
     ```
 
     Example usage specifying a subset of edges:
@@ -33,7 +33,7 @@ class ExportToSparse(Command):
     def add_arguments(self, command_parser):
         command_parser.add_argument("graph", help="Path to the graph (a .PT file) or a config file defining the graph.")
         command_parser.add_argument("output_path", help="Path to store the inspection results.")
-        command_parser.add_argument("--edge_attribute_name", default=None, help="Name of the edge attribute to export.")
+        command_parser.add_argument("--edge-attribute-name", default=None, help="Name of the edge attribute to export.")
         command_parser.add_argument(
             "--edges-name",
             nargs=2,
@@ -46,8 +46,7 @@ class ExportToSparse(Command):
         kwargs = vars(args)
         edges_name = kwargs.get("edges_name", None)
         if edges_name is not None:
-            # Convert list of lists to list of tuples
-            kwargs["edges_name"] = [tuple(pair) for pair in edges_name]
+            kwargs["edges_name"] = [(pair[0], "to", pair[1]) for pair in edges_name]
 
         GraphExporter(
             graph=kwargs["graph"],
--- a/graphs/src/anemoi/graphs/export.py
+++ b/graphs/src/anemoi/graphs/export.py
@@ -63,9 +63,35 @@ class GraphExporter:
 
     @staticmethod
     def get_sparse_matrix(edge_index, edge_attribute, num_source_nodes, num_target_nodes):
-        # Create sparse matrix
+        """Create sparse matrix for y = A x.
+
+        x is defined on source nodes, and output is on target nodes.
+
+        Arguments
+        ---------
+        edge_index : torch.Tensor
+            (2, E) tensor with rows: target nodes, cols: source nodes
+        edge_attribute : torch.Tensor
+            Edge weights/attributes
+        num_source_nodes : int
+            Number of source nodes (n_in)
+        num_target_nodes : int
+            Number of target nodes (n_out)
+
+        Returns
+        -------
+        torch.sparse_coo_tensor
+            Sparse COO tensor on target nodes
+        """
+        rows = edge_index[1]
+        cols = edge_index[0]
+        indices = torch.stack([rows, cols])
+
         A = torch.sparse_coo_tensor(
-            edge_index, edge_attribute, (num_source_nodes, num_target_nodes), device=edge_index.device
+            indices,
+            edge_attribute,
+            (num_target_nodes, num_source_nodes),
+            device=edge_index.device,
         )
         return A.coalesce()
 
@@ -73,9 +99,10 @@ class GraphExporter:
     def convert_to_scipy_sparse(A):
         """Convert PyTorch sparse tensor to SciPy sparse matrix and save.
 
-        Args:
-            A: PyTorch sparse COO tensor
-            filename: Output filename (.npz extension)
+        Arguments
+        ---------
+        A : torch.sparse_coo_tensor
+            Sparse tensor
         """
         # Get indices and values from PyTorch sparse tensor
         indices = A.indices().cpu().numpy()


models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,206 @@
...
+from torch import nn
+from torch_geometric.data import HeteroData
+
+from anemoi.models.distributed.graph import gather_channels
+from anemoi.models.distributed.graph import shard_channels
+from anemoi.models.distributed.shapes import apply_shard_shapes
+from anemoi.models.layers.sparse_projector import build_sparse_projector
+
+
+class BaseResidualConnection(nn.Module):
```
HCookie: This should subclass or metaclass from ABCMeta

models/src/anemoi/models/layers/residual.py
HCookie, line 1: License?

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,206 @@
...
+    """Base class for residual connection modules."""
+
+    def __init__(self, *args, **kwargs):
```
HCookie: What's the reason for these wild card ignores?

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,206 @@
...
+
+class SkipConnection(BaseResidualConnection):
+    """Skip connection module
+
+    It selects the most recent timestep from the input sequence.
+
+    This module is used to bypass processing layers and directly pass the latest input forward.
+    """
+
+    def __init__(self, step: int = -1, *args, **kwargs):
```
HCookie: Again, wild card ignores.
This can lead to hard to debug states.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        """Return a zero tensor with the same shape as the last timestep."""
+        return torch.zeros_like(x[:, 0, ...], device=x.device, dtype=x.dtype)
+
+
+class TruncatedConnection(nn.Module):
+    """Truncated skip connection
+
+    It applies a coarse-graining and reconstruction of input features using sparse projections to
```
HCookie: Is 2nd person best here?
I would expect more like the form of `This applies'

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,206 @@
...
+
+    This module returns a zero tensor with the same shape as the last timestep.
+    """
+
+    def forward(self, x: torch.Tensor, *args, **kwargs) -> torch.Tensor:
+        """Return a zero tensor with the same shape as the last timestep."""
+        return torch.zeros_like(x[:, 0, ...], device=x.device, dtype=x.dtype)
+
+
+class TruncatedConnection(nn.Module):
```
HCookie: If this is a residual, why does it not subclass from the base?

models/src/anemoi/models/layers/residual.py
HCookie, line 1: Overall the interface feels a bit clunky with massive wild card ignores, if this is inline with other parts of models I can overlook it, but I'd like to see some more thought on the abstraction to prevent these wildcards

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -0,0 +1,155 @@
...
+    norm = total.scatter_add_(0, edge_index[1].long(), weights)
+    norm = norm[edge_index[1]]
+    return weights / (norm + 1e-8)
+
+
+class SparseProjector(torch.nn.Module):
+    """Constructs and applies a sparse projection matrix for mapping features between grids.
+
+    The projection matrix is constructed from edge indices and edge attributes (e.g., distances),
+    with optional row normalization.
```
HCookie:
```suggestion
    with optional row normalisation.
```
British english correction

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -0,0 +1,155 @@
...
+        edges_name: str,
+        edge_weight_attribute: Optional[str] = None,
+        src_node_weight_attribute: Optional[str] = None,
+        **kwargs,
+    ) -> "SparseProjector":
+        """Build a SparseProjection from a graph.
+
+        Parameters
+        ----------
+        graph : networkx.Graph
```
HCookie: Is this the correct type hint?

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -0,0 +1,155 @@
...
+            **kwargs,
+        )
+
+    def forward(self, x, *args, **kwargs):
+        # This has to be called in the forward because sparse tensors cannot be registered as buffers,
+        # as they can't be broadcast correctly when using DDP.
+        self.projection_matrix = self.projection_matrix.to(x.device)
+
+        out = []
+        device_type = "cuda" if torch.cuda.is_available() else "cpu"
```
HCookie: This should use the type of the `x.device`?

models/src/anemoi/models/layers/residual.py
JPXKQX: No reason

models/src/anemoi/models/layers/sparse_projector.py
JPXKQX: No

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,215 @@
...
+    autocast : bool, default False
+        Whether to use automatic mixed precision for the projections.
+    truncation_up_file_path : str, optional
+        File path (.npz) to load the up-projection matrix from.
+    truncation_down_file_path : str, optional
+        File path (.npz) to load the down-projection matrix from.
+
+    Example
+    -------
+    >>> from torch_geometric.data import HeteroData
```
anaprietonem, line 88: Very nice docstring!

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,215 @@
...
+
+class NoConnection(BaseResidualConnection):
+    """No-op connection
+
+    This module returns a zero tensor with the same shape as the last timestep.
+    """
+
+    def forward(self, x: torch.Tensor, grid_shard_shapes=None, model_comm_group=None) -> torch.Tensor:
+        """Return a zero tensor with the same shape as the last timestep."""
+        return torch.zeros_like(x[:, 0, ...], device=x.device, dtype=x.dtype)
```
anaprietonem: Not too critical but in the future it could be nice to think about how to define the NoConnection more efficiently and then avoid instantiating a big array in the GPU that's not used

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,215 @@
...
+        return up_edges, down_edges
+
+    def forward(self, x: torch.Tensor, grid_shard_shapes=None, model_comm_group=None) -> torch.Tensor:
+        """Apply truncated skip connection."""
+        batch_size = x.shape[0]
+        x = x[:, -1, ...]  # pick latest step
+        shard_shapes = apply_shard_shapes(x, 0, grid_shard_shapes) if grid_shard_shapes is not None else None
+
+        x = einops.rearrange(x, "batch ensemble grid features -> (batch ensemble) grid features")
+        x = self._to_channel_shards(x, shard_shapes, model_comm_group)
```
anaprietonem, line 189: what's the difference between `_to_grid_shards`  and `_to_channel_shards`  - as in I understand once would sharding across the grid dimension and the other across the num channels - but if the shard_shapes is the same between the two calls - then ? maybe @japols can chime in

models/src/anemoi/models/layers/residual.py
anaprietonem: or just adding 0 maybe?
````
import torch
a=torch.ones(3,3)
a=a.to('cuda')
assert a==a+0
```
seems fine

JPXKQX: The current code needs the residual connection output to have multiple dimensions. For example,

```
        x_out[..., self._internal_output_idx] += x_skip[..., self._internal_input_idx]
```

japols: could we just return None and only apply skip connection if not None?

models/src/anemoi/models/layers/residual.py
ssmmnn11: we go from full channels but sharded grid points to full grid points but sharded channels. shard_shapes only carry information about the geometry.

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,215 @@
+# (C) Copyright 2025 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+from abc import ABC
```
HCookie, line 11:
```suggestion

from abc import ABC
```

models/src/anemoi/models/layers/residual.py
```diff
@@ -0,0 +1,215 @@
...
+    grid to a truncated (coarse) grid, and another to project back to the original resolution.
+
+    Parameters
+    ----------
+    graph : HeteroData
+        The graph containing the subgraphs for down and up projections.
+    data_nodes : str
+        Name of the nodes representing the data nodes.
+    truncation_nodes : str
+        Name of the nodes representing the truncated (coarse) nodes.
```
HCookie: From the code, these are also optional

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -0,0 +1,153 @@
...
+    edges_name: Optional[tuple[str, str, str]] = None,
+    edge_weight_attribute: Optional[str] = None,
+    **kwargs,
+) -> SparseProjector:
+    """Factory method to build a SparseProjector."""
+    assert (file_path is not None) ^ (
+        graph is not None and edges_name is not None
+    ), "Either file_path or graph and edges_name must be provided."
+
+    if file_path is not None:
```
HCookie, line 145: Arguably this should also check for the setting of the graph in addition to / instead of the callers of this function

models/src/anemoi/models/layers/residual.py
JPXKQX: ?

HCookie: ?

HCookie (dismissed):

PR #673, opened 2025-11-13
pass edges to mapper and processor directly

ssmmnn11:
pass edges and edge attributes to mapper and processor


models/src/anemoi/models/layers/graph_providers.py
```diff
@@ -272,7 +292,13 @@ def get_edges(
 
         Raises
         ------
+        ValueError
+            If coordinates are not provided
         NotImplementedError
-            This functionality is not yet implemented
+            If build_graph() is not yet implemented
```
HCookie, line 298: Technically this function does not raise a  `NotImplementedError`, but still worth mentioning in the docs

PR #677, opened 2025-11-14
feat: time_interpolator_callbacks

mc4117:
## Description
Introduce deterministic callback plots for the time interpolator. These were previously broken because they only showed the first step of the time interpolator

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
   - id: python-check-blanket-noqa # Check for # noqa: all
   - id: python-no-log-warn # Check for log.warn
 - repo: https://github.com/psf/black-pre-commit-mirror
-  rev: 25.9.0
+  rev: 25.11.0
   hooks:
   - id: black
     args: [--line-length=120]
@@ -41,7 +41,7 @@ repos:
     - --profile black
     - --project anemoi
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.3
+  rev: v0.14.7
   hooks:
   - id: ruff
     args:
@@ -50,7 +50,7 @@ repos:
     - --exit-non-zero-on-fix
     - --exclude=docs/**/*_.py
 - repo: https://github.com/sphinx-contrib/sphinx-lint
-  rev: v1.0.1
+  rev: v1.0.2
   hooks:
   - id: sphinx-lint
 # For now, we use it. But it does not support a lot of sphinx features
@@ -65,7 +65,7 @@ repos:
   - id: docconvert
     args: ["numpy"]
 - repo: https://github.com/tox-dev/pyproject-fmt
-  rev: "v2.11.0"
+  rev: "v2.11.1"
   hooks:
   - id: pyproject-fmt
     args: ["--max-supported-python", "3.12"]
--- a/models/src/anemoi/models/utils/compile.py
+++ b/models/src/anemoi/models/utils/compile.py
@@ -8,6 +8,7 @@
 # nor does it submit to any jurisdiction.
 
 import logging
+from functools import reduce
 from importlib.util import find_spec
 
 import torch
@@ -72,7 +73,6 @@ def mark_for_compilation(model: Module, compile_config: DictConfig | None) -> Mo
 
     default_compile_options = {}
     compiled_modules = []
-
     # Loop through all modules
     for name, module in model.named_modules():
         entry = _get_compile_entry(module, compile_config)
@@ -81,9 +81,16 @@ def mark_for_compilation(model: Module, compile_config: DictConfig | None) -> Mo
             options = entry.get("options", default_compile_options)
 
             LOGGER.debug("%s will be compiled with the following options: %s", str(module), str(options))
-            module.forward = torch.compile(module.forward, **options)  # Note: the function is not compiled yet
             # It is just marked for JIT-compilation later
             # It will be compiled before its first forward pass
+
+            module.compile(**options)
+
+            parts = name.split(".")
+            parent = reduce(getattr, parts[:-1], model)
+            LOGGER.debug("Replacing %s with a compiled version", str(parts[-1]))
+            setattr(parent, parts[-1], module)
+
             compiled_modules.append(entry.module)
 
     LOGGER.info("The following modules will be compiled: %s", str(unique(compiled_modules)))
--- a/models/tests/utils/test_compile.py
+++ b/models/tests/utils/test_compile.py
@@ -7,7 +7,9 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+import io
 import logging
+import os
 
 import torch
 from omegaconf import DictConfig
@@ -108,9 +110,6 @@ def test_compile() -> None:
 
     result_compiled = ln_compiled.forward(x_in, cond)
 
-    # check the function was compiled
-    assert hasattr(ln_compiled.forward, "_torchdynamo_orig_callable")
-
     # check the result of the compiled function matches the uncompiled result
     assert torch.allclose(result, result_compiled)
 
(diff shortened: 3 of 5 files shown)


mc4117:
This is an example of the plots in mlflow https://mlflow.ecmwf.int/#/experiments/54/runs/c203922b9ffc41b49671eac5b13b81d7/artifacts

Rilwan-Adewoyin:
Just to clarify on the plots such as "pred_val_sample_istep04_batch0000_rank0_epoch000" in https://mlflow.ecmwf.int/#/experiments/54/runs/c203922b9ffc41b49671eac5b13b81d7/artifacts are the images showing [pred-input] referring to that time step t - the input at step=0 or at the last interpolated step= t-1 ?

mc4117:
> Just to clarify on the plots such as "pred_val_sample_istep04_batch0000_rank0_epoch000" in https://mlflow.ecmwf.int/#/experiments/54/runs/c203922b9ffc41b49671eac5b13b81d7/artifacts are the images showing [pred-input] referring to that time step t - the input at step=0 or at the last interpolated step= t-1 ?

They're referring to that time step t and the last interpolated time step t-1. Note that this model is not properly trained - I just ran for a couple of steps to get the plot

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -30,6 +30,7 @@ model:
   latent_skip: False # True/False for skip connection on latent mesh
 
 training:
+  multistep_input: 1
```
VeraChristina: Do we need the multistep_input in the interpolator config? It seems confusing to me as we then define explicit_times

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -858,6 +879,7 @@ def _plot(
         batch: torch.Tensor,
         batch_idx: int,
         epoch: int,
+        time_out: tuple,
```
VeraChristina: Small nitpick: my brain parses `time_out` into timeout; something like `output_times` seems more readable to me. (Feel free to ignore this one)

training/src/anemoi/training/config/interpolator.yaml
mc4117: I included this to make the process step in the callbacks work https://github.com/ecmwf/anemoi-core/blob/b45daf4bee52341fba166d32f4789486efbcfad5/training/src/anemoi/training/diagnostics/callbacks/plot.py#L973 but I agree maybe it's not the best name? But if I change the name that would be a breaking config change

training/src/anemoi/training/diagnostics/callbacks/plot.py
mc4117: Good idea!

training/tests/unit/diagnostics/test_callbacks.py
```diff
@@ -28,9 +28,15 @@
 NUM_FIXED_CALLBACKS = 3  # ParentUUIDCallback, CheckVariableOrder, RegisterMigrations
 
 default_config = """
+training:
```
anaprietonem, line 31: I think you just need this entry here and not again inside callbacks right?

mc4117: yes you're right!

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
```diff
@@ -151,13 +153,19 @@ def on_validation_batch_end(
                     post_processor.nan_locations = pl_module.allgather_batch(post_processor.nan_locations)
             self.post_processors = self.post_processors.cpu()
 
+            if self.config["training"]["model_task"] == "anemoi.training.train.tasks.GraphInterpolator":
```
anaprietonem: Could this be moved to the init of the BasePlotCallback and then we can reuse it across all the functions. At lest output_times comes from the config so that ones should be fine. We could have in BasePlotCallback something like
```
def _get_output_times(config,pl_module):
    if config["training"]["model_task"] == "anemoi.training.train.tasks.GraphInterpolator":
        output_times = (len(config.training.explicit_times.target), "time_interp")
    else:
        output_times = (getattr(pl_module, "rollout", 0), "forecast")
    return output_times
```
and then that one can be used in all the `on_validation_epoch_end` calls like `output_times=self.__get_output_times(config,pl_module)`

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1043,17 +1065,19 @@ def _plot(
...
 
-        data, output_tensor = self.process(pl_module, outputs, batch)
+        data, output_tensor = self.process(pl_module, outputs, batch, output_times)
 
         local_rank = pl_module.local_rank
-        rollout = getattr(pl_module, "rollout", 0)
-        for rollout_step in range(rollout):
+
+        for rollout_step in range(output_times[0]):
+            init_step = rollout_step if output_times[1] == "time_interp" else 0
```
anaprietonem:
```suggestion
            # Reference step - 0 step fo forecast model, previous time index for temporal interpolator
            init_step = rollout_step if output_times[1] == "time_interp" else 0
```

feel free to adapt the comment but I think a line explaining the logic could be useful

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1132,10 +1156,12 @@ def _plot(
                 for name in self.parameters
             }
 
+            init_step = rollout_step if output_times[1] == "time_interp" else 0
```
anaprietonem: Propose to move this to a small function in BasePlot or PlotAdditionalMetrics

```
    def _get_init_step(self, rollout_step: int, model: tuple) -> int:
        """Return index of initial step for plotting."""
        return rollout_step if mode == "time_interp" else 0

    init_step = self._get_init_step(rollout_step,output_times[1])

```
then it's easier to debug or fix in the future across all classes.

training/src/anemoi/training/config/interpolator.yaml
anaprietonem: Is a bit forward looking and yet to be decided - but any chance we need this because of the work Dieter was doing with multistep_input?

If you don't pass it in the config what value does this get initialised with? We could also put this in the `output_times` function. I mentioned below

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
anaprietonem: Also you we could check if the pl_module isinstance(GraphInterpolator) rather than the string of the config? thinking that would be less prone to breaking if the code moves

training/src/anemoi/training/config/interpolator.yaml
VeraChristina: @anaprietonem, I think Dieter's work is related to multistep *output*, and for the forecaster not the time interpolator (where explicit times are configurable).

@mc4117, if we only need this value for plotting, I would suggest to set a default value wherever we use this in the plots, or to add an internal method to the plotting callbacks that gets this value or the default, similar to what Ana suggests below for the `output_times`

mc4117: I've checked and it's 2 because this is the default in training. I still think this should be overwritten as it feels like a mistake waiting to happen in different parts of the code

training/src/anemoi/training/diagnostics/callbacks/plot.py
mc4117: added!

training/src/anemoi/training/diagnostics/callbacks/plot.py
mc4117: added

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
mc4117: add and made change for pl module

training/src/anemoi/training/config/interpolator.yaml
mc4117: I've removed this from the config and added it to the interpolator task init

Issue #599, opened 2025-10-10
Spectral loss


theissenhelen:
### Is your feature request related to a problem? Please describe.

Anemoi-training already provides a spectral loss in the form of a 2d FFT, mostly relevant for limited area models. A spectral loss for global models requires a differentiable implementation of accurate spherical harmonics in torch. 

### Describe the solution you'd like

We need the following initial  steps for this work:

1. An accurate differentiable harmonic transform in torch that 

- deals with loading pre-computed and/or computing the Legendre polynomials

- sufficiently efficient to allow computations on standard grid resolutions both with respect to runtime and memory 

Our baseline here is the spherical harmonic implementation of ectrans.

2. A loss module in anemoi-training that handles the spectral transform and computes a loss by e.g. using the power spectrum. 

### Describe alternatives you've considered

[Torch-harmonics](https://github.com/NVIDIA/torch-harmonics): 

Torch-harmonics does not currently support reduced gaussian grids.

frazane:
Hi @theissenhelen, I am working on the spectral losses that are currently implemented in anemoi. I am still applying them on a regional domain, so using the 2D FFT. However, I am considering a refactor of the code, and I would like to do it in away that paves the way for the future integration of spherical harmonics. Is this something you are already working on or can I move forward with this? Also, would be nice to have a chat in any case!

theissenhelen:
Hi @PortillaS-Predictia,

I had a look at your code, very neat! It is very similar to what I am doing except that I am using ECMWF’s ectrans to compute the polynomials and adhere to ectrans conventions for the SHT. Yes, that paper is exactly what we have been looking at as well.

PR #678, opened 2025-11-14
feat: introduce spectral losses module

frazane:
Following #599 and after discussing with @theissenhelen I am opening this PR to start the work on a generic spectral losses module. The focus of this PR is to establish the design of the module and aims to define the right abstractions and tests. Only very basic spectral losses will be actually implemented here for testing purposes. The work will also build on the existing spectral losses implemented in [spatial.py](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/losses/spatial.py). Feedback is of course more than welcome! 

## Design proposal

A main concern when designing spectral losses is the separation of the logic of the transformation to spectral domain from the logic for computing the error metric itself. Therefore, we introduce two respective abstract classes, `SpectralTransform` and `SpectralLoss` for this goal. The former is then assigned as the `transform` attribute of the latter.

- `SpectralTransform`: This class requires all subclasses to implement the `__call__()`  method. As an example, I already implemented `FFT2D` by inheriting from the `SpectralTransform` class. `SHT` is added as a placeholder. Note that this is temporarily implemented inside the spectral losses module, but might be moved at some point into `anemoi-models`. An important assumption of this class is that the input and output tensors will always be assumed to be of shape [batch, time, [ensemble], points, variable], so any reshaping required for the transform happens inside and is undone before returning the value. It's still unclear to me whether this generalizes well (see second point in open questions).

- `SpectralLoss`: All subclasses that directly inherit from this class will have to implement the `forward()` method, which will contain the entire logic for the loss computation, including scaling and reduction. For now, only one loss has been implemented (ported from the `spatial.py` module).

- `FunctionalSpectralLoss`: Following an existing pattern in the `losses` package, we also introduce a `FunctionalSpectralLoss` class that inherits from both `FunctionalLoss` and `SpectralLoss` and allows for substantial code deduplication. An example implementation is implemented as `SpectralL2Loss`. 

## Open questions
- What will be the general signature of the SHT? Shapes of input and output tensors? Arguments?
- What would be the correct way to exclude some scalers, e.g. scalers applied on the nodes which become meaningless in the spectral domain?
- What do we do where the logic of the transform and the error metric cannot be strictly decoupled? For instance, I imagine some losses might only be computed on 2D arrays in the spectral domain so with FFT2D. 
- Shall we implement correctness tests?

### DIFF
--- a/models/src/anemoi/models/layers/graph_provider.py
+++ b/models/src/anemoi/models/layers/graph_provider.py
@@ -454,6 +454,13 @@ class ProjectionGraphProvider(BaseGraphProvider):
         super().__init__()
 
         if file_path is not None:
+            if src_node_weight_attribute is not None:
+                msg = f"Building ProjectionGraphProvider from file, so src_node_weight_attribute='{src_node_weight_attribute}' will be ignored."
+                LOGGER.warning(msg)
+
+            if edge_weight_attribute is not None:
+                msg = f"Building ProjectionGraphProvider from file, so edge_weight_attribute='{edge_weight_attribute}' will be ignored."
+                LOGGER.warning(msg)
             self._build_from_file(file_path, row_normalize)
         else:
             assert (
--- a/models/src/anemoi/models/schemas/residual.py
+++ b/models/src/anemoi/models/schemas/residual.py
@@ -86,6 +86,8 @@ class TruncatedConnectionSchema(BaseModel):
                     "When using graph-based projection, do not specify truncation_up_file_path or truncation_down_file_path."
                 )
 
+        return self
+
 
 ResidualConnectionSchema = Annotated[
     SkipConnectionSchema | TruncatedConnectionSchema,


OpheliaMiralles:
> > I wouldn't say it starts the work on spectral loss since there are some already implemented in the anemoi-core losses...
> 
> Starts the work for a _generic_ module for spectral losses, which has a broader goal of defining abstractions. But yes absolutely, it also draws on what's already implemented in the `spatial.py`, you're right to point that out. I adapted the PR message!

I assigned myself as a reviewer so I'm willing to contribute and give feedback :)

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,220 @@
...
+        \mathrm{FourierLoss}(X, \hat X)
+            = \lVert F - \hat F \rVert_p^2
+    By default uses L2 loss in spectral domain.
+    """
+
+    def calculate_difference(self, pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        return (pred - target) ** 2
+
+
+class FourierCorrelationLoss(SpectralLoss):
```
OpheliaMiralles, line 190: I would also add the log version of the FFT2 norm

TobiasGoeckeDWD:
Great initiative. Some implementation of spectral weights would also be usefeul. Maybe a some funcion of wavenumber, some power law.

TobiasGoeckeDWD:
`What would be the correct way to exclude some scalers, e.g. scalers applied on the nodes which become meaningless in the spectral domain?`

The node weights could be applied before the transform, no? 
$\hat{x}=T(W\otimes x)$

frazane:
> `What would be the correct way to exclude some scalers, e.g. scalers applied on the nodes which become meaningless in the spectral domain?`
> 
> The node weights could be applied before the transform, no? x ^ = T ( W ⊗ x )

I am not sure what would be the influence of this on the distance in the spectral domain. Do you have more details?

training/src/anemoi/training/losses/spectral.py
frazane: Yes definitely! Do you want to do it in this PR or later?

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,203 @@
...
+        grid_shard_slice: slice | None = None,
+        group: ProcessGroup | None = None,
+    ) -> torch.Tensor:
+
+        is_sharded = grid_shard_slice is not None
+        group = group if is_sharded else None
+        eps = torch.finfo(pred.dtype).eps
+
+        # temporary fix for https://github.com/ecmwf/anemoi-core/issues/725
+        if without_scalers is not None and 2 not in without_scalers and not isinstance(without_scalers[0], str):
```
sahahner: Thank you for this nice work! 

Do I understand correctly that you are applying the loss scalers in the spectral space?

Are you here excluding the grid dimension from the scalers?

Then I would recommend to use the TensorDim:

`from anemoi.training.utils.enums import TensorDim`

frazane: That's right. It's a bit of a dirty fix but hopefully we will find a better long-term solution. Yes that's what I am doing there and indeed I should have used the enum. Thanks!

OpheliaMiralles:
> It could be good to check whether the refactored version behaves exactly as the previous implementation like @OpheliaMiralles mentioned, but correctness tests could be beneficial as well.
> 
> Could you also delete the `spatial.py`, please?

Done :)

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+def _ensure_without_scalers_has_2(without_scalers: list[str] | list[int] | None) -> list[str] | list[int]:
+    """Temporary fix for https://github.com/ecmwf/anemoi-core/issues/725.
+
+    Some pipelines pass numeric scaler indices and rely on excluding scaler index 2
+    by default. Ensure this exclusion is present for numeric lists.
+    """
+    if without_scalers is None:
+        return [2]
```
sahahner: Can we please use the TensorDim for the node dimension here, to avoid hard coding the second dimension. See https://github.com/ecmwf/anemoi-core/pull/678#discussion_r2602993632

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -0,0 +1,270 @@
...
+import torch.fft
+
+from anemoi.models.layers.spectral_helpers import CartesianRealSHT
+from anemoi.models.layers.spectral_helpers import EcTransOctahedralSHTModule
+from anemoi.models.layers.spectral_helpers import OctahedralRealSHT
+
+LOGGER = logging.getLogger(__name__)
+
+
+class SpectralTransform:
```
theissenhelen: This should inherit from ABC.

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,360 @@
...
+        group = group if is_sharded else None
+        eps = torch.finfo(pred.dtype).eps
+
+        pred_spectral = self._to_spectral_flat(pred)
+        target_spectral = self._to_spectral_flat(target)
+
+        power_pred = torch.abs(pred_spectral) ** 2
+        power_tgt = torch.abs(target_spectral) ** 2
+
+        log_diff = torch.log(power_tgt + eps) - torch.log(power_pred + eps)
```
PortillaS-Predictia, line 179: Why is this loss applied "entry-wise" to the spectral coefficients (instead of aggregating the coefficients and applying the loss to the resulting power spectrum)? The way I see it, this only makes sense for "2D/LAM" models (flat geometry), in which case the Fourier amplitudes would be "translationally invariant" (i.e., a slightly misplaced forecast would only differ in phase, but not amplitude, and this loss would correctly evaluate to zero) ... but this isn't true for spherical geometry (the amplitudes change too, within each energy band, given by the Wigner _d_-matrices, and this loss would not evaluate to zero ... I guess). Do you have any reference for this?

OpheliaMiralles: I personally only used it with local models, but in principle you are right and there should be an aggregation of the coefficients. However, the way anemoi applies scalers to the loss is for now preventing us to do that, because each scaler applies on a dim which shape corresponds to the output/predicted tensor. Since the spectral transforms change the shape of the vectors on the grid dim, we should think about a way to apply the scalers before. But maybe out of scope for this PR and for the moment other losses can be used for spherical geometry (including the one you introduced :)).

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,248 @@
...
+from anemoi.training.losses.base import BaseLoss
+from anemoi.training.utils.enums import TensorDim
+
+LOGGER = logging.getLogger(__name__)
+
+
+def _ensure_without_scalers_has_2(without_scalers: list[str] | list[int] | None) -> list[str] | list[int]:
+    """Temporary fix for https://github.com/ecmwf/anemoi-core/issues/725.
+
+    Some pipelines pass numeric scaler indices and rely on excluding scaler index 2
```
sahahner:
```suggestion
    Some pipelines pass numeric scaler indices and rely on excluding scalers over grid dimension
```

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,248 @@
...
+
+        pred_spectral = self._to_spectral_flat(pred)
+        target_spectral = self._to_spectral_flat(target)
+
+        diff = torch.abs(pred_spectral - target_spectral) ** 2
+
+        result = self.scale(
+            diff,
+            scaler_indices,
+            without_scalers=_ensure_without_scalers_has_2(without_scalers),
```
sahahner:
```suggestion
            without_scalers=_ensure_without_scalers_has_grid_dimension(without_scalers),
```

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,248 @@
...
+from anemoi.models.layers.spectral_transforms import FFT2D
+from anemoi.models.layers.spectral_transforms import SHT
+from anemoi.models.layers.spectral_transforms import SpectralTransform
+from anemoi.training.losses.base import BaseLoss
+from anemoi.training.utils.enums import TensorDim
+
+LOGGER = logging.getLogger(__name__)
+
+
+def _ensure_without_scalers_has_2(without_scalers: list[str] | list[int] | None) -> list[str] | list[int]:
```
sahahner:
```suggestion
def _ensure_without_scalers_has_grid_dimension(without_scalers: list[str] | list[int] | None) -> list[str] | list[int]:
```

sahahner:
Thank you for this nice refactor. Has someone ensured that the previous version of the code and the version after this refactor behave the same?

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        self.x_dim = x_dim
+        self.y_dim = y_dim
+        self.nodes_slice = slice(*nodes_slice)
+
+    def forward(
+        self,
+        data: torch.Tensor,
+    ) -> torch.Tensor:
+        data = data[:, :, self.nodes_slice, :]
```
sahahner: Do we want to include an assert somewhere that catches the case that the data is actually not rectangular and of size x_dim, y_dim? 
Possibly this already is checked somewhere else and I am not able to find it.

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,248 @@
...
+
+        power_pred = torch.abs(pred_spectral) ** 2
+        power_tgt = torch.abs(target_spectral) ** 2
+
+        log_diff = torch.log(power_tgt + eps) - torch.log(power_pred + eps)
+
+        result = self.scale(
+            log_diff**2,
+            scaler_indices,
+            without_scalers=_ensure_without_scalers_has_2(without_scalers),
```
OpheliaMiralles:
```suggestion
            without_scalers=_ensure_without_scalers_has_grid_dimension(without_scalers),
```

training/src/anemoi/training/losses/spectral.py
```diff
@@ -0,0 +1,248 @@
...
+
+        pred_spectral = self._to_spectral_flat(pred)
+        target_spectral = self._to_spectral_flat(target)
+
+        cross = torch.real(pred_spectral * torch.conj(target_spectral))
+
+        cross = self.scale(
+            cross,
+            scaler_indices,
+            without_scalers=_ensure_without_scalers_has_2(without_scalers),
```
OpheliaMiralles:
```suggestion
            without_scalers=_ensure_without_scalers_has_grid_dimension(without_scalers),
```

models/src/anemoi/models/layers/spectral_transforms.py
frazane: How do we know this information a priori? If the shape information provided by the user is wrong, an error will be raised by the einops operation and it's pretty informative in my experience.

PortillaS-Predictia: In the `BaseGraphModule` constructor (where the loss is instantiated) we have access to the "graph_data" ... so we could do `n_nodes = graph_data[config.graph.data].x.shape[0]` and then check the relation `x_dim * y_dim = n_nodes`.

frazane: This transform is meant to be applied to a subset of the nodes in the graph (as specified by `nodes_slice`), for instance in the case of the SGM approach, so the relation `x_dim * y_dim = n_nodes` will not hold.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -0,0 +1,103 @@
...
+        batch_size, ens, _, var = data.shape
+        # [batch, ens, y*x, variables] -> [batch*ens*variables, y, x]
+        data = einops.rearrange(data, "b e (y x) v -> (b e v) y x", x=self.x_dim, y=self.y_dim)
+        fft_data = torch.fft.fft2(data)
+        # [batch**ens*variables, y, x] -> [batch, ens y, x, variables]
+        return einops.rearrange(fft_data, "(b e v) y x -> b e y x v", b=batch_size, e=ens, v=var)
```
PortillaS-Predictia: Do we really need to collapse the `b e v` dimensions? I believe this would work just fine:

```
data = data[:, :, self.nodes_slice, :]
data = einops.rearrange(data, "b e (y x) v -> b e y x v", x=self.x_dim, y=self.y_dim)
return torch.fft.fft2(data, dim=(2, 3))
```

Or even agnostic of trailing dimensions:
```
data = data[:, :, self.nodes_slice, :]
data = einops.rearrange(data, "... (y x) v -> ... y x v", x=self.x_dim, y=self.y_dim)
return torch.fft.fft2(data, dim=(-3, -2))
```

models/src/anemoi/models/layers/spectral_transforms.py
PortillaS-Predictia: Then I guess we could combine `n_nodes` and `nodes_slice` to build the appropriate check (?) ... but maybe it's not worth the hassle, and the error raised by `einops` will be enough (if Sara agrees).

models/src/anemoi/models/layers/spectral_transforms.py
frazane: Good observation. If the results are equivalent, it would make more sense indeed. Also, maybe `dim` could be computed as `(TensorDim.GRID, TensorDim.GRID + 1)`?

OpheliaMiralles: We could also use torch.index_select and TensorDim.GRID instead of data[:, :, self.nodes_slices, :] if we want to anticipate adding the time dimension

OpheliaMiralles: And I would vote for this 
data = einops.rearrange(data, "... (y x) v -> ... y x v", x=self.x_dim, y=self.y_dim)
return torch.fft.fft2(data, dim=(-3, -2)) because -3 and -2 are likely going to remain grid dims as vars is always last

OpheliaMiralles:
> Thank you for this nice refactor. Has someone ensured that the previous version of the code and the version after this refactor behave the same?

Tests for previous implemented losses in spatial.py did not change and they still pass. The tests are not really checking behaviour but the logic and computations are exactly the same so I would say we are safe

models/src/anemoi/models/layers/spectral_transforms.py
sahahner: What about something like this in line 74? 

```
try:
    data = einops.rearrange(data, "b e (y x) v -> (b e v) y x", x=self.x_dim, y=self.y_dim,)
except Exception as e:
    raise RuntimeError(
        f"Possible dimension mismatch in einops.rearrange in FFT2D layer: "
        f"expected (y * x) == last spatial dim with y={self.y_dim}, x={self.x_dim}"
    ) from e
```
    
Then the einops error is printed, but it is also cleary stated that this comes from the 2d spectral layer. 
Additionally, we could add a test by providing incorrect `x_dim` and `y_dim`.

models/src/anemoi/models/layers/spectral_transforms.py
OpheliaMiralles: I'll do that if you want @frazane and add the warning / test @sahahner mentioned, then I think it could be good to go?

PR #679, opened 2025-11-15
feat(graphs): add scalar attributes

matschreiner:
This PR introduces four new scalar node attributes (SinLatitude, CosLatitude, SinLongitude, CosLongitude) that compute trigonometric transformations of node coordinates. Previously, these features were obtained from the dataset, but moving them to graph attributes makes them available for all node types, including hidden nodes.


graphs/src/anemoi/graphs/nodes/attributes/scalar_attributes.py
```diff
@@ -0,0 +1,50 @@
+# (C) Copyright 2024- Anemoi contributors.
```
HCookie:
```suggestion
# (C) Copyright 2025- Anemoi contributors.
```

I believe this is to be the date the file began existing

PR #680, opened 2025-11-17
feat(graphs): add postprocessor for stacking features

matschreiner:
# Description

This PR introduces the StackScalarFeatures post-processor, which consolidates multiple scalar node features into a single stacked feature vector. This is useful when scalar attributes (e.g., SinLatitude, CosLongitude, derived node attributes) are generated independently by the attribute system but must be combined into a dense feature tensor before being consumed by a model.

The processor:
- takes a list of scalar feature names,
- reads them from the specified NodeStorage in a HeteroData graph,
- concatenates them along the last dimension,
- stores the result under a configurable output name (default: "features").


matschreiner:
> Hi Mathias, I only have one comment on the PR Iwe should use node attributes (instead of scalar features) to be consistent with the naming used within the package. Apart from that, I wonder if we need this in anemoi-graphs. Currently, we support this in anemoi-models. A list of edge attributes is specified, and the mapper reads and concatenates them all

Hi Mario, I was actually working on a separate project where I’m using anemoi-graphs on its own because of its flexibility. In that context, I don't have access to the concatenation logic that lives in anemoi-models. 
Maybe it can be done in another way, but I think that having this functionality inside anemoi-graphs improves the package's utility as a standalone outside the anemoi ecosystem!

JPXKQX:
The functionality has been moved to a utility function that can be shared across packages -> #761

PR #682, opened 2025-11-17
fix(training,tasks): abstract RolloutForecasting task

JPXKQX:
## Description

This PR reduces code duplication between `GraphForecaster` and `GraphEnsForecaster`.

- Rename references to `rollout_step` in the base class. 
- Plotting callbacks: Get latlons from the model graph instead of using a task attribute to store them.
- Create `BaseRolloutGraphModule` class to abstract forecasting rollout functionality.


training/src/anemoi/training/train/tasks/rollout.py
```diff
@@ -0,0 +1,166 @@
...
+            config=config,
+            graph_data=graph_data,
+            statistics=statistics,
+            statistics_tendencies=statistics_tendencies,
+            data_indices=data_indices,
+            metadata=metadata,
+            supporting_arrays=supporting_arrays,
+        )
+
+        self.rollout = config.training.rollout.start
```
floriankrb, line 77: I wonder if `self.rollout = config.training.rollout.start` contains the start of the rollout or its length.
As suggested below with`LOGGER.debug("Rollout window length: %d", self.rollout)`


Perhaps it could be clarified with:
```suggestion
        self.rollout_start = config.training.rollout.start
```

training/src/anemoi/training/train/tasks/ensforecaster.py
```diff
@@ -142,89 +126,25 @@ def set_ens_comm_subgroup(
         self.ens_comm_subgroup_num_groups = ens_comm_subgroup_num_groups
         self.ens_comm_subgroup_size = ens_comm_subgroup_size
 
-    def gather_and_compute_loss(
+    def forward(self, x: torch.Tensor, fcstep: int) -> torch.Tensor:
+        return super().forward(x, fcstep=fcstep)
```
floriankrb: This is just calling the mother class's method. Remove it?
```suggestion
```

training/src/anemoi/training/train/tasks/rollout.py
JPXKQX: This is a placeholder for the rollout step. It's initially set to `rollout.start` and then increased after each epoch

training/src/anemoi/training/train/tasks/ensforecaster.py
JPXKQX: The reason is that the `forward()` method in `BaseGraphModule` only takes x as a positional argument, propagating all keyword arguments

training/src/anemoi/training/train/tasks/rollout.py
floriankrb: I see, it could be called `self.j_rollout`, and set to `None` here.  Then, when starting the loops, we would have a `self.j_rollout = self.rollout_start` ?

This `j_rollout` is confusing with the `fcstep`

JPXKQX: Another option could be `self.rollout_step` but this would overload the term currently used for the `_rollout_step()` method

JPXKQX: Perhaps we can change this later, once we have found a more appropriate alternative

PR #687, opened 2025-11-19
feat: plot multi datasets

VeraChristina:
## Description

Adaptation of plotting callbacks to multiple datasets. We add a parameter `datasets` to the callback configuration that allows specifying which datasets to plot. All plots will be based on the same configuration, including the choice of parameters to plot.

**In order to plot different datasets with different parameters, the user can configure multiple callbacks of the same type with different parameters.**

Callbacks included:
- PlotLoss
- PlotSample
- PlotSpectrum
- PlotHistogram
- GraphTrainableFeaturesPlot

Callbacks not included in this PR:
- ensemble plots
- LongRolloutPlots

Here are MLFlow runs for a [single dataset](https://mlflow.ecmwf.int/#/experiments/420/runs/8dca1935a3194edea16a21d4ed4d6a13/artifacts) and a [two dataset](https://mlflow.ecmwf.int/#/experiments/420/runs/ea59f113b0d34061b487038552d9467c/artifacts) use case.

## The config interface for plotting
Most datasets will have somewhat different parameters, and hence different parameters to plot. The config interface implemented on this PR means that users will either need to plot parameters shared by all datasets, or configure multiple callbacks of the same type for different datasets.
 
The reasons for this interface as a first draft is not that is it optimal for multiple datasets but rather that:
 - it keeps plots and plotting configs for the single dataset case backwards compatible. Avoiding regression on plotting for existing use cases was the main requirement.
- it requires less changes to the callbacks than making them more configurable
- it provides _some functionality_ to plot multiple datasets.
- we don't know yet how multiple datasets will mostly be used and it might make sense to delay the design of the interface a bit. This will give us more time to rethink plotting callbacks more generally, rather than settling on an interface now.

## Other decisions
- We will adapt pydantic schemas for plotting (and set a default for the new parameter `datasets = ["data"]`) as part of fixing schemas on the multiple datasets branch
- I'd suggest to update the remaining plotting callbacks separately but open to thoughts.


training/src/anemoi/training/config/debug_single.yaml
```diff
@@ -50,8 +50,8 @@ data:
   timestep: 6h
 
 diagnostics:
-  plot:
-    callbacks: []
+  # plot:
```
mc4117, line 52: what is the idea behind commenting this out? Is it because we think callbacks should run as default? In which case I would just delete these lines

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -939,32 +1000,38 @@ class BasePlotAdditionalMetrics(BasePerBatchPlotCallback):
...
         pl_module: pl.LightningModule,
+        dataset_name: str,
         outputs: list,
         batch: torch.Tensor,
     ) -> tuple[np.ndarray, np.ndarray]:
 
         if self.latlons is None:
-            self.latlons = np.rad2deg(pl_module.latlons_data.clone().detach().cpu().numpy())
+            self.latlons = {}
+        if dataset_name not in self.latlons:
```
mc4117, line 1016: Should this not be `if dataset_name in self.latlons`?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1117,37 +1196,41 @@ def _plot(
...
+                    []
+                    if self.config.data.datasets[dataset_name].diagnostic is None
+                    else self.config.data.datasets[dataset_name].diagnostic
+                )
+                plot_parameters_dict_spectrum = {
+                    pl_module.data_indices[dataset_name].model.output.name_to_index[name]: (
+                        name,
+                        name not in diagnostics,
+                    )
+                    for name in self.parameters
```
mc4117, line 1218: Do these self.parameters not need to be split up by dataset too?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -862,47 +914,51 @@ def _plot(
...
-            metadata_variables=self.metadata_variables,
-        )
-        self.parameter_names = [self.parameter_names[i] for i in argsort_indices]
-        if not isinstance(self.loss, BaseLoss):
-            LOGGER.warning(
-                "Loss function must be a subclass of BaseLoss, or provide `squash`.",
-                RuntimeWarning,
+        for dataset_name in dataset_names:
+            data_indices = pl_module.data_indices[dataset_name]
+            parameter_names = list(data_indices.model.output.name_to_index.keys())
```
mc4117, line 925: this is more a question than a comment: if I have [tp] in dataset_1 and [10u, tp] in dataset_2, how can I specify that I want only tp from dataset_1 to be plotted and 10u from dataset_2 to be plotted? if I understand the current code correctly, this would not be possible as the parameter_names to plot are not defined by dataset?

mc4117: or would I define two separate callbacks?

training/src/anemoi/training/config/debug_single.yaml
anaprietonem: Yes I think that's the case, for context we are aiming to merge @VeraChristina's PR into the 'feature/multi-dataset-integration' where the other features related to multiple datasets lives. There are some things missing here like for example the pydantic schemas but we agreed with Mario this can be added and reviewed properly when we prepare the merge of `[feature/multi-dataset-integration](https://github.com/ecmwf/anemoi-core/tree/feature/multi-dataset-integration)` to main

VeraChristina: Yes, agreed, they should run by default. This config file is more for the purpose of debugging various parts of multiple datasets, and we'll make sure callbacks are enabled in all the configuration templates that we add before merging `feature/multi-dataset-integration` into main.

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: Yes, this is a good question. The main aim of this implementation is to not introduce a regression for single dataset use cases. Therefore, we decided to not "dictionarify" the configuration of the callbacks in the user config at this point.

In order to produce plots with different configurations for different datasets, the current approach would be to add two callbacks. (We might revisit this in the refactor of the plotting callbacks.)

I'll make this more explicit in the PR description, and perhaps add the config I used for the era_cerra plots as an example.

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: These are the parameters from the callback config, so in the current implementation no (see below)

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: that's a good question - I guess we could consider having a dictionary for the dataset names so for each name we pass the variables but I fear then we would need to add also the parameter groups or users will end up requesting something like that - and I'd say for those cases it would recommended they define another callback - what do you think @VeraChristina ?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -238,6 +239,7 @@ def __init__(self, config: OmegaConf, every_n_batches: int | None = None):
         """
         super().__init__(config)
         self.every_n_batches = every_n_batches or self.config.diagnostics.plot.frequency.batch
+        self.dataset_names = dataset_names if dataset_names is not None else ["data"]
```
anaprietonem: could it be worth adding something in the future that if dataset_names are['all'] we try to get them for the list of datasets specified in the data config? I'd assume we will be saving such list or similar as part of the trainer metadata

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: I don't think so — this block is to collect the latlons from the Lightning module and to store them on the callback. (At this point self.latlons doesn’t necessarily contain an entry for this dataset yet, so we shouldn’t check dataset_name in self.latlons; we’re initializing it here.)

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -670,16 +700,27 @@ def __init__(self, config: OmegaConf, every_n_epochs: int | None = None) -> None
...
-    def get_edge_trainable_modules(self, model: torch.nn.Module) -> dict[tuple[str, str], torch.Tensor]:
+    def get_edge_trainable_modules(
+        self,
+        model: torch.nn.Module,
+        dataset_name: str,
+    ) -> dict[tuple[str, str], torch.Tensor]:
         trainable_modules = {
-            (model._graph_name_data, model._graph_name_hidden): model.encoder,
-            (model._graph_name_hidden, model._graph_name_data): model.decoder,
+            (model._graph_name_data, model._graph_name_hidden): model.encoder[dataset_name],
```
anaprietonem, line 720: _graph_name_hidden I understand that if we have one processor would be the same - but for _graph_name_data I would have imagined this would also be dependant of the dataset as in each encoder would have a different input graph if they use different datasets?

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: I think we either dictionarify the entire config of the callback or none of it? Just because it might be confusing otherwise. (Or because people will want to change it, like Ana says.)

I also think it'll likely be a breaking change to the config if we do that, which we are trying to avoid in other parts. (We can add a pydantic default for the single key `dataset_names` to have value `["data"]` but it's harder to do this for keys of dicts. -- I will need to check what the solution for this is in other parts of the config.)

VeraChristina: I think this is part of a bigger question around what the aim of this PR is for callbacks. My stance is: keep the things that were previously working, enable some amount of plotting for multi datasets now (potentially having to add multiple callbacks). In a later PR as part of the refactor: rethink the interface and enable better plotting for multi dataset use cases when we know a bit better what the plotting requirements for those use cases are

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1016,6 +1084,8 @@ def __init__(
         self.per_sample = per_sample
         self.colormaps = colormaps
 
+        self.dataset_names = dataset_names if dataset_names is not None else ["data"]
```
anaprietonem: could we somehow move this to the `BasePlot` so we do this already when calling the super?

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: I am okey if we want to keep this PR scope more narrowed and as a first pass. I think still Mariana's raised a good point but we can tackle it in the 'feature/multi-dataset-integration' with more curated configs examples

VeraChristina: I agree it's a good point -- it's the main question I have about the config interface. I think if we want to have a different interface for plotting multiple datasets, we should do it on this PR and not on the multi datasets branch. (Maybe I am not quite understanding your point?)

mc4117: I'm also happy to leave this to a later PR. I guess my question is a bit of an edge case and unlikely to happen

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: Yes, I think that's a good idea. Will make a note of this as part of the broader interface question

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: Good question. I had to look this up: The graph names here are just the names given by the config to identify the graphs. (For the default config that's "data" and "hidden".) These names are not dictionarified, but they point to dictionaries of graphs -- I guess this comes down to at which level of granularity we dictionarify and here it is just one below.

I am happy to add the dataset_name as a third entry to the key, if we think that makes things more clear. But really all that the key is used for is to keep track of the different parts of the model (to label the plots I think), and the plots should already have the dataset_name in the file name because I added that separately.

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: done!

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: Yes okey so I think as you point out these are just names, and mostly affects this for the label plots. 
Screenshot 2025-11-21 at 12 17 50

I find a bit unclear but this might be need tackling in the multiple-datasets PR and not here (as in what those two settings means for Mutilple-ds). I would recommend maybe just adding a comment for that function about why we don't need to apply the key to `model._graph_name_data` (so we don't get back to this question in the future)
Screenshot 2025-11-21 at 12 24 38

anaprietonem (approved):
LGTM to me. I'd recommend updating the description to capture the design decisions and adding a comment for clarification on the trainable_features plot.

PR #693, opened 2025-11-20
feat(model): add interpolation mappers

JPXKQX:
## Description

This PR adds `InterpolationForwardMapper` and `InterpolationBackwardMapper`. These encoder & decoder blocks use the recently introduced `SparseProjector` to operate as an interpolation layer.

- ✅ Tests added.
- ✅ Schemas updated.
- ✅ Sample config provided.

### DIFF
--- a/models/src/anemoi/models/layers/mapper.py
+++ b/models/src/anemoi/models/layers/mapper.py
@@ -15,7 +15,6 @@ from abc import abstractmethod
 from typing import Optional
 
 import einops
-import numpy as np
 import torch
 from torch import Tensor
 from torch import nn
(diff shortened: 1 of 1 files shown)


models/src/anemoi/models/layers/mapper.py
```diff
@@ -1438,3 +1443,150 @@ def pre_process(self, x, shard_shapes, model_comm_group=None, x_src_is_sharded=F
...
+        self.hidden_dim = hidden_dim
+        super().__init__(
+            sub_graph=sub_graph,
+            edge_weight_attribute=edge_weight_attribute,
+            src_grid_size=src_grid_size,
+            dst_grid_size=dst_grid_size,
+            autocast=autocast,
+            **kwargs,
+        )
+        self.emb_nodes = nn.Linear(in_channels_src, hidden_dim) if in_channels_src != hidden_dim else nn.Identity()
```
HCookie, line 1537: If the channels src and hidden were the same would we not still want some sort of embedding via a linear layer?

models/src/anemoi/models/layers/mapper.py
```diff
@@ -1438,3 +1443,150 @@ def pre_process(self, x, shard_shapes, model_comm_group=None, x_src_is_sharded=F
...
+        x, _ = x
+
+        if shard_shapes is None:
+            shard_shapes = apply_shard_shapes(x, 0, grid_shard_shapes) if grid_shard_shapes is not None else None
+
+        x = einops.rearrange(x, "(batch grid) features -> batch grid features", batch=batch_size)
+        x = self._to_channel_shards(x, shard_shapes, model_comm_group)
+        x = self.project(x)
+        x = self._to_grid_shards(x, shard_shapes, model_comm_group)
+        x = self.emb_nodes(x)
```
HCookie, line 1496: This is currently undefined, I would suggest making this an abstract method?

models/src/anemoi/models/layers/mapper.py
JPXKQX: That's a valid point, but when thinking about interpolation, I wouldn't expect linear mapping. For example, imagine wanting to use this layer as a residual connection for observations.

PR #695, opened 2025-11-21
feat: edge pre mlp for gt conv

ssmmnn11:
Allow for edge feature mixing in graph transformer conv

### DIFF
--- a/models/src/anemoi/models/layers/ensemble.py
+++ b/models/src/anemoi/models/layers/ensemble.py
@@ -1,23 +1,101 @@
 import logging
+from abc import ABC
+from abc import abstractmethod
 from typing import Optional
 
+import einops
 import torch
 from torch import Tensor
 from torch import nn
 from torch.distributed.distributed_c10d import ProcessGroup
 from torch.utils.checkpoint import checkpoint
 
+from anemoi.models.distributed.graph import gather_channels
 from anemoi.models.distributed.graph import shard_tensor
 from anemoi.models.distributed.shapes import change_channels_in_shape
+from anemoi.models.distributed.shapes import get_shard_shapes
 from anemoi.models.layers.mlp import MLP
+from anemoi.models.layers.sparse_projector import build_sparse_projector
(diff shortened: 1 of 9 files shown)


models/src/anemoi/models/layers/block.py
```diff
@@ -442,6 +442,7 @@ def __init__(
         bias: bool = True,
         qk_norm: bool = False,
         update_src_nodes: bool = False,
+        edge_pre_mlp: bool = True,
```
HCookie: If it is optional, why is it on by default?

HCookie: Will also need to add some info in the docstring

models/src/anemoi/models/layers/block.py
```diff
@@ -488,6 +491,15 @@ def __init__(
         self.lin_self = Linear(in_channels, num_heads * self.out_channels_conv, bias=bias)
         self.lin_edge = Linear(edge_dim, num_heads * self.out_channels_conv)  # , bias=False)
 
+        # Optional edge preprocessing MLP
+        if edge_pre_mlp:
+            self.edge_pre_mlp = nn.Sequential(
+                Linear(edge_dim, edge_dim),
+                layer_kernels.Activation(),
```
JPXKQX: `layer_kernels.Activation` seems to be used later too (L513), what about putting it into a variable next to Linear(L486) and LayerNorm (L487)?

models/src/anemoi/models/layers/block.py
```diff
@@ -488,6 +491,15 @@ def __init__(
         self.lin_self = Linear(in_channels, num_heads * self.out_channels_conv, bias=bias)
         self.lin_edge = Linear(edge_dim, num_heads * self.out_channels_conv)  # , bias=False)
 
+        # Optional edge preprocessing MLP
+        if edge_pre_mlp:
```
ssmmnn11: we could move the edge_pre_mlp below the last attribute that has learnable parameters. e.g. before graph_attention_backend, but not important.

PR #701, opened 2025-11-26
refactor(training,tasks): abstract diffusion task

JPXKQX:
## Description

This PR refactors some of the diffusion tasks:
- It creates a `BaseDiffusionForecaster` abstract class, instead of inheriting from `GraphForecaster`.
- The references to `rollout_step` are deleted as we are not currently supporting (or planning to support) rollout diffusion.

Now, all `GraphForecaster`, `GraphEnsForecaster`, `GraphDiffusionForecaster` and `GraphDiffusionTendForecaster` use the same exact `_get_input()` and `_get_target`.

Screenshot 2025-12-02 at 14 18 39

### DIFF
--- a/training/src/anemoi/training/train/tasks/ensforecaster.py
+++ b/training/src/anemoi/training/train/tasks/ensforecaster.py
@@ -127,6 +127,7 @@ class GraphEnsForecaster(BaseRolloutGraphModule):
         self,
         y_pred: torch.Tensor,
         y: torch.Tensor,
+        step: int | None = None,
         validation_mode: bool = False,
     ) -> tuple[torch.Tensor | None, dict[str, torch.Tensor]]:
         y_pred_ens = gather_tensor(
@@ -136,17 +137,19 @@ class GraphEnsForecaster(BaseRolloutGraphModule):
             mgroup=self.ens_comm_subgroup,
         )
 
-        loss = self.loss(
+        loss = self._compute_loss(
             y_pred_ens,
             y,
-            squash=True,
-            grid_shard_slice=self.grid_shard_slice,
-            model_comm_group=self.model_comm_group,
             grid_dim=self.grid_dim,
             grid_shard_shape=self.grid_shard_shapes,
         )
 
-        return loss, y_pred_ens if validation_mode else None
+        # Compute metrics if in validation mode
+        metrics_next = {}
+        if validation_mode:
+            metrics_next = self._compute_metrics(y_pred_ens, y, step=step, grid_shard_slice=self.grid_shard_slice)
+
+        return loss, metrics_next, y_pred_ens
 
     def _rollout_step(
         self,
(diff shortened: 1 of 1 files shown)


models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
```diff
@@ -618,7 +618,7 @@ def compute_tendency(
         x_t1 : torch.Tensor
             The state at time t1 with full input variables.
         x_t0 : torch.Tensor
-            The state at time t0 with full input variables.
+            The state at time t0 with prognostic input variables.
```
anaprietonem, line 621: why do have this change now? (ie x_t0 just having the prognostic variables?)

models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
```diff
@@ -635,24 +635,14 @@ def compute_tendency(
...
-            x_t0 = input_post_processor(
-                x_t0[..., self.data_indices.data.output.full],
-                in_place=False,
-                data_index=self.data_indices.data.output.full,
-            )
-        else:
-            x_t1 = x_t1[..., self.data_indices.data.output.full]
-            x_t0 = x_t0[..., self.data_indices.data.output.full]
+            x_t1 = input_post_processor(x_t1, in_place=False, data_index=self.data_indices.data.output.full)
+            x_t0 = input_post_processor(x_t0, in_place=False, data_index=self.data_indices.data.output.prognostic)
```
anaprietonem, line 639: no longer need else?

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -462,13 +464,17 @@ def compute_loss_metrics(
             Current step
         validation_mode : bool, optional
             Whether to compute validation metrics
+        y_pred_state : torch.Tensor, optional
```
anaprietonem: Would this mean we should be able to train tendency based forecaster models or any of the other models than inherits from the base class?

anaprietonem: I think the names are getting a bit confusing. The docstring could be improved to explain clearly that y_pred, y are tendencies and now y_state and y_pred_state are full state tensors

training/src/anemoi/training/train/tasks/rollout.py
```diff
@@ -82,6 +82,25 @@ def __init__(
...
 
+    def _get_input(self, batch: torch.Tensor) -> torch.Tensor:
+        """Get input tensor shape for diffusion model."""
+        x = batch[
+            :,
+            0 : self.multi_step,
+            ...,
+            self.data_indices.data.input.full,
+        ]  # (bs, multi_step, ens, latlon, nvar)
+        msg = f"Batch length not sufficient for requested multi_step length!, {batch.shape[1]} !>= {self.multi_step}"
```
anaprietonem: is this the same error message as line 57 in forecaster.py? do we need to keep the two?

models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
JPXKQX: This relates to using the `batch` only to obtain `x` and `y`, and then not using the `batch` anymore

models/src/anemoi/models/models/diffusion_encoder_processor_decoder.py
JPXKQX: no, related to the previous comment. Before, we were passing 
```
                batch[:, self.multi_step + rollout_step, ...],
                x_ref,
```
and now, we pass
```
            y,
            x_ref,
```

training/src/anemoi/training/train/tasks/base.py
JPXKQX: I am keeping the same names as before, but happy to update them (open for suggestions)

JPXKQX: > Would this mean we should be able to train tendency based forecaster models or any of the other models than inherits from the base class?

not right now, but the split between the tendency computation and the diffusion itself will be clearer, so it would be easier to implement in the future

training/src/anemoi/training/train/tasks/rollout.py
JPXKQX: yes, thanks for catching this

training/src/anemoi/training/train/tasks/base.py
ssmmnn11: I played around recently with tendency prediction. We could also move the "skipped" connection out of the model, instead it would then be explicitly applied in the forecaster and predict step. But of course not in this PR.

JPXKQX:
[Integration tests are passing](https://github.com/ecmwf/anemoi-core/actions/runs/20370086678)

dietervdb-meteo:
Ok, it will be nice to streamline this. Agreed that at the moment it is probably better to keep it for after a few of the structural PRs coming up.

PR #723, opened 2025-12-05
fix(training): handle CombinedLoss in print_variable_scaling

### LINKED ISSUE #720: Missing scaler attribute on CombinedLoss
Getting the scalers attribute from the CombinedLoss results in an error:

```
[omitted repeated traceback for brevity]
  File ".../anemoi-core/training/src/anemoi/training/train/train.py", line 201, in model
    model = model_task(**kwargs)  # GraphForecaster -> pl.LightningModule
            ^^^^^^^^^^^^^^^^^^^^
  File ".../anemoi-core/training/src/anemoi/training/train/tasks/rollout.py", line 67, in __init__
    super().__init__(
  File ".../anemoi-core/training/src/anemoi/training/train/tasks/base.py", line 211, in __init__
    self._scaling_values_log = print_variable_scaling(
                               ^^^^^^^^^^^^^^^^^^^^^^^
  File ".../anemoi-core/training/src/anemoi/training/losses/utils.py", line 41, in print_variable_scaling
    variable_scaling = loss.scaler.subset_by_dim(TensorDim.VARIABLE.value).get_scaler(len(TensorDim)).reshape(-1)
                       ^^^^^^^^^^^
  File ".../torch/nn/modules/module.py", line 1928, in __getattr__
    raise AttributeError(
AttributeError: 'CombinedLoss' object has no attribute 'scaler'. Did you mean: 'scale'?
```

this happens because the attribute is actually deliberately deleted here

https://github.com/ecmwf/anemoi-core/blob/dba42684794422e46b16c76f4c17a6db4859fd2c/training/src/anemoi/training/losses/combined.py#L139

is this absolutely necessary? If yes, why, and what are possible solutions to this? It seems that this is related to a logging utility function but the functionality of the class itself is correct.

### Version

latest

Linux x86

### Organisation

MeteoSwiss

frazane:
Closes #720

### Summary of changes:
- added handling of CombinedLoss class in the print_variable_scaling function by recursively call the function on the sub-losses and aggregating the results into a single dictionary

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-    "training": "0.7.0",
-    "graphs": "0.7.2",
-    "models": "0.10.0"
+    "training": "0.8.0",
+    "graphs": "0.8.0",
+    "models": "0.11.0"
 }
--- a/graphs/CHANGELOG.md
+++ b/graphs/CHANGELOG.md
@@ -8,6 +8,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Please add your functional changes to the appropriate section in the PR.
 Keep it human-readable, your future self will thank you!
 
+## [0.8.0](https://github.com/ecmwf/anemoi-core/compare/graphs-0.7.2...graphs-0.8.0) (2025-12-05)
+
+
+### ⚠ BREAKING CHANGES
+
+* **edges:** Edge feature revision #643 ([#727](https://github.com/ecmwf/anemoi-core/issues/727))
+
+### Features
+
+* **edges:** Edge feature revision ([#643](https://github.com/ecmwf/anemoi-core/issues/643)) ([720f4d8](https://github.com/ecmwf/anemoi-core/commit/720f4d8938f0b71425788940547cca23dd1345be))
+* **edges:** Edge feature revision [#643](https://github.com/ecmwf/anemoi-core/issues/643) ([#727](https://github.com/ecmwf/anemoi-core/issues/727)) ([d1372cf](https://github.com/ecmwf/anemoi-core/commit/d1372cfd89cb1052fd1e9da2a9a90c785ec9ddd9))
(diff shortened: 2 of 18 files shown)


training/src/anemoi/training/losses/utils.py
```diff
@@ -38,8 +39,14 @@ def print_variable_scaling(loss: BaseLoss, data_indices: IndexCollection) -> dic
         Dictionary mapping variable names to their scaling values. If max_variables is specified,
         only the top N variables plus 'total_sum' will be included.
     """
+    if isinstance(loss, CombinedLoss):
```
anaprietonem, line 44: What about trying to generalise the losses so the BaseLoss records the sub_loss.__class__.__name__ as a list, and then just iterate on that as part of this function? Then we don't need have any check for the `CombinedLoss` 

@HCookie

frazane: The way I see it the CombinedLoss is indeed a special case of a loss, so I see no issues in treating it as such. And it's only a one-line check (as opposed to a a larger change applied on the base class).

That's for this particular case though. There might be other places where this generalization would bring additional benefits, but I am not aware of them at the moment.

anaprietonem: My concern is that's easy to end up with many special cases that translate into many 'if statements' and then are difficult to understand or remember all. I was thinking if we could have similar or other cases coming from other losses, like the MultiScaleLoss in the future, and where generalising the BaseLoss to make it easier to log the scaling used for each loss. 

I get this might not be yet the case so let's get this one in, but @frazane to check we don't break this could you extend the tests in https://github.com/ecmwf/anemoi-core/blob/main/training/tests/unit/train/test_print_variable_scaling.py to also capture the `CombinedLoss` please?

frazane: Yes that's a valid concern. I agree we should do this in the future, but it will require a bit more thought. And now I need this fix to make progress with #678. I'll add the tests, thanks for the feedback and maybe let's have a chat at some point to see how we can get this done properly 👍

Issue #703, opened 2025-11-26
Update hardware references (docs & configs)


JPXKQX:
### What maintenance does this project need?

There are some references to `hardware` configuration that may need to be addressed:
- training/docs/user-guide
- graphtransformer_ens.yaml & transformer_ens.yaml (in training/src/anemoi/training/config/model)
- default & diffusion & ensemble & interpolator & lam -> line 5 (in training/src/anemoi/training/config/training)
- training/src/anemoi/training/schemas/{training.py & diagnostics.py}
- test_cicd_aicon_04

### Organisation

ECMWF

PR #737, opened 2025-12-09
fix(docs,schemas): update hardware reference in docstrings, documentation and schemas

anaprietonem:
## Description
When updating the configs for the system section, we forgot to update some details regarding the docs. This PR fixes that and some other things that Mario reported in https://github.com/ecmwf/anemoi-core/issues/703

[
](https://github.com/ecmwf/anemoi-core/issues/703)

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--737.org.readthedocs.build/en/737/

----
📚 Documentation preview 📚: https://anemoi-graphs--737.org.readthedocs.build/en/737/

----
📚 Documentation preview 📚: https://anemoi-models--737.org.readthedocs.build/en/737/

### DIFF
--- a/graphs/src/anemoi/graphs/generate/tri_icosahedron.py
+++ b/graphs/src/anemoi/graphs/generate/tri_icosahedron.py
@@ -86,17 +86,29 @@ def create_stretched_tri_nodes(
         Order of the node coordinates to be sorted by latitude and longitude.
     """
     assert area_mask_builder is not None, "AOI mask builder must be provided to build refined grid."
-    # Get the low resolution nodes outside the AOI
+    # Get the low resolution nodes
     base_coords_rad = get_latlon_coords_icosphere(base_resolution)
+
+    # Define the low resolution outside AOI mask
     base_area_mask = ~area_mask_builder.get_mask(base_coords_rad)
 
-    # Get the high resolution nodes inside the AOI
-    lam_coords_rad = get_latlon_coords_icosphere(lam_resolution)
-    lam_area_mask = area_mask_builder.get_mask(lam_coords_rad)
+    # Get the high resolution nodes
+    coords_rad = get_latlon_coords_icosphere(lam_resolution)
 
-    coords_rad = np.concatenate([base_coords_rad[base_area_mask], lam_coords_rad[lam_area_mask]])
-
+    # Get the node ordering for all high resolution nodes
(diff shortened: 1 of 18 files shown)


training/docs/user-guide/hydra-intro.rst
```diff
@@ -72,19 +73,22 @@ added to the training command like so:
 
 The following missing config options which must be overridden by users:
 
--  ``hardware.paths.data``: Location of base directory where datasets
-   are stored
+-  ``system.output.root``: Location of output directory
 
--  ``hardware.paths.graph``: Location of graph directory
+-  ``system.input.dataset``: Filename(s) of datasets used for training
```
frazane, line 78: So one can specify more than one dataset? If yes, I would document how this is done.

dietervdb-meteo (approved):
I browsed and searched through the new docs, looks like it was changed wherever it should. 
Also nice to have the test on warm start. Do we have similar tests for resuming via `run_id` and/or transfer learning? [Not for this PR, just curious]
LGTM, thanks @anaprietonem !

dietervdb-meteo:
Thanks @JPXKQX , so the configs needed to be fixed as well. How come tests pass while the configs are still not updated? Shouldn't we have some tests where we validate the default configs??

frazane (commented):
Only a minor comment/question. Otherwise looks really good! Thank you.

training/docs/user-guide/hydra-intro.rst
JPXKQX: Do you mean for the multiple datasets?

dietervdb-meteo: Hi means like a regional + global dataset for SGM or LAM i think.

frazane: I guess what @dietervdb-meteo says primarily. But I am not sure, that's why I am asking.

anaprietonem: Yes it's as before (so what Dieter mentioned) - just that now there is no need to pass a path and a filename. Rather you just have one entry ``system.input.dataset`` - where you define your dataset. And you can then also have ````system.input.dataset_forcing`` or ````system.input.land_dataset``, or other examples depending on your use case

PR #743, opened 2025-12-10
feat(training)!: Configurable weight averaging (exponential moving average (EMA), ...)

jakob-schloer:
Weight averaging is a technique to improve model generalization by averaging model weights during training. In the new pytorch lightning version (2.6.0) different versions of weight averaging are now supported. 

Here, I refactor the existing SWA implementation to support general weight averaging methods through the config. For example, EMA training can be used by adding the following to the `config.training`: 
```
         weight_averaging:
            _target_: pytorch_lightning.callbacks.EMAWeightAveraging
            decay: 0.999
```

The impact of EMA for smoother training and less fluctuation between checkpoints is especially important for the biases at longer rollout: 

image

Figure: Validation loss over training iterations for 2t. Upper panel shows validation loss at step 1 (1 day) and lower panel step 14 (14 days). 

**Note:** 
- I've deliberately deleted the previous version of weight initialization to keep the code cleaner. 
- Old configs should still work, since defaults are null if `weight_averaging` does not exist, but this is still a breaking change if swa was used before.

----
📚 Documentation preview 📚: https://anemoi-training--743.org.readthedocs.build/en/743/

----
📚 Documentation preview 📚: https://anemoi-graphs--743.org.readthedocs.build/en/743/

----
📚 Documentation preview 📚: https://anemoi-models--743.org.readthedocs.build/en/743/

### DIFF
--- a/models/src/anemoi/models/migrations/scripts/1776237003_rename_swa_to_weight_averaging.py
+++ b/models/src/anemoi/models/migrations/scripts/1776237003_rename_swa_to_weight_averaging.py
@@ -54,19 +54,3 @@ def migrate(ckpt: CkptType) -> CkptType:
     training.pop("swa", None)
     training.weight_averaging = None
     return ckpt
-
-
-def rollback(ckpt: CkptType) -> CkptType:
-    """Rollback the checkpoint.
-
-    Parameters
-    ----------
-    ckpt : CkptType
-        The checkpoint dict.
-
-    Returns
-    -------
-    CkptType
-        The rollbacked checkpoint dict.
-    """
-    return ckpt
--- a/models/src/anemoi/models/migrations/scripts/1776237003_rename_swa_to_weight_averaging.py
+++ b/models/src/anemoi/models/migrations/scripts/1776237003_rename_swa_to_weight_averaging.py
@@ -1,0 +1,72 @@
+# (C) Copyright 2025 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from anemoi.models.migrations import CkptType
+from anemoi.models.migrations import MigrationContext
+from anemoi.models.migrations import MigrationMetadata
+
+# DO NOT CHANGE -->
+metadata = MigrationMetadata(
+    versions={
+        "migration": "1.0.0",
+        "anemoi-models": "%NEXT_ANEMOI_MODELS_VERSION%",
+    },
+)
+# <-- END DO NOT CHANGE
+
+
+def migrate_setup(context: MigrationContext) -> None:
+    """Migrate setup callback to be run before loading the checkpoint.
+
+    Parameters
+    ----------
+    context : MigrationContext
+       A MigrationContext instance
+    """
+    context.delete_attribute("anemoi.training.schemas.training.SWA")
+
+
+def migrate(ckpt: CkptType) -> CkptType:
+    """Migrate the checkpoint.
+
+    Renames the removed ``training.swa`` config key to ``training.weight_averaging``.
+    Any previously-configured SWA is silently disabled; users who want weight
+    averaging on a migrated checkpoint should re-enable it via the new
+    ``weight_averaging`` Hydra-instantiate config.
+
+    Parameters
+    ----------
+    ckpt : CkptType
+        The checkpoint dict.
+
+    Returns
+    -------
+    CkptType
+        The migrated checkpoint dict.
+    """
+    training = ckpt["hyper_parameters"]["config"].training
+    training.pop("swa", None)
+    training.weight_averaging = None
+    return ckpt
+
+
+def rollback(ckpt: CkptType) -> CkptType:
+    """Rollback the checkpoint.
+
+    Parameters
+    ----------
+    ckpt : CkptType
+        The checkpoint dict.
+
+    Returns
+    -------
+    CkptType
+        The rollbacked checkpoint dict.
+    """
+    return ckpt
--- a/training/src/anemoi/training/config/training/diffusion.yaml
+++ b/training/src/anemoi/training/config/training/diffusion.yaml
@@ -29,11 +29,14 @@ gradient_clip:
   val: 1.
   algorithm: norm
 
-# stochastic weight averaging
-# https://pytorch.org/blog/stochastic-weight-averaging-in-pytorch/
-swa:
-  enabled: False
-  lr: 1.e-4
+# weight averaging only supported in Lightning 2.6+
+weight_averaging: null
+# For EMA:
+#  _target_: pytorch_lightning.callbacks.EMAWeightAveraging
+#  decay: 0.999
+# For SWA:
+#  _target_: pytorch_lightning.callbacks.StochasticWeightAveraging
+#  swa_lrs: 1.e-4
 
 optimization:
   optimizer:
--- a/training/src/anemoi/training/config/training/diffusion.yaml
+++ b/training/src/anemoi/training/config/training/diffusion.yaml
@@ -2,6 +2,7 @@
 defaults:
   - scalers: global
   - optimization: default
+  - weight_averaging: null
 
 # resume or fork a training from a checkpoint last.ckpt or specified in system.input.warm_start
 run_id: null
@@ -28,15 +29,6 @@ num_sanity_val_steps: 6
 gradient_clip:
   val: 1.
   algorithm: norm
-
-# weight averaging only supported in Lightning 2.6+
-weight_averaging: null
-# For EMA:
-#  _target_: pytorch_lightning.callbacks.EMAWeightAveraging
-#  decay: 0.999
-# For SWA:
-#  _target_: pytorch_lightning.callbacks.StochasticWeightAveraging
-#  swa_lrs: 1.e-4
 
 optimization:
   optimizer:
--- a/training/src/anemoi/training/config/training/weight_averaging/ema.yaml
+++ b/training/src/anemoi/training/config/training/weight_averaging/ema.yaml
@@ -1,0 +1,3 @@
+_target_: pytorch_lightning.callbacks.EMAWeightAveraging
+decay: 0.999
+update_starting_at_step: ${training.lr.warmup}
--- a/training/src/anemoi/training/config/training/weight_averaging/ema.yaml
+++ b/training/src/anemoi/training/config/training/weight_averaging/ema.yaml
@@ -1,3 +1,2 @@
 _target_: pytorch_lightning.callbacks.EMAWeightAveraging
 decay: 0.999
-update_starting_at_step: ${training.lr.warmup}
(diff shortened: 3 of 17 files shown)


anaprietonem:
@jakob-schloer  - How you're getting on with this PR? Are you planning to work on this soon?

models/src/anemoi/models/migrations/scripts/1776237003_rename_swa_to_weight_averaging.py
```diff
@@ -0,0 +1,72 @@
...
+    CkptType
+        The migrated checkpoint dict.
+    """
+    training = ckpt["hyper_parameters"]["config"].training
+    training.pop("swa", None)
+    training.weight_averaging = None
+    return ckpt
+
+
+def rollback(ckpt: CkptType) -> CkptType:
```
anaprietonem: This functionality was deprecated, so migration checkpoints do not need to provide this now. You can remove the rollback function

training/src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -58,6 +57,96 @@ def nestedget(config: DictConfig, key: str, default: Any) -> Any:
...
+    for name, current_buf in pl_module.named_buffers():
+        if name not in avg_buffers:
+            continue
+        avg_buf = avg_buffers[name]
+        if avg_buf.shape != current_buf.shape:
+            continue
+        current_buf.data.copy_(avg_buf.data)
+
+
+def _get_weight_averaging_callback(config: DictConfig) -> list[Callback]:
```
anaprietonem: I'd suggest to move those to it's own file and then just import this function in the callbacks __init__ so that we don't fill the __init__ with function related to this

training/src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -58,6 +57,96 @@ def nestedget(config: DictConfig, key: str, default: Any) -> Any:
...
+    if weight_averaging_config is None:
+        return []
+    if not isinstance(weight_averaging_config, dict | DictConfig):
+        return []
+    if "_target_" not in weight_averaging_config:
+        return []
+
+    callback = instantiate(weight_averaging_config)
+
+    # Patch swap/copy methods to use name-based matching. Needed for dynamic scalers like NaNMaskScaler.
```
anaprietonem: Jakob could you explain a bit more in detail why this is needed? and when/where it's used? Have to tested this when sharding the model?

training/src/anemoi/training/losses/scaler_tensor.py
```diff
@@ -336,7 +336,7 @@ def update_scaler(self, name: str, scaler: torch.Tensor, *, override: bool = Fal
         dimension = self._tensors[name][0]
 
         original_scaler = self._tensors.pop(name)
-        original_scaler_buffer = self._buffers.pop(name, None)
+        original_scaler_buffer = self._buffers.get(name)
```
anaprietonem, line 339: why is this change required?

training/src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -58,6 +57,96 @@ def nestedget(config: DictConfig, key: str, default: Any) -> Any:
...
+
+
+def _get_weight_averaging_callback(config: DictConfig) -> list[Callback]:
+    """Get weight averaging callback.
+
+    Supported are ExponentialMovingAverage and StochasticWeightAveraging.
+
+    Example config:
+        weight_averaging:
+          _target_: anemoi.training.diagnostics.callbacks.optimiser.ExponentialMovingAverage
```
anaprietonem: I think this example needs updating, to use pytorch_lightning.callbacks.EMAWeightAveragin right?

training/src/anemoi/training/config/training/diffusion.yaml
```diff
@@ -39,11 +39,14 @@ gradient_clip:
   val: 1.
   algorithm: norm
 
-# stochastic weight averaging
-# https://pytorch.org/blog/stochastic-weight-averaging-in-pytorch/
-swa:
-  enabled: False
-  lr: 1.e-4
+# weight averaging only supported in Lightning 2.6+
```
anaprietonem: is this both for EMA and SWA? If so we should add a check for the torch version in the code before those gets instantiated so the error is properly caught.

training/src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -58,6 +57,96 @@ def nestedget(config: DictConfig, key: str, default: Any) -> Any:
...
+    Returns
+    -------
+    list[Callback]
+        List containing the weight averaging callback, or empty list if not configured.
+    """
+    weight_averaging_config = nestedget(config, "training.weight_averaging", None)
+    if weight_averaging_config is None:
+        return []
+    if not isinstance(weight_averaging_config, dict | DictConfig):
+        return []
```
anaprietonem: if this is returning an empty list, then why not using in the config as default: 
weight_averaging: [ ] as default in the config for consistency?

it's also not super clear why we return a list here, if one should just be able to use just one WA technique right?

matschreiner:
Hi Jacob! Looks really nice, made a small PR with my suggestions from the ATS metting - https://github.com/ecmwf/anemoi-core/pull/1047

training/src/anemoi/training/diagnostics/callbacks/__init__.py
```diff
@@ -58,6 +57,96 @@ def nestedget(config: DictConfig, key: str, default: Any) -> Any:
...
+    if "_target_" not in weight_averaging_config:
+        return []
```
anaprietonem: I'd suggest to add some logging statements clear to help debug if the callback it's getting loaded correctly or getting skipped

training/src/anemoi/training/schemas/training.py
```diff
@@ -66,16 +66,42 @@ class GradientClip(BaseModel):
...
+          decay: 0.999
+          update_starting_at_step: 1000
     """
 
-    enabled: bool = Field(example=False)
-    "Enable stochastic weight averaging."
-    lr: NonNegativeFloat = Field(example=1.0e-4)
-    "Learning rate for SWA."
+    target_: Literal[
+        "pytorch_lightning.callbacks.EMAWeightAveraging",
```
anaprietonem: Regarding those schemas, since we are using pytorch_lightning classes directly, my recommendation would be to keep these consistent with the OptimizerSchema (lenient one) so that then we can pass as many args as we need and we rely on the API defined by lightning, and do not risk getting differences in the defaults with validation on/off that we saw before for the optimiser.

In the future if we have a custom anemoi WA class that we build on top of any of the PTL ones, then that one should include a schema.

anaprietonem (changes requested):

models/src/anemoi/models/migrations/scripts/1776237003_rename_swa_to_weight_averaging.py
jakob-schloer: Good to know. I've removed the rollback function.

training/src/anemoi/training/diagnostics/callbacks/__init__.py
jakob-schloer: The patch swap method is needed because the lightning `AveragedModel` uses positional pairing but our `ScaleTensors` are stored as `register_buffer(name, scaler, persistent=False)`. So once the AverageModel has been created and a dynamical scaler later adds/removes a buffer they no longer line up in position. 
The `_safe_swap_models` and `_safe_copy_average_to_current` pair entries by name instead of position.

training/src/anemoi/training/diagnostics/callbacks/__init__.py
jakob-schloer: Yes, changed that. I had a version with my own EMA class before :)

training/src/anemoi/training/losses/scaler_tensor.py
jakob-schloer: It's part of the same problem as above with the dynamic scalers. `_buffers` is an `OrderedDict`. If you pop and later call `register_buffer` again, the buffer gets inserted at the end. So every time a dynamic scaler updated itself, that buffer jumped to the end of the iteration order. So over the course of training, `named_buffers()` kept reshuffling.
We need both fixes because the pop -> get change fixes the EMA "live module" (`AveragedModel.update_parameters`) buffer during training and the `_safe_swap_models` and `_safe_copy_average_to_current` fix ensures that the buffers of the EMA "averaged module" and the "live module" matches.

training/src/anemoi/training/config/training/diffusion.yaml
jakob-schloer: Sure, I've added the check for lightning>=2.6.0.

training/src/anemoi/training/diagnostics/callbacks/__init__.py
jakob-schloer: Every `_get_*_callback` method returns a list and the caller in the `__init__.py` extends them. I've just sticked to the convention there. 
Changing the default to an empty list would suggest that multiple weight averaging techniques made sense - which they don't. 
For those two reasons, I leave the code as it is.

training/src/anemoi/training/diagnostics/callbacks/__init__.py
jakob-schloer: Sure, I've added some: 
```
  if weight_averaging_config is None:
        LOGGER.debug("No weight averaging configured. Skipping.")
        return []
    if not isinstance(weight_averaging_config, dict | DictConfig):
        LOGGER.warning(
            "training.weight_averaging has unexpected type %s; expected a dict with '_target_'. Skipping.",
            type(weight_averaging_config).__name__,
        )
        return []
    if "_target_" not in weight_averaging_config:
        LOGGER.warning("training.weight_averaging is set but has no '_target_' field. Skipping.")
        return []

    callback = instantiate(weight_averaging_config)
    LOGGER.info("Loaded weight averaging callback: %s", weight_averaging_config["_target_"])
```

training/src/anemoi/training/schemas/training.py
jakob-schloer: That's a much better solution.

training/src/anemoi/training/diagnostics/callbacks/__init__.py
jakob-schloer: Good idea. I've created a `weight_averaging.py`.

training/src/anemoi/training/config/training/weight_averaging/ema.yaml
```diff
@@ -0,0 +1,3 @@
+_target_: pytorch_lightning.callbacks.EMAWeightAveraging
+decay: 0.999
+update_starting_at_step: ${training.lr.warmup}
```
jakob-schloer:
```suggestion
```

jakob-schloer: Refactor of optimizers would make that error prone.

jakob-schloer:
Test with model sharding: 
https://mlflow.ecmwf.int/#/experiments/413/runs/cedd76dd82974a3ba3423b62b5974381/artifacts

Integration test:
https://github.com/ecmwf/anemoi-core/actions/runs/24857081535

anaprietonem (approved):
Thanks Jakob for addressing the comments and doing those extra checks! PR LGTM, feel free to merge!

PR #747, opened 2025-12-11
feat: add ensemble plots for multi datasets

VeraChristina:
## Description
Adapt the ensemble plotting callbacks to the multiple datasets setup

### DIFF
--- a/models/src/anemoi/models/models/ens_encoder_processor_decoder.py
+++ b/models/src/anemoi/models/models/ens_encoder_processor_decoder.py
@@ -95,7 +95,7 @@ class AnemoiEnsModelEncProcDec(AnemoiModelEncProcDec):
             x_data_latent = torch.cat(
                 (
                     x_data_latent,
-                    einops.rearrange(x_skip, "batch time ensemble grid vars -> (batch ensemble grid) vars"),
+                    einops.rearrange(x_skip, "bse grid vars -> (bse grid) vars"),
                 ),
                 dim=-1,
             )
--- a/training/docs/modules/diagnostics.rst
+++ b/training/docs/modules/diagnostics.rst
@@ -72,6 +72,33 @@ sharper and more detailed visuals but may be slower for large datasets.
 **Note** - this asynchronous behaviour is only available for the
 plotting callbacks.
 
+**Progress Bar**
+
+The progress bar callback can be configured to control how training
+progress is displayed. This is particularly useful on HPC systems with
+SLURM where output is written to files, as the default RichProgressBar
+in PyTorch Lightning 2.6+ may not work correctly.
+
+The progress bar is controlled by two configuration options:
+
+-  ``enable_progress_bar``: A boolean flag to enable or disable the
+   progress bar entirely
+-  ``progress_bar``: Configuration for which progress bar callback to
+   use
+
+.. code:: yaml
+
+   enable_progress_bar: True
+   progress_bar:
+     _target_: pytorch_lightning.callbacks.TQDMProgressBar
+     refresh_rate: 1
+
+Lightning 2.6+ supports the
+(https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.RichProgressBar.html#lightning.pytorch.callbacks.RichProgressBar)[RichProgressBar],
+which is recommended for interactive terminals and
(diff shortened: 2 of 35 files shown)


training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
```diff
@@ -368,15 +407,31 @@ def __init__(
         sample_idx: int,
         parameters: list[str],
         precip_and_related_fields: list[str] | None = None,
+        log_scale: bool = False,
```
anaprietonem, line 410: was this missing from before or a new feature? (as in it doesn't seem Multiple-datasets related)

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
```diff
@@ -305,11 +340,12 @@ def on_validation_batch_end(
         batch: torch.Tensor,
         batch_idx: int,
     ) -> None:
+        first_member_batch = {dataset: data[:, :, 0, :, :] for dataset, data in batch.items()}
         super().on_validation_batch_end(
             trainer,
             pl_module,
             outputs,
-            batch[:, :, 0, :, :],
+            first_member_batch,
```
anaprietonem: type hint for batch should be corrected? (as in not in the plot_ens but maybe something we missed in the first PR)

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
VeraChristina: I changed this when debugging. Guessing that the PlotHistogram API changed and that nobody changed it for the ens version of PlotHistogram

VeraChristina: Not sure why the tests don't catch it, though

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
VeraChristina: done! (for both `plot_ens.py` and `plot.py`)

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
anaprietonem: Right okey, I missed we had that also for the deterministic. So you mean https://github.com/ecmwf/anemoi-core/blob/main/training/tests/unit/diagnostics/test_callbacks.py? I guess   there was a default for log_scale False so maybe that was doing the trick

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -260,8 +260,8 @@ def on_validation_batch_end(
         self,
         trainer: pl.Trainer,
         pl_module: pl.LightningModule,
-        output: list[torch.Tensor],
-        batch: torch.Tensor,
+        output: tuple[torch.Tensor, list[dict[str, torch.Tensor]]],
```
mc4117, line 263: Are these changes you need to make for the ensemble or something that was left over from the last PR?

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
mc4117: This is not actually the first member. The loss doesn't actually have an ensemble dimension (this is something that should be refactored at some point)

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: last PR + generally outdated type hints (which Ana noticed because we reuse the classes here)

training/src/anemoi/training/diagnostics/callbacks/plot_ens.py
VeraChristina: Okay, good to know. I think this applies more generally to many comments/decisions in the ensemble plotting code. So maybe worth making a ticket?

VeraChristina: Here's the issue: https://github.com/ecmwf/anemoi-core/issues/757
And I have renamed this in the code for clarity

Issue #750, opened 2025-12-15
Warm Start Path Type Issue (str vs pathlib.Path)


timothyas:
### What happened?

I wanted to fine tune from a previous run, specifying `config.training.fork_run_id` along with `config.system.input.warm_start`. However, I ran into the error reported in the "log" section of this issue (sorry the formatting got messed up while copying that over).

To me, this looks like the `config.system.input.warm_start` is still a string, and needs to be converted to a `pathlib.Path` object.

### What are the steps to reproduce the bug?

Pick up from a previous run, specifying `config.system.input.warm_start` and `config.training.fork_run_id`. 

### Version

training-0.8.1

### Platform (OS and architecture)

GNU Linux x86_64

### Relevant log output

```shell
File "/global/u2/t/timothys/aneml/anemoi-core/training/src/anemoi/training/train/train.py", line 298, in last_checkpoint                                                          
    checkpoint = self._get_warm_start_checkpoint() or self._get_checkpoint_directory(fork_id)                                                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                  
  File "/global/u2/t/timothys/aneml/anemoi-core/training/src/anemoi/training/train/train.py", line 284, in _get_warm_start_checkpoint                                              
    assert Path.is_file(warm_start_path), msg                                                                                                                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                            
  File "/global/homes/t/timothys/.conda/envs/eagle/lib/python3.12/pathlib.py", line 892, in is_file                                                                                
    return S_ISREG(self.stat().st_mode)                                                                                                                                                                ^^^^^^^^^                                                                                                                                                        
AttributeError: 'str' object has no attribute 'stat'
```

### Organisation

NOAA Physical Sciences Laboratory

timothyas:
Thanks @anaprietonem ! Yep all good on my end

PR #751, opened 2025-12-15
fix(training): convert warm_start_path to pathlib.Path object

timothyas:
## Description

#750 

Relatively simple bugfix, that solves the issue described in #750, at least for me.

#750

##  Additional notes ##

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,12 +53,6 @@ repos:
   rev: v1.0.2
   hooks:
   - id: sphinx-lint
-# For now, we use it. But it does not support a lot of sphinx features
-- repo: https://github.com/dzhu/rstfmt
-  rev: v0.0.14
-  hooks:
-  - id: rstfmt
-    exclude: '(cli|schemas)/.*' # Because we use argparse and pydantic sphinx directives
 - repo: https://github.com/b8raoult/pre-commit-docconvert
   rev: "0.1.5"
   hooks:
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-    "training": "0.8.1",
+    "training": "0.8.2",
     "graphs": "0.8.1",
-    "models": "0.11.1"
+    "models": "0.11.2"
 }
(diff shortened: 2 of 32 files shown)


training/src/anemoi/training/train/train.py
```diff
@@ -278,6 +278,8 @@ def tensorboard_logger(self) -> pl.loggers.TensorBoardLogger:
     def _get_warm_start_checkpoint(self) -> Path | None:
         """Returns the warm start checkpoint path if specified."""
         warm_start_path = self.config.system.input.warm_start
+        if isinstance(warm_start_path, str):
```
anaprietonem: @timothyas thanks for catching this.  Would we need the if statement? you could just simply do:
```
warm_start_path = Path(self.config.system.input.warm_start)
```

timothyas: Yeah that makes sense to me, I'll get rid of the if statement. I guess I was being overly cautious

timothyas: Done in  f43fe5a

timothyas:
Hi @anaprietonem, I got an additional email notification that you commented asking if I could put this in the "expand_paths" function in base_schema, but now I can't find that comment. Would you prefer it if I moved this conversion there? Happy to do it

anaprietonem (dismissed):
Thanks @timothyas, LGTM!

anaprietonem:
> Hi @anaprietonem, I got an additional email notification that you commented asking if I could put this in the "expand_paths" function in base_schema, but now I can't find that comment. Would you prefer it if I moved this conversion there? Happy to do it

all good, that was the initial comment but then I updated to the suggestion above!

timothyas:
Sounds good! Thanks for clarifying @anaprietonem. Let me know if you'd like help here - it looks like a check for "None" type is needed. Happy to jump in as needed, but don't want to redo anything you may or may not have done.

anaprietonem (dismissed):
@timothyas apologies for the oversight. Indeed we needed the if statement! 
Changes LGTM, the unit tests are failing but it's just a matter of updating `test_warm_start_file_not_found` under `training/tests/unit/train/test_restarting_run.py` so that it uses the right type of error pytest.raises(FileNotFoundError, match=r"Warm start checkpoint not found"):

if you can add that to your PR then it's good to be merged!

PR #752, opened 2025-12-15
feat(models): triton gt more shapes

cathalobrien:
## Description
Triton GT can now run where number of heads and/or number of channels != a power of 2. e.g. 1526 channels or 12 heads

 This is achieved by padding the matrix offsets to reach the power of 2 size, and then masking out the extra values before loading or storing. This makes the code a bit uglier to follow, but i added some comments. Also I had to allocate some temp arrays with shape (H_pad, C_pad), so the memory usage isnt ideal for non-power-of-2 inputs

i split the pytests into forward and backward reference tests, and used a different function to test correctness which tells you if you're 1% off or 99% off the reference (which can be very motivating when debugging :D)

### DIFF
--- a/training/pyproject.toml
+++ b/training/pyproject.toml
@@ -79,6 +79,9 @@ optional-dependencies.docs = [
   "sphinx-argparse",
   "sphinx-rtd-theme",
 ]
+optional-dependencies.mlflowsync = [
+  "mlflow-export-import @ git+https://github.com/mlflow/mlflow-export-import.git@master",
+]
 optional-dependencies.plotting = [
   "distinctipy>=1",
 ]
--- a/training/src/anemoi/training/commands/mlflow.py
+++ b/training/src/anemoi/training/commands/mlflow.py
@@ -255,7 +255,6 @@ class MlFlow(Command):
                 run_id,
                 config,
                 expand_keys=config.diagnostics.log.mlflow.expand_hyperparams,
-                log_hyperparams=True,
                 clean_params=False,
             )
 
--- a/training/src/anemoi/training/train/train.py
+++ b/training/src/anemoi/training/train/train.py
@@ -436,7 +436,7 @@ class AnemoiTrainer(ABC):
         if self.config.diagnostics.log.mlflow.enabled:
             # Check if the run ID is dry - e.g. without a checkpoint
             self.dry_run = (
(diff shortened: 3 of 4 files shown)


models/src/anemoi/models/triton/gt.py
```diff
@@ -35,44 +37,76 @@ def _gt_fwd(
...
-    acc = tl.zeros((H, C), dtype=tl.float32)  # output accumulator, pending normalization by l_i
-    l_i = tl.zeros((H,), dtype=tl.float32)  # sum of attention weights
-    m_i = tl.full((H,), value=-float("inf"), dtype=tl.float32)  # running max for stability
+    q = tl.load(Q_ptr + dst_off, mask=H_C_mask_1d).to(tl.float32).reshape((H_pad, C_pad))
+    acc = tl.zeros((H_pad, C_pad), dtype=tl.float32)  # output accumulator, pending normalization by l_i
+    l_i = tl.zeros((H_pad,), dtype=tl.float32)  # sum of attention weights
+    m_i = tl.full((H_pad,), value=-float("inf"), dtype=tl.float32)  # running max for stability
 
     # helpers to avoid repeated computations/indexing:
-    qk_scale = 1 / tl.sqrt(float(C))
```
japols, line 64: I would keep this line here instead of having it be a function argument. It's already a compile time constant since C is a constexpr. We can make it a bit more explicit via:
`qk_scale: tl.constexpr = 1.0 / tl.sqrt(float(C))`

models/src/anemoi/models/triton/gt.py
```diff
@@ -130,52 +165,81 @@ def _gt_bwd_dst_pass(
     H: tl.constexpr,
     C: tl.constexpr,
     out_dtype: tl.constexpr,
+    qk_scale: tl.constexpr,
 ):
     dst_idx = tl.program_id(0)
     if dst_idx >= N_dst:
         return
 
-    qk_scale = 1.0 / tl.sqrt(float(C))
```
japols, line 138: same here

japols (changes requested):
Nice work, thanks for taking this on :)

models/src/anemoi/models/triton/gt.py
cathalobrien: sure

japols (approved):
looks good now, well done :)

PR #756, opened 2025-12-17
refactor: move trajectory_ids to new class

JPXKQX:
- We move the `trajectory_ids` specification from an independent (optional) field, model_run_info, to a new dataset class.

Now, we specify the dataset class in the config. There are 2 options
- `NativeGridDataset` current dataset.
- `TrajectoryDataset` dataset with different non-overlapping trajectories. 

This design aligns with incoming changes that could require new dataset classes (e.g. observations)

For each dataset, we can specify the following configuration:
```
      dataset: ${dataloader.dataset}
      start: 2021
      end: 2021
      frequency: ${data.frequency}
      drop: []
```
And if we want the dataset to have the trajectory information, we should add it as follows,
```
      dataset: ${dataloader.dataset}
      start: 2021
      end: 2021
      frequency: ${data.frequency}
      drop: []
      trajectory:
           start: 2020-02-05T12:00:00
           length: 18
```

Open for suggestions for names!

----
📚 Documentation preview 📚: https://anemoi-training--756.org.readthedocs.build/en/756/

----
📚 Documentation preview 📚: https://anemoi-graphs--756.org.readthedocs.build/en/756/

----
📚 Documentation preview 📚: https://anemoi-models--756.org.readthedocs.build/en/756/

### DIFF
--- a/models/src/anemoi/models/models/__init__.py
+++ b/models/src/anemoi/models/models/__init__.py
@@ -8,12 +8,15 @@
 # nor does it submit to any jurisdiction.
 
 from .base import BaseGraphModel
-from .diffusion_encoder_processor_decoder import AnemoiDiffusionModelEncProcDec
-from .diffusion_encoder_processor_decoder import AnemoiDiffusionTendModelEncProcDec
 from .encoder_processor_decoder import AnemoiModelEncProcDec
 from .ens_encoder_processor_decoder import AnemoiEnsModelEncProcDec
 from .hierarchical import AnemoiModelEncProcDecHierarchical
 from .interpolator import AnemoiModelEncProcDecInterpolator
+
+from .autoencoder import AnemoiModelAutoEncoder  # isort: skip
+from .hierarchical_autoencoder import AnemoiModelHierarchicalAutoEncoder  # isort: skip
+from .diffusion_encoder_processor_decoder import AnemoiDiffusionModelEncProcDec  # isort: skip
+from .diffusion_encoder_processor_decoder import AnemoiDiffusionTendModelEncProcDec  # isort: skip
 
 __all__ = [
     "BaseGraphModel",
@@ -23,4 +26,7 @@ __all__ = [
     "AnemoiDiffusionTendModelEncProcDec",
     "AnemoiModelEncProcDecHierarchical",
     "AnemoiModelEncProcDecInterpolator",
+    "AnemoiEnsModelEncProcDec",
+    "AnemoiModelAutoEncoder",
+    "AnemoiModelHierarchicalAutoEncoder",
 ]
--- a/models/src/anemoi/models/models/autoencoder.py
+++ b/models/src/anemoi/models/models/autoencoder.py
@@ -1,0 +1,265 @@
+# (C) Copyright 2025 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+import logging
+from typing import Optional
+
+import einops
+import torch
+from hydra.utils import instantiate
+from torch import Tensor
+from torch.distributed.distributed_c10d import ProcessGroup
+
+from anemoi.models.distributed.graph import shard_tensor
+from anemoi.models.distributed.shapes import get_or_apply_shard_shapes
+from anemoi.models.distributed.shapes import get_shard_shapes
+from anemoi.models.layers.graph_provider import create_graph_provider
+from anemoi.models.models import BaseGraphModel
+from anemoi.utils.config import DotDict
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AnemoiModelAutoEncoder(BaseGraphModel):
+
+    def _build_networks(self, model_config: DotDict) -> None:
+        """Builds the model components."""
+
+        self.encoder_graph_provider = torch.nn.ModuleDict()
+        self.encoder = torch.nn.ModuleDict()
+
+        for dataset_name in self._graph_data.keys():
+            # Create graph providers
+            self.encoder_graph_provider[dataset_name] = create_graph_provider(
+                graph=self._graph_data[dataset_name][(self._graph_name_data, "to", self._graph_name_hidden)],
+                edge_attributes=model_config.model.encoder.get("sub_graph_edge_attributes"),
+                src_size=self.node_attributes[dataset_name].num_nodes[self._graph_name_data],
+                dst_size=self.node_attributes[dataset_name].num_nodes[self._graph_name_hidden],
+                trainable_size=model_config.model.encoder.get("trainable_size", 0),
+            )
+
+            self.encoder[dataset_name] = instantiate(
+                model_config.model.encoder,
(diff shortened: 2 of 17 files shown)


training/src/anemoi/training/data/dataset.py
```diff
@@ -1,4 +1,4 @@
-# (C) Copyright 2024 Anemoi contributors.
+# (C) Copyright 2025 Anemoi contributors.
```
mchantry:
```suggestion
# (C) Copyright 2024- Anemoi contributors.
```

training/src/anemoi/training/data/dataset.py
```diff
@@ -121,7 +119,53 @@ def __repr__(self) -> str:
...
+        tree = super().tree(prefix)
+        tree.add(f"Trajectory start: {self.trajectory_start}")
+        tree.add(f"Trajectory length: {self.trajectory_length} steps")
+        return tree
+
+
+def create_dataset(dataset_config: dict) -> BaseAnemoiReader:
+    """Factory function to create dataset based on dataset configuration."""
+    trajectory_config = dataset_config.pop("trajectory", {})
+    if trajectory_config is not None and hasattr(trajectory_config, "start") and hasattr(trajectory_config, "length"):
```
anaprietonem, line 180: I would add a log statement here in to state what dataset is being created and if for some reason the trajectory config is not well defined that it returns a meaningful error message

training/src/anemoi/training/data/dataset.py
```diff
@@ -121,7 +119,53 @@ def __repr__(self) -> str:
...
         tree = Tree(prefix + " 💾 " + f"{self.__class__.__name__}")
         tree.add(f"Dataset: {self.data}")
-        tree.add(f"Timestep: {self.timestep}")
+        tree.add(f"Frequency: {self.frequency}")
         tree.add(f"Resolution: {self.resolution}")
         tree.add(f"Num variables: {len(self.name_to_index)}")
         return tree
+
+
+class NativeGridDataset(BaseAnemoiReader):
```
anaprietonem, line 132: do we need this? or it's more for backward compatibility?

training/src/anemoi/training/data/dataset.py
```diff
@@ -7,96 +7,95 @@
...
             return None
 
-    @cached_property
+    @property
     def variables(self) -> list[str]:
         """Return dataset variables."""
         return self.data.variables
 
-    @cached_property
+    @property
```
anaprietonem, line 62: why the change from @cached_property to @property?

training/src/anemoi/training/data/dataset.py
```diff
@@ -7,96 +7,95 @@
...
     def resolution(self) -> str:
         """Return dataset resolution."""
         return self.data.resolution
 
-    def get_sample(self, time_indices: int | list[int] | slice, reader_group_rank: int) -> torch.Tensor:
-        # NOTE: this is temporary until anemoi datasets allows indexing with arrays or lists
-        # data[start...] will be replaced with data[self.relative_date_indices + i]
+    @property
+    def trajectory_ids(self) -> list[str] | None:
```
anaprietonem: should this method just be specific to the trajectory dataset class?

training/src/anemoi/training/config/dataloader/non_analysis.yaml
```diff
@@ -0,0 +1,91 @@
...
+# ============
+# Dataloader definitions
+# These follow the anemoi-datasets patterns
+# You can make these as complicated for merging as you like
+# See https://anemoi-datasets.readthedocs.io
+# ============
+
+dataset: ${system.input.dataset}
+
+model_run_info:  # Add for non-analysis training
```
mc4117: maybe `forecast_data_info` would be a bit clearer?

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -1,6 +1,6 @@
 defaults:
 - data: zarr
-- dataloader: native_grid
+- dataloader: non_analysis
```
mc4117: not sure I would change this here. For the interpolator we will almost definitely do pre-training on analysis

JPXKQX: would it be better if I moved the `model_run_info` into the native_grid.yaml and set `null` as the default value?

training/src/anemoi/training/config/dataloader/non_analysis.yaml
JPXKQX: I used the same names for backward compatibility at the config level, but I'm happy to use different names.

training/src/anemoi/training/config/interpolator.yaml
mc4117: yes I think so

training/src/anemoi/training/config/dataloader/non_analysis.yaml
mc4117: I was just worried that what `model` means here might be a bit confusing

training/src/anemoi/training/data/dataset.py
JPXKQX: And I think it is a good practice too, in order to avoid inherithing from a concrete class.

training/src/anemoi/training/data/dataset.py
JPXKQX: I have added the log statements with level=info, but maybe they should be level=debug, what do you think?

mc4117: This log statement will only appear when making a dataset with trajectories, right? If so then I think log info is fine

training/src/anemoi/training/data/dataset.py
JPXKQX: My intuition is that these are always called by the properties of the MultiDataset, which they are still @cached_property, so this should be equivalent (or even better?)

training/src/anemoi/training/data/dataset.py
JPXKQX: I have added a `has_trajectories` attribute to the base class, and moved the trajectory_ids to the `TrajectoryDataset` class. Thanks for the feedback

training/src/anemoi/training/data/dataset.py
JPXKQX: I have added two log statements, but they are not very informative. They just print the class used

training/src/anemoi/training/config/interpolator.yaml
JPXKQX: Done!

training/src/anemoi/training/config/dataloader/non_analysis.yaml
JPXKQX: I'm fine with either option, @Magnus-SI, what do you think? I think you're one of the main users

training/src/anemoi/training/data/dataset.py
mchantry: Do we need what?

anaprietonem: 'NativeGridDataset' is a 'mirror' class that just inherits from `BaseAnemoiReader`. I found that strange, I understand   from Mario's comment that is mostly a best practice to not use directly the base class and in case this one would need extension compared to the base in the future, so all good

training/src/anemoi/training/data/dataset.py
anaprietonem: it's still good for me, as it's mostly a way to sanity check which type of dataset is being returned

PR #760, opened 2025-12-18
fix: wrong model for ens in benchmarking tests

anaprietonem:
## Description
Current benchmarking tests for the crps use the graphtransformer config while it should be graphtransformer_ens.
This PR fixes that.

##  Additional notes ##


cathalobrien:
good catch, an oversight on my part. there is a performance degradation from qk_norm in the processor. and a memory cost from the ensemble loss. But i have updated the benchmark tests to allow this to pass.

cathalobrien (approved):

PR #770, opened 2025-12-19
feat(models): Add fused qk_norm to triton GT kernel

cathalobrien:
## Description
This PR adds an option to the Triton GT kernel to normalise Q and K.

Currently, when QK normalisation is done in Anemoi (e.g. ensemble processor), a layernorm is computed over q and k before calling the GT. This is inefficient as it forces us to load all of Q and K twice (once for normalisation and once for computing GT.conv), and increases the memory usage as more intermediate inputs must be saved for the backward pass.

This PR fuses QK normalisation into the GT conv, by applying normalisation right after Q and K are loaded in the GT kernel. Since loading the elements from memory is the expensive part, and that is done anyway, the runtime cost of enabling QK_norm is quite low.  Benchmarking results are shown below, but it is faster and uses less memory then applying a compiled RMS norm followed by the triton GT kernel.

The normalisation method implemented is RMSNorm, which is different to the LayerNorm used currently but still widely used across ML. RMSNorm was implemented because LayerNorms arent suited to the tiled memory access pattern of the GT conv (due to maintaining global mean and variance).

The RMSNorm triton kernels are split into a different file, for clarity and future reuse. The GT pytests have been extended to test the RMSNorm option

```


=== Benchmarking graph: era_to_h fwd+bwd ===

rms_norm then pyg (non-compiled):

Compiled rms_norm x pyg:
OptimizedModule
        elapsed time: 39.23 ms / iter
        peak memory usage: 12709.78 MB


 standalone Triton GraphTransformer:
apply
        elapsed time: 9.14 ms / iter
        peak memory usage: 9886.27 MB


 compiled rms norm then standalone Triton GraphTransformer:
apply
        elapsed time: 14.84 ms / iter
        peak memory usage: 11024.06 MB


 fused rms_norm x Triton GraphTransformer:
apply
        elapsed time: 12.12 ms / iter
        peak memory usage: 9886.24 MB



=== Benchmarking graph: h_to_h fwd+bwd ===

rms_norm then pyg (non-compiled):

Compiled rms_norm x pyg:
OptimizedModule
        elapsed time: 10.83 ms / iter
        peak memory usage: 5916.90 MB


 standalone Triton GraphTransformer:
apply
        elapsed time: 1.43 ms / iter
        peak memory usage: 3092.68 MB


 compiled rms norm then standalone Triton GraphTransformer:
apply
        elapsed time: 5.19 ms / iter
        peak memory usage: 3252.27 MB


 fused rms_norm x Triton GraphTransformer:
apply
        elapsed time: 2.19 ms / iter
        peak memory usage: 3092.65 MB



=== Benchmarking graph: h_to_era fwd+bwd ===

rms_norm then pyg (non-compiled):

Compiled rms_norm x pyg:
OptimizedModule
        elapsed time: 52.97 ms / iter
        peak memory usage: 19149.75 MB


 standalone Triton GraphTransformer:
apply
        elapsed time: 13.76 ms / iter
        peak memory usage: 14724.34 MB


 compiled rms norm then standalone Triton GraphTransformer:
apply
        elapsed time: 19.84 ms / iter
        peak memory usage: 15863.98 MB


 fused rms_norm x Triton GraphTransformer:
apply
        elapsed time: 14.39 ms / iter
        peak memory usage: 14724.31 MB



=== Benchmarking graph: dop_enc_large fwd+bwd ===

rms_norm then pyg (non-compiled):

Compiled rms_norm x pyg:
OptimizedModule
        elapsed time: 44.08 ms / iter
        peak memory usage: 16828.71 MB


 standalone Triton GraphTransformer:
apply
        elapsed time: 13.04 ms / iter
        peak memory usage: 13126.87 MB


 compiled rms norm then standalone Triton GraphTransformer:
apply
        elapsed time: 19.45 ms / iter
        peak memory usage: 14955.59 MB


 fused rms_norm x Triton GraphTransformer:
apply
        elapsed time: 12.06 ms / iter
        peak memory usage: 13126.84 MB
```


models/src/anemoi/models/schemas/encoder.py
```diff
@@ -32,8 +32,16 @@ class GraphTransformerEncoderSchema(TransformerModelComponent):
     "Size of trainable parameters vector. Default to 8."
     sub_graph_edge_attributes: list[str] = Field(examples=["edge_length", "edge_dirs"])
     "Edge attributes to consider in the encoder features."
-    qk_norm: bool = Field(example=False)
-    "Normalize the query and key vectors. Default to False."
+    qk_norm: Union[str, bool] = Field(examples=["None", "layer_norm", "rms_norm", True, False])
+    "Normalize the query and key vectors. Options are 'None', 'layer_norm', 'rms_norm', True, False. Bools are supported for backward compatibility. If True, 'layer_norm' is used for normalization."
+
+    # convert qk_norm from bool to str for backward compatibility
```
cathalobrien: this doesn't seem to work when config_validation is True, I still get 'False' passed down as qk_norm, where it should be converted to 'None'. Any ideas @anaprietonem ?

cathalobrien: changing to mode='after' seems to get it working

anaprietonem: Could we just pass null to be consistent with other parts of the config? that'd should be equivalent to 'none'

cathalobrien: good idea, I'll try that. thanks :)

models/src/anemoi/models/layers/block.py
```diff
@@ -531,11 +530,25 @@ def __init__(
...
+                "rms_norm",
+                "layer_norm",
+            ], f"Invalid qk_norm {self.qk_norm} for triton backend. Valid options are 'none', 'rms_norm' and 'layer_norm'."
+
+            self.conv = GraphTransformer(self.out_channels_conv, self.qk_norm)
         else:
             self.conv = GraphTransformerConv(out_channels=self.out_channels_conv)
 
+        # Triton graph attention has a fused QK norm which will be used instead to perform QK normalisation
+        if self.qk_norm != "none" and self.graph_attention_backend != "triton":
```
japols: maybe add this to the else path just above?

models/src/anemoi/models/layers/block.py
```diff
@@ -808,7 +821,7 @@ def forward(
         else:
             query, key, value, edges = self.prepare_qkve_edge_sharding(query, key, value, edges)
 
-        if self.qk_norm:
+        if self.qk_norm != "none" and self.graph_attention_backend != "triton":
```
japols: can we set qk_norm to None in the non-fused path in `__init__` and check for `is not None` here?

models/src/anemoi/models/layers/block.py
```diff
@@ -918,7 +931,8 @@ def forward(
 
         query, key, value, edges = self.shard_qkve_heads(query, key, value, edges, shapes, batch_size, model_comm_group)
 
-        if self.qk_norm:
+        # Triton graphtransformer has a fused qk_norm option which is used instead
+        if self.qk_norm != "none" and self.graph_attention_backend != "triton":
```
japols: same here

models/src/anemoi/models/triton/gt.py
```diff
@@ -388,14 +405,15 @@ def __init__(self):
             )
 
     @staticmethod
-    def forward(ctx, q, k, v, e, csc, reverse):
+    def forward(ctx, q, k, v, e, csc, reverse, qk_norm):
```
japols, line 408: can we make this a string here instead for readability that defaults to no norm? 

I think the hardcoding of constants `0=no_norm, 1=rms_norm, 2=ln` is fine for triton but for this part I would stick to strings

models/src/anemoi/models/triton/gt.py
```diff
@@ -440,14 +451,74 @@ def backward(ctx, d_out):
...
         )
 
-        return dQ, dK, dV, dE, None, None
+        return dQ, dK, dV, dE, None, None, None
+
+
+class GraphTransformer(torch.nn.Module):
+    def __init__(self, dim: int, qk_norm: str = "none"):
+        super().__init__()
+        self.dim = dim
```
japols, line 505: probably remove self.dim as we don't use it anywhere

models/src/anemoi/models/triton/norm.py
```diff
@@ -0,0 +1,451 @@
...
+def _norm_bwd_standalone(
+    x_ptr,
+    dout_ptr,
+    w_ptr,
+    dx_ptr,
+    dw_partial_ptr,
+    H: tl.constexpr,
+    C: tl.constexpr,
+    norm_type: tl.constexpr,
+    elementwise_affine: tl.constexpr,
```
japols: do we need to support elementwise_affine in general? if yes, why not also a trainable bias? I think this adds a lot of code that is currently unused (as it's not supported for the fused qk norm in triton/gt.py)

models/src/anemoi/models/triton/norm.py
```diff
@@ -0,0 +1,451 @@
...
+
+    acc = tl.zeros((C_pad,), dtype=tl.float32)
+    for curr in range(0, N):
+        acc += tl.load(
+            dw_partial_ptr + curr * C + C_pad_off,
+            mask=C_mask,
+        )
+
+    tl.store(dw_ptr + C_pad_off, acc, mask=C_mask)
+
```
japols, line 241: I think we could potentially remove everything below this line and also the elementwise_affine part above (`w, dw` etc)

japols (changes requested):
Very nice work and speedups! :) Also added some nitpick comments.

models/src/anemoi/models/layers/block.py
cathalobrien: good idea

models/src/anemoi/models/triton/norm.py
cathalobrien: done

models/src/anemoi/models/triton/norm.py
cathalobrien: yeah I ripped it out

models/src/anemoi/models/layers/block.py
cathalobrien: good idea, I replaced it with a bool

japols (changes requested):
Seems like we actually do want the fused LN to support elementwise_affine via trainable parameters (also bias probably?) to match the behaviour of standard LayerNorms. Should be possible by registering parameters in the torch.nn.Module wrapper of the triton GT and manually computing the gradient update.

anaprietonem:
As discussed with @cathalobrien we are closing this PR for now as more work is required to match the standard LayerNorms. This is still a useful feature for Anemoi, so once that work is finished a new PR is welcome.

PR #771, opened 2025-12-20
test: Add online dry run training workflow test

jjlk:
## Description
This PR adds an integration test to improve the robustness against code changes. It runs a training configured to use a MLFlow dry run. It is particularly useful for training workflow orchestration.

The code to generate an MLFlow run ID has been improved to be able to be use also in tests.

Test the sequence:
- Generate a MLFLow run ID
- Run a training using this run ID

In addition it tests also:
- Log metrics in the MLFLow test server

Training workflow orchestration. 

##  Additional notes ##
My test code is pretty raw. The test uses only the GNN configuration which is arbitrary.

### DIFF
--- a/training/tests/integration/conftest.py
+++ b/training/tests/integration/conftest.py
@@ -393,6 +393,13 @@ def architecture_config_with_checkpoint(
 ) -> tuple[OmegaConf, str]:
     # Reuse the same overrides that architecture_config gets
     overrides = request.param
+
+    # ✅ Skip ONLY gnn on Python 3.10
+    import sys
+
+    if sys.version_info[:2] == (3, 10) and any("model=gnn" in o for o in overrides):
+        pytest.skip("GNN checkpoint incompatible with Python 3.10")
+
     cfg, dataset_url, model_architecture = build_architecture_config(
         overrides,
         testing_modifications_with_temp_dir,
@@ -401,7 +408,7 @@ def architecture_config_with_checkpoint(
     # rest of your logic...
     if "gnn" in model_architecture:
         existing_ckpt = get_test_data(
-            "anemoi-integration-tests/training/checkpoints/testing-checkpoint-gnn-global-2025-07-31.ckpt",
+            "anemoi-integration-tests/training/checkpoints/testing-checkpoint-gnn-global-2026-01-12.ckpt",
         )
     elif "graphtransformer" in model_architecture:
         existing_ckpt = get_test_data(
--- a/models/docs/introduction/overview.rst
+++ b/models/docs/introduction/overview.rst
@@ -104,6 +104,20 @@ process the data on the hidden grid. The `Processors` use a series of
 checkpointed chunks via `num_chunks` to reduce memory usage during
 training.
 
+Graph Providers
+===============
+
+Graph providers encapsulate the logic for supplying edge indices and
+attributes to mapper and processor layers. This separation allows for
+different graph types:
+
+-  **StaticGraphProvider**: For fixed graph structures with optional
+   trainable edge parameters
+-  **NoOpGraphProvider**: For edge-less architectures (e.g., pure
+   Transformers)
+-  **DynamicGraphProvider**: For on-the-fly graph construction (future)
+-  **ProjectionGraphProvider**: For sparse projection matrices
+
 **************
  Data Indices
 **************
--- a/models/docs/modules/layers.rst
+++ b/models/docs/modules/layers.rst
@@ -74,6 +74,15 @@ This environment variable controls the number of chunks used in the
    :no-undoc-members:
    :show-inheritance:
 
+*****************
+ Graph Providers
+*****************
+
+.. automodule:: anemoi.models.layers.graph_provider
+   :members:
+   :no-undoc-members:
+   :show-inheritance:
+
 ******
  Conv
 ******
--- a/models/docs/usage/create_model.rst
+++ b/models/docs/usage/create_model.rst
@@ -203,15 +203,23 @@ In this example, ``model_interface.model`` is the following:
 .. code:: python
 
    AnemoiModelEncProcDec(
-     (encoder): GraphTransformerForwardMapper(
+     (encoder_graph_provider): StaticGraphProvider(
        (trainable): TrainableTensor()
+     )
+     (encoder): GraphTransformerForwardMapper(
        (proc): GraphTransformerMapperBlock(
          (lin_key): Linear(in_features=1024, out_features=1024, bias=True)
          ...
        )
      )
+     (processor_graph_provider): StaticGraphProvider(
+       (trainable): TrainableTensor()
+     )
      (processor): TransformerProcessor(
        ...
+     )
+     (decoder_graph_provider): StaticGraphProvider(
+       (trainable): TrainableTensor()
      )
      (decoder): GraphTransformerBackwardMapper(
        (proc): GraphTransformerMapperBlock(
@@ -219,6 +227,11 @@ In this example, ``model_interface.model`` is the following:
          ...
      )
    )
+
+Note that each encoder, processor, and decoder has a corresponding
+``*_graph_provider`` that manages the graph edges and trainable edge
+parameters. The graph providers supply edge attributes and indices to
+their corresponding mappers/processors during the forward pass.
 
 .. _layer-kernels:
 
--- a/models/src/anemoi/models/distributed/khop_edges.py
+++ b/models/src/anemoi/models/distributed/khop_edges.py
@@ -100,6 +100,44 @@ def sort_edges_1hop_sharding(
     return edge_attr, edge_index, [], []
 
 
+def shard_edges_1hop(
+    edge_attr: Tensor,
+    edge_index: Adj,
+    src_size: int,
+    dst_size: int,
+    model_comm_group: Optional[ProcessGroup],
+) -> tuple[Tensor, Adj, tuple[list[torch.Size], list[torch.Size]]]:
+    """Sort and shard edges for 1-hop sharding.
+
+    Parameters
+    ----------
+    edge_attr : Tensor
+        Edge attributes
+    edge_index : Adj
+        Edge index
+    src_size : int
+        Number of source nodes
+    dst_size : int
+        Number of destination nodes
+    model_comm_group : ProcessGroup, optional
+        Model communication group
+
+    Returns
+    -------
+    tuple[Tensor, Adj, tuple[list[torch.Size], list[torch.Size]]]
+        Sharded edge_attr, sharded edge_index, and tuple of (shapes_edge_attr, shapes_edge_idx)
+    """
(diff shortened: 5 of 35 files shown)


training/tests/integration/test_training_cycle.py
```diff
@@ -230,3 +231,29 @@ def test_training_cycle_diffusion(diffusion_config: tuple[DictConfig, str], get_
 def test_config_validation_diffusion(diffusion_config: tuple[DictConfig, str]) -> None:
     cfg, _ = diffusion_config
     BaseSchema(**cfg)
+
+
+@skip_if_offline
+@pytest.mark.slow
+def test_training_cycle_dry_run(gnn_config: tuple[DictConfig, str], get_test_archive: GetTestArchive) -> None:
```
VeraChristina: Can we call the test something with MLflow? Like `test_mlflow_dry_run` or `test_training_cycle_mlflow_dry_run`

training/tests/integration/test_training_cycle.py
```diff
@@ -230,3 +231,29 @@ def test_training_cycle_diffusion(diffusion_config: tuple[DictConfig, str], get_
...
+    cfg, url = gnn_config
+
+    # Override graph attribute
+    cfg["system"]["input"]["graph"] = "existing"
```
VeraChristina: Why do we need this? As in, should this go in the use case modifications for the gnn config; or is this related to MLflow somehow?

training/tests/integration/test_training_cycle.py
```diff
@@ -230,3 +231,29 @@ def test_training_cycle_diffusion(diffusion_config: tuple[DictConfig, str], get_
...
+
+    # Override config for MLFlow logging using an dry run ID
+    cfg["diagnostics"]["log"]["mlflow"]["enabled"] = True
+    cfg["diagnostics"]["log"]["mlflow"]["tracking_uri"] = "https://mlflow-test.ecmwf.int"
+    cfg["diagnostics"]["log"]["mlflow"]["offline"] = False
```
VeraChristina: Does this config setting work if no access to an MLflow server is configured? If not, I would suggest adding a marker `mlflow` or `hpc` (like we did [here](https://github.com/ecmwf/anemoi-core/blob/e6b7dbb3b200a4117fd51de02b201e6669414317/training/tests/conftest.py#L54) for tests requiring multiple gpus), such that this test is skipped by default and can be manually enabled for the ci on the hpc runners by adding a respective flag `--mlflow` or `--hpc`

training/src/anemoi/training/commands/mlflow.py
```diff
@@ -140,11 +209,6 @@ def add_arguments(command_parser: argparse.ArgumentParser) -> None:
             type=Path,
             help="Output file path.",
         )
-        prepare.add_argument(
-            "--verbose",
-            "-v",
-            action="store_true",
-        )
```
VeraChristina, line 147: Why remove this? Seems unrelated to the PR

training/tests/integration/test_training_cycle.py
VeraChristina: Also, for this to run in the ci, do we need any set up steps in the workflow, e.g. for authentication to the MLflow server?

jjlk: The HPC user has a cron job that runs the authentication every 15 days. So it can access the MLFlow server

training/src/anemoi/training/commands/mlflow.py
jjlk: This option was not used. Minor clean up!

training/tests/integration/test_training_cycle.py
jjlk: Added to the use case modifications

training/tests/integration/test_training_cycle.py
jjlk: To make the test more exhaustive I added a step for the MLFlow authentication

training/tests/integration/config/test_config.yaml
```diff
@@ -2,6 +2,7 @@
 system:
   input:
     dataset: anemoi-integration-tests/training/datasets/aifs-ea-an-oper-0001-mars-o48-1979-19-6h-v6-testset.zarr
+    graph: first_graph_o48.pt
```
anaprietonem: Was this needed? I think we didn't provide a graph on purpose so the graph would generated every time. If this is failing it might be a bug unrelated to this PR that needs fixing

training/tests/integration/test_training_cycle.py
```diff
@@ -230,3 +230,33 @@ def test_training_cycle_diffusion(diffusion_config: tuple[DictConfig, str], get_
...
+@skip_if_offline
+@pytest.mark.slow
+@pytest.mark.mlflow
+def test_training_cycle_mlflow_dry_run(gnn_config: tuple[DictConfig, str], get_test_archive: GetTestArchive) -> None:
+    from anemoi.training.commands.mlflow import prepare_mlflow_run_id
+    from anemoi.utils.mlflow.auth import TokenAuth
+
+    cfg, url = gnn_config
+
+    # Override config for MLFlow logging using an dry run ID
```
anaprietonem: Some of these settings could go into it's own fixture, so if we want to reuse them in the future it's easier. I would also suggest to hardcode the experiment name to something like 'ci_hpc_mlflow'?

training/tests/integration/test_training_cycle.py
```diff
@@ -230,3 +230,37 @@ def test_training_cycle_diffusion(diffusion_config: tuple[DictConfig, str], get_
...
+    cfg["diagnostics"]["log"]["mlflow"]["enabled"] = True
+    cfg["diagnostics"]["log"]["mlflow"]["tracking_uri"] = mlflow_server
+    cfg["diagnostics"]["log"]["mlflow"]["offline"] = False
+
+    # Log in and acquire a token from keycloak
+    TokenAuth(url=cfg.diagnostics.log.mlflow.tracking_uri).login()
```
gmertes: login should not be called from the CI, only from interactive sessions. The token will already be available on the system. If for whatever reason it's not, login will ask for user input and the job will hang indefinitely. We want it to crash instead (which the AnemoiMLflowClient and AnemoiMLflowLogger will do if the token is missing).

training/tests/conftest.py
```diff
@@ -59,17 +61,33 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+        dest="mlflow",
+        default=False,
+        help="enable tests marked as requiring MLFlow test server",
+    )
+    parser.addoption(
+        "--mlflow-server",
+        dest="mlflow_server",
+        default=None,
+        help="MLFlow server for tests requiring MLFlow (only if --mlflow is set)",
+    )
```
gmertes, line 81: Can you make it crash if --mlflow is set but the server is empty ?

training/tests/conftest.py
```diff
@@ -59,17 +61,33 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
 
 
 def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item]) -> None:
     """Automatically skip @pytest.mark.multigpu tests unless --multigpu is used."""
-    if not config.getoption("--multigpu"):
-        skip_marker = pytest.mark.skip(reason="Skipping tests requiring multipe GPUs, use --multigpu to enable")
-        for item in items:
-            if item.get_closest_marker("multigpu"):
-                item.add_marker(skip_marker)
+    for option_name in ["multigpu", "mlflow"]:
```
gmertes: nit: Maybe worth it to move this list to a constant value at the top of the file? (also update the docstring)

training/tests/integration/test_training_cycle.py
gmertes: +1 on the hardcoded experiment name, or in the config?

training/tests/integration/conftest.py
```diff
@@ -482,3 +482,15 @@ def diffusion_config(
     cfg = OmegaConf.merge(template, testing_modifications_callbacks_on_with_temp_dir, use_case_modifications)
     OmegaConf.resolve(cfg)
     return cfg, dataset_urls[0]
+
+
+@pytest.fixture
+def mlflow_dry_run_config(gnn_config: tuple[DictConfig, str], mlflow_server: str) -> tuple[DictConfig, str]:
+    cfg, url = gnn_config
+
+    cfg["diagnostics"]["log"]["mlflow"]["enabled"] = True
```
gmertes, line 497: Is `authentication` already True in the gnn_config? If not, it should also be set here.

training/tests/conftest.py
jjlk: I have this now:
```
python -m pytest -v test_training_cycle.py --slow -k "test_training_cycle_mlflow_dry_run" --mlflow
============================================================================================================================================================================================================= short test summary info ==============================================================================================================================================================================================================
ERROR test_training_cycle.py::test_training_cycle_mlflow_dry_run - ValueError: MLFlow server must be provided via --mlflow-server when using --mlflow
=================================================================================================================================================================================================== 29 deselected, 2 warnings, 1 error in 0.91s ====================================================================================================================================================================================================
```

training/tests/conftest.py
jjlk: Done

training/tests/integration/config/test_config.yaml
jjlk: If I do not add the option, I get the following error:
```
FAILED test_training_cycle.py::test_training_cycle_mlflow_dry_run - omegaconf.errors.MissingMandatoryValue: Missing mandatory value: system.input.graph
```
when I resolve the configuration in `prepare_mlflow_run_id`

training/tests/integration/conftest.py
jjlk: Shall I changed the default src/anemoi/training/config/diagnostics/evaluation.yaml file?
```
cat  src/anemoi/training/config/diagnostics/evaluation.yaml
log:
  wandb:
    ...
  tensorboard:
    ...
  mlflow:
    enabled: False
    _target_: anemoi.training.diagnostics.mlflow.logger.AnemoiMLflowLogger
    offline: False
    authentication: False
```
?

jjlk: the gnn config is based on this config.

gmertes: I think False in the config makes sense, and only set it to True in the fixture

training/tests/integration/config/test_config.yaml
anaprietonem: That seems then a bug introduced in the system config refactor and should be fixed outside of this PR 😅. Should be a   'small' thing. In general I don't think we want to pass the graph path but rather recompute this every time.

training/tests/integration/config/test_config.yaml
```diff
@@ -2,9 +2,16 @@
...
```
anaprietonem:
```suggestion
    graph: null
```

PR #782, opened 2026-01-07
feat: Reconstruction plots callback

icedoom888:
## Description
Introducing focus area to plot only specific regions.
Introducing reconstruction plots to compare input and output and their difference.

----
📚 Documentation preview 📚: https://anemoi-training--782.org.readthedocs.build/en/782/

----
📚 Documentation preview 📚: https://anemoi-graphs--782.org.readthedocs.build/en/782/

----
📚 Documentation preview 📚: https://anemoi-models--782.org.readthedocs.build/en/782/

### DIFF
--- a/training/src/anemoi/training/diagnostics/callbacks/plot.py
+++ b/training/src/anemoi/training/diagnostics/callbacks/plot.py
@@ -1191,7 +1191,6 @@ class PlotSample(BasePlotAdditionalMetrics):
             )
 
 
-
 class PlotReconstruction(BasePlotAdditionalMetrics):
     """Plots a post-processed sample: input, reconstruction and error. Used in Autoencoder training."""
 


training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -100,6 +117,23 @@ class LongRolloutPlotsSchema(BaseModel):
     "Delay between frames in the animation in milliseconds, by default 400."
     colormaps: dict[str, ColormapSchema] | None = Field(default=None)
     "List of colormaps to use, by default None."
+    focus_area: FocusAreaSchema | None = Field(default=None)
+    "Region of interest to restrict plots to, specified by 'spatial_mask' or 'latlon_bounds'."
+
+
+class FocusAreaSchema(BaseModel):
```
JPXKQX: Isn't this repeated ? (L36-59)

icedoom888: You are right! Just fixed it ;) thanks for looking!

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -605,6 +605,257 @@ def plot_flat_sample(
...
+      tall/narrow or crosses the dateline, you may want to choose the
+      ``central_longitude`` or ``standard_parallels`` explicitly.
+    - Input is not validated; ensure ``latlon`` has at least two points and a
+      non-zero latitude span for meaningful standard parallels.
+    """
+    try:
+        import cartopy.crs as ccrs
+
+    except ModuleNotFoundError as e:
+        error_msg = "Module cartopy not found. Install with optional-dependencies.plotting."
```
anaprietonem: Nice compromise, just the change to add this to the pyproject.toml is not part of this PR?

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -605,6 +605,257 @@ def plot_flat_sample(
...
+    ModuleNotFoundError
+        If ``cartopy`` is not installed. Install via the
+        ``optional-dependencies.plotting`` extra.
+
+    Notes
+    -----
+    - This heuristic works well for many regional plots. If your domain is very
+      tall/narrow or crosses the dateline, you may want to choose the
+      ``central_longitude`` or ``standard_parallels`` explicitly.
+    - Input is not validated; ensure ``latlon`` has at least two points and a
```
anaprietonem: could be worth adding some asserts for this?

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -605,6 +605,257 @@ def plot_flat_sample(
             )
 
 
+def lambert_conformal_from_latlon_points(latlon: np.ndarray) -> object:
```
anaprietonem: `training/src/anemoi/training/diagnostics/maps.py` contains utility classes and functions related to projections. I thin this function could be moved there so `plots.py` just contains strictly plotting functions.

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -55,17 +56,22 @@
 class BasePlotCallback(Callback, ABC):
     """Factory for creating a callback that plots data to Experiment Logging."""
 
-    def __init__(self, config: BaseSchema) -> None:
+    def __init__(self, config: BaseSchema, focus_area: dict | None = None) -> None:
```
anaprietonem, line 59: if the focus_area is defined via the config, why not extracting it from there rather than passing it to this class in the init?

training/docs/modules/diagnostics.rst
```diff
@@ -109,28 +127,31 @@ which is recommended for interactive terminals and
...
-            # one-parameter groups are possible to highlight individual parameters
-            parameter_groups:
-               moisture: [tp, cp, tcw]
-               sfc_wind: [10u, 10v]
+           # group parameters by categories when visualizing contributions to the loss
+           parameter_groups:
+              moisture: [tp, cp, tcw]
+              sfc_wind: [10u, 10v]
+           # Example focusing loss on a predefined mask
+           focus_area:
```
anaprietonem, line 148: probably I am being slow, but is this then a key you pass to 'graph config' but then you reference here or how does it work?

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -677,10 +933,27 @@ def single_plot(
             ax=ax,
         )
 
-    xmin, xmax = max(lon.min(), -np.pi), min(lon.max(), np.pi)
-    ymin, ymax = max(lat.min(), -np.pi / 2), min(lat.max(), np.pi / 2)
-    ax.set_xlim((xmin - 0.1, xmax + 0.1))
-    ax.set_ylim((ymin - 0.1, ymax + 0.1))
+    if transform is not None:
```
anaprietonem, line 876: some of this seems to be could be better abstracted and integrated into `training/src/anemoi/training/diagnostics/maps.py`

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -677,10 +933,27 @@ def single_plot(
...
+
+        if hasattr(ax, "add_feature"):
+            ax.add_feature(cfeature.COASTLINE.with_scale("50m"), zorder=1, alpha=0.8)
+            ax.add_feature(cfeature.BORDERS.with_scale("50m"), linestyle=":", zorder=1)
+        else:
+            # If it's a regular Axes, add_feature doesn't exist
+            LOGGER.warning("Axis is not a GeoAxes; skipping cartopy features.")
+
+    except ModuleNotFoundError:
+        LOGGER.warning("Module cartopy not found. Coastlines and borders will not be plotted.")
```
anaprietonem, line 895: Coastlines are being plotted below 
`continents.plot_continents(ax)`

that's why my comment above about trying to harmonise this with the introduction of cartopy

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -605,6 +605,257 @@ def plot_flat_sample(
...
+    reconstruction: np.ndarray,
+    difference: np.ndarray,
+    vname: str,
+    clevels: float,
+    datashader: bool = False,
+    precip_and_related_fields: list | None = None,
+    cmap: Colormap | None = None,
+    error_cmap: Colormap | None = None,
+) -> None:
+    """Plot one variable's input, reconstruction, and error map on a flat lat/lon grid.
```
anaprietonem, line 715: Is it fair to say this is a special case for the plot_predicted_multilevel_flat_sample where we don't have a target but rather the input and the pred? so it would like the plotting a diagnostic variable but where target is input?

Thinking if would dramatically reduce the amount of code if we could generalise the `plot_predicted_multilevel_flat_sample` to work with that case

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -84,6 +90,35 @@ def __init__(self, config: BaseSchema) -> None:
...
+            self.latlons = np.rad2deg(self.latlons.cpu().numpy())
+
+        # Compute focus mask
+        focus_mask = np.ones(self.latlons.shape[0], dtype=bool)
+        self.tag = None
+
+        if self.focus_area is not None:
+            if self.focus_area["spatial_mask"] is not None:
+                focus_mask = np.zeros(self.latlons.shape[0], dtype=bool)
+                spatial_mask_idxs = pl_module.model.graph_data["data"][self.focus_area["spatial_mask"]]
```
anaprietonem: could have an assert and raise error if the given spatial mask doesn't exist in graph_data?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1160,6 +1357,14 @@ def _plot(
         local_rank = pl_module.local_rank
         data, output_tensor = self.process(pl_module, outputs, batch, output_times)
 
+        # Compute focus mask
+        focus_mask = self.get_focus_mask(pl_module)
```
anaprietonem, line 1366: These needs  to be repeated for all callbacks - we should have a function to clip the data as part of the basePlot and then reuse it

training/src/anemoi/training/diagnostics/plots.py
icedoom888: Mmm weird, added it back :D

training/src/anemoi/training/diagnostics/plots.py
icedoom888: Done ✅

training/src/anemoi/training/diagnostics/plots.py
icedoom888: DOne!

training/src/anemoi/training/diagnostics/callbacks/plot.py
icedoom888: You will probably have a list of different plots in the config, each with it's own focus area (you might want to plot full europe as well as your LAM or the global too). Each plot object gets initialised independently, so focus area behaves exactly like other parameters like "every_n_batches" which are passed in the init of the specific plot rather than the config itself.

training/docs/modules/diagnostics.rst
icedoom888: Yes exactly! Eg. in stretched grid, you have the "cutout_mask" defined during graph creation, if you pass it here, it will plot only that region ;)

training/src/anemoi/training/diagnostics/plots.py
icedoom888: Removed coastlines and kept borders

training/src/anemoi/training/diagnostics/plots.py
icedoom888: True, but the function was already quite complex and had many different columns. 
As it is a different layout, I preferred to just create a new one rather than extending this...

training/src/anemoi/training/diagnostics/callbacks/plot.py
icedoom888: Will do!

training/src/anemoi/training/diagnostics/plots.py
icedoom888: Not sure which parts you refer too here?

training/src/anemoi/training/diagnostics/callbacks/plot.py
icedoom888: What do you mean? various callbacks can have different focus_areas, you won't know until you reach every callback plot, no?

JPXKQX:
Hi Alberto, first of all, thanks for the PR. I think this focus_area functionality can be very useful for many users. Below are a few thoughts and suggestions.

- It seems that this PR contains two distinct contributions: a) the introduction of the focus_area argument, b) the reconstruction maps functionality. As the latter depends on the autoencoder PR being merged first, it might be worth splitting this into two separate PRs, so that the focus_area feature can be merged and made available as soon as possible.

- The need for a focus_area has already been raised by users, so I agree it would be a very valuable. Regarding implementation, one possible improvement would be to move this functionality into a dedicated class in a new file. In this setup, the focus_area could be created in the __init__ method (possibly via a factory method) and then applied in forward(). This could help keep the logic more modular and easier to test.

- A few open questions around naming, where consistency with existing code might help:

  - `focus_area`. In anemoi-graphs, we usually refer to an area/region of interest. It might be worth considering whether to keep a similar name for coherence (e.g. area_of_interest) or to deliberately use a different term to distinguish concepts. Not sure what is preferred.
  - `spatial_mask` naming. In the existing configs, we typically refer to this kind of object as `node_attribute_name` or `mask_attr_name`
  -  `latlon_bounds`. We already have a similar argument in graphs called `area`. While that name may be a bit generic, common alternatives in other libraries include: extent (used in matplotlib), bbox (bounding box), clip, ... Something more explicit like `latlon_bbox`, which specifies that it has to be lat/lon, could also work well.

- In anemoi-graphs and anemoid-dataset, the `area` (`latlon_bounds`) is usually stored as a flat tuple of four values rather than a tuple of tuples. This also seems consistent with many other python packages

These comments are mainly intended to raise discussion rather than to suggest strict changes. Happy to hear your thoughts and suggestions. Thanks again for your contribution!

training/src/anemoi/training/diagnostics/plots.py
anaprietonem: thank you!

training/src/anemoi/training/diagnostics/plots.py
anaprietonem: I replied on the other PR, let me know if it's unclear

training/src/anemoi/training/diagnostics/plots.py
anaprietonem: The issue that I see at the moment is that projection and transform are specific and will just work if one has cartopy installed. So I would like to find a way we can define this and reuse it for the other plots so that we reuse the functionality we currently have but with the option of customising the projections and the transform.

anaprietonem: If we don't want to make it completely configurable from the config I think a clearer design would be something like 

```
@dataclass(frozen=True)
class MapContext:
    """Encapsulates map projection, data CRS, and extent handling."""
    projection: object | None
    data_crs: object | None
    pad: float = 0.1

    @classmethod
    def from_latlons(
        cls,
        latlons: np.ndarray,
        *,
        datashader: bool = False,
        pad: float = 0.1,
    ) -> MapContext:
        try:
            import cartopy.crs as ccrs
        except ModuleNotFoundError:
            return cls(projection=None, data_crs=None, pad=pad)

        if datashader:
            return cls(projection=None, data_crs=None, pad=pad)

        projection = lambert_conformal_from_latlon_points(latlons)
        data_crs = ccrs.PlateCarree()

        return cls(projection=projection, data_crs=data_crs, pad=pad)

    def set_extent(
        self,
        ax,
        lon: np.ndarray,
        lat: np.ndarray,
    ) -> None:
        """Set axis extent consistently for projected or flat axes."""
        if self.data_crs is not None and hasattr(ax, "set_extent"):
            ax.set_extent(
                [
                    lon.min() - self.pad,
                    lon.max() + self.pad,
                    lat.min() - self.pad,
                    lat.max() + self.pad,
                ],
                crs=self.data_crs,
            )
        else:
            xmin, xmax = max(lon.min(), -np.pi), min(lon.max(), np.pi)
            ymin, ymax = max(lat.min(), -np.pi / 2), min(lat.max(), np.pi / 2)

            ax.set_xlim(xmin - self.pad, xmax + self.pad)
            ax.set_ylim(ymin - self.pad, ymax + self.pad)
```

and we then could pass this to the plots like

```
from map_context import MapContext

map_ctx = MapContext.from_latlons(
    latlons,
    datashader=datashader,
)

fig, ax = plt.subplots(
    n_vars,
    n_plots_per_sample,
    figsize=(n_plots_per_sample * 4, n_vars * 3),
    layout="constrained",
    subplot_kw={"projection": map_ctx.projection},
)
```
so the MapContext will handle all the logic and code related to cartopy

anaprietonem: Also some of those settings could be defined at the level of callback ie if you have a specific `AutoencoderPlotSample` then there we can define those and try to reuse the `plot_predicted_multilevel_flat_sample`

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1085,13 +1153,21 @@ def _plot(
 
         local_rank = pl_module.local_rank
 
+        # Compute focus mask
+        focus_mask = self.get_focus_mask(pl_module)
+
         for rollout_step in range(output_times[0]):
             init_step = self._get_init_step(rollout_step, output_times[1])
 
+            # Apply spatial mask
```
anaprietonem, line 1162: this clipping could be one outside of the loop so then it's reused for all rollout_steps right?

training/src/anemoi/training/diagnostics/plots.py
anaprietonem: On the topic of creating a new function vs using the existing one, from what I see it's mostly the definition of the projection and the fact that for the autoencoder you just one 3 subplots, vs the forecasting one includes 6 subplots right? So I think this could be configured to define this at the callback level and then reuse the functions that we have. 

I understand having the functions separated might make it easier to understand but bugs and features in the future would then need to be copied across both of them, making dev workflow more cumbersome so I would prefer we homogenise it.

training/src/anemoi/training/diagnostics/callbacks/plot.py
icedoom888: Done!

PR #729, opened 2025-12-08
feat(models): Spherical Harmonics Transform(s) - Octahedral & HEALPix Grids

PortillaS-Predictia:
## Description
Draft, to be coordinated with https://github.com/ecmwf/anemoi-core/pull/678 and https://github.com/ecmwf/anemoi-core/pull/709.

Introduces the groundwork for a "general" Spherical Harmonics Transform. Currently, four grids are supported: Gaussian, Equiangular, Octahedral and HEALPix (inverse). This approach can be extended to any "isolatitudinal grid" (i.e., a grid in which the nodes are distributed along rings of constant latitude). It can't, however, be extended to other grids, such as the "triangular refined icosahedron".

The "base" transform, Gaussian and Equiangular (exact quadrature), is from [torch-harmonics](https://github.com/NVIDIA/torch-harmonics) ([arxiv](https://arxiv.org/abs/2306.03838)). The extensions to both Octahedral and HEALPix (inexact) were inspired by [S2FFT](https://astro-informatics.github.io/s2fft/) ([arxiv](https://arxiv.org/abs/2311.14670)).

Basic example - forward transform from a Gaussian grid, followed by an inverse transform to an Octahedral grid.
```
sht_gaussian = CartesianRealSHT(
    nlat=NLAT,
    nlon=NLON,
    grid="legendre-gauss",
)

isht_octahedral = OctahedralInverseRealSHT(
    nlat=(2 * N),
    lmax=sht_gaussian.lmax,
    mmax=sht_gaussian.mmax,
)

spectral = sht_gaussian(spatial_gaussian)
spatial_octahedral = isht_octahedral(spectral)
```

Basic example - three plots of the same field after consecutive forward/inverse transforms to different grids (Gaussian, Octahedral and HEALPix, respectively). The reconstruction error (measured in the spectral domain) is MSE ~1e-14. The field is bandwidth-limited to 16.
gaussian
octahedral
healpix

## What problem does this change solve?

Spectral losses and other "operations" defined on the spectral domain - for instance, the learnable spectral filter in a SFNO-type module ([example](https://github.com/Predictia/anemoi-core/blob/0019972fc74cd18fd368adb98f2ca2bc3e982560/models/src/anemoi/models/layers/sfno.py#L41)).

## What issue or task does this change relate to?

https://github.com/ecmwf/anemoi-core/pull/678, https://github.com/ecmwf/anemoi-core/pull/709 and https://github.com/ecmwf/anemoi-core/issues/599.

##  Additional notes ##

[...]

PR #709, opened 2025-11-28
feat: octhedral spherical harmonic transform - ectrans-based

theissenhelen:
## Description
This PR implements a spherical harmonic transform for octahedral reduced gaussian grids based on conventions used in [ectrans.](https://github.com/ecmwf-ifs/ectrans)

## What issue or task does this change relate to?
Part 1 of  #599 

##  Additional notes ##

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

PR #788, opened 2026-01-08
feat: Consolidate and expand spectral losses

OpheliaMiralles:
## Description
On top of #678, this PR introduces SpectralCRPS, adapt Octahedral losses by @theissenhelen and @PortillaS-Predictia (PRs #709 and #729) to match skeleton introduced in #678 by @frazane. The PR also adds tests for all of the losses. We can decide later which Octahedral loss we want or leave both options. 

New feature/Consolidation

#678 #709 #729

##  Additional notes ##
To be merged after #678 and if/when we decide to harmonise something about #709 and #729.

----
📚 Documentation preview 📚: https://anemoi-training--788.org.readthedocs.build/en/788/

----
📚 Documentation preview 📚: https://anemoi-graphs--788.org.readthedocs.build/en/788/

----
📚 Documentation preview 📚: https://anemoi-models--788.org.readthedocs.build/en/788/


training/src/anemoi/training/losses/spectral.py
```diff
@@ -63,7 +67,13 @@ class SpectralLoss(BaseLoss):
...
-        transform: Literal["fft2d", "sht"] = "fft2d",
+        transform: Literal[
+            "fft2d",
+            "cartesian_sht",
+            "octahedral_sht",
+            "ectrans_octahedral_sht",
+            "dct2d",
+        ] = "fft2d",
         *,
         x_dim: int | None = None,
```
sahahner: x_dim and y_dim should not be here, please, as this is only needed for the 2D spectral loss and therefore be added as additional arguments in the 2D versions.

OpheliaMiralles: I broke something while merging #678 so I need to redo the merge anyway

OpheliaMiralles: wait no this is part of #678 I think maybe you mixed SpectralLoss (which accepts x_dim/y_dim) and SpectralTransform (in models/layers) which doesn't have x_dim y_dim as default dims?

frazane: x_dim and y_dim can still be passed to the 2D transform as `kwargs`, so we don't really need them explicitly in the function signature no?

training/tests/unit/losses/test_loss_function.py
```diff
@@ -385,7 +428,174 @@ def test_fcl_loss() -> None:
     assert loss_total.numel() == 1, "Expected a single aggregated loss value"
 
     wrong = (torch.ones((6, 1, 710 * 640 + 1, 2)), torch.zeros((6, 1, 710 * 640 + 1, 2)))
-    with pytest.raises(einops.EinopsError):
```
frazane, line 403: Tests will fail unless the expected error is specifically `EinopsError`

OpheliaMiralles: Yes should be good now

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -84,13 +88,44 @@ def forward(
...
+            x,
+            "(b e v) y x -> b e y x v",
+            b=b,
+            e=e,
+            v=v,
+        )
+
+
 class SHT(SpectralTransform):
     """Placeholder for Spherical Harmonics Transform."""
```
sahahner:
```suggestion
    """Bsaeclass for Spherical Harmonics Transform."""
```

training/src/anemoi/training/losses/spectral.py
sahahner: thanks for changing this.

training/pyproject.toml
```diff
@@ -94,6 +94,7 @@ optional-dependencies.tests = [
   "pytest-mock>=3",
   "pytest-skip-slow>=0.0.5",
   "sshfs",
+  "torch-dct>=0.1.6",        # required for spectral loss tests
```
anaprietonem: where is this being used? I couldn't spot any direct import but probably I missed something

training/tests/integration/test_training_cycle.py
```diff
@@ -128,6 +128,8 @@ def test_training_cycle_ensemble(ensemble_config: tuple[DictConfig, str], get_te
     AnemoiTrainer(cfg).train()
 
 
+@skip_if_offline
+@pytest.mark.slow
```
anaprietonem: the config validation for the ensemble it's supposed to run as a unit-test, any reason why this is changing here?

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py
```diff
@@ -102,6 +102,7 @@ def assert_metadatakeys(metadata: dict, *metadata_keys: tuple[str, ...]) -> None
 
 
 @typechecked
+@pytest.mark.slow
```
anaprietonem: same comment as before - those should run as unit tests

training/src/anemoi/training/losses/spectral.py
```diff
@@ -246,3 +256,98 @@ def __init__(
             scalers=scalers,
             **kwargs,
         )
+
+
+class SpectralCRPSLoss(SpectralLoss, KernelCRPS):
```
anaprietonem: Is the same as the SpectrCRPSLoss used for the Streched-grid models like Bris? (as in is this something you have tried already with any CRPS model?)

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -7,24 +7,27 @@
...
+from anemoi.models.layers.spectral_helpers import CartesianRealSHT
+from anemoi.models.layers.spectral_helpers import EcTransOctahedralSHTModule
+from anemoi.models.layers.spectral_helpers import OctahedralRealSHT
 from anemoi.training.utils.enums import TensorDim
 
 LOGGER = logging.getLogger(__name__)
 
 
-class SpectralTransform(torch.nn.Module):
+class SpectralTransform(abc.ABC):
```
anaprietonem: Is this motivated by the fact that those 'layers' don't have any trainable parameters so you reckon we shouldn't be using torch.nn.Module?

training/src/anemoi/training/losses/spectral.py
```diff
@@ -94,26 +102,28 @@ def __init__(
...
-            kwargs.setdefault("y_dim", y_dim)
-
         if transform == "fft2d":
             self.transform = FFT2D(**kwargs)
-            # expose dims on the loss (legacy API + tests)
-            self.x_dim = int(kwargs.get("x_dim"))
-            self.y_dim = int(kwargs.get("y_dim"))
-        elif transform == "sht":
-            self.transform = SHT()
+        elif transform == "dct2d":
```
anaprietonem, line 107: It might be earlier, but we would need to decide the config interface for this as some of those statements could be replaced if we decide to use the hydra instantiate pattern we are already using in other areas of anemoi-core. Any thoughts?

@frazane @sahahner  it would be also to know your take here

anaprietonem (commented):
Left a few comments, thanks for opening a PR and looking into this Ophelia

training/pyproject.toml
OpheliaMiralles: In DCT2D spectral transform (models/layers)

training/tests/integration/test_training_cycle.py
OpheliaMiralles: No this can be reverted

training/src/anemoi/training/losses/spectral.py
OpheliaMiralles: Yes should be

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py
OpheliaMiralles:
```suggestion
@typechecked
```

training/tests/integration/test_training_cycle.py
OpheliaMiralles:
```suggestion
```

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -84,13 +88,44 @@ def forward(
...
+        return einops.rearrange(
+            x,
+            "(b e v) y x -> b e y x v",
+            b=b,
+            e=e,
+            v=v,
+        )
+
+
 class SHT(SpectralTransform):
```
radiradev: What is the point of this class? It just creates an additional level of inheritance, all the below classes that inherit from `SHT` can inherit directly from `SpectralTransform`.

training/src/anemoi/training/losses/spectral.py
sahahner: agree, that we should evaluate that eventually.

models/src/anemoi/models/layers/spectral_transforms.py
radiradev: @anaprietonem there are a lot of `torch.nn.Module`-s that don't have trainable params e.g [`torch.nn.modules.activation.Sigmoid`](https://docs.pytorch.org/docs/stable/generated/torch.nn.modules.activation.Sigmoid.html). I feel like these transform classes do something similar, so it might be instructive to keep it.

training/src/anemoi/training/losses/spectral.py
sahahner: This comment is not ready to be resolved yet. 
My previous comment may not have been clear enough.
The octahedral spectral loss still depended on `x_dim` and `y_dim`, which is incorrect and potentially misleading. I have now removed this dependency so that the octahedral spectral losses can be used independently of providing unused spatial grid dimensions. We will also need to adapt the corresponding tests to reflect this change.

models/src/anemoi/models/layers/spectral_transforms.py
OpheliaMiralles: True!

frazane: It was originally a placeholder, waiting for the actual implementation

training/src/anemoi/training/losses/spectral.py
```diff
@@ -246,3 +251,98 @@ def __init__(
...
+        SpectralLoss.__init__(
+            self,
+            transform=transform,
+            x_dim=x_dim,
+            y_dim=y_dim,
+            ignore_nans=ignore_nans,
+            scalers=scalers,
+            **kwargs,
+        )
+        KernelCRPS.__init__(self, ignore_nans=ignore_nans)
```
evenmn: Would it make sense to inherit from AlmostFairKernelCRPS instead of KernelCRPS? In the recent versions of Bris, we use AlmostFairCRPS also for the spectral term. It does not impact the results much, but makes it easier to explain in papers

training/src/anemoi/training/losses/spectral.py
```diff
@@ -246,3 +251,98 @@ def __init__(
...
+        transform: Literal[
+            "fft2d",
+            "dct2d",
+            "cartesian_sht",
+            "octahedral_sht",
+        ] = "fft2d",
+        *,
+        x_dim: int | None = None,
+        y_dim: int | None = None,
+        cutoff_ratio: float | None = None,
```
evenmn: We used the cutoff ratio in an earlier implementation, but I'm not sure if it makes sense to set this to anything else than 1. Reducing this number would effectively filter out the highest frequencies, but I don't see where this is useful

training/src/anemoi/training/losses/spectral.py
```diff
@@ -246,3 +251,98 @@ def __init__(
...
+        cutoff_ratio: float,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Keep only the lowest spectral modes."""
+        if not (0.0 < cutoff_ratio <= 1.0):
+            msg = "cutoff_ratio must be in (0, 1]."
+            raise ValueError(msg)
+
+        n_modes = pred.shape[-2]
+        keep = max(1, int(cutoff_ratio * n_modes))
+        return pred[:, :, :keep, :], tgt[:, :, :keep, :]
```
evenmn: I'm not sure if I understand this part. Are pred and tgt sorted with descending frequency? How do you ensure that when doing 2d transforms (e.g. Fourier)? In my original implementation, setting cutoff=1.0 would correspond to filter out frequencies higher than the Nyquist limit, but I cannot see if that's the case here

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -84,13 +88,44 @@ def forward(
...
+class DCT2D(SpectralTransform):
+    """2D Discrete Cosine Transform."""
+
+    def __init__(self, x_dim: int, y_dim: int) -> None:
+        self.x_dim = x_dim
+        self.y_dim = y_dim
+
+    def __call__(self, data: torch.Tensor) -> torch.Tensor:
+        try:
+            from torch_dct import dct_2d
```
sahahner, line 196: Do we need to include this package in the toml file under a dependency?

OpheliaMiralles: Yes that would be cleaner

training/src/anemoi/training/losses/spectral.py
OpheliaMiralles: I just removed the cutoff factor. If it's not used by anyone, it complicates things for no added value

training/src/anemoi/training/losses/spectral.py
frazane: I proposed it this way because of my profound dislike for hydra's instantiate pattern, but it can definitely still be subject to change.

models/src/anemoi/models/layers/spectral_transforms.py
OpheliaMiralles: I don't know if because we do it someplace else we should do it there too, if this is the only reason

training/src/anemoi/training/losses/spectral.py
OpheliaMiralles: Maybe in a later PR then because I do like the fact that we are not using hydra instantiate for once

models/src/anemoi/models/layers/spectral_transforms.py
frazane: I believe subclassing from `nn.Module` would also enable better integration in torch's computational graph tracking.

sahahner: agree that this should be a torch module. we probably want to use the register buffer functionality.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -81,30 +102,156 @@ def forward(
...
+class OctahedralSHT(SpectralTransform):
+    """SHT on an octahedral reduced grid."""
+
+    def __init__(
+        self,
+        nlat: int,
+        lmax: int | None = None,
+        mmax: int | None = None,
+        folding: bool = False,
+        nodes_slice: tuple[int, int | None] | None = None,
```
sahahner: Do I understand correctly that node_slide is something that we do not need for global grids? Then I think this should be removed from the global methods, please.

OpheliaMiralles: Which ones are the global methods? You mean the OctahedralSHT?

models/src/anemoi/models/layers/spectral_transforms.py
OpheliaMiralles: OK we'll change back to torch module

models/src/anemoi/models/layers/spectral_transforms.py
OpheliaMiralles: Maybe please remove it so I don't pick the wrong methods

sahahner: will remove it from octahedral and cartesian SHTs

models/src/anemoi/models/layers/spectral_transforms.py
sahahner: done. this was already part of the last commit, as otherwise I got device mismatches.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -81,30 +107,129 @@ def forward(
...
+        return einops.rearrange(coeffs, "(b e v) yF xF -> b e yF xF v", b=b, e=e, v=v)
+
+
+class OctahedralSHT(SpectralTransform):
+    """SHT on an octahedral reduced grid."""
+
+    def __init__(
+        self,
+        nlat: int,
+        lmax: int | None = None,
```
sahahner: We need to document these options better in a docstring here. Also, a section in the documentation for the spectral losses would be necessary.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -81,30 +107,129 @@ def forward(
...
+        x = einops.rearrange(data, "b e p v -> (b e v) p")
+        coeffs = self._sht(x)  # complex: (b*e*v, L, M)
+        tmp = einops.rearrange(coeffs, "(b e v) yF xF -> b e yF xF v", b=b, e=e, v=v)
+        return tmp
+
+
+class EcTransOctahedralSHT(SpectralTransform):
+    def __init__(
+        self,
+        truncation: int,
```
sahahner: We need to document these options better in a docstring here. Also, a section in the documentation for the spectral losses would be necessary.

training/src/anemoi/training/losses/spectral.py
```diff
@@ -94,26 +99,26 @@ def __init__(
...
+            # expected additional args: lmax/mmax/folding
+            self.transform = OctahedralSHT(**kwargs)
+        elif transform == "ectrans_octahedral_sht":
+            # expected args: truncation (+ optional dtype, filepath)
+            self.transform = EcTransOctahedralSHT(**kwargs)
         else:
             msg = f"Unknown transform type: {transform}"
             raise ValueError(msg)
 
     def _to_spectral_flat(self, x: torch.Tensor) -> torch.Tensor:
```
sahahner, line 124: For the global losses, we need to make sure that this loss function is not applied on a batch, which is still sharded over the grid dimension. Do I assume correctly that this is also the case for the 2D spectral losses?

evenmn: Yes, when doing the transform, we will need the full fields. I think we should force setting "keep_batch_sharded=False" when dealing with spectral loss

sahahner: There is a flag in the loss function, that enforces to gather the batch before calculating the loss. I can try that option first before forcing the entire model to keep the batch sharded.

sahahner: Just pushed the changes such that the loss function will not be sharded when a spectral loss is used.

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,746 @@
...
+
+    t = np.arange(0, 4 * nside - 1)
+
+    offsets = healpix_rings_offset(t, nside)
+    offsets = np.einsum("t, m -> tm", offsets, np.arange(0, mmax))
+
+    return offsets
+
+
+class HEALPixInverseRealSHT(Module):
```
sahahner: Is there also a forward (so non-inverse) class necessary for this?

PortillaS-Predictia: Hi Sara. I only "implemented" (mostly copied from [S2FFT](https://astro-informatics.github.io/s2fft/)!) the inverse transform, because we only needed as much back then (we used it for a completely different purpose, unrelated to "spectral losses"). We can have the forward transform if we ultimately use this approach (instead of the ecTrans's).

models/src/anemoi/models/layers/spectral_transforms.py
PortillaS-Predictia: Should we harmonize it with the `FFT2D` class? The `apply_filter` argument serves the same (or similar) purpose (compared to the `lmax`/`mmax` arguments here) - "apply low-pass filter to ignore frequencies beyond the Nyquist limit".

Please notice that the `folding` argument isn't even implemented, it will raise `NotImplementedError` (supposedly, it should account for the frequency [extension](https://astro-informatics.github.io/s2fft/api/utility/healpix_ffts.html#s2fft.utils.healpix_ffts.spectral_periodic_extension) and [folding](https://astro-informatics.github.io/s2fft/api/utility/healpix_ffts.html#s2fft.utils.healpix_ffts.spectral_folding), but I don't fully understand it).

models/src/anemoi/models/layers/spectral_helpers.py
sahahner: sounds good. we can leave this comment unresolved until we have figured that out then.

models/src/anemoi/models/layers/spectral_helpers.py
samhatfield, line 1: @sahahner, @PortillaS-Predictia and I had a chat earlier, and we wondered if we should just delete all HEALPix-related stuff from this file. We don't think there's a use case for it yet, and we have our hands full with the other spherical harmonic transform implementations. On top of that, it's currently a "stub" with no wrapper class in spectral_transforms.py, and also only the inverse is implemented so it can't even be used for spectral loss calculations. Shall we get rid of it and potentially resurrect it later when we have a use case?

models/src/anemoi/models/layers/spectral_transforms.py
sahahner: We have decided to finalise this harmonisation when we have decided which method should go into main.

models/src/anemoi/models/layers/spectral_helpers.py
anaprietonem: Yes I think that sounds sensible!

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,746 @@
...
+    Method of computation follows
+    [1] Schaeffer, N.; Efficient spherical harmonic transforms aimed at pseudospectral numerical simulations, G3: Geochemistry, Geophysics, Geosystems.
+    [2] Rapp, R.H.; A Fortran Program for the Computation of Gravimetric Quantities from High Degree Spherical Harmonic Expansions, Ohio State University Columbus; report; 1982; https://apps.dtic.mil/sti/citations/ADA123406.
+    [3] Schrama, E.; Orbit integration based upon interpolated gravitational gradients.
+    """
+
+    return legpoly(mmax, lmax, np.cos(t), norm=norm, inverse=inverse, csphase=csphase)
+
+
+class CartesianRealSHT(Module):
```
samhatfield: What's the meaning of `Real` in this context, and below for `OctahedralRealSHT`? If it's referring to the fact that the inputs must be real valued, I would say this is not necessary to mention as that will always be the case.

samhatfield: Also, following discussion with @PortillaS-Predictia, I would recommend to rename this and the inverse, along with the wrapper `CartesianSHT` to `RegularSHT` to denote the fact that it takes a regular grid (i.e. a latitude-longitude grid with no reduction towards the poles) as input.

models/src/anemoi/models/layers/spectral_helpers.py
sahahner: done

samhatfield:
Two more ideas that would simplify this PR:
1. I have looked at the two implementations of octahedral SHT (`OctahedralRealSHT` and `EcTransOctahedralSHTModule`) and they are essentially the same, algorithmically. The major difference is in how the Legendre polynomials are generated: the former uses a native pytorch approach, whereas the latter retrieves them from the ecTrans library. Therefore I suggest we choose just one of these two rather than implementing both.
  @sahahner and I have compared the computational performance of the two and found that `OctahedralRealSHT` has a slight advantage, so I propose that we delete `EcTransOctahedralSHTModule`. I would still like to have the option to use ecTrans to generate Legendre polynomials, as I believe the algorithm there to be more numerically stable at high resolutions (e.g. O2000 and above) (not confirmed, just a hunch). But that can be done with a later PR.
2. Before merging we also need a generic reduced grid transform in addition to the existing regular grid (`CartesianRealSHT`) and the octahedral grid. This should be a trivial extension of `OctahedralRealSHT`. Then we will be able to handle e.g. the N320 grid of ERA5. I can do this work, but only after 1. is resolved.

models/src/anemoi/models/layers/spectral_helpers.py
PortillaS-Predictia: Should we also remove the "Cartesian" (Regular) transforms and just add `torch-harmonics` as a dependency? These classes are _mostly_ the original code from said library. I only ever added type hints and some minor changes (I also solved a minor bug with some `dtypes`, that's why there's a type change in the `forward` method of the inverse transform). However, in the meantime, they also added some other features, like properly guessing the truncation frequency for these Grids (@samhatfield).

models/src/anemoi/models/layers/spectral_helpers.py
PortillaS-Predictia: Yes, `Real` only refers to the reality of the input fields, we can remove it. Renaming the class is also fine by me, but I left another comment related to this ([here](https://github.com/ecmwf/anemoi-core/pull/788#discussion_r2736566330)). Let me know what you think.

models/src/anemoi/models/layers/spectral_helpers.py
sahahner: I think we could possibly remove the cartesian/regular transform, as I am not aware that anyone is actively using these grids. The code would still be available on the closed PR #729. We could also point to `torch-harmonics` in the documentation for implementations for other grids and add the dependency when someone starts working with these grids.

samhatfield: So in fact this PR would only implement octahedral and reduced SHTs?

OpheliaMiralles: And spectral CRPS used in bris models, it’s already a decent expansion

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,228 @@
...
+
+        self.n_grid_points = sum(self.lons_per_lat)
+
+        self.slon = [0] + list(np.cumsum(self.lons_per_lat))[:-1]
+        self.rlon = [nlat + 8 - nlon // 2 for nlon in self.lons_per_lat]
+
+        theta, weight = legendre_gauss_weights(nlat)
+        theta = np.flip(np.arccos(theta))
+
+        pct = precompute_legpoly(self.mmax, self.lmax, theta)
```
samhatfield:
```suggestion
        pct = precompute_legpoly(self.mmax, self.lmax, theta, norm="ectrans", csphase=False)
```
I suggest we do this so the polynomials match the ones from ecTrans, in case we want to use ecTrans to generate polynomials. Note that `norm="ectrans"` doesn't actually activate a particular code path, but we have to give it _some_ non-`None` value so the default isn't used.

samhatfield: Any thoughts @PortillaS-Predictia ?

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,213 @@
...
+            torch.cat([x, torch.zeros((*x.shape[:-1], rlon), device=x.device)], dim=-1)
+            for x, rlon in zip(rfft, self.rlon)
+        ]
+
+        return torch.stack(
+            tensors=rfft,
+            dim=-2,
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
```
samhatfield, line 236:
```suggestion
    def forward(self, x: Tensor) -> Tensor:
        """Performs direct SHT transform (Fourier transform followed by Legendre transform).

        Parameters
        ----------
        x : torch.Tensor
            field [..., grid]

        Returns
        -------
        torch.Tensor
            spectral representation of field [..., total wavenumber l, zonal wavenumber m]
        """
```

models/tests/layers/test_sht.py
```diff
@@ -0,0 +1,128 @@
...
+Random array of complex spectral coefficients.
+
+By definition arranged on an upper triangular matrix of width and height (truncation + 1), but with
+values below the diagonal just set to zero. The m = 0 coefficients are also purely real, to ensure
+that inverse transformed fields are also real.
+"""
+
+
+def random_spectral_array(truncation, dtype):
+    # Shape: [batch index, ensemble member, l, m, variable]
```
samhatfield:
```suggestion
    # Shape: [batch index, ensemble member, l, m]
```

models/tests/layers/test_sht.py
samhatfield, line 1: I'm not a pytest expert. Could someone have a look and see if I did it right? I wasn't sure what's the best way to pass variables from the fixture to the test so I returned a dictionary, but it seems a bit untidy. Could you have a look @OpheliaMiralles?

models/tests/layers/test_sht.py
```diff
@@ -0,0 +1,128 @@
...
+    return spectral_array
+
+
+class TestRegularSphericalHarmonicTransform:
+    """Test suite for OctahedralSphericalHarmonicTransform."""
+
+    @pytest.fixture
+    def init(self):
+        # Choose GPUs if available
+        device = "cuda" if torch.cuda.is_available() else "cpu"  # Spectral truncation
```
samhatfield:
```suggestion
        device = "cuda" if torch.cuda.is_available() else "cpu"
```

models/tests/layers/test_sht.py
```diff
@@ -0,0 +1,128 @@
...
+        assert maxdiff < tolerance
+
+
+class TestOctahedralSphericalHarmonicTransform(TestRegularSphericalHarmonicTransform):
+    """Test suite for OctahedralSphericalHarmonicTransform."""
+
+    @pytest.fixture
+    def init(self):
+        # Choose GPUs if available
+        device = "cuda" if torch.cuda.is_available() else "cpu"  # Spectral truncation
```
samhatfield:
```suggestion
        device = "cuda" if torch.cuda.is_available() else "cpu"
```

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,213 @@
...
+    def irfft_rings(self, x: Tensor) -> Tensor:
+
+        irfft = [self.irfft(x[..., t, :], nlon) for t, nlon in enumerate(self.lons_per_lat)]
+
+        return torch.cat(
+            tensors=irfft,
+            dim=-1,
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
```
samhatfield, line 366:
```suggestion
    def forward(self, x: Tensor) -> Tensor:
        """Performs inverse SHT transform (inverse Legendre transform followed by inverse Fourier transform).

        Parameters
        ----------
        torch.Tensor
            spectral representation of field [..., total wavenumber l, zonal wavenumber m]

        Returns
        -------
        x : torch.Tensor
            field [..., grid]
        """
```
```

File: models/src/anemoi/models/layers/spectral_helpers.py (hunk 21 above)
Comment (PortillaS-Predictia) [reply models/src/anemoi/models/layers/spectral_helpers.py]: Thanks Sam,

If this change doesn't break anything (it shouldn't), then I believe it combines both approaches nicely!

One word of caution - the normalization will affect the magnitude of the loss. That is, if the normalization is `4pi` instead of `sqrt(4pi)`, then all losses will be smaller by a factor of `sqrt(4pi)`. This is relevant when combining losses (for instance, `spectral loss` and `mse loss`) and adequately weighting them.

Comment (samhatfield) [reply models/src/anemoi/models/layers/spectral_helpers.py]: Yes good point. I thought that now is the right time to fix the definition of the Legendre polynomials before we start using spectral losses in earnest.

File: models/tests/layers/test_sht.py
Comment (OpheliaMiralles) [reply models/tests/layers/test_sht.py:1]: If you confirm that the below refactor is similar to your previous version only without classes, it might be neater to use it instead. 

```
# (C) Copyright 2025 Anemoi contributors.
#
# This software is licensed under the terms of the Apache Licence Version 2.0
# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
#
# In applying this licence, ECMWF does not waive the privileges and immunities
# granted to it by virtue of its status as an intergovernmental organisation

import pytest
import torch

from anemoi.models.layers.spectral_helpers import InverseSphericalHarmonicTransform
from anemoi.models.layers.spectral_helpers import SphericalHarmonicTransform

"""
Random array of complex spectral coefficients.

By definition arranged on an upper triangular matrix of width and height (truncation + 1), but with
values below the diagonal just set to zero. The m = 0 coefficients are also purely real, to ensure
that inverse transformed fields are also real.
"""


def random_spectral_array(truncation: int, dtype: torch.dtype) -> torch.Tensor:
    # Shape: [batch index, ensemble member, l, m]
    shape = (1, 1, truncation + 1, truncation + 1)
    spectral_array = torch.complex(torch.randn(shape, dtype=dtype), torch.randn(shape, dtype=dtype))
    spectral_array[0, 0, :, 0].imag = 0.0  # m = 0 modes must be real

    # Zero the lower triangle, which has no meaning
    for i in range(truncation + 1):
        spectral_array[0, 0, :i, i] = 0.0 + 0.0j

    return spectral_array


def _lons_per_lat(nlat: int, grid_kind: str) -> list[int]:
    if grid_kind == "regular":
        return [2 * nlat] * nlat
    if grid_kind == "octahedral":
        lons = [20 + 4 * i for i in range(nlat // 2)]
        return lons + list(reversed(lons))
    raise ValueError(f"Unknown grid_kind={grid_kind!r}")


@pytest.fixture(params=["regular", "octahedral"])
def sht_setup(request):
    # Choose GPUs if available
    device = "cuda" if torch.cuda.is_available() else "cpu"
    torch.set_default_device(device)

    truncation = 39  # T39 corresponding to O40 grid
    dtype = torch.float64  # float64 for numerical correctness checking
    torch.manual_seed(0)
    tolerance = 1e-11

    nlat = 2 * (truncation + 1)
    lons_per_lat = _lons_per_lat(nlat=nlat, grid_kind=request.param)

    direct = SphericalHarmonicTransform(
        nlat, lons_per_lat=lons_per_lat, lmax=truncation + 1, mmax=truncation + 1
    ).to(device)
    inverse = InverseSphericalHarmonicTransform(
        nlat, lons_per_lat=lons_per_lat, lmax=truncation + 1, mmax=truncation + 1
    ).to(device)

    return {
        "grid_kind": request.param,
        "truncation": truncation,
        "dtype": dtype,
        "tolerance": tolerance,
        "direct": direct,
        "inverse": inverse,
    }


def test_idempotency_direct_inverse(sht_setup):
    """direct followed by inverse returns the original (band-limited) field."""
    truncation = sht_setup["truncation"]
    dtype = sht_setup["dtype"]
    tolerance = sht_setup["tolerance"]
    direct = sht_setup["direct"]
    inverse = sht_setup["inverse"]

    before_spectral = random_spectral_array(truncation, dtype)

    # Ensure the direct input is band-limited by constructing it via inverse.
    before = inverse(before_spectral)

    after = inverse(direct(before))
    assert torch.allclose(before, after, rtol=tolerance)


def test_idempotency_inverse_direct(sht_setup):
    """inverse followed by direct returns the original spectral coefficients."""
    truncation = sht_setup["truncation"]
    dtype = sht_setup["dtype"]
    tolerance = sht_setup["tolerance"]
    direct = sht_setup["direct"]
    inverse = sht_setup["inverse"]

    before = random_spectral_array(truncation, dtype)
    after = direct(inverse(before))

    # Compute max relative diff over the meaningful upper triangle (including diagonal)
    maxdiff = 0.0
    for m in range(truncation + 1):
        ref = before[0, 0, m:, m]
        got = after[0, 0, m:, m]
        maxdiff = max(maxdiff, torch.abs((ref - got) / ref).max().item())

    assert maxdiff < tolerance

```

Comment (samhatfield) [reply models/tests/layers/test_sht.py:1]: Yes, excellent. Very neat. Many thanks! :)

File: models/src/anemoi/models/layers/spectral_helpers.py [hunk 27]
```diff
@@ -0,0 +1,237 @@
...
+        self.lmax = lmax or nlat
+        self.mmax = mmax or nlat
+
+        self.nlat = nlat
+        self.lons_per_lat = lons_per_lat
+
+        self.n_grid_points = sum(self.lons_per_lat)
+
+        self.slon = [0] + list(np.cumsum(self.lons_per_lat))[:-1]
+        self.rlon = [nlat + 8 - nlon // 2 for nlon in self.lons_per_lat]
```
PortillaS-Predictia: This is bit trickier @samhatfield. Let me explain what was the purpose of the `rlon` variable,

The _ring-wise_ rFFT would return a "jagged array" because each ring has a different number of nodes. Thus, for a ring with `nlon` nodes, the rFFT would return a vector of length `nlon // 2 + 1`. Now, if we want to stack all rings onto the same array, we need to _pad_ the shorter rings with zeros _to the right_ (that is, adding zeros above each ring's Nyquist frequency).

In an octahedral grid, the longest ring has `4N + 16` nodes and it's rFFT would have `2N + 9` frequencies (that is, `nlat + 9`). However, the _i_-th ring only has `nlon[i]` nodes and `nlon[i] // 2 + 1` frequencies. Thus, it needs to be _padded_ with `nlat + 9 - (nlon[i] // 2 + 1) = nlat + 8 - nlon[i] // 2` zeros.

Of course, me being me, I never documented any of this.

Now, if we extend this class' support to other grids, we should change this too. The math should look like this - the longest ring has `max(lons_per_lat)` nodes and `max(lons_per_lat) // 2 + 1` frequencies, and the remaining rings have `lons_per_lat[i] // 2 + 1` frequencies, so they should be _padded_ with `max(lons_per_lat) // 2 - lons_per_lat[i] // 2` zeros. Thus, this line should look like `self.rlon = [max(lons_per_lat) // 2 - nlon // 2 for nlon in self.lons_per_lat]`, but please double-check this :)

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,237 @@
...
+    [3] Schrama, E.; Orbit integration based upon interpolated gravitational gradients.
+    """
+
+    return legpoly(mmax, lmax, np.cos(t), norm=norm, inverse=inverse, csphase=csphase)
+
+
+class SphericalHarmonicTransform(Module):
+
+    def __init__(self, nlat: int, lons_per_lat: list[int], lmax: int | None = None, mmax: int | None = None) -> None:
+
```
PortillaS-Predictia, line 152: I guess there would be an _edge case_ in which all rings have the same number of nodes (that is, a regular grid). Then, the `for` loops in the `rfft_rings` method could be substituted with a single `torch.fft.rfft` call, which should be faster. We would be "trading" code clarity for performance. Just a heads-up, it looks good to me the way it's right now 👍.

models/src/anemoi/models/layers/spectral_helpers.py
samhatfield: Very good observation. This means that for the regular grid we are currently padding every longitude by 8 points for no good reason. I'll make some changes as you suggest.

models/src/anemoi/models/layers/spectral_helpers.py
samhatfield: Yeah, I think that would make a big difference in performance. I suppose we could have another class which extends `SphericalHarmonicTransform` and overloads `rfft_rings`?

PortillaS-Predictia: I would use the same class,

I would rename the method `rfft_rings` to `rfft_rings_jagged`, create a new method `rfft_rings_regular` which just returns `torch.fft.rfft(x.reshape(*x.shape[:-1], self.nlat, self.lons_per_lat[0]), norm="forward")` and add the following to the `__init__`:

```
self.rfft_rings = (
    self.rfft_rings_jagged if len(set(lons_per_lat)) > 1
    else self.rfft_rings_regular
)
```

This way, when the code calls `self.rfft_rings` in the `forward` method, it will appropriately point to either the "jagged transform" or the "regular transform". Does this make sense?

PS: This is just my code style, a new class would do the job as well :)

training/src/anemoi/training/losses/base.py
```diff
@@ -171,14 +171,15 @@ def reduce(
                 msg = f"Invalid squash_mode '{squash_mode}'. Supported modes are: 'avg', 'sum'"
                 raise ValueError(msg)
 
+        grid_dim = -2 if not squash else -1
```
OpheliaMiralles:
```suggestion
```

OpheliaMiralles:
```suggestion
```

training/src/anemoi/training/losses/base.py
```diff
@@ -171,14 +171,15 @@ def reduce(
...
         # here the grid and time dimension are summed because
         # 1. the normalisation over grid points is handled in the node weighting
         # 2. the normalization over output steps is handled by the time_step scaler
         space_time_summed = self.sum_function(
             out,
             dim=(
                 TensorDim.TIME,
-                TensorDim.GRID,
+                grid_dim,
```
OpheliaMiralles:
```suggestion
                TensorDim.GRID,
```

models/src/anemoi/models/layers/spectral_helpers.py
samhatfield: Good suggestion - I've implemented something along these lines, can you take a look?

PortillaS-Predictia: Thanks Sam! Looks good to me!

samhatfield (approved):

samhatfield:
Ah yes of course, N320. I need a bit of help there. Sara showed me that you store N grid information (e.g. longitudes per latitude) on the cloud and retrieve them when necessary. I'll need to implement some logic for that, but I'm not sure where to start.

Also I seem to have broken idempotency tests with the last commit which I wasn't expecting! I'll look into that now.

samhatfield:
That was really dumb.

samhatfield:
I've started playing around with N320 grids. So far so good. Looks like we'll have to tweak the error tolerance on the SHT tests though. For some reason the N320 grid is less tolerant (might just be because we can _only_ test N320, whereas for other grids we still test with 48 lats pole-to-equator - lower resolution, less rounding error accumulation?).

samhatfield:
> So now the N320 has been tested too, would you say that all points mentioned by @sahahner are complete?

I don't think so. Here's my take on Sara's points:

- [ ] Test spectral loss in a combined loss: here we should have a close eye on the loss function scalers and their correct application. _I'm not sure what this entails. Probably @sahahner is best placed to do this._
- [ ] As Sam and Sergio are refactoring the SHT classes into one, it would be important to test the losses again after the refactor. _Same as above._
- [x] We need to enable and test the SHT classes on n320 grids. _Done._
- [ ] Documentation: all newly introduced spectral losses and until now undocumented 2D versions need to be documented. _Not sure what counts as "officially documented". Are we talking about putting more comments in the code, or adding readthedocs pages?_
- [ ] Scientific evaluation of SHT transforms (runtime, performance, default values for truncation, ...). _We still don't really know what impact an additional spectral transform will have on runtime during training. If a forward pass takes 10x longer (unlikely) then what we have here is not usable and needs to be refactored and optimised before benchmarking. I am talking with colleagues at ECMWF about how we can best measure this._

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -81,30 +105,135 @@ def forward(
...
+        self._sht = SphericalHarmonicTransform(
+            nlat=self.nlat, lons_per_lat=self.lons_per_lat, lmax=self.nlat // 2, mmax=self.nlat // 2
+        )
+        self.y_freq = self._sht.lmax
+        self.x_freq = self._sht.mmax
+
+    def forward(self, data: torch.Tensor) -> torch.Tensor:
+        b, t, e, p, v = data.shape
+        assert p == self._sht.n_grid_points, f"Input points={p} does not match expected nlat*nlon={self.nlat*self.nlon}"
+        x = einops.rearrange(data, "b t e (y x) v -> (b t e v) y x", y=self.nlat, x=self.nlon)
```
samhatfield:
```suggestion
        x = einops.rearrange(data, "b t e p v -> b t e v p")
```

I think it should be this, but not 100% sure as I'm still learning einops syntax.

Note that `SphericalHarmonicTransform` could be modified so that this reshaping is unnecessary.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -81,30 +105,135 @@ def forward(
...
+        self.x_freq = self._sht.mmax
 
-        Parameters
-        ----------
-        data : torch.Tensor
-            Input data in the spatial domain.
+    def forward(self, data: torch.Tensor) -> torch.Tensor:
+        b, t, e, p, v = data.shape
+        assert p == self._sht.n_grid_points, f"Input points={p} does not match expected nlat*nlon={self.nlat*self.nlon}"
+        x = einops.rearrange(data, "b t e (y x) v -> (b t e v) y x", y=self.nlat, x=self.nlon)
```
samhatfield:
```suggestion
        x = einops.rearrange(data, "b t e p v -> b t e v p")
```
Same issue as below.

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,301 @@
...
+        self.n_grid_points = sum(self.lons_per_lat)
+
+        # Set offsets to start of each latitude in flattened grid dimension
+        self.slon = [0] + list(np.cumsum(self.lons_per_lat))[:-1]
+
+        # Set padding for each latitude so every rFFT output ring has the same length
+        self.rlon = [max(self.lons_per_lat) // 2 - nlon // 2 for nlon in self.lons_per_lat]
+
+        # Use more efficient batched rfft for regular grids
+        self.rfft_rings = self.rfft_rings_reduced if len(set(self.lons_per_lat)) > 1 else self.rfft_rings_regular
```
anaprietonem: what's the difference in performance between these two? I think it would be to have a logger statement printing which one is being picked up

samhatfield: There's a massive difference in performance because for full grids we can execute all ffts in a single kernel (they're all the same length), whereas for reduced grids (including octahedral) we have to execute one latitude at a time (as they're different lengths). I haven't measured the exact difference in speed though.

Happy to add an info logger.

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,301 @@
...
+
+        # Compute Gaussian latitudes and quadrature weights
+        theta, weight = legendre_gauss_weights(nlat)
+        theta = np.flip(np.arccos(theta))
+
+        # Precompute associated Legendre polynomials
+        pct = precompute_legpoly(self.mmax, self.lmax, theta, norm="ectrans", csphase=False)
```
anaprietonem: is `norm=ectrans` doing something now? @samhatfield I had understood that the ecctrans backend for the polynomials was not part of the PR so  this something different?

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,301 @@
...
+    Expansions, Ohio State University Columbus; report; 1982; https://apps.dtic.mil/sti/citations/ADA123406.
+    [3] Schrama, E.; Orbit integration based upon interpolated gravitational gradients.
+    """
+
+    return legpoly(mmax, lmax, np.cos(t), norm=norm, inverse=inverse, csphase=csphase)
+
+
+class SphericalHarmonicTransform(Module):
+
+    def __init__(self, nlat: int, lons_per_lat: list[int], lmax: int | None = None, mmax: int | None = None) -> None:
```
anaprietonem: this would need docstrings and also the functions above. You can see some examples of the style here https://github.com/ecmwf/anemoi-core/blob/main/models/src/anemoi/models/data_indices/tensor.py (and also some more info here https://anemoi.readthedocs.io/en/latest/contributing/documentation.html#docstrings)

models/src/anemoi/models/layers/spectral_helpers.py
```diff
@@ -0,0 +1,301 @@
...
+    """
+
+    xlg, wlg = np.polynomial.legendre.leggauss(n)
+    xlg = (b - a) * 0.5 * xlg + (b + a) * 0.5
+    wlg = wlg * (b - a) * 0.5
+
+    return xlg, wlg
+
+
+def legpoly(
```
anaprietonem, line 49: couldn't it be faster to already compute this with torch?  (as in you then need don't need the torch.from_numpy)

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -61,18 +64,39 @@ def __init__(
...
         self.y_dim = y_dim
+        nodes_slice = nodes_slice or (0, None)  # we don't want einops to silently fail
+        # by slicing random parts of the input
         self.nodes_slice = slice(*nodes_slice)
+        self.apply_filter = apply_filter
+        if apply_filter:
+            self.filter = self.lowpass_filter(x_dim, y_dim)
+
+    @staticmethod
+    def lowpass_filter(x_dim, y_dim):
```
anaprietonem:
```suggestion
    def lowpass_filter(x_dim: torch.Tensor, y_dim: torch.Tensor)->torch.Tensor:
```

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -61,18 +64,39 @@ def __init__(
...
+        fx = torch.fft.fftfreq(x_dim)
+        fy = torch.fft.fftfreq(y_dim)
+
+        KX, KY = torch.meshgrid(fx, fy, indexing="ij")
+        k = torch.sqrt(KX * KX + KY * KY)
+
+        mask = k < 0.5  # torch.where(k < 0.5, 1.0 - 2.0 * k, 0.0)
```
anaprietonem, line 137: why 0.5?

models/src/anemoi/models/layers/spectral_helpers.py
samhatfield: Yes this is maybe a bit misleading. There are three choice for polynomial normalisation:
1. "schmidt" (`norm="schmidt"`)
2. "ortho" (`norm="ortho"`) **default**
3. other (I'm not sure if it has a name) (*this is the normalisation ecTrans uses*)

I want the third option so that I match ecTrans, but because the default value is `norm="ortho"`, I have to pass _something_ so that that option isn't chosen. I chose `"ectrans"` to give a hint that I'm following the same convention as ecTrans, but in fact I could pass any string not equal to `"schmidt"` or `"ortho"`.

training/pyproject.toml
```diff
@@ -99,6 +99,7 @@ optional-dependencies.remote = [
 ]
 optional-dependencies.tests = [
   "anemoi-graphs[tri]",                         # required to load the gnn checkpoint as part of the tests
+  "anemoi-models[spectral]",                    # required to test the DCT2D spectral loss
```
anaprietonem, line 96: much nicer, thanks

models/src/anemoi/models/layers/spectral_helpers.py
samhatfield: We retrieve the Gaussian weights from `np.polynomial.legendre.leggauss`. I don't think there is a torch equivalent?

training/docs/modules/losses.rst
```diff
@@ -112,25 +112,105 @@ The config for the multiscale loss functions is the following:
...
+Spectral losses rely on a transform that maps grid-point fields to spectral coefficients.
+
+Supported transforms include:
+
+* ``FFT2D``: 2D FFT for regular latitude/longitude grids (or any regular 2D field) with
+  known ``x_dim`` and ``y_dim``.
+* ``DCT2D``: 2D Discrete Cosine Transform for regular 2D fields. This transform requires
+  the optional dependency ``torch-dct``.
+* ``EcTransOctahedralSHT``: Spherical Harmonic Transform (SHT) on the *octahedral reduced
+  Gaussian grid* using ecTrans assets (via precomputed ``npz`` assets, or generated with
```
anaprietonem: @samhatfield if ecTrans is not yet part of the PR, this probably needs updating?

training/src/anemoi/training/losses/spectral.py
```diff
@@ -246,3 +256,82 @@ def __init__(
             scalers=scalers,
             **kwargs,
         )
+
+
+class SpectralCRPSLoss(SpectralLoss, AlmostFairKernelCRPS):
```
anaprietonem, line 264: If this is recommended for ensemble >1 shouldn't we have an assert and a log warning or error if someone tries to use this with the wrong model?

training/tests/unit/losses/test_loss_function.py
```diff
@@ -373,6 +418,9 @@ def test_logfft2dist_loss() -> None:
 
 
 def test_fcl_loss() -> None:
+    import einops
+
+    # TODO (Ophelia): edit this when multi ouptuts get merged
```
anaprietonem: Ophelia is this something that you can address now? or have been addressed and the comment can be removed?

models/src/anemoi/models/layers/spectral_helpers.py
anaprietonem: That's fine I would then have ectrans as an option and execute the part of the code being executed now and then raise an error for other norms to make it more clear.

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -533,7 +533,7 @@ def _prepare_tensors_for_loss(
 
         is_sharded = grid_shard_slice is not None
 
-        sharding_supported = (self.loss_supports_sharding or validation_mode) and (
+        sharding_supported = (self.loss_supports_sharding) and (  # loss calculated in training and validation mode
```
anaprietonem, line 556: why is this change needed?

training/docs/modules/losses.rst
```diff
@@ -112,25 +112,105 @@ The config for the multiscale loss functions is the following:
...
+   training_loss:
+     datasets:
+       your_dataset_name:
+         _target_: anemoi.training.losses.spectral.SpectralCRPSLoss
+         transform: ectrans_octahedral_sht
+         truncation: 127
+         # Path to precomputed Legendre polynomials / weights.
+         # If the file does not exist, assets can optionally be generated when `ectrans4py`
+         # is installed and the path is writable.
+         filepath: /path/to/ectrans_assets_T127.npz
```
anaprietonem: is this implemented?

training/tests/unit/losses/test_loss_function.py
OpheliaMiralles: this is addressed I'll remove the todo

models/src/anemoi/models/layers/spectral_helpers.py
samhatfield: How about I just delete the non-ecTrans normalisation options? I guess we're not planning on using them.

OpheliaMiralles:
> > > So now the N320 has been tested too, would you say that all points mentioned by @sahahner are complete?
> > 
> > 
> > I don't think so. Here's my take on Sara's points:
> > 
> > * [x]   Test spectral loss in a combined loss: here we should have a close eye on the loss function scalers and their correct application. _I'm not sure what this entails. Probably @sahahner is best placed to do this._
> > * [x]   As Sam and Sergio are refactoring the SHT classes into one, it would be important to test the losses again after the refactor. _Same as above._
> > * [x]   We need to enable and test the SHT classes on n320 grids. _Done._
> > * [x]   Documentation: all newly introduced spectral losses and until now undocumented 2D versions need to be documented. _Not sure what counts as "officially documented". Are we talking about putting more comments in the code, or adding readthedocs pages?_
> > * [ ]   Scientific evaluation of SHT transforms (runtime, performance, default values for truncation, ...). _We still don't really know what impact an additional spectral transform will have on runtime during training. If a forward pass takes 10x longer (unlikely) then what we have here is not usable and needs to be refactored and optimised before benchmarking. I am talking with colleagues at ECMWF about how we can best measure this._
> 
> related to this. I agree that it seems some of those points are still to be addressed. I think docs and SHT refactor could though be marked as done?
> 
> * The combined loss, I think Sara can help once she's back but @OpheliaMiralles  is this something you have checked before or could be worth just adding a test to confirm it works?
> 
> It would also be good to check the integration tests. At the moment this would just check that at least no changes here are breaking existing tests.

There is a test already

training/src/anemoi/training/train/tasks/base.py
OpheliaMiralles: good question, might be related to the fft2d loss hard to interpret when grid is sharded (x_dim and y_dim are not the grid dimensions anymore) so I think there should be an assert in any loss using a 2D transform that keep_batch_sharded should be set to False to compute the loss on the full grid

training/src/anemoi/training/losses/spectral.py
OpheliaMiralles: There is an assertion from the AmlostFairKernelCRPS init if the ensemble size is 1

models/src/anemoi/models/layers/spectral_transforms.py
OpheliaMiralles: @PortillaS-Predictia might know

PortillaS-Predictia: This filter removes the "corners of the square", because these have a "radial frequency" higher than the "per-axis Nyquist frequency".

The function `torch.fft.fftfreq` creates frequencies normalized to the interval [-0.5, 0.5), where 0.5 would represent the "Nyquist limit", however if you have a `2D meshgrid` (a "square") then the "corners of the square" would have radial frequency equal to `sqrt(0.5^2 + 0.5^2) = 0.707`, which is above the "per-axis limit" of 0.5.

I'm not sure why we would want to remove these frequencies. My guess is that this creates an "isotropic filter" (i.e., it looks the same in every direction, like a circle instead of a square), such that the highest frequency in the filtered field is 0.5, regardless of how we choose the `x` and `y` directions. I think this is just a design choice, not a compulsory requirement to prevent aliasing (the "corners" shouldn't be any more aliased than the "sides").

evenmn: I guess the idea here is that the frequencies larger than 0.5 are above the Nyquist limit, and thus unphysical. However, it might not be a problem if we include them, as the same "unphysical" frequencies are included both for the targets and predictions. Our experiments without a low-pass filter indicate that the filter is not crucial

training/docs/modules/losses.rst
samhatfield: Docs were out of date.

PortillaS-Predictia:
Hi @samhatfield,

You could also try [compiling](https://anemoi.readthedocs.io/projects/training/en/latest/user-guide/models.html#compilation) the module (just the transform, not the loss function). However, my understanding is that `torch.compile`'s support for complex numbers is limited.

training/tests/unit/losses/test_combined_loss.py
```diff
@@ -103,3 +105,36 @@ def test_combined_loss_seperate_scalers() -> None:
     assert isinstance(loss.losses[1], MAELoss)
     assert "test" not in loss.losses[1].scaler
     assert "test2" in loss.losses[1].scaler
+
+
+def test_combined_loss_with_spectral_crps_backward() -> None:
```
sahahner, line 123: Thank you for adding a test for the combined loss with the spectral CRPS loss. I would like to test this next week for training a model with the deterministic model.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -7,14 +7,14 @@
...
-
 import abc
 import logging
 
 import einops
 import torch
 import torch.fft
 
+from anemoi.models.layers.spectral_helpers import SphericalHarmonicTransform
 from anemoi.training.utils.enums import TensorDim
```
anaprietonem: This shouldn't be here - we can't import anemoi-training in anemoi-models. @frazane @OpheliaMiralles  this is from the previous skeleton PR I believe, and could be worth addressing in a separate PR quickerif you have some time please.

OpheliaMiralles: You mean we should not import the TensorDim?

anaprietonem: yes exactl, Dieter created an issue https://github.com/ecmwf/anemoi-core/issues/945

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -81,30 +105,146 @@ def forward(
...
-        -------
-        torch.Tensor
-            Data transformed to the spectral domain.
+        if grid != "n320":
+            raise ValueError("Only the N320 reduced Gaussian grid SHT is supported.")
+        else:
+            self.nlat = 640
+
+        # Fetch regular grid data
+        from anemoi.transform.grids.named import lookup
```
sahahner: Here, we introduce a dependency of anemoi-transform in anemoi-models. Is that okay?

OpheliaMiralles: If it stays it needs to be a dependency somewhere, added it to models spectral optional dependencies

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -81,30 +154,171 @@ def forward(
...
+        super().__init__()
 
-        Returns
-        -------
-        torch.Tensor
-            Data transformed to the spectral domain.
+        if grid != "n320":
+            raise ValueError("Only the N320 reduced Gaussian grid SHT is supported.")
+        else:
+            self.nlat = 640
```
anaprietonem: @samhatfield  can't we get this from the grid we are loading rather than hardcoding it?

models/src/anemoi/models/layers/spectral_transforms.py
anaprietonem: My suggestion is, add  anemoi-transform as part of the anemoi-models[spectra] dependencies group. This is just required when people will be trying to use the Reduced Gaussian spectra not for anything else. Then I would have a proper error catch here, where if anemoi-transform is not installed when trying to use this class then a error message tells you to install the additional dependencies doing `pip install anemoi-models[spectra]`

models/src/anemoi/models/layers/spectral_transforms.py
samhatfield: I suppose we might as well yes. I thought it would be simpler to hard code it since we only support N320 atm, but I'll make it more generic for future compatibility.

pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[project]
+name = "anemoi-core"
```
sahahner: Why is this additional toml file needed or was it added by mistake?

training/pyproject.toml
```diff
@@ -63,6 +63,7 @@ optional-dependencies.azure = [ "azure-ai-ml", "azure-identity", "azureml-mlflow
 optional-dependencies.dev = [
   "anemoi-training[all,docs,tests]",
   "pre-commit>=3.3.3",
+  "ruff>=0.14.10",
```
anaprietonem: why is ruff being added here? Ruff is defined at the pre-commit leve for the mono repo

models/src/anemoi/models/layers/spectral_transforms.py
sahahner: Sam and I found a way around that.

models/pyproject.toml
```diff
@@ -35,6 +35,7 @@ classifiers = [
 ]
 dynamic = [ "version" ]
 dependencies = [
+  "anemoi-transform>=0.1.2",
```
sahahner:
```suggestion
  
```

pyproject.toml
OpheliaMiralles: I think it's really useful for developers to have an env with all anemoi-core dependencies at once to run for models/training/graphs,  I think it should stay but indeed not really part of this PR

training/pyproject.toml
OpheliaMiralles: Can revert this

pyproject.toml
sahahner: I see where this comes from now. Maybe you can propose it in a separate PR then, please? I think then this change would be more visible to others as well.

OpheliaMiralles: but it can be removed

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -295,9 +295,7 @@ def __init__(
 
         self.shard_shapes, self.grid_sizes = {}, {}
         for dataset_name in self.dataset_names:
-            self.grid_sizes[dataset_name] = graph_data[dataset_name][
-                "data"
-            ].num_nodes  # TODO(Mario): Replace by dataset.grid_size
+            self.grid_sizes[dataset_name] = graph_data[dataset_name].num_nodes
```
JPXKQX: Hi, I wouldn't remove the TODO, as this was introduced to remove the dependency on the graph in the training (needed for observations)

sahahner:
```suggestion
            self.grid_sizes[dataset_name] = graph_data[dataset_name][
                "data"
            ].num_nodes  # TODO(Mario): Replace by dataset.grid_size
```

JPXKQX:
```suggestion
            self.grid_sizes[dataset_name] = graph_data[dataset_name][
                "data"
            ].num_nodes  # TODO(Mario): Replace by dataset.grid_size
```

JPXKQX: I would revert completely this change. It is changing the expected behaviour. Not specifying the "data" nodes will return the total number of nodes (data + hidden + any other nodes like truncation nodes).

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -929,9 +944,6 @@ def calculate_val_metrics(
                     scaler_indices=(..., indices),
                     grid_shard_slice=grid_shard_slice,
                     group=self.model_comm_group,
-                    model_comm_group_size=self.model_comm_group_size,
```
sahahner, line 932: Why is it necessary to remove this line?

japols: don't think it should be, also wouldn't remove the 2 lines below

training/src/anemoi/training/train/tasks/base.py
OpheliaMiralles: Agreed it shouldn't be there I am on a branch mixing changes from the multi graph PR and spectral losses so it ended up there

models/pyproject.toml
```diff
@@ -60,7 +60,8 @@ optional-dependencies.migrations = [
   "jinja2>=3.1",
   "rich",
 ]
-optional-dependencies.tests = [ "hypothesis>=6.11", "pytest>=8" ]
+optional-dependencies.spectral = [ "torch-dct>=0.1.6" ]
```
OpheliaMiralles: So here we should add anemoi-transform dependency I guess @sahahner @anaprietonem ?

training/src/anemoi/training/losses/kcrps.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         self,
         fair: bool = True,
         ignore_nans: bool = False,
-        **kwargs,
+        **kwargs,  # noqa: ARG002
```
sahahner, line 29: Do we need these changes in the kcrps file? Is there a reason whe we are not passing kwargs to the init of the base loss anymore?

sahahner:
> > > > The scalers work now for the combined loss. Thank you for merging your fix, @OpheliaMiralles.
> > > > I have noticed, though, that model sharding is broken, so I need to check that still.
> > > 
> > > 
> > > We are not supposed to shard batches at the loss level with spectral losses (FFT2D etc don't have the right xdim and ydim otherwise), config should include keep_batch_sharded = False when we apply spectral losses. But I don't know if you are mentioning something else that is broken
> > 
> > 
> > It is possible to shard the model and I have enforced the spectral losses to unshard the batch over grid dimension in this line: https://github.com/ecmwf/anemoi-core/pull/788/changes#diff-1201c839d742594c8c5d191e83b09598c3ec47fb2109719f89ed52c94d6c7c0fR102
> > Thankfully this option was already implemented for the loss functions.
> 
> Yes that's what I meant! So what is broken? Mario hinted some lines were deleted in previous commits in task base.py related to sharding but I reverted this. Maybe now sharding works again?

I just pushed a fix. The `supports_sharding` flag from the loss function was not passed for the losses contributing to the combined loss correctly to the task, where the batches have to be gathered before passing the data to the loss function.

training/src/anemoi/training/losses/kcrps.py
OpheliaMiralles: Yes we need it because we init two classes in SpectralCRPS with different kwargs, but here the change is only a comment

sahahner: Yes, your are right. I picked the wrong line and ignore_nan is the only argument the init of the base loss takes :)

PR #797, opened 2026-01-13
feat(training): make activation checkpoints configurable

ssmmnn11:
Make gradient checkpoints configurable. 

----
📚 Documentation preview 📚: https://anemoi-training--797.org.readthedocs.build/en/797/

----
📚 Documentation preview 📚: https://anemoi-graphs--797.org.readthedocs.build/en/797/

----
📚 Documentation preview 📚: https://anemoi-models--797.org.readthedocs.build/en/797/

### DIFF
--- a/training/src/anemoi/training/data/datamodule.py
+++ b/training/src/anemoi/training/data/datamodule.py
@@ -135,11 +135,7 @@ class AnemoiDatasetsDataModule(pl.LightningDataModule):
     @cached_property
     def ds_train(self) -> MultiDataset:
         """Create multi-dataset for training."""
-        return self._get_dataset(
-            self.train_dataloader_config,
-            shuffle=True,
-            label="training",
-        )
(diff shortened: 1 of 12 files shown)


models/src/anemoi/models/layers/utils.py
```diff
@@ -32,6 +32,27 @@ def forward(self, *args, **kwargs):
...
+        Whether to apply gradient checkpointing
+    *args, **kwargs
+        Arguments to pass to the function
+
+    Returns
+    -------
+    The result of calling func with the provided arguments
+    """
+    if enabled:
+        return checkpoint(func, *args, **kwargs, use_reentrant=False)
```
anaprietonem, line 52: out of curiosity - what's use_reentrant doing and why do we keep it to false?

ssmmnn11: use_reentrant is the modern way to do grad checkpoints - and at some point it was new and the default was to have it true iirc. We could check if the current default is False anyway (probably).

PR #806, opened 2026-01-16
fix: filteringlosswrapper

### LINKED ISSUE #803: FilteringLossWrapper (CombinedLoss) not correctly functioning
### What happened?

I could not use the FilteringLossWrapper in a fairly easy setup. 

I have a fairly easy setup where the only loss I compute during training is an MSE loss comparing the tp output of my model against tp data from IMERG. FilteredLossWrapper is designed for these types of scenarii.

```python

  training_loss:
    # loss class to initialise

    losses:
      - _target_: anemoi.training.losses.filtering.FilteringLossWrapper
        predicted_variables: ["tp"]
        target_variables: ["imerg"]
        loss:
          _target_: anemoi.training.losses.MSELoss
          ignore_nans: False

``` 

First, FilteredLossWrapper breaks because it needs the set_data_indices [here](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/losses/filtering.py#L71)
 method to be performed. We have a set_data_indices hook in the BaseLoss [here](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/losses/base.py#L74)
, but this hook is not initialized anywhere else, so when we use FilteredLossWrapper within CombinedLoss, the method is just empty.

We could define the method inside CombinedLoss, but this is to be discussed:

```python

    def set_data_indices(self, data_indices: IndexCollection) -> None:

        for loss in self.losses:
            if hasattr(loss, "set_data_indices"):
                loss.set_data_indices(data_indices)

``` 

Second, the mechanics behind the set_data_indices coded in the FilteredLossWrapper might not be correct. The method is currently defined like this:

```python
    def set_data_indices(self, data_indices: IndexCollection) -> None:
        """Hook to set the data indices for the loss."""
        name_to_index = data_indices.data.output.name_to_index
        model_output = data_indices.model.output
        output_indices = model_output.full

        if self.predicted_variables is not None:
            predicted_indices = [model_output.name_to_index[name] for name in self.predicted_variables]
        else:
            predicted_indices = output_indices
        if self.target_variables is not None:
            target_indices = [name_to_index[name] for name in self.target_variables]
        else:
            target_indices = output_indices

        assert len(predicted_indices) == len(
            target_indices,
        ), "predicted and target variables must have the same length for loss computation"

        self.predicted_indices = predicted_indices
        self.target_indices = target_indices

``` 

To understand what is going on it's helpful to print the target_indices, predicted indices and shapes of the prediction and the target in the very simply example I mentioned before: 

(inside the forward method of FilteringLossWrapper)

```python
        
print(pred.shape, target.shape, self.predicted_indices, self.target_indices)

torch.Size([1, 1, 542080, 103]) torch.Size([1, 1, 542080, 103]) [35] [115]

``` 
The predicted indices seem to be correct, [35], which corresponds to tp, but the target indices are completely wrong. [115] is out of bounds in the target and this produces a very messy cuda error.

The issue seems to come from obtaining the indices from data_indices.data.output.name_to_index. In the data space, the index for imerg, the target variable, is [115], but the target tensor is already transformed with these indices well before the computation of losses is done, [here](https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/train/tasks/forecaster.py#L72).

```python

            y = batch[:, self.multi_step + rollout_step, ..., self.data_indices.data.output.full]


``` 

We cannot use self.data_indices.data.output on "y", the target, again since it's already been used. The problem gets a little bit more messy when you consider that in the case I described, imerg could be defined as a "target" variable.

```python

data:

  target:
  - imerg


``` 

in which case 

```python
        
print(pred.shape, target.shape, self.predicted_indices, self.target_indices)

torch.Size([1, 1, 542080, 102]) torch.Size([1, 1, 542080, 103]) [35] [115]

``` 

so we cannot use the model output space for target (102 vs 103). In conclusion, we need to work on the logic of set_data_indices. 

### Version

recent main version

linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

ecmwf

OpheliaMiralles:
## Description
Originally fixes #803. Has evolved to provide any loss the flexibility to be applied to a subset of chosen variables. This PR wraps all the losses in FilteringLossWrapper (default to output.full), allowing any loss to be called with predicted_variables and target_variables without explicitely mentionning the FilteringLossWrapper, see e.g. below config. 

```
variables = ["tp", "t2m", "u10", "v10", "msl"]
DictConfig(
                {
                    "_target_": "anemoi.training.losses.CombinedLoss",
                    "losses": [
                        {
                            "_target_": "anemoi.training.losses.MSELoss",
                            "predicted_variables": variables,
                            "target_variables": variables,
                            "scalers": ["pressure_level"],
                        },
                        {
                            "_target_": "anemoi.training.losses.MSELoss",
                            "predicted_variables": ["tp"],
                            "target_variables": ["imerg"],
                            "scalers": ["pressure_level"],
                        },
                    ],
                    "scalers": ["*"],
                },
            )
```

Includes
- [x]  Integration tests passing for all tasks
- [x]  Unit tests passing
- [x] New integration test for @gabrieloks usecase comparing tot prec to imerg 
- [x] New unit tests adapted to new logic of variable filtering at the loss forward level and not in each task
- [x] Works for @gabrieloks usecase, loss behaves as expected

FilteringLossWrapper's method set_data_indices was never called during training. 

##  Additional notes ##
To be tested by @gabrieloks 

----
📚 Documentation preview 📚: https://anemoi-training--806.org.readthedocs.build/en/806/

----
📚 Documentation preview 📚: https://anemoi-graphs--806.org.readthedocs.build/en/806/

----
📚 Documentation preview 📚: https://anemoi-models--806.org.readthedocs.build/en/806/


training/tests/integration/conftest.py
```diff
@@ -532,6 +532,30 @@ def multi_output_interpolator_config(
     return cfg, url_dataset
 
 
+@pytest.fixture
+def imerg_target_config(
+    testing_modifications_with_temp_dir: DictConfig,
+    get_tmp_paths: GetTmpPaths,
```
OpheliaMiralles:
```suggestion
    get_tmp_paths: GetTmpPath,
```

training/src/anemoi/training/losses/base.py
```diff
@@ -115,7 +118,7 @@ def scale(
                 "Scaler tensor must be at least applied to the GRID dimension. "
                 "Please add a scaler here, use `UniformWeights` for simple uniform scaling.",
             )
-            raise RuntimeError(error_msg)
+            LOGGER.warning(error_msg)
```
ssmmnn11: What is the reason for this change?

OpheliaMiralles:
```suggestion
            raise RuntimeError(error_msg)
```

training/src/anemoi/training/train/tasks/diffusionforecaster.py
```diff
@@ -489,6 +491,9 @@ def _step(
         x_ref = {dataset_name: (ref[:, -1] if ref.ndim == 5 else ref) for dataset_name, ref in x_ref.items()}
 
         tendency_target = self._compute_tendency_target(y, x_ref)
+        tendency_target_full = {d: data.narrow(1, self.n_step_input, self.n_step_output) for d, data in batch.items()}
+        for d in batch:
+            tendency_target_full[d][..., self.data_indices[d].data.output.full] = tendency_target[d]
```
ssmmnn11: I think this will be a view and then you mutate the batch, what we probably don't want.

training/src/anemoi/training/losses/base.py
```diff
@@ -115,7 +118,7 @@ def scale(
                 "Scaler tensor must be at least applied to the GRID dimension. "
                 "Please add a scaler here, use `UniformWeights` for simple uniform scaling.",
             )
-            raise RuntimeError(error_msg)
+             raise RuntimeError(error_msg)
```
OpheliaMiralles:
```suggestion
            raise RuntimeError(error_msg)

```

models/src/anemoi/models/data_indices/collection.py
```diff
@@ -27,12 +27,28 @@
 class IndexCollection:
     """Collection of data and model indices."""
 
+    @staticmethod
+    def _contiguous_span(indices: list[int]) -> tuple[bool, int, int]:
```
HCookie, line 31: Please add a docstring, it is not clear from the name what the tuple even is supposed to be

training/src/anemoi/training/losses/base.py
```diff
@@ -312,9 +315,8 @@ def forward(
...
             Weighted loss
         """
-        del kwargs  # not used in this base implementation, but may be used in child classes
         is_sharded = grid_shard_slice is not None
         out = self.calculate_difference(pred, target)
         out = self.scale(out, scaler_indices, without_scalers=without_scalers, grid_shard_slice=grid_shard_slice)
-
-        return self.reduce(out, squash, group=group if is_sharded else None)
+        squash_mode = kwargs.get("squash_mode", "avg")
+        return self.reduce(out, squash, group=group if is_sharded else None, squash_mode=squash_mode)
```
HCookie: Why not add it as an actual arg rather than popping?

training/src/anemoi/training/losses/combined.py
```diff
@@ -120,13 +121,28 @@ def __init__(
         if loss_weights is None:
             loss_weights = (1.0,) * len(losses)
 
+        data_indices = kwargs.pop("data_indices", None)
+        scalers = kwargs.pop("scalers", {})
```
HCookie: Again, why not add to `__init__` then pop?

training/src/anemoi/training/utils/index_space.py
```diff
@@ -0,0 +1,18 @@
...
+
+from enum import StrEnum
+
+
+class IndexSpace(StrEnum):
+    """Tensor index-space for variable resolution in loss wrappers."""
+
+    MODEL_OUTPUT = "model_output"
+    DATA_OUTPUT = "data_output"
+    DATA_FULL = "data_full"
```
HCookie, line 18: If this is useful outside of the losses, we might need to move it

training/src/anemoi/training/losses/loss.py
```diff
@@ -77,21 +82,47 @@ def get_loss_function(
...
+        if data_indices is not None:
+            loss_config.update(
+                {"data_indices": data_indices},
+            )
+            data_indices = None  # for combined loss we want the individual losses to handle data indices
+        loss_config.update(
+            {"scalers": combined_scalers},
+        )
+        scalers_to_include = []
+        scalers = {}  # for combined loss we want the individual losses to handle scalers
```
HCookie: Why are we reverting the generalisability of this function to require a specific branch for the combinedloss. We have discussed this in the past, I do not think this a good design choice, one that I made intentionally to avoid.

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -923,22 +974,14 @@ def calculate_val_metrics(
...
-                    model_comm_group=self.model_comm_group,
-                    model_comm_group_size=self.model_comm_group_size,
-                    grid_dim=self.grid_dim,
-                    grid_shard_shapes=self.grid_shard_shapes,
-                )
-                continue
+            assert isinstance(
+                metric,
+                BaseLoss,
+            ), f"Validation metric {metric_name!r} must inherit BaseLoss, got {type(metric)}"
```
HCookie, line 988: Can you explain why we cannot use other metrics now?

HCookie (changes requested):

HCookie (dismissed):
Thanks for the contribution, overall I think it adds some nice new functionality, I'm a bit concerned about some of the implementation details, and think we might need to discuss this more

training/src/anemoi/training/train/tasks/base.py
ssmmnn11: The issue is that now full targets enter the metric / loss computation and hence filtering needs to take place. Though I think we already reject non BaseLoss losses in current main.

training/src/anemoi/training/losses/base.py
ssmmnn11: good point

training/src/anemoi/training/utils/index_space.py
OpheliaMiralles: For now it is very specific to the loss, we can leave it there and if another usecase pops up move it

VeraChristina:
Would it be possible to split this PR into more readable chunks, e.g. to separate out the tests that were added but do not seem to be related to the filtering loss wrapper? That would be super helpful for reviewing.

models/src/anemoi/models/data_indices/tensor.py
```diff
@@ -52,13 +52,27 @@ def __init__(
         self.target = self._build_idx_from_list(target)
         self.full = self._build_idx_from_list(includes)
         self.excludes = sorted(list(set(self.name_to_index.keys()) - set(self.includes)))
+        self.full_index_to_name = {int(index): name for name, index in self.name_to_index.items()}
+        self.ordered_names = [self.full_index_to_name[int(index)] for index in self.full.tolist()]
+        self.name_to_position = {name: position for position, name in enumerate(self.ordered_names)}
```
VeraChristina, line 57: what is the difference to `name_to_index`? what is the information we will need in inference? We are currently using `name_to_index` to gather inference metadata. If this only affects logic used for the losses, that should be okay. Just checking if we need more information on indices for inference

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1010,10 +1011,19 @@ def _plot(
             for rollout_step in range(adapter.loss_plot_times):
                 y_hat = outputs[1][rollout_step][dataset_name]
                 start = adapter.get_loss_plot_batch_start(rollout_step)
-                y_time = batch[dataset_name].narrow(1, start, pl_module.n_step_output)
-                var_idx = data_indices.data.output.full.to(device=batch[dataset_name].device)
-                y_true = y_time.index_select(-1, var_idx)
```
VeraChristina, line 658: Is this now handled via pred_layout and target_layout?

training/src/anemoi/training/losses/combined.py
```diff
@@ -16,21 +16,29 @@
...
 from anemoi.training.losses.scaler_tensor import ScaleTensor
+from anemoi.training.utils.enums import TensorDim
 
 
 class CombinedLoss(BaseLoss):
     """Combined Loss function."""
 
+    # CombinedLoss builds child losses itself, so it needs the filtered scaler
+    # set and data indices during construction.
+    factory_context_keys = frozenset({"available_scalers", "data_indices"})
```
VeraChristina: small preference to not use strings but constants or enums as flags that modify downstream logic, e.g. from a `LossContextKey` class -- you use this pattern elsewhere

training/src/anemoi/training/losses/kcrps.py
```diff
@@ -191,7 +191,7 @@ def forward(
         kcrps_ = einops.rearrange(kcrps_, "bs t v latlon -> bs t 1 latlon v")
         kcrps_ = self.scale(kcrps_, scaler_indices, without_scalers=without_scalers, grid_shard_slice=grid_shard_slice)
 
-        return self.reduce(kcrps_, squash=squash, squash_mode="sum", group=group if is_sharded else None)
+        return self.reduce(kcrps_, squash=squash, squash_mode=squash_mode, group=group if is_sharded else None)
```
VeraChristina, line 194: does this modify default behaviour, or do we set "sum" somewhere as a default?

training/src/anemoi/training/losses/variable_mapper.py
```diff
@@ -0,0 +1,336 @@
...
+                target_variables,
+            ), "predicted and target variables must have the same length for loss computation"
+
+        super().__init__()
+
+        self._loss_scaler_specification = {}
+        assert isinstance(
+            loss,
+            BaseLoss,
+        ), f"Invalid loss type provided: {type(loss)}. Expected a str or dict or BaseLoss."
```
VeraChristina: seems like the assert is not in line with error message and type hint above

training/src/anemoi/training/losses/variable_mapper.py
```diff
@@ -0,0 +1,336 @@
...
+            return scaler_indices, False
+
+        pred_index_to_local = {index: pos for pos, index in enumerate(pred_indices)}
+        mapped_indices = [pred_index_to_local[index] for index in requested_indices if index in pred_index_to_local]
+        remapped_scaler_indices = (*scaler_indices[:-1], self._restore_indexer_type(mapped_indices, variable_indexer))
+        return remapped_scaler_indices, len(mapped_indices) == 0
+
+    def forward(self, pred: torch.Tensor, target: torch.Tensor, **kwargs) -> torch.Tensor:
+        pred_layout = kwargs.pop("pred_layout", None)
+        target_layout = kwargs.pop("target_layout", None)
```
VeraChristina: is mutating kwargs directly intended?

training/src/anemoi/training/losses/variable_mapper.py
```diff
@@ -0,0 +1,336 @@
...
+        if len(scaler_indices) == 0:
+            return scaler_indices, False
+
+        variable_indexer = scaler_indices[-1]
+        requested_indices = self._maybe_to_index_list(variable_indexer)
+        if requested_indices is None:
+            return scaler_indices, False
+
+        pred_index_to_local = {index: pos for pos, index in enumerate(pred_indices)}
+        mapped_indices = [pred_index_to_local[index] for index in requested_indices if index in pred_index_to_local]
```
VeraChristina, line 290: maybe add a warning or log if indices are dropped here?

training/src/anemoi/training/losses/variable_mapper.py
```diff
@@ -0,0 +1,336 @@
...
+    def needs_shard_layout_info(self) -> bool:
+        """Whether the wrapped loss requires explicit shard-layout metadata."""
+        return getattr(self.loss, "needs_shard_layout_info", False)
+
+    def _get_predicted_indices_for_scaler_variable_axis(self, variable_size: int) -> list[int] | None:
+        if variable_size == 1:
+            # Broadcast scalers do not need filtering.
+            return None
+        if not self.predicted_indices_by_layout:
+            msg = "LossVariableMapper data indices must be set before adding variable scalers."
```
VeraChristina: This is only set when set_data_indices is called. Can this happen in the constructor? Or can we make more clear when the object is "fully initialised"?

training/src/anemoi/training/losses/loss.py
```diff
@@ -26,11 +30,62 @@
...
+        takes_scalers = "available_scalers" in context_keys
+        takes_data_indices = self.data_indices is not None and "data_indices" in context_keys
+
+        if takes_scalers:
+            constructor_kwargs["available_scalers"] = self.available_scalers
+
+        if takes_data_indices:
+            constructor_kwargs["data_indices"] = self.data_indices
+
+        return constructor_kwargs, takes_scalers, takes_data_indices
```
VeraChristina, line 58: Thanks for adding the context to recover the polymorphism! We could consider not returning a tuple but a more extensible context object but also happy with the current version

VeraChristina (commented):
Thanks for starting to make the tests more concise! I'll have a look at them later. Generally, this is looking good and Harrison's concerns regarding polymorphism of the loss class seems to be addressed. A couple of comments below

training/tests/integration/config/test_filtering.yaml
```diff
@@ -0,0 +1,84 @@
...
+        sample_idx: ${diagnostics.plot.sample_idx}
+        output_steps: 1
+        every_n_batches: ${diagnostics.plot.frequency.batch}
+        precip_and_related_fields: ${diagnostics.plot.precip_and_related_fields}
+        parameters: [tp]
+
+training:
+  max_epochs: 1
+  max_steps: 2
+
```
VeraChristina: Can we remove this? Or is there a reason to not use the general testing set-up as defined in testing_modifications?

training/tests/integration/config/test_filtering.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - dataset: ${system.input.dataset}   # dataset B (same file)
+    select: [tp]
+    rename:
+      tp: imerg
+  batch_size:
+    training: 1
+    validation: 1
+  limit_batches:
+    training: 1
+    validation: 1
```
VeraChristina: see above

training/tests/unit/diagnostics/mlflow/test_mlflow_utils.py
```diff
@@ -49,43 +49,3 @@ def test_expand_iterables_with_nested_dict() -> None:
     dictionary = {"a": {"b": ["a", "b", "c"]}}
     expanded = expand_iterables(dictionary)
     assert expanded == {"a": {"b": ["a", "b", "c"]}}
-
-
-def test_expand_iterables_with_nested_list() -> None:
```
VeraChristina, line 54: why remove these tests?

models/src/anemoi/models/data_indices/tensor.py
gabrieloks: **name_to_index** is the raw dataset mapping
**full_index_to_name** is the inverse of name_to_index - > index_to_name
**ordered_names** is the variables that are in the defined subspace (includes) and in the order they appear. The subspace could be diagnostic+prognostic variables for example
**self.name_to_position** new name_to_index mapping that reflects the real position of variables in **ordered_names**

training/src/anemoi/training/diagnostics/callbacks/plot.py
gabrieloks: Yes, the responsibility for variable-space alignment shifted from the caller (plot callback manually doing index_select) to the loss function (via IndexSpace layout). The batch tensor is now passed raw and the loss does the alignment.

models/src/anemoi/models/data_indices/tensor.py
OpheliaMiralles: I would not expand inference metadata yet, `name_to_index` seems right for inference

training/src/anemoi/training/losses/kcrps.py
OpheliaMiralles: We set "sum" as the default in forward args, the only change is that it is an explicit argument

training/src/anemoi/training/losses/variable_mapper.py
```diff
@@ -0,0 +1,336 @@
...
+            if indices and max(indices) >= variable_size:
+                continue
+            matches[layout] = indices
+
+        for preferred_layout in (IndexSpace.MODEL_OUTPUT, IndexSpace.DATA_OUTPUT, IndexSpace.DATA_FULL):
+            if preferred_layout in matches:
+                indices = matches[preferred_layout]
+                if indices == list(range(variable_size)):
+                    return None
+                return indices
```
VeraChristina, line 111: Are we confident that matching variable_size against the layouts in this order leads to the correct set of indices in cases where the different layout sizes coincide? (Are they the same indices when they have the same number of indices?) Do scalers match a unique index layout by number?

training/src/anemoi/training/losses/variable_mapper.py
OpheliaMiralles: rectified

training/src/anemoi/training/losses/variable_mapper.py
OpheliaMiralles: pred_layout / target_layout have now been set as explicit kwargs in forward() instead of mutating

training/src/anemoi/training/losses/combined.py
OpheliaMiralles: added LossFactoryContextKey in losses/base.py, switched the factory plumbing there, and added a test in test_filtered_loss.py

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1010,10 +1011,19 @@ def _plot(
...
-                loss = reduce_to_last_dim(self.loss[dataset_name](y_hat, y_true, squash=False).detach().cpu().numpy())
+                y_true = batch[dataset_name].narrow(1, start, pl_module.n_step_output)
+                loss = reduce_to_last_dim(
+                    self.loss[dataset_name](
+                        y_hat,
+                        y_true,
+                        squash=False,
+                        pred_layout=IndexSpace.MODEL_OUTPUT,
+                        target_layout=IndexSpace.DATA_FULL,
```
VeraChristina, line 664: Does this work for all different tasks, or does this need to be handled by the PlotAdapters?

VeraChristina (commented):
As far as I can see, tests look good. A few more comments below

training/src/anemoi/training/diagnostics/callbacks/plot.py
OpheliaMiralles: I never tried myself so maybe @gabrieloks ?

training/tests/integration/config/test_filtering.yaml
OpheliaMiralles:
```suggestion

```

training/tests/integration/config/test_filtering.yaml
```diff
@@ -0,0 +1,80 @@
...
```
OpheliaMiralles:
```suggestion
```

training/tests/unit/diagnostics/mlflow/test_mlflow_utils.py
gabrieloks: @ssmmnn11 @OpheliaMiralles ?

training/src/anemoi/training/losses/variable_mapper.py
gabrieloks: Yes. Layout sizes can only coincide when the variable sets are identical (e.g. MODEL_OUTPUT == DATA_OUTPUT only when there are no target-only variables). Since all layouts build their positions by sorting against the same global [name_to_index], identical variable sets produce identical indices.

    def _build_idx_from_list(self, var_list):
        sorted_variables = torch.Tensor(sorted(i for name, i in self.name_to_index.items() if name in var_list)).to(
            torch.int
        )
        return sorted_variables

training/tests/unit/diagnostics/mlflow/test_mlflow_utils.py
gabrieloks: I would say we can keep these tests. I see no reason for removing them

training/tests/integration/test_training_cycle.py
```diff
@@ -371,6 +371,15 @@ def test_training_cycle_mlflow_dry_run(
...
+    imerg_target_config: tuple[DictConfig, str],
+    get_test_archive: GetTestArchive,
+) -> None:
+    cfg, url = imerg_target_config
+    get_test_archive(url)
+    AnemoiTrainer(cfg).train()
+
+
 def test_training_cycle_multidatasets_diffusion(
     multidatasets_diffusion_config: tuple[DictConfig, list[str]],
```
VeraChristina, line 386:
```suggestion

@skip_if_offline
@pytest.mark.slow
def test_training_cycle_multidatasets_diffusion(
multidatasets_diffusion_config: tuple[DictConfig, list[str]],
```

VeraChristina (approved):
Thanks for addressing the feedback! LGTM

PR #811, opened 2026-01-21
feat(training): downscaling

JPXKQX:
## Description

This PR adds support for downscaling in anemoi. After multiple datasets, the downscaling task can be added to the supported use cases.

Thanks to @MeraX @TobiasGoeckeDWD, who co-developed this branch during their visit to Reading in October 2025.


training/src/anemoi/training/train/tasks/downscaler.py
```diff
@@ -0,0 +1,78 @@
...
+    ) -> tuple[torch.Tensor, Mapping[str, torch.Tensor]]:
+        loss = torch.zeros(1, dtype=next(iter(batch.values())).dtype, device=self.device, requires_grad=False)
+
+        x = self.get_inputs(batch, sample_length=self.multi_step)
+        y_pred = self(x)
+
+        y = self.get_targets(batch, lead_step=self.multi_step - 1)
+
+        # y includes the auxiliary variables, so we must leave those out when computing the loss
+        # Compute loss for each dataset and sum them up
```
MeraX: is this comment still up to date?

training/src/anemoi/training/train/tasks/downscaler.py
```diff
@@ -0,0 +1,78 @@
...
+                self.data_indices[dataset_name].data.output.full,
+            ]
+        return y
+
+    def _step(
+        self,
+        batch: dict,
+        validation_mode: bool = False,
+    ) -> tuple[torch.Tensor, Mapping[str, torch.Tensor]]:
+        loss = torch.zeros(1, dtype=next(iter(batch.values())).dtype, device=self.device, requires_grad=False)
```
MeraX:
```suggestion
```
This seems unused.

MeraX (commented):
Hi Mario, thanks for bringing this in! I was just thinking about the name of this feature/task. I was wondering if "downscaling" is the most generic term to describe what it actually provides.
If I'm not mistaken, I belief, the new task could basically be used for anything that predicts Y(t) from X(t) and former X if using multi step input. X and Y could be different datasets and grids, but they could also be the same. I.e. this new task also supports various possibilities of statistical post processing and generation of derived products. I wouldn't call all of these "downscaling". But to be honest, I have no good name in mind. Maybe something in the direction of "diagnosing task"?

training/src/anemoi/training/train/tasks/downscaler.py
JPXKQX: not sure, `y` should contain the prognostic and diagnostics variables here. Do you remember where this comment came from?

MeraX: I was more referring to "and sum them up". The `sum` went somewhere else.

I can't judge the `y` comment, but I have the feeling, that it could go to `compute_loss_metrics`.

training/src/anemoi/training/config/training/downscaling.yaml
```diff
@@ -0,0 +1,153 @@
...
+
+# Dataset-specific loss and metrics configuration
+training_loss:
+  datasets:
+    cerra:
+      # CERRA uses additional scalers
+      _target_: anemoi.training.losses.MSELoss
+      scalers: ['pressure_level', 'general_variable', 'nan_mask_weights', 'node_weights']
+      ignore_nans: false
+
```
MeraX, line 122: I'm trying to understand the implications of the new code and how https://github.com/ecmwf/anemoi-core/blob/9f5006693b51adc0a96295c509f39ec9fb62e0f0/training/src/anemoi/training/train/tasks/base.py#L648 will work together with this config.

* I assume that `self.dataset_names == ['ear5', 'cerra']` in  https://github.com/ecmwf/anemoi-core/blob/9f5006693b51adc0a96295c509f39ec9fb62e0f0/training/src/anemoi/training/train/tasks/base.py#L648 . So the loop `        for dataset_name in self.dataset_names:` would iterate over `era5`.
* However, from the above config, I would also conclude that `self.loss.keys == ['cerra'].
* But this would mean that `self.loss[dataset_name]` must fail with a KeyError for `era5` in https://github.com/ecmwf/anemoi-core/blob/9f5006693b51adc0a96295c509f39ec9fb62e0f0/training/src/anemoi/training/train/tasks/base.py#L529

What am I overseeing?

JPXKQX: I made some changes after rebasing the latest version of multiple-datasets, but I haven't run the code yet. I think the priority is to merge the autoencoder PR now

models/src/anemoi/models/interface/__init__.py
```diff
@@ -75,28 +76,77 @@ def __init__(
         self.supporting_arrays = supporting_arrays if supporting_arrays is not None else {}
         self.data_indices = data_indices
         self._build_model()
+        self._update_metadata()
 
-    def _build_model(self) -> None:
-        """Builds the model and pre- and post-processors."""
-        # Instantiate processors
+    def _build_processors_for_dataset(
```
mchantry, line 82: Why is the model interface changing?

JPXKQX: I will delete it now. The PR was pointing to the multiple-datasets branch, so I think this is a leftover from the merge

training/src/anemoi/training/train/tasks/forecaster.py
```diff
@@ -81,6 +81,7 @@ def _rollout_step(
                     ...,
                     self.data_indices[dataset_name].data.output.full,
                 ]
+
```
JPXKQX:
```suggestion
```

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -646,6 +650,10 @@ def compute_loss_metrics(
         # Prepare tensors for loss/metrics computation
         total_loss, metrics_next, y_preds = None, {}, {}
         for dataset_name in self.dataset_names:
+            if dataset_name not in self.target_dataset_names:
+                # Loss and metrics not defined for this dataset
+                continue
```
MeraX:
```suggestion
        for dataset_name in self.target_dataset_names:
```
would this be reasonable?

training/src/anemoi/training/train/tasks/single_step.py
```diff
@@ -96,29 +51,76 @@ def _step(
...
-
         for dataset_name, dataset_batch in batch.items():
             y_time = dataset_batch.narrow(1, 0, self.n_step_output)
-            var_idx = self.data_indices[dataset_name].data.output.full.to(device=dataset_batch.device)
-            y[dataset_name] = y_time.index_select(-1, var_idx)
+            var_indices = self.data_indices[dataset_name].data.output.full.to(device=dataset_batch.device)
+            y[dataset_name] = y_time.index_select(-1, var_indices)
+        return y
+
+    def _step(
```
anaprietonem, line 70: two functions called _step?

training/src/anemoi/training/train/tasks/single_step.py
```diff
@@ -96,29 +51,76 @@ def _step(
...
-    def on_train_epoch_end(self) -> None:
-        pass
+
+class GraphDownscaler(BaseSingleStepGraphModule):
+    """Graph neural network downscaler for PyTorch Lightning."""
+
+    task_type = "downscaler"
+
+
+class GraphAutoEncoder(BaseSingleStepGraphModule):
```
anaprietonem, line 98: likely this need some work to harmonise bring the implementation provided by Alberto here?

training/src/anemoi/training/schemas/training.py
```diff
@@ -352,6 +352,7 @@ def check_length_of_weights_and_losses(self) -> Self:
     | SpectralLossSchema
     | MultiScaleLossSchema
     | None
+    | None
```
anaprietonem, line 355: why double None?

training/src/anemoi/training/train/tasks/single_step.py
anaprietonem: Also we should check if we are breaking the autoencoder checkpoints if we move the class here. Not sure if we even have a test at moment to check for this one particularly

training/src/anemoi/training/config/downscaling.yaml
```diff
@@ -0,0 +1,79 @@
...
+  resolution: o96
+  datasets:
+    era5:  # low-res dataset: all variables should be forcing variables
+      diagnostic: []
+      processors:
+        normalizer:
+          _target_: anemoi.models.preprocessing.normalizer.InputNormalizer
+          config:
+            std: null
+    cerra:  # high-res dataset: all variables should be diagnostic or forcing variables
```
anaprietonem, line 23: should we have specific data modules for downscaling to assert these conditions or how we could we do this?

training/tests/integration/config/test_downscaling.yaml
```diff
@@ -0,0 +1,58 @@
...
+        start: 2017-01-08 18:00:00
+      cerra:
+        start: 2017-01-08 18:00:00
+
+data:
+  datasets:
+    cerra:
+      diagnostic: ["10u", "10v", "2t", "tp"]
+
+diagnostics:
```
anaprietonem, line 26: Is this working? I would have imagined downscaling plotting is similar to the autoencoder where you're xt and yt is the same and hence you need some adjustments for the plotting functions. (I was looking at this here https://github.com/ecmwf/anemoi-core/tree/feat/plotting_autoencoder)

PR #816, opened 2026-01-22
feat: diffusion tendency forecasting support for multi output steps

ssmmnn11:
Support for diffustion tendency forecaster and multiple output steps


ssmmnn11:
the datasets have tendency statistics for multiple timesteps, such as 6h, 12h - but it needs to be specified per dataset.

PR #824, opened 2026-01-23
feat: Add first tests for spectral helpers

samhatfield:
#678 introduced the initial infrastructure for computing spectral transforms. #788 and #709 will expand on this and implement several types of transform, including spherical harmonic transforms.

All of these transforms should have tests implemented. One crucial test for a transform and its inverse is idempotency, i.e. do I get the same thing when I perform the inverse followed by the direct transform, and vice versa? This PR implements idempotency tests for `EcTransOctahedralSHTModule` and its inverse. This will hopefully serve as a reference for implementing idempotency tests for other transforms.

`test_idempotency_inverse_direct` is hopefully self-explanatory. `test_idempotency_direct_inverse` has one nasty gotcha, which I've tried to work around in a somewhat suboptimal way. This test starts in grid point space and transforms to spectral space and back. The input is a random field on the octahedral grid. This field will contain high-frequency modes above the spectral band limit (set by the truncation) and which will therefore be stripped out when the direct transform is computed. When the inverse is then computed, these modes will not be restored and the before and after arrays will therefore never match.

To work around this, for this test I actually randomise a spectral field and then inverse transform it to grid point space. This inverse transformed field is the input to the idempotency test, and will by definition not contain any modes above the band limit. It's not ideal, because we are using the inverse transform in order to initialise a test _for_ the inverse transform. But I can't think of any other simple way to initialise the input to this test.

I'm not a pytest expert so please feel free to suggest improvements to the test suite.

Also note that this test requires an installation of [ectrans4py](https://sites.ecmwf.int/docs/ectrans/page/ectrans4py.html) in order to initialise `EcTransOctahedralSHTModule`. This dependency will eventually be removed, once I can figure out how to cache the matrices which are computed by ectrans4py.

### DIFF
--- a/models/tests/layers/test_spectral_helpers.py
+++ b/models/tests/layers/test_spectral_helpers.py
@@ -7,8 +7,8 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-import torch
 import pytest
+import torch
 
 from anemoi.models.layers.spectral_helpers import EcTransOctahedralSHTModule
 from anemoi.models.layers.spectral_helpers import InverseEcTransOctahedralSHTModule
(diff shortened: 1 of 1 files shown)


samhatfield:
Update: we have decided not to implement the ecTrans version of the octahedral SHT, but rather a generalised SphericalHarmonicTransform, which in future may have the option of deriving its Legendre polynomials from ecTrans. This PR is no longer necessary, but the tests have been taken and used in PR https://github.com/ecmwf/anemoi-core/pull/788.

PR #827, opened 2026-01-23
feat: obsinterpolator

OpheliaMiralles:
## Description
 Imbalanced version of the interpolator 

New feature, allows some inputs to be known in the future (e.g. NWP forecastst) and some not (observations) in order to "interpolate" between two dense atmospheric states.

#636 


dietervdb-meteo (changes requested):
I think you still need to add the pydantic schema's?

Also would be great to have some integration tests. First would maybe be just to have a default config and do the config integration check as for other tasks. Second could be to provide a toy dataset and then have an integration test checking an actual training run. @VeraChristina can probably help you with this.

PR #830, opened 2026-01-26
feat(migrations): remove migration rollback

bdvllrs:
## Description
Migration rollbacks are not used. Worse, the rollback function are defined in migration script but without being implemented.
This means that the migrator will think that all of the migration script can safely be rollbacked when actually they cannot.

I should have probably made the rollback function opt-in in the migration scrip generation CLI rather than opt-out.

Anyway, because the functionality is not used, this PR removes it: less code to maintain and less confusion with what rollbacks are and avoid weird bugs with Python 3.13 and cloudpickle.

This also removes an additional dependency for migrations.

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--830.org.readthedocs.build/en/830/

----
📚 Documentation preview 📚: https://anemoi-graphs--830.org.readthedocs.build/en/830/

----
📚 Documentation preview 📚: https://anemoi-models--830.org.readthedocs.build/en/830/


anaprietonem (approved):
LGTM, Thanks @bdvllrs !

I have triggered the integration tests https://github.com/ecmwf/anemoi-core/actions/runs/21623094478 - after those pass it should be good to get this one in

PR #834, opened 2026-01-26
chore: multi output steps fix interpolator [step 1]

dietervdb-meteo:
This is a first step in adding incremental changes after a revert of the multi output steps PR #636 


models/src/anemoi/models/models/interpolator.py
```diff
@@ -49,14 +49,18 @@ def __init__(
...
-        target_forcing_config = get_multiple_datasets_config(model_config.training.target_forcing)
-
-        self.num_target_forcings = {}
-        for dataset_name, target_forcing in target_forcing_config.items():
-            self.num_target_forcings[dataset_name] = len(target_forcing.data) + target_forcing.time_fraction
-
         self.input_times = model_config.training.explicit_times.input
         self.output_times = model_config.training.explicit_times.target
+
+        self.multi_out = model_config.training.multistep_output
```
dietervdb-meteo: This can probably be removed as it is set in identical fashion by the super init? "If" statement below can then go after super init?

mc4117: for some reason if I move this it breaks

mc4117: I fixed this by subclassing now

PR #839, opened 2026-01-27
fix(models): merged main with autoencoder

ssmmnn11:
Merged main with autoencoder multi out support

----
📚 Documentation preview 📚: https://anemoi-training--839.org.readthedocs.build/en/839/

----
📚 Documentation preview 📚: https://anemoi-graphs--839.org.readthedocs.build/en/839/

----
📚 Documentation preview 📚: https://anemoi-models--839.org.readthedocs.build/en/839/


models/src/anemoi/models/models/autoencoder.py
```diff
@@ -101,10 +104,11 @@ def _assemble_input(self, x, batch_size, grid_shard_shapes=None, model_comm_grou
             )
             node_attributes_data = shard_tensor(node_attributes_data, 0, shard_shapes_nodes, model_comm_group)
 
+        x_input = x[:, : self.multi_step, ...]
```
mc4117, line 107: Should there be a flag if multi_step and multi_out are not equal to each other? I feel this is important for the autoencoder? Maybe this flag already exists?

dietervdb-meteo: @icedoom888 , what do you think?

models/src/anemoi/models/models/autoencoder.py
```diff
@@ -0,0 +1,271 @@
...
+    def _assemble_forcings(self, x, batch_size, grid_shard_shapes=None, model_comm_group=None, dataset_name=None):
+        assert dataset_name is not None, "dataset_name must be provided when using multiple datasets."
+        node_attributes_target = self.node_attributes[dataset_name](self._graph_name_data, batch_size=batch_size)
+        if grid_shard_shapes is not None:
+            shard_shapes_nodes = get_or_apply_shard_shapes(
+                node_attributes_target, 0, grid_shard_shapes, model_comm_group
+            )
+            node_attributes_target = shard_tensor(node_attributes_target, 0, shard_shapes_nodes, model_comm_group)
+
+        x_forcing = x[:, : self.multi_out, ...]
```
dietervdb-meteo, line 150: Bit confused about this line... If such a selection of time steps was necessary, why isn't it present in main? (In that case there is only one (output) timestep and one does not seem to need to select that timestep?) Maybe `x` already has the shape (batch, n_output_steps, ... ) or maybe I'm missing something?

mc4117: I think this is because these are the forcings that get passed to the decoder. In main we don't have this time dimension in the output but now we do?

models/src/anemoi/models/models/autoencoder.py
mc4117: I added a flag so you can see how it would be implemented @icedoom888

models/src/anemoi/models/models/autoencoder.py
dietervdb-meteo: Right! Working so long on multi-out now I forgot main still doesn't have the extra time dimension :)

models/src/anemoi/models/models/autoencoder.py
icedoom888: Sorry i don't think i get this: by definition AE has input=output, any change to this premise transforms this to a mapper. 

multi_step and multi_out should always be equal to each other!

mc4117: yes this is why I put an assert statement in to make sure multi_step and multi_out are always equal to each other

mc4117: https://github.com/ecmwf/anemoi-core/blob/feat/multi-output-steps-onwards_autoencoder/training/src/anemoi/training/train/tasks/autoencoder.py#L75

icedoom888: Oh you meant an assert since the beginning?

mc4117: yes from the beginning for the autoencoder. We also have a similar assert for the GraphInterpolator to differentiate from the GraphMultiOutInterpolator

PR #840, opened 2026-01-28
fix: filteringlosswrapper with target variables

gabrieloks:
## Description
Working fix for loss filtering with target variables
##  Additional notes ##

### DIFF
--- a/training/tests/unit/losses/test_filtering_with_target_only_variables.py
+++ b/training/tests/unit/losses/test_filtering_with_target_only_variables.py
@@ -173,7 +173,7 @@ class TestCombinedLossWithTargetOnlyVariables:
 
     @pytest.fixture
     def data_indices_with_imerg(self):
-        """IndexCollection multiple variables and imerg (obs) as target-only."""
+        """IndexCollection with multiple variables and imerg as target-only."""
         data_config = {
             "data": {
                 "forcing": [],
@@ -194,7 +194,7 @@ class TestCombinedLossWithTargetOnlyVariables:
 
     @pytest.fixture
     def scalers_custom(self):
-        """Create scalers for weather variables."""
+        """Create scalers with custom values per variable."""
         n_vars = 6
         return {
             "pressure_level": (3, torch.ones(n_vars) * 2.0),
@@ -238,7 +238,7 @@ class TestCombinedLossWithTargetOnlyVariables:
         assert isinstance(loss.losses[0], FilteringLossWrapper)
         assert isinstance(loss.losses[1], FilteringLossWrapper)
 
-        # First loss: 5 weather variables
+        # First loss: 5 variables
         first_loss_scaler = loss.losses[0].loss.scaler.tensors["pressure_level"][1]
         assert first_loss_scaler.shape[0] == 5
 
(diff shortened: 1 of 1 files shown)


training/src/anemoi/training/train/tasks/base.py
```diff
@@ -849,7 +867,9 @@ def calculate_val_metrics(
 
         y_postprocessed = post_processor(y, in_place=False)
         y_pred_postprocessed = post_processor(y_pred, in_place=False)
+        if y_postprocessed.shape[3] != y_pred_postprocessed.shape[3]:  # maybe 3 shouldn't be hardcoded
 
+            y_postprocessed = y_postprocessed[..., self.mask_target]
```
HCookie, line 872: Definitely not, is it to index the variable dimension? If so we have an enum for this

training/tests/unit/losses/test_filtering_with_target_only_variables.py
```diff
@@ -0,0 +1,516 @@
...
+    @pytest.fixture
+    def weather_scalers(self):
+        """Create scalers for weather variables."""
+        n_vars = 6
+        return {
+            "pressure_level": (3, torch.ones(n_vars) * 2.0),
+            "general_variable": (3, torch.tensor([1.0, 0.5, 0.5, 0.5, 0.8, 10.0])),
+        }
+
+    def test_combined_loss_with_target_only_variable(self, data_indices_weather_with_imerg, weather_scalers):
```
HCookie: Did you tell the LLM to make the tests weather related?

gabrieloks: hahahah bad claude

gabrieloks:
> I am not sure to be a fan of the idea of inserting an additional indexing layer when all losses can be wrapped by default with predicted indices by the model. I think if this is an additional flexibility to compare predicted variables to any variable in the input data it should be somehow embedded in the way we compute the loss and not a quick fix at the task level. Are you sure there is no way to make it work in a closer way to the original PR?


Thanks for the feedback. I understand the concern about quick fixes at the task level, but I think having a working solution with good test coverage is a reasonable first step. It makes any deeper refactoring, if needed, safer to attempt in the future.


The issue isn't really limited to the losses, it's more about whether the target gets reduced from the start or not. The architecture seems to have been designed with early reduction in mind, and several things break without it. I already had to fix the scalers and normalizers to handle the potential shape discrepancy between targets and predictions when target-only variables are used, so I preferred to keep things on the safe side rather than changing the core behaviour. So I would prefer not to reduce the targets, but if we can show that the scalers, normalizers and validation can work without big changes, even if the targets are not reduced, I'm down for it. I would prefer more a surgical approach here. The idea I had here was to try and fix the filtering but minimising the changes everywhere else.

The reindexing isn't really an additional layer, it's just accounting for how target variables get reduced via [data_indices.data.output.full]. When you have original indices like [0, 1, 2, 3, 8, ...] after reduction, they need to be mapped to sequential positions [0, 1, 2, 3, 4, ...] to match the actual tensor shape. That's all this does.

The tests I added document the expected behaviour, but I'm open to discussion. Also if you think there's an easier way of doing this that I might have missed, feel free to go with it

OpheliaMiralles:
> > I am not sure to be a fan of the idea of inserting an additional indexing layer when all losses can be wrapped by default with predicted indices by the model. I think if this is an additional flexibility to compare predicted variables to any variable in the input data it should be somehow embedded in the way we compute the loss and not a quick fix at the task level. Are you sure there is no way to make it work in a closer way to the original PR?
> 
> Thanks for the feedback. I understand the concern about quick fixes at the task level, but I think having a working solution with good test coverage is a reasonable first step. It makes any deeper refactoring, if needed, safer to attempt in the future.
> 
> The issue isn't really limited to the losses, it's more about whether the target gets reduced from the start or not. The architecture seems to have been designed with early reduction in mind, and several things break without it. I already had to fix the scalers and normalizers to handle the potential shape discrepancy between targets and predictions when target-only variables are used, so I preferred to keep things on the safe side rather than changing the core behaviour. So I would prefer not to reduce the targets, but if we can show that the scalers, normalizers and validation can work without big changes, even if the targets are not reduced, I'm down for it. I would prefer more a surgical approach here. The idea I had here was to try and fix the filtering but minimising the changes everywhere else.
> 
> The reindexing isn't really an additional layer, it's just accounting for how target variables get reduced via [data_indices.data.output.full]. When you have original indices like [0, 1, 2, 3, 8, ...] after reduction, they need to be mapped to sequential positions [0, 1, 2, 3, 4, ...] to match the actual tensor shape. That's all this does.
> 
> The tests I added document the expected behaviour, but I'm open to discussion. Also if you think there's an easier way of doing this that I might have missed, feel free to go with it

Of course I did not mean to be dismissive of your changes or anything, I’m actually glad someone cares about this :). I’d just be grateful if you could send the exact config you are using in a flatten format by dm on slack and I’ll try to reproduce the issue see if I can see why wrapping at the loss level is trickier than it looks (I can use different data no problem). If your changes are actually more minimal and they pass, we would go your way. Is this ok for you?

training/src/anemoi/training/train/tasks/base.py
OpheliaMiralles: Var dim could also just be -1, this one rarely changes

PR #842, opened 2026-01-28
fix: set min version for azureml-mlflow

cathalobrien:
## Description
This PR sets a minimum version for the azureml-mlflow. 

This PR is to fix a failing integration test (see [here](https://github.com/ecmwf/anemoi-core/actions/runs/21439322518/job/61737956627#step:3:1020)). if this minimum dependency is not set, pip resolves to use a very old version. This pulls in an old version of mlflow which isn't supported on the python version used in the integration tests.

When the version is pinned, the integration tests pass (see [here](https://github.com/ecmwf/anemoi-core/actions/runs/21439567578/job/61738788024#step:3:694))

I picked the newest azure-mlflow major release v1.6.0. The failing version was v1.0.74. Probably there is a lower minimum version which would work. But the package isnt open source so I dont know how to check which version of mlflow it would pull in without running an integration test for each version, trial-and-error style. We can always revert to a lower version if there is a problem later :)


anaprietonem:
@phinate  @timothyas  we need to do this change for some errors on the CI we are seeing is it okey to bump this  minimum version on your side?

PR #848, opened 2026-01-30
refactor(training)!: remove `grid_indices` inside Dataset

JPXKQX:
## Description

This PR removes the dependency of `grid_indices` in the graphs. The current dependency graph is defined below:

image

This will allow graphs to be created later (deeper) in the workflow.

The recommended way to mask the LAM configurations is to use the 'max_radius_km: 999.9' option in the dataset, rather than using the post-processor in the graph and referencing it later in the data loader configuration.

The PR is marked as breaking because the LAM config files will change. `grid_indices` is no longer used in the dataloader configuration. The LAM checkpoints (trained with `MaskedGrid`) won't be available for transfer learning / checkpoint migration.

#### main
```
avThroughputIterPerS: 0.66iter/s (date: 2026-02-06, commit: 99aaf715bd423acfc21a1794e231acdd7ee37ac4)
peakMemoryMB: 15430.44MB (date: 2026-02-06, commit: 99aaf715bd423acfc21a1794e231acdd7ee37ac4)
```
#### refactor/remove-grid-indices
```
avThroughputIterPerS: 0.65iter/s (date: 2026-02-09, commit: 3fe8ebe760b052b7867964bb30b883dabf562629 )
peakMemoryMB: 15430.44MB (date: 2026-02-09, commit: 3fe8ebe760b052b7867964bb30b883dabf562629)
```

----
📚 Documentation preview 📚: https://anemoi-training--848.org.readthedocs.build/en/848/

----
📚 Documentation preview 📚: https://anemoi-graphs--848.org.readthedocs.build/en/848/

----
📚 Documentation preview 📚: https://anemoi-models--848.org.readthedocs.build/en/848/

### DIFF
--- a/training/tests/integration/conftest.py
+++ b/training/tests/integration/conftest.py
@@ -243,8 +243,12 @@ def lam_config_with_graph(
     cfg, urls = lam_config
     cfg.graph = existing_graph_config
 
+    dataset_name = "data"  # default dataset name
     url_graph = "anemoi-integration-tests/training/graphs/lam-graph-2026-02-19.pt"
-    cfg.system.input.graph = Path(get_test_data(url_graph))
+    tmp_path_graph = Path(get_test_data(url_graph))
+    dataset_graph_filename = tmp_path_graph.name.replace(".pt", f"_{dataset_name}.pt")
+    tmp_path_graph.rename(tmp_path_graph.parent / dataset_graph_filename)
+    cfg.system.input.graph = tmp_path_graph
     cfg.diagnostics.plot.callbacks = []  # remove plotting callbacks as they are tested in lam training cycle test
     return cfg, urls
 


training/src/anemoi/training/data/multidataset.py
```diff
@@ -87,7 +78,10 @@ def __init__(
             ", ".join(self.dataset_names),
             len(self.valid_date_indices),
         )
+        self._lazy_init_model_and_reader_group_info()
 
+    def _lazy_init_model_and_reader_group_info(self) -> None:
```
anaprietonem, line 82: was this missing before or why the change now?

training/src/anemoi/training/data/dataset.py
```diff
@@ -89,6 +103,21 @@ def resolution(self) -> str:
         """Return dataset resolution."""
         return self.data.resolution
 
+    @cached_property
+    def cutout_mask(self) -> np.ndarray:
+        """Return cutout mask."""
+        cutout_mask = np.zeros(self.grid_size, dtype=bool)
+        if len(self.data.grids) <= 1:
```
anaprietonem, line 122: can't we check here simp,y if grids is None or empty?

training/src/anemoi/training/data/dataset.py
```diff
@@ -89,6 +103,21 @@ def resolution(self) -> str:
...
+            err_msg = "Dataset `cutout_mask` property requires a cutout grid but does not have one."
+            raise ValueError(err_msg)
+        cutout_mask[: self.data.grids[0]] = True
+        return cutout_mask
+
+    @cached_property
+    def boundary_mask(self) -> np.ndarray:
+        """Return boundary mask."""
```
anaprietonem: think this could use a more descriptive docstring of where is that boundary mask used or why it's useful

training/src/anemoi/training/data/multidataset.py
JPXKQX: I was moving those lines into a method, but happy to leave as it was if you think it doesn't improve clarity

training/src/anemoi/training/data/dataset.py
JPXKQX: `grids` property is never None, or empty. If you `ds=era5-o96`, then `ds.grids = (40000, )`.

training/src/anemoi/training/config/lam.yaml
```diff
@@ -26,12 +26,7 @@ dataloader:
       - dataset: ${system.input.forcing_dataset}
     adjust: all
     min_distance_km: 0
-  grid_indices:
-    datasets:
-      data:
-        _target_: anemoi.training.data.grid_indices.MaskedGrid
-        nodes_name: data
-        node_attribute_name: indices_connected_nodes
+    max_distance_km: 300
```
anaprietonem, line 29: this probably need to better documented

training/src/anemoi/training/distributed/strategy.py
```diff
@@ -226,6 +228,23 @@ def _setup_communication_groups(self) -> int:
 
         return model_comm_group_id
 
+    def _setup_shard_shapes(self, trainer: pl.Trainer) -> dict:
```
anaprietonem, line 231: TODO - in the future create base strategy class to reduce this

mchantry:
Please can you clarify the nature of the breaking change.
Configs, checkpoints (can they be ported), ....

training/src/anemoi/training/distributed/strategy.py
JPXKQX: Should I start with something minimal in this PR?

JPXKQX: proposal in #882

training/src/anemoi/training/config/lam.yaml
JPXKQX: Added to `docs/`

dietervdb-meteo, line 29: I think it would make sense to put this closer to the actual boundary size used by the graph. In the graph config `margin_radius_km=10`. I'm not sure what the cutoff factor provides as lenght (I assume a few tens of kms?), say `l`. I would then put `max_distance_km = ceil(10+l)`

JPXKQX: That should take into account the `CutOffEdges` configuration, right?

dietervdb-meteo: Yes, when building the graph you can see to what length `l` the cutoff factor translates.

dietervdb-meteo: I.e. total size of LAM boundary is extent of hidden (`margin_radius_km`) plus length of cuttoff edges outside regional domain (`l`).

dietervdb-meteo: We could possibly mention a suggestion along those lines in the docs?

JPXKQX: yes, sorry, I went too fast over your comment

JPXKQX: Should we use something like 120? This is just in case someone uses the o96 boundary so that it includes at least one set of nodes

dietervdb-meteo: Ok, it is a good point this now will depend on what dataset one actually plans to use. I think it makes sense to put it to a value such that one includes indeed at least a (few) row(s) of o96 boundary nodes. So maybe first fix `max_distance_km` like that, then see if the cutoff factor chosen in the lam graph config connects to most of these boundary nodes and adapt cutoff factor to that. (Rather than adapating `max_distance_km` to cutoff factor as I first suggested)

anaprietonem (approved):

Issue #546, opened 2025-09-16
Missing W&B config converstion


monte-flora:
### What happened?

When wandb logger is enabled, an issue with OmegaConf arises. 

### What are the steps to reproduce the bug?

Looking over anemoi.training.diagnostics.logger.py (on this line https://github.com/ecmwf/anemoi-core/blob/caf3478389de78b8a6bd1c24acf6e96ebc283fe1/training/src/anemoi/training/diagnostics/logger.py#L174), it can use the same `convert_to_omegaconf` as for the mlflow package on line above (https://github.com/ecmwf/anemoi-core/blob/caf3478389de78b8a6bd1c24acf6e96ebc283fe1/training/src/anemoi/training/diagnostics/logger.py#L94). When I copied over that conversion, I was able to successfully log to my W&B account. 

This bug is timely with the recent issue raised in #538 

### Version

latest 

### Platform (OS and architecture)

Linux 

### Relevant log output

```shell

```

### Organisation

The Weather Company

floriankrb:
Well spotted, a PR implementing this change would improve the current code and would be welcome.

Issue #538, opened 2025-09-10
Review support for W&B and Tensorboard


anaprietonem:
### What maintenance does this project need?

We currently support 3 loggers:
- W&B
- tensorboard
- mlflow

Mlflow is the one we consider the majority of the anemoi community is using. We would like then to revisit the need to support the 3 of them, where the proposal could be:
- focus on mlflow - try adding better test for that logger
- 2nd tier ideally targeting research/academia users for which w&B can be useful
- deprecate tensorboard

Feedback or thoughts are welcomed!

PR #850, opened 2026-01-30
feat(training): Deprecate tensorboard and Harmonise loggers interface

anaprietonem:
## Description
PR to clean up loggers:
- Deprecate tensorboard as proposed in https://github.com/ecmwf/anemoi-core/issues/538
- Fix Weight Biases and harmonise the interface - https://github.com/ecmwf/anemoi-core/issues/546
- Add specific catch for Mlflow Postgresql error 

Changes here include proposed changes in https://github.com/ecmwf/anemoi-core/pull/530 (cc: @floriankrb )

[](https://github.com/ecmwf/anemoi-core/issues/538)

[
](https://github.com/ecmwf/anemoi-core/issues/538)

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/schemas/training.py
+++ b/training/src/anemoi/training/schemas/training.py
@@ -16,6 +16,7 @@ from typing import Self
 
 from pydantic import AfterValidator
 from pydantic import BaseModel as PydanticBaseModel
+from pydantic import ConfigDict
 from pydantic import Discriminator
 from pydantic import Field
 from pydantic import NonNegativeFloat
@@ -82,6 +83,8 @@ class LR(BaseModel):
 
 class OptimizerSchema(PydanticBaseModel):
     """Choosing the PydanticBaseModel to allow extra inputs."""
+
+    model_config = ConfigDict(extra="allow")
 
     target_: str = Field(..., alias="_target_")
     """Full path to the optimizer class, e.g. `torch.optim.AdamW`."""
--- a/training/src/anemoi/training/train/tasks/base.py
+++ b/training/src/anemoi/training/train/tasks/base.py
@@ -1046,7 +1046,14 @@ class BaseGraphModule(pl.LightningModule, ABC):
         if hasattr(opt_cfg, "model_dump"):
             opt_cfg = opt_cfg.model_dump(by_alias=True)
 
-        return instantiate(opt_cfg, params=params, lr=self.lr)
+        optimizer = instantiate(opt_cfg, params=params, lr=self.lr)
+
+        # Log the actual optimizer settings to help users verify configuration
+        defaults_to_log = {k: v for k, v in optimizer.defaults.items() if k != "params"}
+        LOGGER.info("Optimizer initialized: %s", type(optimizer).__name__)
+        LOGGER.info("Optimizer settings: %s", defaults_to_log)
+
+        return optimizer
 
     def _create_scheduler(self, optimizer: torch.optim.Optimizer) -> dict[str, Any]:
         """Helper to create the cosine LR scheduler."""
--- a/training/tests/unit/schemas/test_allowed_extra_keys.py
+++ b/training/tests/unit/schemas/test_allowed_extra_keys.py
@@ -1,0 +1,33 @@
+# (C) Copyright 2026- Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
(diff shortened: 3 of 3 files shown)


training/src/anemoi/training/config/diagnostics/evaluation.yaml
HCookie, line 1: Given the shared config here, would this make sense another layer down?

training/src/anemoi/training/diagnostics/logger.py
```diff
@@ -97,33 +80,44 @@ def get_wandb_logger(config: DictConfig, model: pl.LightningModule) -> pl.logger
...
+
+    if not wandb_config.enabled:
         LOGGER.debug("Weights & Biases logging is disabled.")
         return None
 
+    wandb_config = OmegaConf.to_container(wandb_config)
+    del wandb_config["gradients"]
+    del wandb_config["parameters"]
+    del wandb_config["enabled"]
+    del wandb_config["interval"]
```
HCookie: That's a bit rough, would conversion and then popping off the dict make more sense?

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -340,13 +344,30 @@ class WandbSchema(BaseModel):
...
+        for extra_field, value in extras.items():
+            if extra_field not in allowed_extras:
+                msg = f"Extra field '{extra_field}' is not allowed. Allowed fields are: {list(allowed_extras.keys())}."
+                raise ValueError(msg)
+            if not isinstance(value, allowed_extras[extra_field]):
+                msg = f"Extra field '{extra_field}' must be of type {allowed_extras[extra_field].__name__}."
+                raise TypeError(msg)
+
+        return self
+
```
HCookie: Why is this needed, the `interval` is set as a default, so not required anyway.

anaprietonem: Config backward compatibility - still allow old config that don't have the internal under the w&b section to work

training/src/anemoi/training/diagnostics/logger.py
anaprietonem: Yes I will try to look at it and see if I can improve that

training/src/anemoi/training/config/diagnostics/evaluation.yaml
anaprietonem: Hmm as in a logger subfolder?

training/src/anemoi/training/schemas/diagnostics.py
HCookie: But it will be already, as `interval` is set by default by pydantic

training/src/anemoi/training/diagnostics/logger.py
anaprietonem: I updated this now!

training/src/anemoi/training/schemas/diagnostics.py
anaprietonem: Point understood, updated it now

training/src/anemoi/training/config/diagnostics/evaluation.yaml
anaprietonem: Done!

anaprietonem:
Yes so the ticket was initially discussed at ATS and that's the one that had the label https://github.com/ecmwf/anemoi-core/issues/538. Still it's true the PR could have used the ATS approved label too to be more consistent. 

Regarding your question, it's an imperfect system at the moment and very much dependant on how you have your configs set up. For example if you a top-level config and just overwrite there some parameters - likely you don't get errors cause you get the defaults directly from the source folder (if you get them if should be broken configs in purpose for the user to be aware something major has changed). If you don't use the repo defaults or have a different set up, I appreciate this process is a painful one at the moment. A way to usually get a feeling of what has changed is look at changes in the AICON integration test. 
Screenshot 2026-02-18 at 09 18 04

The migration system is a good idea. But note with checkpoints we don't migrate them on the fly either (it's still the user responsibility to migrate and understand those migrations).

icedoom888:
@anaprietonem FYI no one (except ECMWF probably) is running with configs built within the anemoi-core config folder.
And inheriting from defaults only, completely defies the purpose of using hydra...
Most of the users have their configs either completely unrolled, or inheriting from a structure of defaults built for the specific center of the experiment at hand. 

Every time a config change is introduced, it triggers a 1/2 hours effort to migrate or change all existing configs (without even mentioning the versioning issue: experiment X was trained using config Y and versions Z, now config Y does not work with version Z.1 etc etc)..

I think we should really prioritize a migration function that the user can trigger:
`anemoi-training update-config <config_path>`

icedoom888:
Looking at AICON integration test it's a nice trick, but you can agree this is not really sustainable and doesn't really help out scaling changes to 100 configs

PR #854, opened 2026-02-02
fix(models, graphs): models and graphs tests on gpu

ssmmnn11:
Test pass when run on GPU node


models/tests/layers/mapper/test_graphconv_mapper.py
```diff
@@ -24,6 +24,11 @@
 from anemoi.utils.config import DotDict
 
 
+@pytest.fixture(scope="module")
```
anaprietonem: wouldn't it be useful to have this on graphs too?

models/tests/layers/mapper/test_graphconv_mapper.py
```diff
@@ -24,6 +24,11 @@
...
+def device():
+    return torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
anaprietonem: And I think if we move the fixture to a conftest.py in models then you should be able to reuse it across the files!

PR #855, opened 2026-02-03
feat(training, models): graph config for skipped noise injector and multiscale loss

ssmmnn11:
Description
This PR adds new configuration support in the training and models components to enable graph-based settings for skipped noise injection and multiscale loss integration.

- Main changed

- Introduces graph configuration options to control:
  - Noise injector behavior when using graph-based models
  - Multiscale loss projections across graph representations

- Adds associated tests covering:
  - Graph-based multiscale loss scales
  - Truncation and projection behavior

- Implements support for noise projection in graph-based models
- Includes additional fixes and refactors to improve graph config handling

- Why?

These changes enhance flexibility in defining how graph-structured models handle noise injection and multiscale loss terms during training, leveraging existing functionality in anemoi-graphs to use sparse matrices generation. Facilitating reproducibility via graph recipes and reducing dependency on truncation files graph not properly tracked. 
In addition, it will enable more involved interpolation workflows in the future. 

----
📚 Documentation preview 📚: https://anemoi-training--855.org.readthedocs.build/en/855/

----
📚 Documentation preview 📚: https://anemoi-graphs--855.org.readthedocs.build/en/855/

----
📚 Documentation preview 📚: https://anemoi-models--855.org.readthedocs.build/en/855/


models/src/anemoi/models/layers/ensemble.py
```diff
@@ -227,8 +247,11 @@ def __init__(
             Layer kernel configurations
         noise_matrix : str, optional
             Optional path to noise truncation matrix
+        graph_data : Optional[HeteroData], optional
+            Unused. Accepted for interface compatibility, by default None.
         """
         super().__init__()
+        _ = graph_data
```
MeraX: What is `_`used for? Is this a convention for unused variables? 

Would it be worth raising an Exception if a dismissed `graph_data` argument is given?

graphs/src/anemoi/graphs/nodes/builders/from_reference.py
```diff
@@ -0,0 +1,50 @@
...
+import torch
+from torch_geometric.data import HeteroData
+
+from anemoi.graphs.nodes.builders.base import BaseNodeBuilder
+from anemoi.graphs.utils import get_grid_reference_distance
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ReferenceNodes(BaseNodeBuilder):
```
MeraX, line 22: Could you provide more context explaining the purpose of this class? It might be obvious the the developer, but I would not understand, why this class is needed.

training/docs/user-guide/models.rst
```diff
@@ -191,15 +197,88 @@ The truncation system supports several types of transformation matrices:
...
+``gauss_weight`` in the projection graph).
+
+.. code:: yaml
+
+   # graph config
+   graph:
+     projections:
+       multiscale:
+         matrices:
+           - edges_name: [smooth_1x, "to", smooth_1x]
```
MeraX:
```suggestion
           - edges_name: ["smooth_1x", "to", "smooth_1x"]
```
I think it would be less confusing if all strings were handled the same.

training/docs/user-guide/models.rst
```diff
@@ -191,15 +197,88 @@ The truncation system supports several types of transformation matrices:
...
+             edge_weight_attribute: gauss_weight
+           - edges_name: [smooth_2x, "to", smooth_2x]
+             edge_weight_attribute: gauss_weight
+         nodes: ...
+         edges: ...
+       residual:
+         down_edges_name: [${graph.data}, "to", truncation]
+         up_edges_name: [truncation, "to", ${graph.data}]
```
MeraX:
```suggestion
         down_edges_name: ["${graph.data}", "to", "truncation"]
         up_edges_name: ["truncation", "to", "${graph.data}"]
```

training/src/anemoi/training/config/graph/projections/multiscale/4scale.yaml
```diff
@@ -0,0 +1,108 @@
+---
+defaults:
+  # Load graph/projections/smooth_knn*_n320.yaml and mount each config under
+  # the matching local path in "smoothers" (e.g. smoothers.smooth_1x).
+  # Use graph/projections/multiscale=4scale_o96 to switch to o96 smoothers.
+  - /graph/projections@smoothers.smooth_1x: smooth_knn16_n320
+  - /graph/projections@smoothers.smooth_2x: smooth_knn32_n320
+  - /graph/projections@smoothers.smooth_4x: smooth_knn64_n320
+  - /graph/projections@smoothers.smooth_8x: smooth_knn128_n320
```
MeraX: Is this Hydra logic?

training/src/anemoi/training/config/graph/hierarchical_2level.yaml
```diff
@@ -12,7 +16,7 @@ nodes:
   data:
     node_builder:
       _target_: anemoi.graphs.nodes.AnemoiDatasetNodes
-      dataset: "PLACEHOLDER_DATASET_WILL_BE_OVERRIDDEN"  # Always overridden from dataloader config
+      dataset: null  # Always overridden from dataloader config
```
MeraX: Is this related to this PR?

MeraX (commented):
Hi, I'm not an ensemble expert but still curious. I left a few comments, just scratching on the surface of your contribution. I would need more time to study the documentation and code to fully understand the configuration.

models/src/anemoi/models/layers/ensemble.py
ssmmnn11: Thank you for spotting this. I should just pass it to noiseconditioning below like the noise matrix.

training/docs/user-guide/models.rst
ssmmnn11: fixed. thanks.

training/src/anemoi/training/config/graph/projections/multiscale/4scale.yaml
ssmmnn11: Yes, I added a comment describing what it does.

training/src/anemoi/training/config/graph/hierarchical_2level.yaml
ssmmnn11: True, not related. Just a clean-up.

training/src/anemoi/training/train/train.py
```diff
@@ -164,7 +165,8 @@ def _create_graph_for_dataset(self, dataset_path: str, dataset_name: str) -> Het
         # Create new graph
         from anemoi.graphs.create import GraphCreator
 
-        graph_config = self.config.graph
+        graph_config = OmegaConf.create(OmegaConf.to_container(self.config.graph, resolve=False))
```
anaprietonem: the motivation to put this in training and not inside graphs is so because of the config system living in training. But is should still work fine to write this in a single graph recipe right? is that what you're trying to test under https://github.com/ecmwf/anemoi-core/pull/855/changes#diff-dd6bfe4a22b7729a2a71ff9c97a166dfcad90e094430e5a455cdba783295919dR83?

training/src/anemoi/training/losses/loss.py
```diff
@@ -22,7 +22,18 @@
...
 
-NESTED_LOSSES = ["anemoi.training.losses.MultiscaleLossWrapper"]
+NESTED_LOSSES = {
+    "anemoi.training.losses.MultiscaleLossWrapper",
+    "anemoi.training.losses.multiscale.MultiscaleLossWrapper",
+}
+GRAPH_DATA_WRAPPER_LOSSES = {
+    "anemoi.training.losses.MultiscaleLossWrapper",
+    "anemoi.training.losses.multiscale.MultiscaleLossWrapper",
+    "anemoi.training.losses.CombinedLoss",
```
anaprietonem: why do we need to specify both? (as in "anemoi.training.losses.multiscale.MultiscaleLossWrapper" and also "anemoi.training.losses.MultiscaleLossWrapper",?)

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
```diff
@@ -47,10 +47,12 @@ def __init__(
 
     def _build_networks(self, model_config):
         super()._build_networks(model_config)
+        first_dataset_name = next(iter(self._graph_data.keys()))
```
anaprietonem: isn't this dangerous? as in what if you're graph is different for the different input datasets?

training/src/anemoi/training/config/model/transformer_ens.yaml
anaprietonem, line 13: shouldn't this be using the noise_example?

models/tests/layers/test_noise_injector.py
```diff
@@ -0,0 +1,113 @@
...
+from torch_geometric.data import HeteroData
+
+from anemoi.models.layers.ensemble import NoiseConditioning
+from anemoi.models.layers.ensemble import NoiseInjector
+
+
+def _build_noise_graph() -> HeteroData:
+    graph = HeteroData()
+    graph["noise"].num_nodes = 2
+    graph["hidden"].num_nodes = 4
```
MeraX: Does this constraint the names of the nodes in the configuration? I.e. Does it assume that the hidden nodes are always defined as `hidden` and data must not be named `noise`?

training/src/anemoi/training/train/train.py
ssmmnn11: yes correct, config lives in training - graph creater will still get a single graph recipe.

training/src/anemoi/training/losses/loss.py
ssmmnn11: hopefully improved now.

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
ssmmnn11: I added asserts to make sure they are consistent similar to how we handle processor hidden nodes etc.

models/tests/layers/test_noise_injector.py
ssmmnn11: good point. changed now.

training/src/anemoi/training/config/model/transformer_ens.yaml
ssmmnn11: I did not want to change the defaults. default is independent noise injection for each processor node.

models/tests/layers/test_noise_injector.py
MeraX: oh, please forget my comment. I was confused. This is just testing code. Should be fine as it is.

ssmmnn11: Great!

training/src/anemoi/training/config/graph/projections/smooth_knn64_o96.yaml
```diff
@@ -0,0 +1,9 @@
+---
+# Self-smoothing projection parameters for o96 (~100 km), KNN=64.
+
+node_name: smooth
+reference_node_name: ${graph.data}
+edge_weight_attribute: gauss_weight
+gaussian_norm: l1
+num_nearest_neighbours: 64
+sigma: 0.06280
```
mchantry: @ssmmnn11 how were these calculated? Could they be calculated on the fly in future? (not this PR)

training/src/anemoi/training/config/graph/projections/noise_example.yaml
```diff
@@ -0,0 +1,32 @@
+noise:
```
JPXKQX: Could this be described with the following?:
```
noise:
    node_builder: 
        _target_: anemoi.graphs.nodes.ReducedGaussianGridNodes
        grid: o32
    edge_builders:
      - _target_: anemoi.graphs.edges.KNNEdges
        num_nearest_neighbours: 32
    edge_attributes:
        gauss_weight:
          _target_: anemoi.graphs.edges.attributes.GaussianDistanceWeights
          norm: l1
          sigma: 1.0
```

training/src/anemoi/training/config/model/transformer_ens.yaml
JPXKQX, line 16: I think that we should include `edge_weight_attribute` and `noise_edge_names` here. It will be specified only sometimes, but it should always be in the config to support the option in the schema (without relying in the extra keywords)

training/src/anemoi/training/config/graph/projections/multiscale/4scale.yaml
```diff
@@ -0,0 +1,110 @@
...
+  smooth_2x:
+    node_name: smooth_2x
+  smooth_4x:
+    node_name: smooth_4x
+  smooth_8x:
+    node_name: smooth_8x
+
+matrices:
+  # Order matters for multiscale loss: smoothest -> ... -> least smooth -> none.
+  - edges_name:
```
JPXKQX: I would already specify under `matrices`: 
```
matrices:
  - edge_builders: 
     - _target_: anemoi.graphs.edges.KNNEdges
       num_nearest_neighbours: ${graph.projections.multiscale.smoothers.smooth_8x.num_nearest_neighbours}
     edge_weight_attribute:
       _target_: anemoi.graphs.edges.attributes.GaussianDistanceWeights
       norm: ${graph.projections.multiscale.smoothers.smooth_8x.gaussian_norm}
       sigma: ${graph.projections.multiscale.smoothers.smooth_8x.sigma}
  - edge_builders: 
     - _target_: anemoi.graphs.edges.KNNEdges
       num_nearest_neighbours: ${graph.projections.multiscale.smoothers.smooth_4x.num_nearest_neighbours}
     edge_weight_attribute:
       _target_: anemoi.graphs.edges.attributes.GaussianDistanceWeights
       norm: ${graph.projections.multiscale.smoothers.smooth_4x.gaussian_norm}
       sigma: ${graph.projections.multiscale.smoothers.smooth_4x.sigma}
....
```
I think we could rebuild the whole graph config from these, what do you think? any concerns?

training/src/anemoi/training/config/graph/projections/smooth_knn128_n320.yaml
```diff
@@ -0,0 +1,9 @@
+---
```
JPXKQX: Should we move these files (smooth_knn_...) elsewhere?  a `smooth` folder? 

I'm concerned about this because these config files are used differently to any other config in anemoi-training. They are referenced from a deeper config file.

We normally select as defaults for a folder name, the string corresponding to a file name in that folder. But in this case, most of the configs cannot be used ( e.g. `projection: smooth_knn128_n320`) if I understood correctly.

training/src/anemoi/training/config/graph/projections/smooth_knn64_o96.yaml
ssmmnn11: hav. distance from grid spacing. but in other situations this might not be as clear and it is in general more like a tuning factor. 1.0 should also work -> scale via KNN

training/src/anemoi/training/config/graph/projections/noise_example.yaml
ssmmnn11: currently we still build one graph object which is then used by noise and multi scale loss. So I think this would be a follow-up where we move the graph building inside the objects.

training/src/anemoi/training/config/graph/projections/multiscale/4scale.yaml
ssmmnn11: pls see reply to noise_example

models/src/anemoi/models/models/ens_encoder_processor_decoder.py
```diff
@@ -32,7 +33,7 @@ class AnemoiEnsModelEncProcDec(AnemoiModelEncProcDec):
     def __init__(
         self,
         *,
-        model_config: DotDict,
+        model_config: Any,
```
anaprietonem, line 36: why is the typing changing?

models/src/anemoi/models/schemas/models.py
```diff
@@ -270,11 +270,24 @@ class NoiseConditioningSchema(BaseModel):
...
+    "Edge type identifier (src, relation, dst) for graph-based noise projection."
+    edge_weight_attribute: Optional[str] = Field(default=None)
+    "Optional edge attribute name for graph-based noise projection weights."
     row_normalize_noise_matrix: bool = Field(default=False)
     "Whether to row-normalize the noise projection matrix weights."
     autocast: bool = Field(default=False)
     "Whether to use autocast for the noise projection matrix operations."
 
+    @model_validator(mode="after")
+    def validate_noise_projection(self) -> "NoiseConditioningSchema":
```
anaprietonem, line 292: you could move this to https://github.com/ecmwf/anemoi-core/blob/992a300e6add4894d2ab2b5d44a50c738cdbdc1b/training/src/anemoi/training/schemas/base_schema.py#L81 so that the checks is then done both when config_validation is on and off

models/src/anemoi/models/models/interpolator.py
```diff
@@ -47,9 +48,13 @@ def __init__(
...
         """
-        model_config = DotDict(model_config)
-        self.input_times = model_config.training.explicit_times.input
-        self.output_times = model_config.training.explicit_times.target
+        if type(model_config) is dict and not OmegaConf.is_config(model_config):
+            model_config_local = OmegaConf.create(model_config)
+        else:
+            model_config_local = model_config
+
+        self.input_times = list(model_config_local.training.explicit_times.input)
```
anaprietonem, line 56: is this related to this PR, or left over/not updated with main? Those seems unrelated so probably best not change this as part of the PR (same for the self.latent_skip)

models/src/anemoi/models/models/base.py
```diff
@@ -56,13 +58,16 @@ def __init__(
             Graph definition
         """
         super().__init__()
+        if type(model_config) is dict and not OmegaConf.is_config(model_config):
```
anaprietonem, line 61: why are we doing this here (and also on the interpolator)?

training/src/anemoi/training/utils/graph_config.py
```diff
@@ -0,0 +1,460 @@
...
+
+from anemoi.models.utils.projection_helpers import DEFAULT_DATASET_NAME
+from anemoi.models.utils.projection_helpers import DEFAULT_EDGE_RELATION_NAME
+from anemoi.models.utils.projection_helpers import DEFAULT_EDGE_WEIGHT_ATTRIBUTE
+from anemoi.models.utils.projection_helpers import dataset_projection_node_name
+from anemoi.models.utils.projection_helpers import projection_node_name
+from anemoi.models.utils.projection_helpers import residual_projection_truncation_node_name
+from anemoi.models.utils.projection_helpers import uses_fused_dataset_graph
+
+
```
anaprietonem, line 25: Why do we need all of these utility functions? How easy is to maintain this?

training/src/anemoi/training/train/graph_data.py
```diff
@@ -0,0 +1,193 @@
...
+from anemoi.models.utils.projection_helpers import uses_fused_dataset_graph
+from anemoi.training.utils.graph_config import merge_projection_and_graph_config
+
+if TYPE_CHECKING:
+    from torch_geometric.data import HeteroData
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TrainerGraphDataFactory:
```
anaprietonem, line 32: is this something that should live in training? I'd would say this seems to me a very graph specific functionality that would be better placed in anemoi-graphs and just simply take the part of the training config related to the dataset names that it needs rather than whole config

training/src/anemoi/training/losses/multiscale.py
```diff
@@ -14,18 +14,23 @@
...
 from torch.distributed.distributed_c10d import ProcessGroup
+from torch_geometric.data import HeteroData
 
 from anemoi.models.distributed.graph import gather_channels
 from anemoi.models.distributed.graph import shard_channels
 from anemoi.models.distributed.shapes import apply_shard_shapes
 from anemoi.models.layers.graph_provider import ProjectionGraphProvider
 from anemoi.models.layers.sparse_projector import SparseProjector
+from anemoi.models.utils.projection_helpers import (
+    multiscale_loss_matrices_graph as derive_multiscale_loss_matrices_graph,
```
anaprietonem, line 25: why changing the name here?

training/src/anemoi/training/config/graph/projections/multiscale/none.yaml
```diff
@@ -0,0 +1,2 @@
+---
```
anaprietonem, line 1: why here we have this and the truncation/none.yaml is just an empty file?

training/src/anemoi/training/config/graph/projections/multiscale/4scale.yaml
```diff
@@ -0,0 +1,26 @@
+---
+defaults:
+  - _self_
+
```
anaprietonem, line 4: could we simplify the definition of these configs by doing something like?
```
num_scales: 4
base_num_nearest_neighbours: 16
base_sigma: 0.00471
scale_factor: 2          # optional, defaults to 2
edge_weight_attribute: gauss_weight
gaussian_norm: l1
```

and then just a helper that builds this?

```
def _expand_geometric_smoothers(projection_cfg: Any) -> dict[str, Any] | None:
    """Build an explicit smoothers dict from a compact geometric progression spec."""
    num_scales = projection_cfg.get("num_scales")
    if num_scales is None:
        return None

    base_neighbours = projection_cfg["base_num_nearest_neighbours"]
    base_sigma = projection_cfg["base_sigma"]
    scale_factor = projection_cfg.get("scale_factor", 2)
    edge_weight_attribute = projection_cfg.get("edge_weight_attribute", 'gauss_weight')
    gaussian_norm = projection_cfg.get("gaussian_norm", "l1")

    smoothers = {}
    for i in range(num_scales):
        factor = scale_factor**i
        smoothers[f"smooth_{factor}x"] = {
            "edge_weight_attribute": edge_weight_attribute,
            "gaussian_norm": gaussian_norm,
            "num_nearest_neighbours": base_neighbours * factor,
            "sigma": round(base_sigma * factor, 5),
        }
    return smoothers
```
If we still want to provide support for other cases that don't have this pattern we could still allow for a dictionary of specific smoothers layer? But do you thing we'd need that case?

training/src/anemoi/training/schemas/training.py
```diff
@@ -272,32 +272,55 @@ class BaseLossSchema(BaseModel):
...
+    loss_matrices_graph: bool | list[GraphLossMatrixSchema | None] = False
 
-    @field_validator("weights")
-    @classmethod
-    def validate_weights_length(cls, v: list[float], info: Any) -> list[float]:
-        if "loss_matrices" in info.data:
-            assert len(v) == len(info.data["loss_matrices"]), "weights must have same length as loss_matrices"
-        return v
+    @model_validator(mode="after")
+    def validate_matrix_source(self) -> Self:
```
anaprietonem, line 322: Similarly this could be a function that gets called in the SchemaCommonMixin so it's checked in cases with config validation on and off

training/src/anemoi/training/losses/multiscale.py
```diff
@@ -100,32 +129,127 @@ def update_scaler(self, name: str, scaler: torch.Tensor, *, override: bool = Fal
...
+        self,
+        loss_matrices_graph: list[dict | None],
+        graph_data: HeteroData | None,
+    ) -> list[ProjectionGraphProvider | None]:
+        """Create graph-backed projection providers from explicit edge references."""
+        assert graph_data is not None, "graph_data must be provided when using loss_matrices_graph."
+
+        smoothing_matrices: list[ProjectionGraphProvider | None] = []
+        for entry in loss_matrices_graph:
+            if entry is None or entry is False or entry == "None":
```
anaprietonem, line 207: could this be simplified with the suggestions of moving some of the validation checks to the SchemaMixin?

training/src/anemoi/training/losses/loss.py
```diff
@@ -22,15 +22,46 @@
 
 METRIC_RANGE_DTYPE = dict[str, list[int]]
 
-NESTED_LOSSES = ["anemoi.training.losses.MultiscaleLossWrapper"]
+NESTED_LOSS_CLASS_NAMES = {
+    "MultiscaleLossWrapper",
+}
+GRAPH_DATA_WRAPPER_CLASS_NAMES = {
```
anaprietonem, line 33: What about an alternative solution here, where we could add a specific `needs_graph_data: bool = True` to the losses and then when instantiating the loss do:

```
    if "per_scale_loss" in loss_config:
        per_scale_loss_config = loss_config.pop("per_scale_loss")
        per_scale_loss = get_loss_function(
            OmegaConf.create(per_scale_loss_config),
            scalers,
            data_indices,
            graph_data=graph_data,
            **kwargs,
        )
        loss_config["per_scale_loss"] = per_scale_loss

    target_cls = hydra.utils.get_class(loss_config["_target_"])
    if getattr(target_cls, "needs_graph_data", False) and graph_data is not None:
        kwargs["graph_data"] = graph_data

    loss_function = instantiate(loss_config, _recursive_=False, **kwargs)
```

models/src/anemoi/models/utils/config.py
```diff
@@ -11,7 +11,7 @@
 from omegaconf import DictConfig
 from omegaconf import OmegaConf
 
-DEFAULT_DATASET_NAME = "data"
+from anemoi.models.utils.projection_helpers import DEFAULT_DATASET_NAME
```
anaprietonem, line 14: I don't think this default name should be defined by the projection_helpers. Why did you choose this? I might be missing something.

training/src/anemoi/training/utils/graph_config.py
anaprietonem: After trying to look at this more, I believe this should be something in anemoi-graphs and projection helpers should also be part of anemoi-graphs. And then in graph_config we could just do one name resolution after the graph is build so that those are then used by the model or the losses when needed. If then the graph_factory returns:
```
graph_data, projection_data = graph_factory.build()
```

and then that object can be consumed by the losses and model with something like

```
model = instantiate(config.model, graph=graph_data, projection_data=projection_data)
loss = get_loss_function(config.training_loss, ..., graph_data=graph_data,
                         loss_matrices_graph=projection_data.multiscale_loss_matrices_graph)
```

Could something like this be considered?
With that then the GraphFactory would live in anemoi-graphs and also the projection helpers that would be significantly simplified since the resolve it's done once. And similarly to the GraphCreator object we could have ProjectionCreator that then it's called in the factory to resolve the projection_metadata

anaprietonem (changes requested):

anaprietonem:
Closing this as this feature have been introduce in another branch. 
cc: @ssmmnn11 @JPXKQX

PR #869, opened 2026-02-07
fix(training): scaler index deprecation warnings, consistend index handling, some doc clean-up

ssmmnn11:
Pytorch 2.9 related deprecation warnings, some doc clean-up

### DIFF
--- a/training/src/anemoi/training/losses/scalers/time_step.py
+++ b/training/src/anemoi/training/losses/scalers/time_step.py
@@ -46,7 +46,7 @@ class TimeStepScaler(BaseScaler):
     def get_scaling_values(self) -> torch.Tensor:
         if isinstance(self.weights, torch.Tensor):
             return self.weights.to(dtype=torch.float32)
-        return torch.tensor(self.weights, dtype=torch.float32)
+        return torch.as_tensor(self.weights, dtype=torch.float32)
 
 
 class UniformTimeStepScaler(TimeStepScaler):


training/src/anemoi/training/losses/scalers/time_step.py
```diff
@@ -44,6 +44,8 @@ def __init__(
         self.weights = weights
 
     def get_scaling_values(self) -> torch.Tensor:
+        if isinstance(self.weights, torch.Tensor):
+            return self.weights.to(dtype=torch.float32)
         return torch.tensor(self.weights, dtype=torch.float32)
```
anaprietonem: Would https://docs.pytorch.org/docs/stable/generated/torch.as_tensor.html make this cleaner? 
so `return torch.as_tensor(self.weights, dtype=torch.float32)`?

training/src/anemoi/training/losses/scaler_tensor.py
```diff
@@ -598,7 +599,8 @@ def scale(
         torch.Tensor
             Scaled tensor
         """
-        x_subset = x[subset_indices] if subset_indices is not None else x
+        subset_tuple = tuple(subset_indices) if subset_indices is not None else None
+        x_subset = x[subset_tuple] if subset_tuple is not None else x
```
anaprietonem, line 603: this would change the shape of x_subset if x is multi-dimension right? was this a bug?
Image

ssmmnn11: yes let me double check what I am doing there ..

PR #872, opened 2026-02-09
fix(training): plot callbacks multi out and noopt graph provider

ssmmnn11:
Fix plotting for

- diffusion model
- trainable plotting callback with noop graph provider

### DIFF
--- a/training/docs/checkpoint_integration.rst
+++ b/training/docs/checkpoint_integration.rst
@@ -1,0 +1,256 @@
+.. _checkpoint_integration:
+
+#################################
+ Checkpoint Pipeline Integration
+#################################
+
+This guide covers the checkpoint pipeline infrastructure for Anemoi
+training. The pipeline provides a foundation for building checkpoint
+loading workflows.
+
+.. note::
+
+   This documents Phase 1 (Pipeline Infrastructure). Sources, loaders,
+   and modifiers are implemented in subsequent phases.
+
+**************
+ Core Classes
+**************
+
+CheckpointContext
+=================
+
+The ``CheckpointContext`` carries state through pipeline stages:
+
+.. code:: python
+
+   from anemoi.training.checkpoint import CheckpointContext
+
+   # Create context with a model
+   context = CheckpointContext(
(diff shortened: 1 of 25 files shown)


JPXKQX (approved):
It looks good to me. The code is way more readable now, too. Thanks!  

[Integration tests](https://github.com/ecmwf/anemoi-core/actions/runs/21831051451) pass.

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1036,10 +1019,10 @@ def _plot(
                     RuntimeWarning,
                 )
 
-            if output_times[1] != "forecast":
+            if pl_module.task_type == "temporal-interpolator":
```
mc4117: I think this is also needed for the autoencoder?

mc4117: Also I'm not sure if we should only plot 1 loss step for the interpolator

training/tests/unit/train/test_tasks.py
```diff
@@ -362,3 +363,148 @@ def test_rollout_advance_input_keeps_latest_steps(
...
+    )
+
+    assert forecaster.output_times == 1
+    for i in range(1, cfg.training.rollout.max + 1):
+        forecaster.rollout = i
+        assert forecaster.get_init_step(i) == 0
+        assert forecaster.output_times == i
+
+
+def test_graphmultioutinterpolator_output_times(monkeypatch: pytest.MonkeyPatch) -> None:
```
mc4117: Nice that you've added this! I think we need one for single out too?

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: Yes good catch about autencoder, I will change it. About your second question, I agree but I have tried to not change anything major here as per Simon's was doing. I think we probably need a bigger refactor of it, and there I think we could challenge/revisit that

PR #879, opened 2026-02-09
fix(training): scaler index handling improvement, doc string clean-up

ssmmnn11:
Pytorch 2.9 related deprecation warnings, some doc clean-up, time step scaler clean-up

### DIFF
--- a/training/src/anemoi/training/train/tasks/interpolator.py
+++ b/training/src/anemoi/training/train/tasks/interpolator.py
@@ -163,6 +163,7 @@ class GraphInterpolator(BaseGraphModule):
                 y[dataset_name] = dataset_batch[
                     :,
                     self.imap[interp_step],
+                    None,
                     :,
                     :,
                     self.data_indices[dataset_name].data.output.full,


training/src/anemoi/training/train/tasks/base.py
```diff
@@ -914,7 +914,7 @@ def calculate_val_metrics(
                 metrics[metric_step_name] = metric(
                     y_pred_postprocessed,
                     y_postprocessed,
-                    scaler_indices=[..., indices],
+                    scaler_indices=(..., indices),
```
anaprietonem, line 917: Why is this needed? it this a bug from before or the fact you reckon using tuple is safer?

ssmmnn11: using lists gives a deprecation warning in pytorch >=2.9

anaprietonem (approved):

training/src/anemoi/training/train/tasks/base.py
anaprietonem: thanks Simon!

Issue #880, opened 2026-02-10
Possibly incorrect loss calculation of GraphInterpolator


aamemiya:
### What happened?

When I used GraphInterpolator, the validation MSE loss value became unexpectedly large. It was about 10 times larger than that of GraphForecaster with the same dataset. 

I found the mismatch of the dimensions between pred and target in anemoi.training.losses.MSELoss when GraphInterpolator is used. It seems that it causes incorrect calculation of MSE loss without raising any error messages. 
I am not sure if this also affects the calculation of training loss. 

when model._target_ is anemoi.models.models.AnemoiModelEncProcDec : 
pred.shape: torch.Size([4, 1, 1, 4096, 2])
target.shape: torch.Size([4, 1, 1, 4096, 2]
-> returns torch.Size([4, 1, 1, 4096, 2])

when model._target_ is anemoi.models.models.AnemoiModelEncProcDecInterpolator : 
pred.shape: torch.Size([4, 1, 1, 4096, 2])
target.shape: torch.Size([4, 1, 4096, 2])
-> returns torch.Size([4, 4, 1, 4096, 2])

### What are the steps to reproduce the bug?

Compare the validation MSE loss between GraphForecaster and GraphInterpolator with the following configuration. 
common settings: 
```
training:
  validation_metrics:
    datasets:
      data:
        mse_inside_lam:
          _target_:  anemoi.training.losses.MSELoss
dataloader:
  batch_size:
    training: 4
    validation: 4
    test: 4
```

GraphForecaster:
```
model:
  model:
    _target_: anemoi.models.models.AnemoiModelEncProcDec
  processor:
    _target_: anemoi.models.layers.processor.GNNProcessor
training:
  multistep_input: 2 
  multistep_output: 1
```

GraphInterpolator:
```
model:
  model:
    _target_: anemoi.models.models.AnemoiModelEncProcDecInterpolator
  processor:
    _target_: anemoi.models.layers.processor.GNNProcessor
training:
  explicit_times:
    input: [0, 6]
    target: [1, 2, 3, 4, 5]
```

### Version

anemoi-training v0.9.1,  anemoi-models 0.12.1

### Platform (OS and architecture)

x86_64 GNU/Linux

### Relevant log output

```shell

```

mc4117:
Good spot! I think I might know why that's happening. Let me try and write a fix, and then I might ask you to test it, if that's ok?

aamemiya:
[...]

PR #881, opened 2026-02-10
fix: single interpolator

mc4117:
## Description
This is a fix related to an issue in the single time-interpolator set-up

This solves a bug that the dimensions were wrong in y_pred in the single time interpolator

#880


mc4117:
The problem is that it does run but the loss just explodes

PR #885, opened 2026-02-10
fix: Optimiser schema was ignoring kwargs

HCookie:
## Description
Allow extra inputs in optimiser schema for use in setting betas, and other keys. Before it was being ignored.

### DIFF
--- a/training/src/anemoi/training/train/tasks/base.py
+++ b/training/src/anemoi/training/train/tasks/base.py
@@ -1043,7 +1043,14 @@ class BaseGraphModule(pl.LightningModule, ABC):
         if hasattr(opt_cfg, "model_dump"):
             opt_cfg = opt_cfg.model_dump(by_alias=True)
 
-        return instantiate(opt_cfg, params=params, lr=self.lr)
+        optimizer = instantiate(opt_cfg, params=params, lr=self.lr)
+
+        # Log the actual optimizer settings to help users verify configuration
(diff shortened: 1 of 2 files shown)


HCookie:
@icedoom888 Yes, likely all of those should be patched. I believe the reasoning here is to allow any kwargs, whereas the utils basemodel explicitly forbids that.

icedoom888:
@HCookie I can confirm that inheriting from the anemoi.utils schema causes issues due to forbedden extra inputs. 
I am now experimenting with applying this fix in all other instances of `PydanticBaseModel `,

training/src/anemoi/training/train/tasks/base.py
```diff
@@ -1045,7 +1045,14 @@ def _create_optimizer_from_config(self, opt_cfg: Any) -> torch.optim.Optimizer:
...
 
-        return instantiate(opt_cfg, params=params, lr=self.lr)
+        optimizer = instantiate(opt_cfg, params=params, lr=self.lr)
+
+        # Log the actual optimizer settings to help users verify configuration
+        defaults_to_log = {k: v for k, v in optimizer.defaults.items() if k != "params"}
+        LOGGER.info("Optimizer initialized: %s", type(optimizer).__name__)
+        LOGGER.info("Optimizer settings: %s", defaults_to_log)
+
+        return optimizer
```
HCookie: @gabrieloks Might need to remove the logger above this

PR #890, opened 2026-02-11
feat(training): Diffusion downscaling with multi datasets [ DO NOT REBASE/UPDATE]

JoffreyDumontLeBrazidec:
## Description
Bring diffusion downscaling into the multids/downscaling branch (residuals, different noise levels, config)
should be possible to refactor, I expect that in the end all the changes should be container in 
diffusiondownscaler_encoder_processor_decoder (model)
diffusiondownscaler (task)
downscalingdatamodule (hack to bring the residual statistics)

##  Additional notes ##

- slight modification of the diffusion_samplers

- manual construction of the encoders and decoders to avoid having as many as there are datasets

An anemoi datasets has been asked for fast testing of the new code.


training/src/anemoi/training/config/system/input/hindcast_o320.yaml
```diff
@@ -0,0 +1,15 @@
+data: ${oc.env:DATA_DIR}/
+grids: ${oc.env:GRID_DIR}/
+
+dataset_in_lres: downscaling-od-cf-eefh-0001-mars-o96-2003-2023-12h-v3
+dataset_in_hres: downscaling-od-cf-enfh-0001-mars-o320-2003-2023-12h-v3-forcings
+dataset_out_hres: downscaling-od-cf-enfh-0001-mars-o320-2003-2023-12h-v3
+
+residual_statistics: ${oc.env:RESIDUAL_STATISTICS_DIR}/o320_dict_0_72.npy
```
anaprietonem, line 8: what are this? are those computed specifically using the anemoi-dataset branch you added in the pyproject?

anaprietonem: Why not using the statistics tendencies used in the tendency diffusion model?

training/src/anemoi/training/train/tasks/diffusiondownscaler.py
```diff
@@ -0,0 +1,476 @@
...
+        """
+        # Call base class to handle sharding
+        y_pred_full, y_full, grid_shard_slice = super()._prepare_tensors_for_loss(
+            y_pred,
+            y,
+            validation_mode,
+            dataset_name,
+        )
+
+        # Squeeze time dimension (always 1) to get 4D tensors for loss computation
```
anaprietonem: why do you need this? Wondering if this would make it difficult to support multi-output downscaling in the future if the time dimension is being squeezed

training/src/anemoi/training/train/tasks/diffusiondownscaler.py
```diff
@@ -0,0 +1,476 @@
...
+            if channel_name in common_channels:
+                input_pos = input_name_to_index[channel_name]
+                channel_mapping.append(input_pos)
+
+        # Convert to tensor for indexing
+        channel_indices = torch.tensor(channel_mapping)
+
+        return channel_indices
+
+    def on_after_batch_transfer(self, batch: torch.Tensor, _: int) -> torch.Tensor:
```
anaprietonem: why not reusing the one from the base class? (you don't need/want batch normalisation?)

training/src/anemoi/training/train/tasks/diffusiondownscaler.py
```diff
@@ -0,0 +1,476 @@
...
+            sigma = torch.full(
+                shape,
+                fill_value=500000.0,
+                device=device,
+            )
+
+        weight = (sigma**2 + sigma_data**2) / (sigma * sigma_data) ** 2
+        return sigma, weight
+
+    def calculate_val_metrics(
```
anaprietonem: I'd assume this could be reused from the parent class

training/src/anemoi/training/train/tasks/diffusiondownscaler.py
```diff
@@ -0,0 +1,476 @@
...
+
+    from torch_geometric.data import HeteroData
+
+    from anemoi.models.data_indices.collection import IndexCollection
+    from anemoi.training.schemas.base_schema import BaseSchema
+
+LOGGER = logging.getLogger(__name__)
+
+
+class GraphDiffusionDownscaler(BaseGraphModule):
```
anaprietonem, line 80: There are parts here like compute_loss that are very similar to the Diffusion task. Is it worth thinking how this needs to be abstracted to reuse them or inherit from them? @JPXKQX  is this something already addressed in the decoupling task PR you put?

training/src/anemoi/training/config/system/input/hindcast_o320.yaml
JoffreyDumontLeBrazidec: they are residual statistics so not equivalent to statistics tendencies (they dont exist in any anemoi.datasets)

they are computed by a personal code, the user has to create a .npy by themselves right now.

also the word tendency doesnt really make sense in this case, I think we cannot use the wording, would be confusing

training/src/anemoi/training/train/tasks/diffusiondownscaler.py
JoffreyDumontLeBrazidec: yes there may be a mistake here, what I want is to squeeze the ensemble dim instead

training/src/anemoi/training/train/tasks/diffusiondownscaler.py
JoffreyDumontLeBrazidec: yes no batch normalisation at this stage

training/src/anemoi/training/train/tasks/diffusiondownscaler.py
JoffreyDumontLeBrazidec: the diffusion parent class ? yes it could

Issue #30, opened 2024-12-03
Disentangle graph and model config.


dietervdb-meteo:
### Is your feature request related to a problem? Please describe.

This is a minor issue, but might still be worth considering.

When an instance of `AnemoiModelEncProcDec` gets initiated the names of the data and hidden nodes of the graph are looked up in the graph config:
```python
self._graph_name_data = model_config.graph.data
self._graph_name_hidden = model_config.graph.hidden
```
This is counter intuitive, especially when one would like to train the model with a pre-created graph stored on disk. In that case one would expect the graph part of the config to simply read:
```yaml
overwrite: False
``` 
Although this will lead to the graph being loaded it will also produce an error, since as described above creation of the model needs the following lines to always be present in the graph config 
```yaml
data: name_of_data_nodes #typically 'data'
hidden: name_of_hidden_nodes #typically 'hidden'
``` 

### Describe the solution you'd like

It would be cleaner if either
a) The names of hidden and data nodes are stored in the graph itself
b) These names are specified in the model config

### Organisation

RMI

anaprietonem:
Hello @dietervdb-meteo thanks for this suggestion! I believe this issues has not yet been resolved?would you like to propose the changes in a PR?

dietervdb-meteo:
Hi @anaprietonem , I'll try to give it a shot in the coming week.

JPXKQX:
Hi, I think moving these arguments under model.model makes a lot of sense. It would fit into the future refactoring of GraphForecaster, but that is far away. Should the change be made now? It will break the config files.

PR #895, opened 2026-02-13
fix(training)!: expands graph config for multiple datasets

JPXKQX:
## Description

Following discussion around configuration usability for multiple datasets, this PR provides support for different encoder/decoder subgraphs for each dataset. 

This aligns with the work described in ["Training with Observations and gridded datasets"](https://github.com/ecmwf/anemoi-core/milestone/5) Milestone. Once the graph creation is moved deep into the model, we expect to create the different data (global, regional, cerra, ...) nodes directly from the dataset removing them from the graph config.

This breaks the configuration files for the multiple dataset use cases, as we have now to explicitly define the full graph.

This PR tackles issue #30 too.

----
📚 Documentation preview 📚: https://anemoi-training--895.org.readthedocs.build/en/895/

----
📚 Documentation preview 📚: https://anemoi-models--895.org.readthedocs.build/en/895/

----
📚 Documentation preview 📚: https://anemoi-graphs--895.org.readthedocs.build/en/895/

### DIFF
--- a/training/docs/user-guide/kcrps-set-up.rst
+++ b/training/docs/user-guide/kcrps-set-up.rst
@@ -146,7 +146,7 @@ members per device needs to be specified.
 .. note::
 
    The total number of ensemble members is the product of the
-   `ensemble_size_per_device` and the `num_gpus_per_ensemble`.
+   `ensemble_size_per_device` and the ratio of `num_gpus_per_ensemble` to `num_gpus_per_model` .
 
 .. literalinclude:: yaml/example_crps_config.yaml
    :language: yaml


models/src/anemoi/models/models/base.py
```diff
@@ -113,16 +106,13 @@ def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
             self.input_dim_latent[dataset_name] = self._calculate_input_dim_latent(dataset_name)
```
havardhhaugen, line 124: Should this be `self.input_dim_latent[dataset_name] = self._calculate_input_dim_latent(self._graph_name_hidden)` ? (Related to the change in `_calculate_input_dim_latent`)

I can see that trainable_parameters.hidden is used for everything, but I think this will break if hidden and data have a different set of node attributes?

models/src/anemoi/models/models/autoencoder.py
```diff
@@ -59,10 +60,10 @@ def _build_networks(self, model_config: DotDict) -> None:
         self.decoder = torch.nn.ModuleDict()
         for dataset_name in self._graph_data.keys():
```
havardhhaugen: should this also be `self.dataset_names`?

JPXKQX: yes

models/src/anemoi/models/models/base.py
JPXKQX: 1) True, updated!
2) Yes, I have updated now. It will raise an error if the trainable parameters are not specified for one of the nodes.

models/src/anemoi/models/models/base.py
```diff
@@ -59,26 +59,19 @@ def __init__(
...
+        self.dataset_names = list(data_indices.keys())
+
         model_config = DotDict(model_config)
-        self._graph_name_data = (
-            model_config.graph.data
-        )  # assumed to be all the same because this is how we construct the graphs
-        self._graph_name_hidden = (
-            model_config.graph.hidden
-        )  # assumed to be all the same because this is how we construct the graphs
+        self._graph_name_hidden = model_config.model.model.hidden_nodes_name
```
icedoom888, line 66: Not the biggest fan of having the hidden_nodes written explicitely in the **model** config.

I suggest we do: 
`self._graph_name_hidden=graph_data.keys() - self.dataset_names`
As you do in models/layers/graph.py and filter out the data nodes by name.
This way you only define the hidden nodes names in the graph config (when creating the nodes) and you don't even have to make sure that those names match the model config.

models/src/anemoi/models/schemas/models.py
```diff
@@ -69,6 +69,8 @@ class DefinedModels(str, Enum):
 class Model(BaseModel):
     target_: DefinedModels = Field(..., alias="_target_")
     "Model object defined in anemoi.models.model."
+    hidden_nodes_name: str | list[str] = Field(examples=["hidden", ["hidden1", "hidden2"]])
```
icedoom888, line 71: Same comment as above: I would not mix graph specific names in the model config.

icedoom888 (changes requested):
LGTM, but I would remove model_config.model.model.hidden_nodes_name.

models/src/anemoi/models/models/base.py
JPXKQX: I don't think that would work:
- For hierarchical models, the order is important.
- For ensemble models (or other), there may be extra sets of nodes used for the multi-scale loss or the residual truncation.

models/src/anemoi/models/schemas/models.py
JPXKQX: Indeed, this relates to some model functionality not to the graph directly. When building a graph, this name doesn't mean anything. If you look at its usage in the code, it's only used in the model class (and some callbacks)

models/src/anemoi/models/models/base.py
icedoom888: - Order is preserved: the keys mantain the order of creation of the nodes -> order of declaration in the graph config, which solveds the issue for us
- Regarding extra nodes that are not hidden nodes per se, I would argue those should be declaired in a separate list in graphs: `additional_nodes: ["model_scale", "residual_truncation"]` which can be removed here form the list

training/src/anemoi/training/config/dataloader/multi.yaml
```diff
@@ -41,10 +41,10 @@ grid_indices:
   datasets:
     era5:
       _target_: anemoi.training.data.grid_indices.FullGrid
-      nodes_name: ${graph.data}
+      nodes_name: era5
     cerra:
       _target_: anemoi.training.data.grid_indices.FullGrid  # Could be different
-      nodes_name: ${graph.data}
```
MeraX: Would it be possible (and beneficial) in this context to relate the `nodes_name` to the parent yaml key?

training/src/anemoi/training/config/graph/encoder_decoder_only.yaml
```diff
@@ -1,15 +1,12 @@
...
-data: "data"
-hidden: "hidden"
-
 nodes:
   # Data nodes
   data:
     node_builder:
       _target_: anemoi.graphs.nodes.AnemoiDatasetNodes # options: AnemoiDatasetNodes, NPZFileNodes
-      dataset: "PLACEHOLDER_DATASET_WILL_BE_OVERRIDDEN"  # Always overridden from dataloader config
+      dataset: ${dataloader.training.datasets.data.dataset}
```
MeraX: 🙏

MeraX (commented):
Thank you, @JPXKQX . I think this makes the code cleaner and more explicit.

training/src/anemoi/training/config/dataloader/multi.yaml
JPXKQX: Good point! It shouldn't be required anymore

training/src/anemoi/training/config/graph/encoder_decoder_only.yaml
JPXKQX: I think the idea is to remove these node builders in line with the Observations milestone, following the work on #848.

models/src/anemoi/models/models/base.py
JPXKQX: I would still feel it is safer to keep it, as we have never enforced any ordering in the graph creation

icedoom888: Well this aslo assumes ordering in :  ` self._graph_name_hidden = model_config.model.model.hidden_nodes_name` which i think it's not enforced?

models/src/anemoi/models/schemas/models.py
icedoom888: I still find it confusing since information is shared across very distant parts of the config.
Can we try to polish this here? And add some checks for the naming and the order?

models/src/anemoi/models/models/base.py
JPXKQX: Sorry for the late reply. I'm getting back to this now. I think it would be a good idea to merge this soon. 
- The order is currently obtained from the '_graph_name_hidden', so the idea is to keep it the same instead of obtaining it from the order of the 'graph.nodes' builders.
- Regarding the `additional_nodes`, this option would need to be specified in all configurations as it will be part of the Pydantic schemas, even if it is `additional_nodes: []`. In that sense, if we have to specify a key in all cases, it makes more sense to use the explicit value of 'hidden_name', as this is what we will use in the code.

JPXKQX: @icedoom888 what do you think?

icedoom888: Hey @JPXKQX sorry but this is still not convincing me.
I still think having the graph hidden names in the model config it's extremely confusing, regardless. 
It completely breaks the concept of swapping graph configurations effortlessly.
Eg. i send you a graph config to try out, you need to know that you also need to update the related section in the model config. Very error prone.

If you don't want to create the list dinamically as I suggested before, you should at least move the `model_config.model.model.hidden_nodes_name` in the graph configurations instead.

models/src/anemoi/models/layers/graph.py
```diff
@@ -68,23 +68,29 @@ class NamedNodesAttributes(nn.Module):
     attr_ndims: dict[str, int]
     trainable_tensors: dict[str, TrainableTensor]
 
-    def __init__(self, num_trainable_params: int, graph_data: HeteroData) -> None:
+    def __init__(self, num_trainable_params: dict[str, int], graph_data: HeteroData) -> None:
```
MeraX: I would recommend aligning the `num_trainable_params` variable name with the config key `trainable_parameters`. (here and in `define_fixed_attributes´)

JPXKQX: I think that makes sense. What would you prefer? We could use num_trainable_parameters in the config or trainable_params here

models/src/anemoi/models/models/base.py
JPXKQX: I think that's not the case, you can still swap the graphs as before. 

This option is not a requirement for creating the graph, but rather an option to define how to use the graph in the model. When creating a graph with the graph config file, these options are not used. They are only used when creating the model.

JPXKQX: If we specify the list dynamically, we still need this extra argument in the model class, `additional_nodes`. It would have the same "problem" you describe now. To me, there are 2 independent decisions here:
- whether this configuration should be in `graph` or in `model`. Following also issue #30, it is clear to me that this option belongs to the model.
- the other question is whether it is more appropriate to have a `hidden_nodes_names` argument or an `additional_nodes`, which I also think makes more sense to use the `hidden_nodes_names` as the additional_nodes will have to be specified in the configs (even if they are null because of the schemas).

JPXKQX: @icedoom888

models/src/anemoi/models/layers/graph.py
MeraX: I would suggest either `num_trainable_parameters` , as long as it is a dict of integers. Otherwise, I would adjust the code to the config and use `trainable_parameters`. This also produces less changes in the configs.

JPXKQX: Great, thanks!

training/src/anemoi/training/train/tasks/base.py
icedoom888: Change docstring according to new changes

icedoom888:
```suggestion
    graph_data : HeteroData
```

models/src/anemoi/models/models/hierarchical_autoencoder.py
ssmmnn11, line 27: bypasses BaseGraphModel.__init__, will that then break?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -471,8 +471,9 @@ def _plot(
             for name in self.parameters
         }
         if self.latlons is None:
```
ssmmnn11, line 473: if self.latlons is None this should break here.

training/src/anemoi/training/diagnostics/callbacks/plot.py
ssmmnn11, line 583: multi dataset ready?

ssmmnn11, line 657: single dataset?

training/src/anemoi/training/diagnostics/callbacks/plot.py
JPXKQX: I haven't tested this, but it wasn't working before because `pl_module.model.model.graph_data` is already a dictionary of graphs in main

training/src/anemoi/training/diagnostics/callbacks/plot.py
JPXKQX: I don't know, same as before

ssmmnn11 (approved):
ok!

Issue #887, opened 2026-02-11
Decouple Task Type from Training Protocol


JPXKQX:
### Is your feature request related to a problem? Please describe.

The current task architecture couples task types (forecasting, time interpolation, autoencoding) with training protocols (deterministic, ensemble, diffusion, tendency-diffusion). This leads to:

- **Class explosion**: Each combination requires a separate class (e.g., `GraphForecaster`, `GraphEnsForecaster`, `GraphDiffusionForecaster`, `GraphDiffusionTendForecaster`, ...)
- **Code duplication**: Similar logic is repeated across classes
- **Difficult extensibility**: Adding new tasks (e.g., downscaling) or protocols requires creating multiple new classes
- **Configuration complexity**: Separate config files needed for each combination

Current classes:
```
tasks/
├── forecaster.py          # Deterministic forecasting
├── ensforecaster.py       # Ensemble forecasting
├── diffusionforecaster.py # Diffusion forecasting (+ tend variant)
├── autoencoder.py         # Autoencoding
├── interpolator.py        # Time interpolation
├── ...
└── base.py               # Base class
```

### Proposed Architecture

Introduce a **task-based composition architecture** that separates: (all names are open for suggestions!)

1. **Training Protocols** (the main orchestrators - current classes inheriting from `LightningModule`): 
   - `DeterministicForecaster` (single prediction with MSE loss)
   - `EnsembleForecaster` (multiple predictions with CRPS loss)
   - `DiffusionForecaster` (denoising diffusion)
   - `TendencyDiffusionForecaster` (diffusion on tendencies)
   - Other  methods

2. **Task Types** (data preparation utilities):
   - **SingleStepTask**:
     - Downscaling (low-res → high-res)
     - Autoencoding (compressed → reconstructed)
     - SingleOutTimeInterpolation (single step temporal interpolation)
   - **MultiStepTask**:
     - RolloutForecasting (multi-step ahead prediction)
     - MultiTimeInterpolation (multi-out)

3. **Task Interface** (methods tasks must implement). The task would communicate with the datamodule to load the samples instead of relying on configuration outside of the `config.data`:
   - `get_inputs(batch)` → prepare input tensor
   - `get_targets(batch)` → prepare target tensor  
   - `advance_input(x, y_pred)` → update state for next step (MultiStep only)
   - and other auxiliary methods ...

4. **Loss Functions** (already somewhat separated but needs alignment)

### Requirements

[...]

PR #896, opened 2026-02-13
feat(training)!: decouple training method & task

JPXKQX:
## Description

This is a work in progress  (#887 ). 

| Method \ Task       | Forecasting | Autoencoding  | TemporalDownscaling |
|--------------------------|------------|--------------|------------------------------|
| Single                                 |       ✅         |       ✅            |                    ✅                           |
| Ensemble                           |      ✅          |       ✅            |                 ✅                           |
| Diffusion                            |       ✅         |       ✅             |                     ✅                          |
| TendencyDiffusion            |       ✅         |      ✅              |                   ✅                          |

Proposed (minor) changes:
- Rollout configuration is moved to the task. It only needs to be specified for forecasting tasks. The `validation_rollout` is now specified too under the task definition. `rollout.max` is renamed to `rollout.maximum` (max is already a built-in operation in Python, so it's better to avoid reusing it for variable names or attributes).
- `model_task` has been renamed to `training_method`.

#### Forecasting

```
anemoi-training train --config-name=debug
anemoi-training train --config-name=ensemble_crps
anemoi-training train --config-name=diffusion
```

#### Autoencoder

```
anemoi-training train --config-name=autoencoder training=single model=transformer
anemoi-training train --config-name=autoencoder training=ensemble model=transformer_ens +system.hardware.num_gpus_per_ensemble=1
anemoi-training train --config-name=autoencoder training=diffusion model=transformer_diffusion
```

#### Temporal downscaling

```
anemoi-training train --config-name=interpolator_multiout training=single model=gnn
anemoi-training train --config-name=interpolator_multiout training=ensemble model=graphtransformer_ens system.hardware.num_gpus_per_ensemble=1
anemoi-training train --config-name=interpolator_multiout training=diffusion model=graphtransformer_diffusion
```

### TODOs

- Make `fcstep` optional in ens model. This doesn't make sense for crps-autoencoder/downscaling/multiouttimeinterpolation.
- Check time_indices passed to the reader.get_sample(). Cast to slice when possible to avoid tensor copies.=
- Update plot adapters
- Update docs

----
📚 Documentation preview 📚: https://anemoi-training--896.org.readthedocs.build/en/896/

----
📚 Documentation preview 📚: https://anemoi-graphs--896.org.readthedocs.build/en/896/

----
📚 Documentation preview 📚: https://anemoi-models--896.org.readthedocs.build/en/896/


training/src/anemoi/training/train/training_task/base.py
```diff
@@ -0,0 +1,187 @@
...
+            md_dict (dict): The metadata dictionary to fill.
+        """
+        md_dict["task"] = self.name
+        # md_dict["timestep"] = self.timestep
+        # Save relative time indices
+        md_dict["relative_input_time_indices"] = self.get_batch_input_time_indices()
+        md_dict["relative_output_time_indices"] = self.get_batch_output_time_indices()
+
+
+class BaseTimelessTask(BaseTask):
```
anaprietonem: BaseStaticTask?

training/src/anemoi/training/train/training_task/general.py
```diff
@@ -0,0 +1,33 @@
...
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from anemoi.training.train.training_task.base import BaseTask
+
+
+class BaseExplicitIndicesTask(BaseTask):
```
anaprietonem: Shall we call this something more explicit like InterpolatingTasks - not sure if ExplicitTime is something meaningful enough (otherwise I'd think the docstrings could be extended to explain it better). But then the Interpolation tasks is not inheriting from this?

training/src/anemoi/training/train/training_task/base.py
JPXKQX: That might cause confusion between static and dynamic models. But I'm not saying 'Timeless' is a good option either

training/src/anemoi/training/train/train.py
```diff
@@ -575,7 +583,7 @@ def train(self) -> None:
         LOGGER.debug("---- DONE. ----")
 
 
-@hydra.main(version_base=None, config_path="../config", config_name="config")
+@hydra.main(version_base=None, config_path="../config", config_name="downscaling_diffusion")
```
anaprietonem: just flagging - this change should be removed

training/src/anemoi/training/train/training_task/general.py
JPXKQX: I'm still not sure what the scope of this class is, but it seems it could be merged with the base class.

matschreiner:
2) This is maybe more a question or reflection. But I think that the models' prediction and training algorithms are too closely related to meaningfully separate them. 
Following this PR, each time we want to implement a new model we need to implement the bespoke TrainingProtocol too. For example here the the DiffusionProtocol is really implementing the EDM diffusion which means that if we want a classical DDPM in anemoi we will also have to implement the ClassicalDiffusionProtocol, and for a VAE we need to implement a VAEProtocol and so on. For each new model we need to implement a TrainingProtocol which to me indicate that they belong together. 

What is the argument for keeping the training protocol and model separate?

training/src/anemoi/training/train/protocols/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from .diffusion import DiffusionProtocol
+from .diffusion import DiffusionTendProtocol
+from .ensemble import EnsembleProtocol
+from .single import SingleProtocol
+
+__all__ = [
+    "DiffusionProtocol",
```
matschreiner: Each of these should be renamed to something that does not indicate that they provide abstraction.

training/src/anemoi/training/train/protocols/single.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+from __future__ import annotations
+
+import logging
+
+import torch
+from torch.utils.checkpoint import checkpoint
+
+from anemoi.training.train.protocols.base import BaseGraphModule
```
matschreiner: BaseGraphModule could be renamed to better reflect its role. The current name only describes implementation details - has a graph and is a plmodule.
If the subclasses are renamed to for example SingleStepTrainingAlgorithm, EnsembleTrainingAlgorithm, etc., I suggest calling it BaseTrainingAlgorithm. This makes the design more readable. The trainer needs "a task and a training algorithm"  instead of a "task and a graph module".

training/src/anemoi/training/config/training/scalers/global.yaml
```diff
@@ -33,13 +33,13 @@ datasets:
     # scale the prognostic losses by the stdev of the variable tendencies (e.g. the 6-hourly differences of the data)
     # useful if including slow vs fast evolving variables in the training (e.g. Land/Ocean vs Atmosphere)
     # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
-    stdev_tendency:
-      _target_: anemoi.training.losses.scalers.StdevTendencyScaler
-      timestep: ${data.timestep}
+    #stdev_tendency:
```
mc4117: why do these scalers need to be commented out?

training/src/anemoi/training/config/training/scalers/stretched.yaml
```diff
@@ -37,13 +37,6 @@ datasets:
     # scale the prognostic losses by the stdev of the variable tendencies (e.g. the 6-hourly differences of the data)
     # useful if including slow vs fast evolving variables in the training (e.g. Land/Ocean vs Atmosphere)
     # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
-    stdev_tendency:
```
mc4117, line 40: Same comment about the scalers here

training/src/anemoi/training/config/autoencoder.yaml
```diff
@@ -4,11 +4,12 @@ defaults:
...
 - graph: encoder_decoder_only
-- model: graphtransformer
-- training: autoencoder
+- model: transformer
+- task: autoencoding
+- training: single
 - _self_
 
-config_validation: True
+config_validation: False
```
mc4117: I think for the default configs we should have config_validation true

training/src/anemoi/training/config/debug.yaml
```diff
@@ -21,15 +22,32 @@ config_validation: True
 diagnostics:
   plot:
     callbacks: []
+  log:
+    mlflow:
+      enabled: False
+      tracking_uri: None
+    wandb:
```
mc4117: I don't think we should need to add more to the diagnostics here?

training/src/anemoi/training/config/debug.yaml
```diff
@@ -21,15 +22,32 @@ config_validation: True
...
+      enabled: False
+      entity: example
+
 system:
   input:
-    graph: ???
+    graph: graph_anemoi_new_${data.resolution}.pt
```
mc4117: also I don't think we should specify paths here right?

training/src/anemoi/training/config/debug.yaml
```diff
@@ -5,10 +5,11 @@ defaults:
 - system: example
 - graph: multi_scale
 - model: gnn
-- training: default
+- task: forecasting
+- training: single
 - _self_
 
-config_validation: True
+config_validation: False
```
mc4117: config_validation should be true I think

training/src/anemoi/training/config/diffusion.yaml
```diff
@@ -5,9 +5,48 @@ defaults:
 - system: example
 - graph: multi_scale
 - model: graphtransformer_diffusion
+- task: forecasting
 - training: diffusion
 - _self_
 
 
 # set to true to switch on config validation
```
mc4117, line 13: again not sure why so much is being added to the config

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -5,13 +5,14 @@ defaults:
...
+- task: timeinterpolating
+- training: single
 - _self_
 
-config_validation: True
+config_validation: False
 
 data:
-  frequency: 1h
+  frequency: 6h
```
mc4117: here the frequency should be 1h. 6h is a bit of a weird interpolator

training/src/anemoi/training/data/dataset.py
```diff
@@ -212,6 +222,58 @@ def tree(self, prefix: str = "") -> Tree:
         return tree
 
 
+class MaskedGridDataset(BaseAnemoiReader):
```
mc4117: what is the use case for this new class? It would be nice to have a bit more detail in the doc string

training/src/anemoi/training/tasks/custom.py
```diff
@@ -0,0 +1,32 @@
...
+class TimeInterpolationTask(BaseSingleStepTask):
+    """Time interpolation task implementation.
+
+    Input and output offsets are specified as duration strings
+    (e.g. ``["0H", "6H"]`` and ``["1H", "2H", "3H", "4H", "5H"]``).
+    """
+
+    name: str = "timeinterpolation"
+
+    def __init__(
```
mc4117: how is the dataloader handled? If this is all that is needed then do we even need to have a task?

models/src/anemoi/models/models/interpolator.py
```diff
@@ -35,6 +35,8 @@ def __init__(
         data_indices: dict,
         statistics: dict,
         graph_data: HeteroData,
+        n_step_input: int,
```
mc4117: I think a lot of these model classes could be unified?

training/src/anemoi/training/config/task/timeinterpolating.yaml
```diff
@@ -0,0 +1,3 @@
+_target_: anemoi.training.tasks.TimeInterpolationTask
+input_offset: ["0H", "6H"]
```
mc4117: I think if we want to use this nomenclature then we have to be very careful with our type setting. and very clear errors if for example the user states "0h" instead of "0H"

training/src/anemoi/training/train/methods/base.py
```diff
@@ -49,6 +49,7 @@
 
     from anemoi.models.data_indices.collection import IndexCollection
     from anemoi.training.schemas.base_schema import BaseSchema
+    from anemoi.training.tasks.base import BaseTask
```
matschreiner, line 54: This couples the training algorithm to the task, which is the opposite of the decoupling this PR aims for.

The fix is simple

define an interface/protocol - probably inside a dedicated file task/protocol.py

```
from typing import Protocol

class TaskProtocol(Protocol):
    def get_inputs(self, xxx) -> yyy: ...
    def get_targets(self, xxx) -> yyy: ... 
    etc..
```

And then replace this import with the protocol.

training/src/anemoi/training/train/methods/base.py
```diff
@@ -137,7 +138,8 @@ def __init__(
         self,
         *,
         config: BaseSchema,
-        graph_data: HeteroData,
+        task: BaseTask,
```
matschreiner, line 144: replace type with abstraction

training/src/anemoi/training/data/dataset.py
JPXKQX: I think this is a leftover from the grid_indices PR. Thanks!

models/src/anemoi/models/interface/__init__.py
```diff
@@ -227,7 +232,7 @@ def predict_step(
             "batch": batch,
             "pre_processors": self.pre_processors,
             "post_processors": self.post_processors,
-            "n_step_input": self.n_step_input,
+            # "n_step_input": self.n_step_input,
```
JPXKQX:
```suggestion
            "n_step_input": self.n_step_input,
```

matschreiner (dismissed):
I'm sorry to be so insisting Mario. But an abstraction is a crucial part of this PR.

Interfaces should be defined first and the code designed around them, not the other way around. Since this PR introduces new concepts, it should also be responsible for defining clearly what those concepts are - it is not something that should be left to be explored in a follow-up PR and discussed with the community after the implementation already exists. The interface is the decoupling, and postponing it means this PR has not achieved its stated goal.

training/src/anemoi/training/tasks/custom.py
JPXKQX: The tasks are defined by the input and output offsets, with this, we are able to have a common abstraction for all tasks that defines how the x & y are sampled from the dataset.

training/src/anemoi/training/train/methods/single.py
```diff
@@ -15,12 +15,12 @@
...
 from torch.utils.checkpoint import checkpoint
 
-from anemoi.training.train.protocols.base import BaseGraphModule
+from anemoi.training.train.methods.base import BaseGraphModule
 
 LOGGER = logging.getLogger(__name__)
 
 
-class SingleProtocol(BaseGraphModule):
+class SingleTraining(BaseGraphModule):
```
matschreiner: Could you rename BaseGraphModule to better reflect its role? BaseTrainingAlgorithm makes the design more readable. The trainer needs "a task and a training algorithm" instead of a "task and a graph module".

JPXKQX: I can rename this, but still unsure whether the training algorithm or the training module is better. I would prefer BaseTrainingModule, because the training method is only one part of the functionality.  The trainer needs a data module and a training module.

matschreiner: Yep, I think  BaseTrainingModule sounds good! :)

training/src/anemoi/training/config/task/timeinterpolating.yaml
MeraX: would it make sense to follow ISO 8601 for time deltas?

On the other hand I would tend to define input in terms of time steps and the `frequence` defined in the config.

training/src/anemoi/training/tasks/base.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def log_extra(self, *_args, **_kwargs) -> None:
+        """Hook to log any task-specific information."""
+
+    def on_train_epoch_end(self, current_epoch: int) -> None:
+        """Hook to update task state at the end of each training epoch (e.g. for curriculum learning)."""
+
+    def fill_metadata(self, md_dict: dict) -> None:
+        """Fill the metadata dictionary with task-specific information."""
+        md_dict["task"] = self.name
```
VeraChristina, line 192: I have updated the names to be consistent with the task names we are currently using in the inference metadata to keep that interface stable

training/src/anemoi/training/data/datamodule.py
```diff
@@ -197,10 +177,18 @@ def fill_metadata(self, metadata: dict) -> None:
...
+        relative_date_indices = sorted(input_relative_date_indices + output_relative_date_indices)
         timesteps = {
-            "relative_date_indices_training": self.relative_date_indices(),
             "timestep": self.config.data.timestep,
+            "relative_date_indices_training": relative_date_indices,  # backwards compatibility with inference
+            "input_relative_date_indices": input_relative_date_indices,  # backwards compatibility with inference
+            "output_relative_date_indices": output_relative_date_indices,  # backwards compatibility with inference
+            "input_offsets": [frequency_to_string(o) for o in self.task.inputs_offsets],
+            "output_offsets": [frequency_to_string(o) for o in self.task.outputs_offsets],
         }
```
VeraChristina: I have moved this information here instead of adding it in the task. I am also happy for it to be added in the task instead, but I'd like to keep the existing structure of the metadata dict to keep the interface with inference stable.

models/src/anemoi/models/models/interpolator.py
VeraChristina: Do we still need the `AnemoiModelEncProcDecMultiOutInterpolator` or should we be able to use the `AnemoiModelEncProcDec` together with the interpolator task, with the changes made in this PR?

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -31,11 +29,5 @@ model:
 
 training:
   model_task: anemoi.training.train.tasks.GraphMultiOutInterpolator
-  # Instead of inferred using multistep and timeincrement, specify time indices explicitly
-  explicit_times:
-    input: [0, 6]
-    target: [1, 2, 3, 4, 5]
-  multistep_input: 1
```
VeraChristina, line 38: I know that this is now defined in the task config but we might want to expose it in the higher-level config as it is one of the main configurations for the interpolator -- mostly for transparency for the user

VeraChristina (commented):
Thanks for adding this! Neat idea to separate out task and training method. Just a few minor comments below.

Two things to think about:
- What are the additional use cases we want to cover in integration tests? Likely not the complete matrix but potentially the ensemble interpolator, maybe other use cases
- Do any of the changes affect inference? I have removed a few references to the training config in the metadata class (which would break with the config changes in this PR) -- but are there other changes? Potentially good to not just run the system-level tests but also to run an interpolator checkpoint from this branch in inference

training/src/anemoi/training/config/interpolator.yaml
dietervdb-meteo: Hi, I should probably have brought this up earlier, but maybe this PR is also the moment to unify time step specification for inteperpolator and forecaster. My preference would be to simply move to (for both forecaster and interpolator]
```yaml
input_steps: [bla, bli]
output_steps: [blo, ble, blu] # or target_steps
```
Which could be in the task part of the config. @JPXKQX, @mc4117  what do you think? I'm happy to help with the adaption of the code to support this new config.

dietervdb-meteo: I.e. and get rid of `explicit_times` and `multistep_in/output`

mc4117: yes I agree this could be a nice place to do this. It seems like it's almost there if you look at the task/time_interpolating.yaml. But it could also be as part of a second PR?

models/src/anemoi/models/models/interpolator.py
mc4117: I think we should aim for this at least

training/src/anemoi/training/config/interpolator.yaml
dietervdb-meteo: I can set it up as a separate PR, for the moment pointing to this branch. We can then see if we merge it here first or later to main. [Will try to get to it tomorrow]

training/src/anemoi/training/data/usable_indices.py
```diff
@@ -36,7 +36,10 @@ def get_usable_indices(
     usable_indices : np.array
         Array of usable indices.
     """
-    usable_indices = np.arange(series_length - max(relative_indices))
+    if isinstance(relative_indices, list):
+        relative_indices = np.array(relative_indices)
+
+    usable_indices = np.arange(series_length)
```
Magnus-SI, line 88: This needs to remain as usable_indices = series_length - max(relative_indices) for the trajectory_ids logic to work. If max(relative_indices) is not subtracted, rel_run will have indices outside the range of trajectory_ids. 
With this change, {series_length} should not be added to the missing indices.
Alternatively, the trajectory_ids logic can be changed.

training/src/anemoi/training/config/interpolator.yaml
dietervdb-meteo: Thanks @JPXKQX for showing me that a lot of this is already there under the hood with the input and output offsets. We have some ideas to move forward. TBC

training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -5,16 +5,14 @@ defaults:
 - system: example
 - graph: multi_scale
 - model: graphtransformer
-- training: default
+- task: timeinterpolating
```
mc4117: I would be tempted to rename this to temporal_downscaling given that we now have spatial_downscaling

training/src/anemoi/training/config/interpolator.yaml
OpheliaMiralles: I have some working code in a branch implementing sparse indices handling, it allows to run this type of configs. Should I open a PR to avoid duplication of work? 
image

JPXKQX: Hi Ophelia,
Sure, it would be helpful to share the existing code. In this case, we have something similar, but in a different order:
```
inputs_offsets: 
  era: ["-6H", "0H"]
  cerra: ["-2H", "-1H", "0H"]
outputs_offsets:
  cerra: ["1H", "2H", "3H"]
```

Would you have any concern with this?  There are still some implementation details to tackle because the proposed tasks are very clean now and don't depend on anything, so introducing a dependency on the datasets could be... undesired? The datamodule already depends on the task

training/src/anemoi/training/config/interpolator.yaml
JPXKQX: This could be a good opportunity to standardise the naming conventions for both spatial and temporal interpolation. If you agree, I will try to rename the current tasks as well.

VeraChristina: Happy for this to be standardised in training. However, if we can avoid changing the names we pass to inference under `metadata_inference.task` that would be great! (just trying to keep the interface stable before we introduce a versioned schema)

JPXKQX: I think these names are still not used by anemoi-inference, so it could be a good opportunity to rename those top

JPXKQX: Anyway, this should be discussed at the @ecmwf/anemoi_technical_subgroup

VeraChristina: Sure, but once we start using them we would need a different version for checkpoints created in the meantime if we change them now. (Inference aims to be backwards compatible for checkpoints.)

JPXKQX: perfect, I have now bumped the metadata version to 3.0

training/src/anemoi/training/data/datamodule.py
```diff
@@ -157,8 +132,7 @@ def _get_dataset(
     ) -> MultiDataset:
         return MultiDataset(
             data_readers=datasets,
-            relative_date_indices=self.relative_date_indices(val_rollout),
-            timestep=self.config.data.timestep,
+            task=self.task,
```
mc4117: I think val_rollout still needs to get passed to the MultiDataset somehow? Maybe through self.task.rollout?

training/tests/unit/data/test_multidataset.py
```diff
@@ -22,7 +22,6 @@ class TestMultiDataset:
     def dataset_config(self) -> dict:
```
mc4117: this needs to be updated given that now the multidataset class reads in the task

training/src/anemoi/training/data/usable_indices.py
JPXKQX: Is this solved now?

JPXKQX: The reason for making this slight change is that relative indices can now be `[ -1, 0, 1 ]`, which should return the same result as `[ 0, 1, 2 ]`.

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -206,18 +206,27 @@ model:
...
+    epoch_increment: 1
+    max: 1
+  validation_rollout: 1
+
 training:
   run_id: null
   fork_run_id: null
   transfer_learning: False # activate to perform transfer learning
-  model_task: anemoi.training.train.tasks.GraphForecaster
+  model_task: anemoi.training.train.methods.SingleTraining
```
MeraX: Pleas apologise, if this has been discussed before or is out of scope for this PR. In either case, please just close this comment.

I was wondering, if `model_task` is still the correct wording here, does it relate to the new "task" section in the config, and whether it rather belongs into the "model" section. Given that the former "tasks" module has been renamed to "methods", perhaps this option should be the "model_method" then?

JPXKQX: Good point! I agree that `model_task` doesn't make sense anymore. Thanks, Marek!

JPXKQX: What about `training_method` ?

training/src/anemoi/training/config/interpolator.yaml
OpheliaMiralles: It looks fine, I just wonder how this works with dataset not aligned in frequency (e.g. one with 5m freq and another with 6h). I know I had issues with this while setting a past window to 1h because the indices don't align so it had to be selected in a configurable way (e.g. last timesteps, or uniform timestepping over the past window)

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
MeraX: yes, `training_method` is much more reasonable than my suggestion. Thank you.

training/src/anemoi/training/data/usable_indices.py
Magnus-SI: It looks good to me. I'm getting some unrelated errors on the newest version of this branch, so I haven't been able to test it there, but this fix runs on my older version without error.

JPXKQX: I haven't tested it with the trajectory_ids yet, so things might still be wrong

training/src/anemoi/training/tasks/temporal_downscaling.py
```diff
@@ -0,0 +1,50 @@
...
+        input_timedelta = as_timedelta(input_timestep)
+        output_timedelta = as_timedelta(output_timestep)
+
+        input_offsets = [datetime.timedelta(hours=0), input_timedelta]
+
+        assert input_timedelta % output_timedelta == datetime.timedelta(
+            0,
+        ), "Input timestep must be an integer multiple of output timestep for temporal downscaling."
+        num_output_steps = input_timedelta // output_timedelta
+        output_offsets = [output_timedelta * (i + 1) for i in range(num_output_steps)]
```
mc4117: output right boundary is not working properly as it's putting it in twice

JPXKQX: It should be 
```
        output_offsets = [output_timedelta * (i + 1) for i in range(num_output_steps - 1)]
```
right?

mc4117: unfortunately this then causes failures in another place, so we need to check where else this is used in the code

mc4117: unfortunately this then causes failures in another place, so we need to check where else this is used in the code

training/src/anemoi/training/tasks/base.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def get_offsets(self, **_kwargs) -> list[datetime.timedelta]:
+        """Get the list of offsets for a given label (e.g. "training", "validation", "test").
+
+        By default, this returns ``self.offsets``, but can be overridden by subclasses to return
+        different offsets per label for example (e.g different rollout in training vs validation).
+        """
+        return self.offsets
+
+    def _offset_to_batch_indices(self, offsets: list[datetime.timedelta]) -> list[int]:
```
mc4117: this needs to be changed, at least in the case of the temporal downscaler

training/docs/modules/diagnostics.rst
```diff
@@ -68,8 +68,7 @@ callbacks can use the same interface across task types:
 - forecaster tasks use ``ForecasterPlotAdapter``;
 - diffusion tasks use ``DiffusionPlotAdapter``;
```
JPXKQX:
```suggestion
```

training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml
```diff
@@ -206,18 +205,27 @@ model:
     - _target_: anemoi.models.layers.bounding.ReluBounding
       variables: []
 
+task:
+  _target_: anemoi.training.tasks.Forecaster
+  multistep_input: 2
+  multistep_output: 1
+  timestep: 6h
```
mc4117, line 212: @MeraX this integration test is now failing because we have moved timestep from data to task. Is it ok if we update your expected metadata in the test? It can either come from this place or the config or the new inference metadata

MeraX: Hi @mc4117, thank you for raising this. I would prefer the new inference metadata, but otherwise, I would suggest just updating the config path in the test.

mc4117: Thanks! @VeraChristina has updated it to the new inference metadata

training/docs/modules/tasks.rst
```diff
@@ -0,0 +1,174 @@
+#######
+ Tasks
+#######
+
+The ``anemoi.training.tasks`` module defines the temporal structure of a
+training sample independently of the Lightning module. Each task
+specifies which time steps are loaded as inputs and which are used as
+targets, and provides helpers for mapping those offsets to batch
```
anaprietonem:
```suggestion
targets, referred as offsets as they are defined as relative positions in time compared to a reference point, and provides helpers for mapping those offsets to batch
```

training/docs/modules/tasks.rst
```diff
@@ -0,0 +1,174 @@
...
+   :members:
+   :no-undoc-members:
+   :show-inheritance:
+
+
+Multistep Input and Output
+==========================
+
+The forecaster task uses ``multistep_input`` and ``multistep_output`` to control how many time
+steps the model injests as input and predicts in a single forward pass.
```
anaprietonem:
```suggestion
steps the model ingests as input and predicts in a single forward pass.
```

training/src/anemoi/training/config/diagnostics/plot/multi.yaml
```diff
@@ -120,7 +117,6 @@ callbacks:
     - v_850
     - 2t
     - tp
-    output_steps: ${training.multistep_output}
```
anaprietonem, line 123: how is this now inferred or handled?

training/src/anemoi/training/config/training/scalers/global.yaml
```diff
@@ -35,11 +35,9 @@ datasets:
     # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
     stdev_tendency:
       _target_: anemoi.training.losses.scalers.StdevTendencyScaler
-      timestep: ${data.timestep}
```
anaprietonem, line 38: why is this now removed ?

training/src/anemoi/training/config/task/autoencoder.yaml
```diff
@@ -0,0 +1 @@
+_target_: anemoi.training.tasks.Autoencoder
```
anaprietonem, line 1: why here we don't specify any MS input or output?

training/src/anemoi/training/config/diagnostics/plot/multi.yaml
JPXKQX: It is a task property

training/src/anemoi/training/config/training/scalers/global.yaml
JPXKQX: extracted from the task

training/src/anemoi/training/data/relative_time_indices.py
```diff
@@ -0,0 +1,92 @@
...
+    if isinstance(relative_indices, int):
+        return reference_index + relative_indices
+
+    if isinstance(relative_indices, np.ndarray):
+        return reference_index + relative_indices
+
+    return [reference_index + offset for offset in relative_indices]
+
+
+def compute_relative_date_indices(
```
anaprietonem, line 15: what's the motivation for having this functions here and not something that the tasks provide? Ie an interface provided as part of the BaseTask?

training/src/anemoi/training/config/task/autoencoder.yaml
JPXKQX: It is implicit in the class. It uses the same timestep in and out. We could add an option to add temporal context, but this can be done in the future too

training/src/anemoi/training/losses/multiscale.py
```diff
@@ -159,12 +159,12 @@ def _prepare_for_smoothing(
                 Target shard shapes for later gathering
         """
         batch_size, out_times, ensemble_size = y_pred_ens.shape[0], y_pred_ens.shape[1], y_pred_ens.shape[2]
-        y_pred_ens_interp = einops.rearrange(y_pred_ens, "b t e g c -> (b e) g (c t)")
+        y_pred_ens_interp = einops.rearrange(y_pred_ens, "b t e g c -> (b e) t g c")
```
anaprietonem, line 162: why this change? this is changing the shape of the tensor to 4D

training/src/anemoi/training/data/relative_time_indices.py
JPXKQX: It could be a method of the task that takes as input the data_reader, but I think this is very similar

training/src/anemoi/training/losses/multiscale.py
mc4117: This is fixing a bug that was left from the multi-out PR. @Magnus-SI and I did some tests to check that this fixes the bug

training/CHANGELOG.md
```diff
@@ -8,7 +8,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Please add your functional changes to the appropriate section in the PR.
 Keep it human-readable, your future self will thank you!
 
-## [0.11.0](https://github.com/ecmwf/anemoi-core/compare/training-0.10.0...training-0.11.0) (2026-03-26)
+## [Unreleased]
```
anaprietonem: this needs to go away. It will get filled with Release-Please

training/src/anemoi/training/train/methods/diffusion.py
```diff
@@ -59,10 +61,6 @@ def __init__(
 
         self.rho = config.model.model.diffusion.rho
 
-        from anemoi.training.diagnostics.callbacks.plot_adapter import DiffusionPlotAdapter
-
-        self._plot_adapter = DiffusionPlotAdapter(self)
-
     def get_input(self, batch: dict[str, torch.Tensor]) -> dict[str, torch.Tensor]:
```
anaprietonem: is this function used now?

training/src/anemoi/training/train/methods/diffusion.py
```diff
@@ -528,11 +526,12 @@ def _step(
...
-        # batch is already normalized in BaseGraphModule._normalize_batch
+        # batch is already normalized in BaseTrainingModule._normalize_batch
         # x: data.input.full (normalized), state_target: data.full (normalized slice view)
-        x = self.get_input(batch)  # (bs, n_step_input, ens, latlon, nvar)
-        state_target = self.get_target(batch)
-        y_data_output = self.get_data_output_target(state_target)
+        x = self.task.get_inputs(batch, data_indices=self.data_indices)  # (bs, n_step_input, ens, latlon, nvar)
+        state_target = self.task.get_targets(batch, data_indices=self.data_indices)
+        y_data_output = self.get_data_output_target(state_target)  # (bs, n_step_output, ens, latlon, nvar)
+        # Move to get_target()?, y = self.task.get_targets(batch, data_indices=self.data_indices)
```
anaprietonem: is this for a follow-up PR or what?

training/src/anemoi/training/train/methods/diffusion.py
JPXKQX: I don't think it exists anymore :(

training/src/anemoi/training/train/methods/diffusion.py
JPXKQX: It shouldn't. I need to check

training/src/anemoi/training/tasks/base.py
```diff
@@ -0,0 +1,219 @@
...
+        data_indices : dict[str, IndexCollection]
+            Data indices per dataset.
+
+        Returns
+        -------
+        dict[str, torch.Tensor]
+            Input tensors per dataset with shape
+            ``(bs, num_inputs, grid, nvar)``.
+        """
+        time_indices = self.get_batch_input_indices()
```
ssmmnn11, line 147: is this a list of ints now? then we use advanced indexing and materialise a new tensor - instead of using narrow. I think we should try to use slices / narrow if these are cont.

ssmmnn11 (changes requested):
a question.

training/src/anemoi/training/tasks/base.py
JPXKQX: `relative_date_indices` should be converted to slice when possible [here](https://github.com/ecmwf/anemoi-core/blob/feat/task-refactor/training/src/anemoi/training/data/multidataset.py#L68). Then, `time_steps` ([here](https://github.com/ecmwf/anemoi-core/blob/feat/task-refactor/training/src/anemoi/training/data/multidataset.py#L300)) will be a `slice` too.

JPXKQX: Fixed now by [commit](https://github.com/ecmwf/anemoi-core/pull/896/commits/dfcc11a2b2c5b21b9d59eb35bb505cec2a712705)

anaprietonem (dismissed):
Thanks for addressing my previous comments, very nice feature @JPXKQX and impressive team work the 3 of you @mc4117 @VeraChristina!

ssmmnn11 (dismissed):
well done

JPXKQX:
[Integration tests](https://github.com/ecmwf/anemoi-core/actions/runs/24723629419)

anaprietonem (approved):
LGTM!

PR #897, opened 2026-02-14
loss variable mapper and filtering fixes

ssmmnn11:
Some ideas how implementation could be slightly adapted to make it compatible with tendency diffusion model. Also we now always have the wrapper, so renamed it LossVariableMapper.

- Reworked filtered-loss handling to use explicit index-space/layout mapping (model_output, data_output, data_full) so prediction/target/scaler variable alignment is consistent across forecaster, ensemble, diffusion, autoencoder, and interpolator tasks.

- Fixed scaler propagation/remapping for filtered losses in CombinedLoss/MultiscaleLossWrapper, updated diagnostics/variable-scaling reporting, and added regression tests for filtering+scaler+layout interactions (including forcing gaps and target-only variables).

Still a bit rough around the edges - we could remove some tests potentially. 

----
📚 Documentation preview 📚: https://anemoi-training--897.org.readthedocs.build/en/897/

----
📚 Documentation preview 📚: https://anemoi-graphs--897.org.readthedocs.build/en/897/

----
📚 Documentation preview 📚: https://anemoi-models--897.org.readthedocs.build/en/897/

### DIFF
--- a/training/tests/unit/train/test_tasks.py
+++ b/training/tests/unit/train/test_tasks.py
@@ -174,6 +174,7 @@ _CFG_FORECASTER = DictConfig(
     },
 )
 
+
 def test_graphinterpolator_preserves_time_dim_in_targets(monkeypatch: pytest.MonkeyPatch) -> None:
     """Regression test: interpolator loss targets must keep singleton time dim."""
     forecaster = GraphInterpolator.__new__(GraphInterpolator)


training/src/anemoi/training/losses/multiscale.py
```diff
@@ -215,8 +214,8 @@ def forward(
             y_pred_ens_tmp, y_tmp = self._smooth_for_loss(y_pred_ens_for_smooth, y_for_smooth, i)
 
             if model_comm_group_size and model_comm_group_size > 1 and self.keep_batch_sharded:
-                y_pred_ens_tmp = gather_channels(y_pred_ens_tmp, shard_shapes, model_comm_group)
-                y_tmp = gather_channels(y_tmp, shard_shapes_y, model_comm_group)
+                y_pred_ens_tmp = gather_channels(y_pred_ens_tmp, shard_shapes, group)
```
OpheliaMiralles, line 217: I guess the model_comm_group changing to "group" is related to the #806 not being up to date with main, if this is the case just confirm by resolving the comment, if it is not I would find it odd to change the naming in this filtering loss wrapper related PRs

training/tests/unit/train/test_tasks.py
```diff
@@ -201,6 +218,284 @@ def _stub_init(
     assert y_pred.shape == (b, 1, e, g, v)
 
 
+def test_graphinterpolator_preserves_time_dim_in_targets(monkeypatch: pytest.MonkeyPatch) -> None:
```
OpheliaMiralles, line 178: Thanks a lot for populating test_tasks :)

training/src/anemoi/training/losses/multiscale.py
ssmmnn11: There is a inconsistency in main between different losses, some take group, the multi scale takes model comm group. but this then falls apart if we wrap several losses. For this reason I had to change this to make multiscale behave as the other ones and we don't get ugly if statements.

PR #898, opened 2026-02-14
test: update integration test settings

ssmmnn11:
modify some integration test settings:

awkward batch size
lr and scheduler to have a realistic learning rate
less processor layers for faster tests


ssmmnn11:
Yes, we would need to update the checkpoint (load existing checkpoint) - from 16 to 2 layers.

training/tests/integration/config/testing_modifications.yaml
```diff
@@ -4,11 +4,11 @@ dataloader:
...
   batch_size:
-    training: 2
-    validation: 2
+    training: 3
+    validation: 3
   limit_batches:
-    training: 2
-    validation: 2
+    training: 3
+    validation: 3
```
VeraChristina, line 11: I think it's fine to keep the limit to two batches with `batch_size=3`

training/tests/integration/config/testing_modifications.yaml
```diff
@@ -38,6 +38,12 @@ graph:
 
 model:
   num_channels: 16
+  processor:
+    num_layers: 2
```
VeraChristina, line 42: wondering if reducing layers will make the tests much faster given that we run very short cycles on reduced data and a lot of time is overhead+plotting -- but that's not to say we shouldn't change this

training/tests/integration/config/testing_modifications.yaml
VeraChristina: I think this is why the test_restart_training test fails

VeraChristina (commented):
Thanks for adding this! Two minor comments below. In the run linked above there seem to be two tests that are failing (in addition to the ones where we would need to update the checkpoints): test_restart_training, and test_training_cycle_diffusion

training/tests/integration/config/testing_modifications.yaml
ssmmnn11: I wanted to go to three because then we have a bit more of a "realistic" learning rate scenario. together with warmup etc.

training/tests/integration/config/testing_modifications.yaml
ssmmnn11: I think on the CPU yes, but increasing the batch size and the number of samples brings us approx. to where we were before I think. Still a more realistic test.

training/tests/integration/config/testing_modifications.yaml
VeraChristina: Makes sense! We might need to tweak the settings of the restart test but that seems fine

Issue #899, opened 2026-02-16
Select Variables missing from Multi-dataset Implementation


icedoom888:
### What happened?

Key: "Select" is missing from the dataloader schema and implementation.
At the moment all variables are selected, even when specified differently in the config.
No error is thrown either.

### What are the steps to reproduce the bug?

Load any dataset with:
```
training:
  datasets:
    data:
      dataset: ${dataloader.dataset}
      start: null
      end: 2023
      frequency: ${data.frequency}
      drop:  []
      select: ${dataloader.selected_vars}
```

### Version

latest

### Platform (OS and architecture)

Linux balfrin-ln004 5.14.21-150500.55.97_13.0.78-cray_shasta_c #1 SMP Thu Mar 13 20:09:44 UTC 2025 (330b47d) x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell

```

### Organisation

Meteoswiss

Issue #891, opened 2026-02-12
be more explicit about open_dataset kwargs


VeraChristina:
### Is your feature request related to a problem? Please describe.

We currently ignore additional keys in the NativeGridDataset schema, and also don't hand them down to the `open_dataset` call in `BaseAnemoiReader`. Additional options to the `open_dataset` call are allowed as part of the `dataset` key in the NativeGridDataset.

We only keep `drop` and `frequency` as additional keys for backswards compatibility.

### Describe the solution you'd like

Make the schema more strict to forbid additional keys in NativeGridDataset (not under `dataset` in NativeGridDataset), so that the user notices that these aren't used.

Perhaps add a comment in the templates to clarify where additional options for the open_dataset call should be added.

PR #900, opened 2026-02-16
fix: integrate select variables in dataloader

icedoom888:
## Description
Solves #899 
##  Additional notes ##

### DIFF
--- a/models/src/anemoi/models/commands/migration.py
+++ b/models/src/anemoi/models/commands/migration.py
@@ -245,7 +245,6 @@ class Migration(Command):
                         "migration_version": "1.0.0",
                         "imports": imports,
                         "final": args.final,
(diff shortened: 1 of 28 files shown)


anaprietonem:
this is also related to the issue @VeraChristina opened last week - https://github.com/ecmwf/anemoi-core/issues/891

MicheleCattaneo:
Nice catch.
It feels like these bugs (there was another about beta's for the optimizers) are a bit sneaky due to the `extra = "ignore"` default value of pydantic's schemas.
What if `extra` was set to either `"allow"` or `"forbid"` everywhere and we let the object's constructor fail if the argument is not expected? Silently dropping arguments sounds too dangerous.
If `"ignore"` is really needed, maybe it's possible to add some kind of hook that at least throws a warning?

PR #904, opened 2026-02-16
feat: Adding features for gridded observation training

pinnstorm:
## Description

Adding features to handle NaN values in training for gridded observations

Fixes back-propagation for NaN values with broken `torch.nansum` and `torch.nanmean` implementation. Also introduces a new imputer that only imputes the input tensors and forcings, not the targets. 

##  Additional notes ##

This is a draft currently for visability

### DIFF
--- a/models/src/anemoi/models/preprocessing/__init__.py
+++ b/models/src/anemoi/models/preprocessing/__init__.py
@@ -10,7 +10,6 @@
 import logging
 from typing import Optional
 
-import torch
 from torch import Tensor
 from torch import nn
 
@@ -200,9 +199,10 @@ class Processors(nn.Module):
         """Run checks on the processed tensor."""
         if not self.inverse:
             # Forward transformation checks:
-            assert not torch.isnan(
-                x
-            ).any(), f"NaNs ({torch.isnan(x).sum()}) found in processed tensor after {self.__class__.__name__}."
+            pass
+            # assert not torch.isnan(
+            #     x
+            # ).any(), f"NaNs ({torch.isnan(x).sum()}) found in processed tensor after {self.__class__.__name__}."
 
 
 class StepwiseProcessors(nn.Module):
--- a/models/src/anemoi/models/preprocessing/imputer.py
+++ b/models/src/anemoi/models/preprocessing/imputer.py
@@ -167,10 +167,12 @@ class BaseImputer(BasePreprocessor, ABC):
         ----------
         x : torch.Tensor
             Input tensor
-        index : list
+        index_x : list[int]
             List of indices for the variables to be imputed
         nan_locations : torch.Tensor
             Tensor with NaN locations
+        index_nl : list[int]
+            List of indices for nan locations
 
         Returns
         -------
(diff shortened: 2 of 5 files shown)


ssmmnn11:
Very nice. Can this replace the current imputer?

training/src/anemoi/training/losses/base.py
```diff
@@ -57,8 +57,9 @@ def __init__(self, ignore_nans: bool = False) -> None:
 
         self.add_module("scaler", ScaleTensor())
 
-        self.avg_function = torch.nanmean if ignore_nans else torch.mean
```
mchantry, line 60: @pinnstorm thanks for the great contribution.
I'm being slow and don't understand why this part needs to change? Please could you explain. Thanks

pinnstorm: Of course no problem! There is a bug in `torch.nanmean` and `torch.nansum` meaning the backward pass with produce a `NaN`: https://github.com/pytorch/pytorch/issues/67180 (and other related issue)

The work around is to leave `torch.mean/sum` and just set the missing values to zero (further down in `forward` method) in both the predictions and targets so they have no impact on the loss.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -546,3 +548,287 @@ def transform(
...
+
+        if not method_config:
+            LOGGER.warning(
+                f"{cls.__name__}: Using default method {default} for all variables not specified in the config.",
+            )
+        for m in method_config:
+            if isinstance(method_config[m], str):
+                method_config[m] = {method_config[m]: f"{m}_{method_config[m]}"}
+            elif isinstance(method_config[m], list):
+                method_config[m] = {method: f"{m}_{method}" for method in method_config[m]}
```
sahahner, line 602: I see that you are bringing the constant and "statistics" imputer together. If we go for this approach, I think it would be nice to do the same for what is the "standard" imputer at the moment to have these two methods aligned.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -546,3 +548,287 @@ def transform(
...
+            # Use nan_to_num with custom nan replacement per variable
+            forcing_slice = x[..., forcing_indices]
+            forcing_mask = nan_mask[..., forcing_indices]
+            forcing_values = imputation_values[forcing_indices]
+            # Broadcast imputation values to match tensor shape
+            forcing_values = forcing_values.view(*([1] * (forcing_slice.ndim - 1)), -1)
+            x[..., forcing_indices] = torch.where(forcing_mask, forcing_values, forcing_slice)
+
+        if len(non_forcing_indices) > 0:
+            if is_training_rollout:
```
sahahner, line 811: I see why you introduced the if/else statements to handle prognostic/forcing/inference. I am wondering, whether this separation using `is_training_rollout` is necessary or if it would be possible to always impute the first self.multi_step indices?

models/src/anemoi/models/preprocessing/__init__.py
```diff
@@ -200,9 +199,10 @@ def _run_checks(self, x):
         """Run checks on the processed tensor."""
         if not self.inverse:
             # Forward transformation checks:
-            assert not torch.isnan(
-                x
-            ).any(), f"NaNs ({torch.isnan(x).sum()}) found in processed tensor after {self.__class__.__name__}."
+            pass
```
sahahner, line 202: Could this function be redefined for the input-only imputer instead of disabling the check entirely?

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -546,3 +548,287 @@ def transform(
...
+                non_forcing_slice = x[..., non_forcing_indices]
+                non_forcing_mask = nan_mask[..., non_forcing_indices]
+                non_forcing_values = imputation_values[non_forcing_indices]
+                # Broadcast imputation values to match tensor shape
+                non_forcing_values = non_forcing_values.view(*([1] * (non_forcing_slice.ndim - 1)), -1)
+                x[..., non_forcing_indices] = torch.where(non_forcing_mask, non_forcing_values, non_forcing_slice)
+
+        return x
+
+    def inverse_transform(self, x: torch.Tensor, in_place: bool = True) -> torch.Tensor:
```
sahahner, line 832: There should be no need to redefine this function as returning `x` is the default behaviour.

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -546,3 +548,287 @@ def transform(
...
+            non_forcing_indices = self.non_forcing_indices_inference
+            imputation_values = self.imputation_values_inference
+
+        # Detect NaNs - only compute mask once
+        nan_mask = torch.isnan(x)
+
+        # Determine which timesteps to impute for non-forcing variables
+        # In training, we only impute the first multi_step timesteps for non-forcing variables
+        # In inference, the tensor typically has only the input timesteps
+        is_training_rollout = x.ndim >= 2 and x.shape[1] > self.multi_step
```
MeraX, line 797: Does this pose any implicit assumption on the chosen task (and "protocol" in the future?)

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -546,3 +548,287 @@ def transform(
...
+        """
+        super().__init__(config, data_indices, statistics)
+
+        if isinstance(statistics, DictConfig):
+            statistics = OmegaConf.to_container(statistics, resolve=True)
+
+        # Get multi_step from config, default to 2 if not specified
+        # This determines how many timesteps are inputs vs targets
+        # Use .get() for DictConfig or getattr for other config types
+        if config is None:
```
MeraX, line 632: Will you move the default into the schemas when this PR matures? I could also imagine that the option could simply be mandatory without needing a default at all. It could also be interesting considering whether the training task (or protocol in the future) could inform the imputer about the number of input steps.

MeraX (commented):
Thank you for sharing this work!

models/src/anemoi/models/preprocessing/imputer.py
```diff
@@ -546,3 +548,287 @@ def transform(
...
+            forcing_mask = nan_mask[..., forcing_indices]
+            forcing_values = imputation_values[forcing_indices]
+            # Broadcast imputation values to match tensor shape
+            forcing_values = forcing_values.view(*([1] * (forcing_slice.ndim - 1)), -1)
+            x[..., forcing_indices] = torch.where(forcing_mask, forcing_values, forcing_slice)
+
+        if len(non_forcing_indices) > 0:
+            if is_training_rollout:
+                # Impute non-forcing variables only in first multi_step timesteps
+                non_forcing_slice = x[:, : self.multi_step, ..., non_forcing_indices]
```
mchantry, line 813: After talking to @sahahner I think this might not work with the interpolator.

sahahner: agree on this one. If we select the input timesteps by "take the first `self.multi_step` timesteps", this will be restricted to forecaster-type model architectures.

training/src/anemoi/training/losses/base.py
HCookie, line 1: @pinnstorm I think this update to the losses should be made into a seperate PR, it would be able to be merged quickly, and address some other issues

pinnstorm: Sounds like a plan!

Issue #886, opened 2026-02-11
Deprecation proposal: Time Interpolator single-out


mc4117:
We currently have two implementations of the time interpolator:

1. Single output step (interpolates forecast 1 timestep at a time)
2. Multiple output steps (interpolates forecast producing all intermediate timesteps at once)

These implementations are currently separated in the code as two different classes in tasks and models.

If single out is not going to be used, we can deprecate the single-out interpolator feature.
If we would like to keep single-out then we should combine these two classes together to make it easier to maintain code.

PR #915, opened 2026-02-18
feat: remove single out time interpolator

mc4117:
## Description
Following discussion on issue #886, this PR deprecates the time interpolator for single outputs

----
📚 Documentation preview 📚: https://anemoi-training--915.org.readthedocs.build/en/915/

----
📚 Documentation preview 📚: https://anemoi-graphs--915.org.readthedocs.build/en/915/

----
📚 Documentation preview 📚: https://anemoi-models--915.org.readthedocs.build/en/915/

### DIFF
--- a/training/src/anemoi/training/config/training/interpolator.yaml
+++ b/training/src/anemoi/training/config/training/interpolator.yaml
@@ -57,8 +57,6 @@ explicit_times:
   input: [0, 6]
   target: [1, 2, 3, 4, 5]
 
-target_forcing: null
-
 # select strategy
 strategy:
   _target_: anemoi.training.distributed.strategy.DDPGroupStrategy
--- a/training/src/anemoi/training/schemas/training.py
+++ b/training/src/anemoi/training/schemas/training.py
@@ -485,15 +485,6 @@ class DiffusionTendForecasterSchema(ForecasterSchema):
     "Training objective."
 
 
-class InterpolationSchema(BaseTrainingSchema):
-    model_task: Literal["anemoi.training.train.tasks.GraphInterpolator"] = Field(..., alias="model_task")
-    "Training objective."
-    explicit_times: ExplicitTimes
-    "Time indices for input and output."
-    target_forcing: DatasetDict[TargetForcing]
-    "Forcing parameters for target output times."
-
-
 class AutoencoderSchema(ForecasterSchema):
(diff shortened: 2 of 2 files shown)


training/tests/unit/diagnostics/test_plotting_callbacks.py
```diff
@@ -397,41 +397,6 @@ def test_process_forecaster_output_shapes():
     assert output_tensor.shape == (output_times, n_step_output, n_ens, nlatlon, nvar), output_tensor.shape
 
 
-def test_process_time_interpolator_output_shapes():
-    """BasePlotAdditionalMetrics.process: time-interpolator task yields expected shapes."""
```
VeraChristina, line 401: Is this test single-out specific?

models/src/anemoi/models/models/interpolator.py
```diff
@@ -20,13 +20,12 @@
 from anemoi.models.distributed.shapes import get_or_apply_shard_shapes
 from anemoi.models.distributed.shapes import get_shard_shapes
 from anemoi.models.models import AnemoiModelEncProcDec
-from anemoi.models.utils.config import get_multiple_datasets_config
```
VeraChristina, line 23: I assume we don't need this because the multi_out interpolator was introduced after multiple datasets?

VeraChristina (approved):
This looks good to me, thanks! Just two small clarification questions for my understanding

models/src/anemoi/models/models/interpolator.py
mc4117: I think the pre-commit hook removed this because we were using this to get the right target forcings before but now we don't need it

training/tests/unit/diagnostics/test_plotting_callbacks.py
mc4117: I think it's now a duplicate of the test below it?

VeraChristina: fair enough. I think we might want to adapt `_make_pl_module_interpolator` to the multi out interpolator, but perhaps in a separate PR?

mc4117: discussed offline and we think that the `_make_pl_module_interpolator` accounts for multi already

VeraChristina (approved):
Thanks for answering the questions and removing the schemas! LGTM

Issue #911, opened 2026-02-18
Bug with DatasetSchema when trying to pass custom Statistics


anaprietonem:
### What happened?

As reported by @icedoom888 current dataset schema doesn't allow to pass a statistics file.

<img width="1195" height="324" alt="Image" src="https://github.com/user-attachments/assets/8daafbd6-aed9-41b8-a098-81d291e9118c" />

To address this we propose to change the schema so that keys such as select, statistics, drop, and frequency would be passed inside the 'dataset' key.  This would be a breaking change in our configs that should be accompanied by docs to explain this to users and developers.

<img width="630" height="183" alt="Image" src="https://github.com/user-attachments/assets/b507b62f-2c1d-45e4-adeb-f266f3724248" />

### What are the steps to reproduce the bug?

Try to pass statistics or other keys accepted by open_dataset and not included in the schema at the moment

### Version

Current

### Platform (OS and architecture)

ATOS

### Relevant log output

```shell

```

cc: @VeraChristina  @JPXKQX  @mchantry

Issue #863, opened 2026-02-05
Redesign of data & dataloader configs


JPXKQX:
### Is your feature request related to a problem? Please describe.

Anemoi training currently supports multiple datasets through manual configuration duplication across different config files after PR #594. As we scale from a handful of datasets to potentially dozens of datasets (e.g., ERA5, CERRA, regional models, reforecast datasets, ...) and different use cases (forecasting, downscaling, ...) the current approach presents several challenges:
- __Limited Scalability__: When increasing the number of datasets, the data.yaml will be huge, containing common datasets that may be reused across different configurations/use cases.  With several use cases, it is common to use the same datasets for different use cases (forecasting, downscaling, time interpolation) where the configuration (forcings & prognostic) may be different for each one. This would add a lot of duplication to the configs too.
- __Separation of Concerns__: Dataset definitions are mixed with dataloader settings and system paths.
- __Hydra Limitation__: Hydra's default mechanism selects one file per config group, making it hard to compose N datasets from M available options.

### Current design

The existing configuration structure uses Hydra's config groups:

```
config/
├── data/
│   ├── zarr.yaml      # Single dataset config
│   └── multi.yaml     # Multi-dataset config
└── dataloader/
    ├── native_grid.yaml
    └── multi.yaml
```

Each dataset configuration includes:
- **Dataset path** (currently in `system.input.dataset`)
- **Variables**:
   - **Forcing variables** (variables only considered as input)
   - **Diagnostic variables** (variables only considered as output)
   - **Prognostic variables**. These are defined as the `selected variables (in dataloader config) - diagnostics - forcings`.
- **Processors** (normalizers, imputers, etc.)
- **Frequency and timestep** settings

### Requirements

An ideal configuration should:
- Dataset definitions should be independent from dataloader and system configurations
- Same dataset config usable across different experiments
- Easily compose experiments with different dataset combinations. Support 1 to N datasets in any combination
- Minimize configuration duplication and errors
- Easy to see what datasets are available
- Validate configurations at load time
- Maintain backward compatibility where possible

### Organisation

ECWMF

dietervdb-meteo:
Hi @anaprietonem , concerning YAML anchors vs Omegaconf. The latter can do most what anchors can and much more. Still the key feature the anchors have, which, afaik, Omegaconf is missing is the ability to overwrite or extend entries. E.g. with omegaconf one can do
```yaml
robot:
  name: Atlas
  manufacturer:
    name: RoboDynamics
[...]

PR #916, opened 2026-02-18
feat!: Rewrite dataset Schema to fix issue 911

anaprietonem:
## Description
Breaking change to dataloader dataset config: per-dataset reader options now use dataset_config with dataset as the source key (inner key), aligned with open_dataset({"dataset": ..., ...}).
This PR updates:
- schema validation (dataset_config/dataset only)
- dataset loading flow
- default dataloader templates (native_grid, multi)
- unit tests
- docs and migration guidance.
Old dataset/name shapes are no longer supported. 

https://github.com/ecmwf/anemoi-core/issues/911
https://github.com/ecmwf/anemoi-core/issues/911
##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--916.org.readthedocs.build/en/916/

----
📚 Documentation preview 📚: https://anemoi-graphs--916.org.readthedocs.build/en/916/

----
📚 Documentation preview 📚: https://anemoi-models--916.org.readthedocs.build/en/916/

### DIFF
--- a/training/tests/unit/data/test_dataset.py
+++ b/training/tests/unit/data/test_dataset.py
@@ -395,6 +395,33 @@ def test_native_dataset_schema_validates_new_dataset_dictionary() -> None:
     assert cfg.end == 2020
 
 
+def test_native_dataset_schema_validates_cutout_dataset_config() -> None:
+    cfg = NativeDatasetSchema(
+        dataset_config={
+            "dataset": {
+                "cutout": [
+                    {
+                        "dataset": "mock-dataset-a.zarr",
+                    },
+                    {
+                        "dataset": "mock-dataset-b.zarr",
+                    },
+                ],
+                "adjust": "all",
+            },
+            "statistics": "mock-statistics.zarr",
+            "frequency": "6h",
+            "drop": [],
+        },
+        start=1985,
+        end=2020,
+    )
+
+    assert cfg.dataset_config is not None
+    assert cfg.start == 1985
+    assert cfg.end == 2020
+
+
 def test_native_dataset_schema_raises_on_invalid_dataset_dictionary() -> None:
     with pytest.raises(ValidationError):
         NativeDatasetSchema(dataset_config={"frequency": "6h"}, start=1985, end=2020)


dietervdb-meteo:
Hi @anaprietonem, thanks. I found peace with leaving start and end as separate arguments :) So far so good. Not sure if a check on forbidding them inside is a good idea? Are we sure this still allows use cases where e.g. a user concatenates two datasets by providing start and end for each of them, then selects sub periods for training and validation by providing start and end in the dataloader?

training/src/anemoi/training/config/dataloader/multi.yaml
```diff
@@ -64,34 +66,38 @@ validation_rollout: 1 # number of rollouts to use for validation
 validation:
   datasets:
     era5:
-      dataset: ${system.input.dataset}
+      dataset_config:
+        dataset: ${system.input.dataset}
+        frequency: ${data.frequency}
+        drop: []
       start: 2021
```
dietervdb-meteo, line 73: Actually, now that I see this... What was the argument again for keeping start an end separate? Taking them inside would simply amount to an extra tab in the config and be conceptually much cleaner?

JPXKQX: I think the argument is that these are the only dataloader configuration options that should be included here (dataloader folder). In the future, we could even remove the dataset (move it to the data/ folder), keeping here only the `start` and `end`.

JPXKQX: We could remove the `frequency` and `drop` from the config and leave a docstring referring to the support of additional anemoi-datasets keys like: drop, select, frequency, statistics, ...

dietervdb-meteo: But isn't the idea that you could in principle provide a completely different dataset for validation than e.g. training? Like training on reanalysis but validating on analysis say? If that is the case, and one needs to at least provide a dataset, then there is always the default to fall back to which is to take as start and end just the start and end of the data [like anemoi-datasets does]. So then it is just an extra spefication the user can make, like drop and any other set of anemoi-dataset args. 

I originally thought that the reason for keeping them separate was because otherwise the config would become more complicated. But as I mention above, with what we have now with this PR putting them in or out simply amounts to a an extra tab...

dietervdb-meteo: Maybe a cleaner solution to go to what you have in mind for the future @JPXKQX would be to have a separate `datasets` part of the config, where you can just list by name a bunch of `anemoi-dataset` configs:
```yaml
datasets:
  training:
  - data1: 
    <anemoi_dataset_config_1a>
  - data2: 
    <anemoi_dataset_config_2a>
  validation:
  - data1: 
    <anemoi_dataset_config_3>
  -  data2: 
    <anemoi_dataset_config_2b>
```
And then refer to them in other parts of the config? In the future these references can then be removed and the relevant datasets can be passed as needed by the code if you want to make e.g. dataloader config dataset independent?

anaprietonem: I found select and drop used very often so I thought from a user perspective it would be clear if those are stated in the config.  @JPXKQX 

I rather than moving the dataset_config to the data I would try to harmonise data and dataloader into just one config section. 

Regarding where start/end should go, @dietervdb-meteo your point about the concat and join, it's a good one so I am adding tests but the check is for the dataset_config, (ie if your dataset is a dict with a start and end which is what you are describing this is fine).

dietervdb-meteo: Ok, thanks @anaprietonem. One more comment to @JPXKQX , if in the future we want to decouple dataset and dataloader further, I think start end could also naturally go into the trajectory specification. In the philosophy that a dataloader should just be about how to sample a given dataset that would make sense. So personally I still don't see any drawback by absorbing start and end into the dataset config for now. But no need to make this a breaking point. If most of you prefer to go this way for now, then that's fine for me.

JPXKQX: I propose that we move this part of the discussion to issue #863 and try to agree on a minimal change, but please comment there so that we can have some discussion around these configs more in depth

Issue #909, opened 2026-02-18
Move shared model functionality to base class to support all use cases


JPXKQX:
### Is your feature request related to a problem? Please describe.

With the increasing number of use cases, new model functionality has been implemented inside specific use-case models. To improve reusability, consistency, and maintainability, we propose moving common functionality into the base model class, so it becomes available across all use cases and avoids further divergence.
- `latent_skip` option. This controls whether a residual (skip) connection is applied around the processor block.
- `processor` is optional. Make the processor component optional at the base class level. If processor is defined, then run encode → process → decode. If processor is not defined, then run encode → decode directly.
- an option to control the `target_forcing`. Add a configuration option to control what information the decoder receives for data nodes. There are several options here: 
    1) Encoded data nodes. Decoder receives only the encoded representation (output of encoder).
    2) Full input data nodes (forcing + prognostic = input of the encoder too)
    3) Custom variable set. Decoder receives a specified subset of variables (typically forcing variables at the target timestep).

PR #922, opened 2026-02-19
feat(models)!: move functionality to base model class

JPXKQX:
## Description

This minor PR brings two functionalities used in several use cases to the base model class. 

- Add `latent_skip` as an option in the base model.
- Make `processor` optional. 
  - B) Create a new `NoOpProcesor` that can be instantiated in the current configs.
 
## Breaking change

This PR is considered a breaking change because a new key `latent_skip` is added to the model config.

The model checkpoints are still compatible. No migration checkpoint is needed.

## Next steps 

- Third point in #909. Refactor all the different options to pass `data` nodes to decoders. This will be done in a follow up PR. It will mean we can drop model classes like: `autoencoder`, `hierarchical_autoencoder`, or `interpolator`.

### DIFF
--- a/training/src/anemoi/training/train/train.py
+++ b/training/src/anemoi/training/train/train.py
@@ -273,7 +273,8 @@ class AnemoiTrainer(ABC):
         """Provide the model instance."""
         assert (
             not (
-                "GLU" in self.config.model.processor.layer_kernels["Activation"]["_target_"]
+                "layer_kernels" in self.config.model.processor
+                and "GLU" in self.config.model.processor.layer_kernels["Activation"]["_target_"]
                 and ".Transformer" in self.config.model.processor.target_
             )
             and not (


models/src/anemoi/models/models/base.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         self.n_step_input = model_config.training.multistep_input
         self.n_step_output = model_config.training.multistep_output
         self.num_channels = model_config.model.num_channels
+        self.latent_skip = model_config.model.get("latent_skip", True)
```
mc4117: I can't find this option in the config? Do you not need to add it there as part of this PR?

models/src/anemoi/models/models/base.py
```diff
@@ -96,34 +98,47 @@ def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
         self.num_input_channels = {}
         self.num_output_channels = {}
         self.num_input_channels_prognostic = {}
+        self.num_input_channels_target_forcings = {}
```
mc4117: you can remove all references to target_forcings as they were removed by PR #915

models/src/anemoi/models/models/base.py
```diff
@@ -96,34 +98,47 @@ def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
...
 
-    def _calculate_output_dim(self, dataset_name: str) -> int:
-        return self.n_step_output * self.num_output_channels[dataset_name]
-
     def _calculate_input_dim_latent(self, dataset_name: str) -> int:
         return self.node_attributes[dataset_name].attr_ndims[self._graph_name_hidden]
 
+    def _calculate_target_dim(self, dataset_name: str) -> int:
+        return (
+            self.n_step_output * self.num_input_channels_target_forcings[dataset_name]
```
mc4117: I don't think this target forcings dimension was ever right. This was part of my problem with them too - they are actually inputs so shouldn't be in the target

models/src/anemoi/models/models/base.py
```diff
@@ -96,34 +98,47 @@ def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
...
             + self.node_attributes[dataset_name].attr_ndims[self._graph_name_data]
         )
 
-    def _calculate_output_dim(self, dataset_name: str) -> int:
-        return self.n_step_output * self.num_output_channels[dataset_name]
-
     def _calculate_input_dim_latent(self, dataset_name: str) -> int:
         return self.node_attributes[dataset_name].attr_ndims[self._graph_name_hidden]
 
+    def _calculate_target_dim(self, dataset_name: str) -> int:
```
mc4117, line 137: why are target_dim and output_dim different?

models/src/anemoi/models/models/encoder_processor_decoder.py
```diff
@@ -91,7 +94,7 @@ def _build_networks(self, model_config: DotDict) -> None:
                 model_config.model.decoder,
                 _recursive_=False,  # Avoids instantiation of layer_kernels here
                 in_channels_src=self.num_channels,
-                in_channels_dst=self.input_dim[dataset_name],
+                in_channels_dst=self.target_dim[dataset_name],
```
mc4117, line 94: not sure why this changed?

JPXKQX: The `in_channels_dst` refers to the input variables from the data nodes in the decoder. I think this should serve as a more general configuration. By default, we pass all the data variables again. In other cases, we may pass the encoded_data_nodes, and in other cases we may pass a subset of variables only (the target_forcings)

models/src/anemoi/models/models/base.py
JPXKQX: I don't know if `target` is the correct naming but it is the one used already. We could probably think about a better option. 
- Output would represent the data that is returned by the model.
- Target would represent data passed as input to the model but referring to the target `timestep`.

models/src/anemoi/models/models/base.py
JPXKQX: added now! thanks

models/src/anemoi/models/models/base.py
```diff
@@ -96,34 +98,48 @@ def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
         self.num_input_channels = {}
         self.num_output_channels = {}
         self.num_input_channels_prognostic = {}
+        self.num_input_channels_decoding_forcings = {}
         self._internal_input_idx = {}
         self._internal_output_idx = {}
+        self._target_forcing_input_idx = {}
```
mc4117: I think this can be removed?

training/src/anemoi/training/config/model/point_wise.yaml
```diff
@@ -5,6 +5,7 @@ keep_batch_sharded: True
 
 model:
   _target_: anemoi.models.models.AnemoiModelEncProcDec
+  latent_skip: True
```
havardhhaugen, line 8: Could this be an instantiate instead of a True False. This would make the config more uniform since that is currently how we specify the grid skip connection. It also opens up the possibility to explore different latent skip connections (although I don't think it's very high on the priority of endless things to explore). 
```
latent_residual:
  _target_: anemoi.models.layers.residual.SkipConnection # options: SkipConnection, TruncatedConnection, NoConnection
  step: -1 # use the latest step as skip connection
```

icedoom888: Yeah this actually makes sense to do here, would just require a couple more abstarct classes but having1/2 more at this point doesn't hurt!

mc4117: I'm not sure the `-1` is the most intuitive for me here. Maybe `forwards` or `backwards` is more intuitive?

JPXKQX: I think this makes sense, but looking at the code now, this may require a significant change because the tensor's size and dimensions are different at this stage (compared to the other residual connection). 
- The current residual connections receive the full batch (bs, ens, time, grid, vars).
- In this case, we have a 2D tensor (num_nodes, nodes_dim)

I would recommend opening an issue documenting this and waiting for the observations milestone to be addressed, keeping this boolean option for now. There is planned work here (in the obs milestone) around carrying more information in the batch, not just the data. This will make it easier to specify the dimension to select the step (if any) in the residual connection.

havardhhaugen: Good point @JPXKQX, I had not looked at the shapes. We can revisit it later :+1:

Issue #925, opened 2026-02-23
Custom area weightings in node attributes


efare04:
### Is your feature request related to a problem? Please describe.

`anemoi-graphs` already has several classes to calculate node area weightings for various grid types. However, our input data is on the ORCA025 grid. Since this grid is tripolar, none of the existing classes are suitable. 

### Describe the solution you'd like

We already have a file giving us the area of each of the grid cells. If we include this in the Anemoi dataset, we can attach it as a node attribute. This will require a class which reads in a variable from an Anemoi dataset and assigns it as a node attribute. This will be similar to the `BaseAnemoiDatasetVariable` class, except it will return attributes with a range of numerical values, rather than a boolean mask. I've already implemented a solution for this, so will create a PR shortly.

### Organisation

UK Met Office

PR #926, opened 2026-02-23
feat(graphs): add custom area weightings

efare04:
## Description
This PR adds a class which allows users to load custom area weightings from a variable in an Anemoi dataset. The behaviour of `AnemoiDatasetVariableWeights`  is based on that of `BaseAnemoiDatasetVariable`, but isn't inherited directly from that class as that would set `dtype = bool`.

None of the existing classes to calculate area weightings are suitable for the tripolar ORCA025 grid, but we already have a file containing the cell areas. This information could be read directly from a variable in an Anemoi dataset into a node attribute, rather than calculating the area weightings within `anemoi-graphs`.

#925 

##  Additional notes ##
I've put this in `area_weights.py`, as we're intending to use it for area weightings, but this class could be used to create a custom node attribute for other purposes. Perhaps it would be worth renaming it (e.g. to `AnemoiDatasetVariableAttribute`) and moving it elsewhere, as its application isn't limited to area weights?

----
📚 Documentation preview 📚: https://anemoi-training--926.org.readthedocs.build/en/926/

----
📚 Documentation preview 📚: https://anemoi-graphs--926.org.readthedocs.build/en/926/

----
📚 Documentation preview 📚: https://anemoi-models--926.org.readthedocs.build/en/926/

### DIFF
--- a/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
+++ b/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
@@ -408,8 +408,6 @@ class AnemoiDatasetVariableWeights(BaseNodeAttribute):
 
     Methods
     -------
-    _read_data(self, nodes)
-        Read the weighting variable from the dataset.
     get_raw_values(self, nodes)
         Extract the data and convert it to a Torch tensor object.
     """


graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -393,3 +394,52 @@ def compute_latitude_weight(self, latitudes: np.ndarray) -> np.ndarray:
...
+    Attributes
+    ----------
+    variable : str
+        Name of the variable to use as weights.
+    norm : str, optional
+        Method to use to normalise the weights.
+
+    Methods
+    -------
+    _read_data(self, nodes)
```
JPXKQX: I would probably remove this method from the docstring as it's a private method

JPXKQX:
Hi @efare04, this PR looks self-contained and provides some new functionality. Thanks for your contribution! I made a minor comment, but I'm happy to approve!

PR #927, opened 2026-02-24
fix: `distributed.all_gather()` support for uneven tensors by explicit padding

fprill:
## Description

`distributed.all_gather()` support for uneven tensors does not apply to all backends
see
- https://docs.pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather
- https://github.com/pytorch/pytorch/issues/156627

##  Additional notes ##

@japols

PR #931, opened 2026-02-25
feat(training): multi-gpu integration tests for model sharding

ssmmnn11:
Adds multi-GPU integration for model-sharding training

including keep_batch_sharded=True and keep_batch_sharded=False.

for example, on an interactive node one can run: pytest -q training/tests/integration/test_training_cycle_multigpu.py --slow --multigpu -k model_sharding -x

This can be quite useful for debugging as well. 

----
📚 Documentation preview 📚: https://anemoi-training--931.org.readthedocs.build/en/931/

----
📚 Documentation preview 📚: https://anemoi-graphs--931.org.readthedocs.build/en/931/

----
📚 Documentation preview 📚: https://anemoi-models--931.org.readthedocs.build/en/931/


cathalobrien:
nice! maybe we can try on CPU too with #927  ?

training/tests/integration/test_training_cycle.py
```diff
@@ -137,6 +157,26 @@ def test_training_cycle_lam(lam_config: tuple[DictConfig, list[str]], get_test_a
     AnemoiTrainer(cfg).train()
 
 
+@skip_if_offline
+@pytest.mark.slow
+@pytest.mark.multigpu
+@pytest.mark.parametrize(
```
anaprietonem: In the integration tests we just run with one gpu - so how would this be used?

ssmmnn11: all the tests now need to be run manually.

VeraChristina: Do we want these to run in the ci? If so, only on demand (with the single-gpu integration and benchmark tests or separately), or also nightly/weekly? Or do we just want to add them for local testing/debugging?

VeraChristina: (In the current set-up they will not be run in the ci workflows)

ssmmnn11: Good question. For now we might just want to run them manually - given that we already test a subset via the benchmark tests. But any ideas very much welcome!

training/tests/unit/distributed/test_strategy.py
```diff
@@ -0,0 +1,193 @@
...
+    graph[("data", "to", "hidden")].edge_index = edge_index_dh
+    graph[("data", "to", "hidden")].edge_attr = edge_attr_dh
+    graph[("hidden", "to", "hidden")].edge_index = edge_index_hh
+    graph[("hidden", "to", "hidden")].edge_attr = edge_attr_hh
+    graph[("hidden", "to", "data")].edge_index = edge_index_hd
+    graph[("hidden", "to", "data")].edge_attr = edge_attr_hd
+    return {"data": graph}
+
+
+def _build_tiny_model() -> AnemoiModelEncProcDec:
```
cathalobrien, line 91: could we replace this with a config file like we do in training/tests/integration/config?

ssmmnn11: Please do! :-)

models/src/anemoi/models/distributed/transformer.py
```diff
@@ -16,49 +16,23 @@
...
+from anemoi.models.distributed.primitives import _alltoallwrapper
 
 
-def _alltoallwrapper(output_list: list, input_list: list, group: ProcessGroup):
-    """Wrapper function for all_to_all across NCCL, MPI and Gloo backends.
-    There is no all_to_all primitive for the Gloo backend. In that case each
-    process broadcasts its tensor asynchronously.
+def _comm_send_chunks(chunks: tuple[Tensor, ...]) -> list[Tensor]:
+    """Normalize collective inputs to always use contiguous format."""
+    return [chunk.clone(memory_format=torch.contiguous_format) for chunk in chunks]
```
japols, line 24: is `torch.clone` necessary here? wouldn't `chunk.to(memory_format=...)` work without a copy in case it is already contiguous?

models/src/anemoi/models/distributed/transformer.py
```diff
@@ -68,34 +42,30 @@ def _headsalltoall(input_: Tensor, shapes: list, group: Optional[ProcessGroup] =
...
         torch.empty(
             (*input_shape[rank][:-3], heads_per_rank[myrank], seq_per_rank[rank], channels_per_rank[rank]),
             dtype=input_.dtype,
             layout=input_.layout,
             device=input_.device,
-            memory_format=input_format,
         )
         for rank in range(comm_size)
     ]
+    output_list = _comm_recv_chunks_like(recv_templates)
```
japols, line 64: similar comment here: do we need to create two copies of empty output_lists? why not just add `memory_format=...` above?

Issue #928, opened 2026-02-24
Stretched-Grid AutoEncoder fails during dataloading


icedoom888:
### What happened?

When training a Stretched Grid AE, the following error occurs:
```
[rank0]:   File "/users/apennino/projects/internal/devml/libs/anemoi-datasets/src/anemoi/datasets/data/indexing.py", line 97, in index_to_slices
[rank0]:     return _tuple_with_slices(_index_to_tuple(index, shape), shape)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/users/apennino/projects/internal/devml/libs/anemoi-datasets/src/anemoi/datasets/data/indexing.py", line 35, in _tuple_with_slices
[rank0]:     result = tuple(slice(*s.indices(shape[i])) for (i, s) in enumerate(result))
[rank0]:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/users/apennino/projects/internal/devml/libs/anemoi-datasets/src/anemoi/datasets/data/indexing.py", line 35, in <genexpr>
[rank0]:     result = tuple(slice(*s.indices(shape[i])) for (i, s) in enumerate(result))
[rank0]:                           ^^^^^^^^^
[rank0]: AttributeError: 'list' object has no attribute 'indices'
[rank0]: . Did you mean: '_return_value'?
srun: error: nid001016: task 0: Exited with exit code 1
srun: Terminating StepId=3948422.0
```
This happens only when training a SG AE and not with the normal AE. The only difference is in the dataset itself.

Issue is related to line: https://github.com/ecmwf/anemoi-core/blob/b198783325bb0a8458be6525cc00e1319de48f4b/training/src/anemoi/training/data/dataset.py#L188

The following fixes the issue:
```
def get_sample(
        self,
        time_indices: slice | int | list[int],
        grid_shard_indices: np.ndarray | slice | None = None,
    ) -> torch.Tensor:
        """Get a sample from the dataset."""
        if isinstance(grid_shard_indices, slice):
            # Load only shards into CPU memory
            # x = self.data[time_indices, :, :, grid_shard_indices] Used to be this
            x = self.data[time_indices]
            x = x[..., grid_shard_indices]
```

### What are the steps to reproduce the bug?

[...]

PR #935, opened 2026-02-26
fix: pytorch slice indexing

anaprietonem:
## Description
This PR is to fix the bug reported in https://github.com/ecmwf/anemoi-core/issues/928. The bug is not easy to trigger as it seems due to torch errors behind non-standard strides in memory. Currently `training/tests/unit/data/test_dataset.py/test_mimic_get_sample_raises_subset_indices_error` fails in main. 
This branch effectively solves this and the test passes. I have also added regression tests to ensure changes in get_sample returns same output as behaviour for already working tensors/lists. 

https://github.com/ecmwf/anemoi-core/issues/928

https://github.com/ecmwf/anemoi-core/issues/928

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/config/stretched.yaml
+++ b/training/src/anemoi/training/config/stretched.yaml
@@ -5,7 +5,7 @@ defaults:
 - system: example
 - graph: stretched_grid
 - model: graphtransformer
-- training: stretched
+- training: autoencoder
 - _self_
 
 config_validation: True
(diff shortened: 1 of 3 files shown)


icedoom888:
Could we also clarify why in multidataset:
https://github.com/ecmwf/anemoi-core/blob/b198783325bb0a8458be6525cc00e1319de48f4b/training/src/anemoi/training/data/multidataset.py#L375

`slice(start, end)` is passed as a grid shard shape  to the normal dataset
https://github.com/ecmwf/anemoi-core/blob/b198783325bb0a8458be6525cc00e1319de48f4b/training/src/anemoi/training/data/dataset.py#L183

PR #937, opened 2026-02-27
chore(configs): Interpolator config cleanup

### LINKED ISSUE #936: Fix and cleanup interpolator training config

Currently, the interpolator's default config does not include a normalization along the time dimension.
This results in a difference in the range of the loss and metrics compared to previous versions of the model, such as the "single step" interpolator, recently deprecated.

The solution simply requires to add the correct scalers in the default configs. As part of this issue, a general clean-up could be done.

_No response_

### Additional context

I am willing to deal with this issue, I will soon open a PR and link it.

### Organisation

ETH / MeteoSwiss

MicheleCattaneo:
## Description

With this PR, the interpolator config gets cleaned up. In particular, the scaler along the time dimension (`UniformTimeStepScaler`) is included in the default configuration.

Closes #936 

### DIFF
--- a/training/src/anemoi/training/schemas/training.py
+++ b/training/src/anemoi/training/schemas/training.py
@@ -495,10 +495,11 @@ class InterpolationMultiSchema(BaseTrainingSchema):
     "Training objective."
     explicit_times: ExplicitTimes
     "Time indices for input and output."
-    
+
     # Needed to allow to override default training configuration
     # Forced to be None (null)
     rollout: Literal[None] = None
+
 
 TrainingSchema = Annotated[
     ForecasterSchema


training/src/anemoi/training/config/interpolator.yaml
```diff
@@ -28,11 +28,3 @@ model:
     _target_: anemoi.models.layers.residual.SkipConnection
     step: 0 # use the previous step as skip connection
   latent_skip: False # True/False for skip connection on latent mesh
-
-training:
-  # Instead of inferred using multistep and timeincrement, specify time indices explicitly
-  explicit_times:
-    input: [0, 6]
-    target: [1, 2, 3, 4, 5]
-  multistep_output: 5
```
mc4117, line 37: I would be tempted to keep this multistep_output override in the main config file. Just because it is such an important parameter for this model. But this is a style choice

dietervdb-meteo: How about removing it from the lower level config then?

MicheleCattaneo: I was not sure where it was best to put it. I'm happy to put it back in the top level config as you suggest, I agree it's nice to see it clearly

MicheleCattaneo: Thinking about it, if I remove it from the lower level config, one can not use the lower level `config/training/interpolator.yaml`, with a base config that is not `config/interpolator.yaml`, unless they remember to add `explicit_times` and `multiste_output` in the training overrides of the other base config.
what do you think?

dietervdb-meteo: Lookign a bit closer, is there any reason to keep a separate `training/interpolator.yaml`? We could jsut call the`training/default.yaml` and if need be do a few more overrides [but actually think they might not even be needed]. @mc4117 ?

mc4117: Yes I like the idea of removing the lower level config and overriding from above

MicheleCattaneo: [8595351](https://github.com/ecmwf/anemoi-core/pull/937/commits/859535160a2b1bf7f65d111d9f48744025b7181f) should implement this

PR #940, opened 2026-03-02
fix(models): use flex attention block mask for sdpa windowed attention 

ssmmnn11:
~~Remove option for sliding window attention with pytorch attention implementation. It tried to materialise the full attention matrix which causes OOM.~~ 

Stop SDPA from OOM-ing when using sliding window. Use the block mask feature from flex attention to avoid materialising the full seq_len^2 attention matrix.


models/src/anemoi/models/layers/attention.py
```diff
@@ -253,26 +284,30 @@ def forward(
...
+            raise NotImplementedError(
                 "Softcap not supported by Pytorchs SDPA. please switch to flash attention or disable softcap."
             )
         if alibi_slopes is not None:
-            NotImplementedError(
+            raise NotImplementedError(
                 "Alibi slopes not supported by Pytorchs SDPA. please switch to flash attention v2 or disable alibi slopes."
             )
+        if window_size is not None and self.attn_mask is None:
+            # build the attention mask for sliding window attention. We build the mask once and reuse it, since it is the same for every forward pass (assuming the sequence length does not change).
```
japols: good to note this, but maybe a bit too long for a single line? ;)

cathalobrien: is that you pre-commit? Ok I'll split it over two :)

models/tests/layers/test_attention.py
```diff
@@ -173,3 +174,83 @@ def test_multi_head_cross_attention_backward_sdpa(batch_size, num_heads, embed_d
...
+    # With a small window on a 16-token sequence, outputs should differ
+    assert not torch.allclose(
+        output, output_global, atol=1e-5
+    ), "Sliding window output should differ from global attention output"
+
+    # Memory usage using sliding window should not be greater then using global attention
+    # Since flex_attentions block mask funciton is used to handle the sliding window,
+    # masking naively with a seq_len^2 array
+    assert (
+        peak_alloc_memory_after_sliding_window_mb <= peak_alloc_memory_after_global_mb
```
ssmmnn11, line 237: should this be <= or something like <= ...  * 1.05, e.g. 5%?

cathalobrien: No i don't think 5% extra tolerance is needed. its not clear to me when the memory usage of sliding window would be more then global

PR #943, opened 2026-03-02
feat!: introduce mlp implementation for gated FFN (swiglu, ...)

ssmmnn11:
Replaces the previous approach of passing GLU-based activation classes via layer_kernels.Activation with a dedicated mlp_implementation parameter on all processor, encoder, and decoder components.

What changed
New mlp_implementation parameter — accepts "mlp" (default), "glu", "swiglu", "geglu", or "reglu". When a gated variant is selected, hidden layers use the new GatedMLPLayer which maintains separate gate and value projections, and layer_kernels.Activation is ignored with a UserWarning.

GLU, SwiGLU, GEGLU, ReGLU removed from activations.py — these classes are superseded by mlp_implementation. Passing a GLU-named class as layer_kernels.Activation now raises a ValueError with a clear message directing users to mlp_implementation.

All model configs updated — processor, encoder, and decoder blocks across all model configs now expose mlp_implementation: mlp (default) with a comment noting the recommended mlp_hidden_ratio of ~2.67 for gated variants.

Migration

# Before (no longer supported)
layer_kernels:
  Activation:
    _target_: anemoi.models.layers.activations.SwiGLU
    dim: 512

# After
processor:
  mlp_implementation: swiglu
  mlp_hidden_ratio: 2.67

This PR is marked as breaking config change since any configs using the old GLU/SwiGLU/GEGLU/ReGLU classes via layer_kernels.Activation will break, they must switch to mlp_implementation

----
📚 Documentation preview 📚: https://anemoi-training--943.org.readthedocs.build/en/943/

----
📚 Documentation preview 📚: https://anemoi-graphs--943.org.readthedocs.build/en/943/

----
📚 Documentation preview 📚: https://anemoi-models--943.org.readthedocs.build/en/943/


mchantry:
@ssmmnn11 it's a great contribution. It would be valuable to add some scientific results to support this great work.

models/src/anemoi/models/schemas/common_components.py
```diff
@@ -35,8 +37,10 @@ class Config:
     "Enable gradient checkpointing to reduce memory usage. Default to True."
     num_chunks: NonNegativeInt = Field(example=1)
     "Number of chunks to divide the layer into. Default to 1."
-    mlp_hidden_ratio: NonNegativeInt = Field(example=4)
-    "Ratio of mlp hidden dimension to embedding dimension. Default to 4."
+    mlp_hidden_ratio: PositiveFloat = Field(example=4)
+    "Ratio of MLP hidden dimension to embedding dimension. Use 4 for `mlp`, and ~2.67 for gated variants (`glu`, `swiglu`, `geglu`, `reglu`) to keep model size similar."
```
anaprietonem, line 41: Could be good to explain why the ~2.67 factor. Also if this is a known factor, could we just hardcoded or raise a warning if the wrong one is passed?

models/src/anemoi/models/layers/block.py
```diff
@@ -505,7 +519,6 @@ def __init__(
 
         Linear = layer_kernels.Linear
         LayerNorm = layer_kernels.LayerNorm
-        Activation = layer_kernels.Activation
```
anaprietonem, line 517: If we are dropping this, is this PR a breaking change as the configs won't be accepting the Activation key as part of the layer kernels or rather won't make sense to any longer specify this?

anaprietonem (changes requested):
Thanks Simon, left some comment I'd like to clarify. Also we'd no longer need the functions here [https://github.com/ecmwf/anemoi-core/blob/992a300e6add4894d2ab2b5d44a50c738cdbdc1b/models/src/anemo…](https://github.com/ecmwf/anemoi-core/blob/992a300e6add4894d2ab2b5d44a50c738cdbdc1b/models/src/anemoi/models/layers/activations.py#L4) so should those be deleted?

models/src/anemoi/models/layers/block.py
JPXKQX: I think the Activation is still used, but inside the MLP

models/src/anemoi/models/schemas/common_components.py
JPXKQX: I think this factor is used to ensure that the number of trainable parameters of the MLP block is identical. In that case, the only way to specify the same number for all combinations would be to include something like the total number of parameters (instead of `mlp_hidden_ratio`), but I don't think this makes sense, as we would need to round to the nearest valid integer.

models/src/anemoi/models/layers/conv.py
```diff
@@ -57,7 +59,8 @@ def __init__(
             out_channels,
             out_channels,
             layer_kernels=layer_kernels,
-            n_extra_layers=mlp_extra_layers,
+            n_extra_layers=mlp_extra_layers + 1,
```
JPXKQX, line 62: why has this changed now?

models/src/anemoi/models/layers/mlp.py
```diff
@@ -17,6 +19,80 @@
...
+def _build_gating_activation(mlp_implementation: MLPImplementation) -> nn.Module:
+    if mlp_implementation == "glu":
+        return nn.Sigmoid()
+    if mlp_implementation == "swiglu":
+        return nn.SiLU()
+    if mlp_implementation == "geglu":
+        return nn.GELU()
+    if mlp_implementation == "reglu":
+        return nn.ReLU()
+    valid = ("mlp", "glu", "swiglu", "geglu", "reglu")
```
JPXKQX: is `mlp` a valid implementation?
```suggestion
    valid = ("glu", "swiglu", "geglu", "reglu")
```

training/README.md
```diff
@@ -32,6 +32,31 @@ Miscellanous tools for training data-driven weather forecasts.
 
 The documentation can be found at https://anemoi-training.readthedocs.io/.
 
+## MLP Implementations
```
JPXKQX: I would move this to the `docs` to its own section under anemoi-models > Layers > ..

training/src/anemoi/training/config/model/point_wise.yaml
```diff
@@ -27,7 +27,7 @@ processor:
   _target_: anemoi.models.layers.processor.PointWiseMLPProcessor
   num_layers: 16
   num_chunks: 2
-  mlp_hidden_ratio: 4
+  mlp_hidden_ratio: 4 # GraphTransformer/Transformer only; use 4 (mlp) or ~2.67 (glu/swiglu/geglu/reglu)
```
JPXKQX: PointWiseMLP should be added to the valid mappers/processor

models/src/anemoi/models/layers/conv.py
JPXKQX: Ok, I see it has been moved from the `for loop` inside the `MLP()`

JPXKQX (approved):
Thanks Ana!

Issue #955, opened 2026-03-05
Default radndom base seed


MeraX:
### Is your feature request related to a problem? Please describe.

It's a chord always having to set `ANEMOI_BASE_SEED` before training when training by hand. Thus I propose adding a fallback to `random.randint()` instead of throwing an error.

### Describe the solution you'd like

update get_base_seed in [training/src/anemoi/training/utils/seeding.py#L39-44](https://github.com/ecmwf/anemoi-core/blob/d71d06525614fa42b5dc6376b7c2d92b5180df85/training/src/anemoi/training/utils/seeding.py#L34-L38) to
```
    base_seed_threshold = 1000
    for env_var in env_var_list:
        if env_var in os.environ:
            base_seed = int(os.environ.get(env_var))
            break
    else:
        base_seed = random.randint(base_seed_threshold, 2**31 - 1)
        LOGGER.info(f"Randomly set {base_seed=}. (Use environment variables {env_var_list} to override.)")
```

### Challenge

seeding parallel processes.

PR #957, opened 2026-03-07
fix(training): add base seed fallback

ssmmnn11:
Removes the requirement to set a manual base seed if training is run without slurm. 

Fix for #955 

----
📚 Documentation preview 📚: https://anemoi-training--957.org.readthedocs.build/en/957/

----
📚 Documentation preview 📚: https://anemoi-graphs--957.org.readthedocs.build/en/957/

----
📚 Documentation preview 📚: https://anemoi-models--957.org.readthedocs.build/en/957/

### DIFF
--- a/models/src/anemoi/models/distributed/primitives.py
+++ b/models/src/anemoi/models/distributed/primitives.py
@@ -54,6 +54,89 @@ def _split(input_: Tensor, dim_: int, shapes_: tuple, group: Optional[ProcessGro
     return output
 
 
+def _gather_into_tensor(
+    input_: Tensor,
+    dim_: int,
+    shapes: tuple,
+    group: ProcessGroup,
+) -> Tensor:
+    input_format = get_memory_format(input_)
+    input_ = input_.contiguous(memory_format=input_format)
+
+    out_shape = list(input_.shape)
+    out_shape[dim_] = sum(shape[dim_] for shape in shapes)
+
+    output = torch.empty(
+        out_shape, dtype=input_.dtype, layout=input_.layout, device=input_.device, memory_format=input_format
+    )
+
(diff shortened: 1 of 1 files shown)


training/src/anemoi/training/utils/seeding.py
```diff
@@ -37,7 +38,8 @@ def get_base_seed(base_seed_env: str | None = None) -> int:
             base_seed = int(os.environ.get(env_var))
             break
 
-    assert base_seed is not None, f"Base seed not found in environment variables {env_var_list}"
+    if base_seed is None:
+        base_seed = 42
```
MeraX, line 42:
```suggestion
        base_seed = 42
        LOGGER.info(f"Set fixed value {base_seed=}. (You can use environment variables {env_var_list} to override.)")
```

MeraX:
@mc4117, please have a look. I know, you suggested using a random fallback. However, @ssmmnn11 pointed out that this would make seeding all parallel processes and workers quite cumbersome. Would you approve this simpler solution?

training/src/anemoi/training/utils/seeding.py
ssmmnn11: Outputting a info / warning here means it will appear everytime get_base_seed is called. I would prefer to rely on the diagnostic seed output we have elswhere in the code: each [worker](https://github.com/ecmwf/anemoi-core/blob/497442888f96ff9cad214b552307a56c8ddfb316/training/src/anemoi/training/data/multidataset.py#L354) / [task](https://github.com/ecmwf/anemoi-core/blob/497442888f96ff9cad214b552307a56c8ddfb316/training/src/anemoi/training/train/train.py#L128) prints out seeds and a sanity random number, so seeds are easily verifiable.

Do you think that is sufficient?

PR #958, opened 2026-03-08
fix(training): multiscale wrapper sharding is broken after multi dataset merge

ssmmnn11:
Fix multiscale loss wrapper sharding after multi dataset merge.

### DIFF
--- a/.github/workflows/integration-tests-hpc.yml
+++ b/.github/workflows/integration-tests-hpc.yml
@@ -75,7 +75,7 @@ jobs:
             source .venv/bin/activate
             pip install --upgrade pip
             pip install -e ./training[all,tests,profile] -e ./models[all,tests] -e ./graphs[all,tests]
-            srun python -m pytest -v training/tests/integration --slow --multigpu -k "test_benchmark_training_cycle"
+            srun python -m pytest -v training/tests/integration/test_benchmark.py --slow --multigpu
             deactivate
             rm -rf $REPO_NAME
           sbatch_options: |
--- a/training/src/anemoi/training/data/multidataset.py
+++ b/training/src/anemoi/training/data/multidataset.py
@@ -371,8 +371,15 @@ class MultiDataset(IterableDataset):
 
         x = {}
         for name, dataset in self.datasets.items():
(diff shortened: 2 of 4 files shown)


training/src/anemoi/training/losses/multiscale.py
```diff
@@ -47,7 +47,10 @@ def __init__(
         weights : list[float]
             Per-scale loss weights
         keep_batch_sharded : bool
```
japols, line 49: naive question: aren't we deciding this based on `grid_shard_shapes/slice is not None` in forward?

training/src/anemoi/training/losses/multiscale.py
```diff
@@ -179,21 +194,25 @@ def forward(
...
-        **_kwargs,
-    ) -> list[torch.Tensor]:
-
+        **kwargs,
+    ) -> torch.Tensor:
         shard_shapes, shard_shapes_y = None, None
-        if model_comm_group_size and model_comm_group_size > 1 and self.keep_batch_sharded:
+        # The sharded smoothing path only applies when the batch already enters
+        # the loss sequence-sharded across the model group.
+        is_model_sharded = self.keep_batch_sharded and group is not None and group.size() > 1
```
japols: I think our pattern should be: `is_model_sharded = grid_shard_shapes is not None` and then the communication primitive handles cases where `group is None` or `group.size() == 1`

training/src/anemoi/training/losses/base.py
```diff
@@ -201,6 +201,11 @@ def name(self) -> str:
         """Used for logging identification purposes."""
         return self.__class__.__name__.lower()
 
+    @property
+    def needs_shard_layout_info(self) -> bool:
+        """Whether the loss needs explicit shard-layout metadata beyond grid_shard_slice/group."""
```
japols, line 217: wondering if we really need this. we could just pass `grid_shard_shapes` instead of `grid_shard_slice` everywhere and derive the slice from that.

maybe worth revisiting in #964

ssmmnn11: yes, I think that would be nice as a follow up

PR #959, opened 2026-03-09
feat(training): support flash attention v4

cathalobrien:
## Description
support flash attention v4 in the transformer processor.

flash attention v4 is optimised for Blackwell, but runs on hoppers and amperes too. the main advantage is that it can simply be pip installed instead of having to be compiled by source.

Unfortunately on Ampere and Hopper GPUs some features are not supported yet in flash attention v4:
1) sliding window
2) global backward pass with head dim < 128

From my testing on Hopper, there is no speedup between fa4 and fa3. Likely on Blackwell this would be different since fa4 could take advantage of new hardware support there

```
=== Benchmarking: o96-2048c fwd+bwd ===
Z=1 H=16 N_CTX=40320 HEAD_DIM=128 window_size=0 causal=False dtype=torch.float16

Flash-Attention 2:
flash_attn_func
        elapsed time: 142.26 ms / iter
        peak memory usage: 2570.42 MB


Flash-Attention 3:
flash_attn_func
        elapsed time: 74.26 ms / iter
        peak memory usage: 2572.91 MB


Flash-Attention 4:
flash_attn_func
        elapsed time: 74.09 ms / iter
        peak memory usage: 2575.34 MB
```

still we should make the interface available in anemoi for when flash attention 4 is upgraded.

I have tested this code with flash attention v2, v3 and v4 and it runs


models/src/anemoi/models/layers/attention.py
```diff
@@ -349,38 +347,74 @@ def _init_rotary_embeddings(self, use_rotary_embeddings: bool, head_dim: int, fl
...
 
         returns:
-            flash attention module
-            use_flash_attention_v3 (bool)
+            tuple of (flash attention function, parsed version)
         """
-        use_flash_attn_v3 = False
+        flash_attn_version = (
+            -1
+        )  # will be set to a valid version if either flash attention v2, v3 or v4 is successfully imported
```
HCookie:
```suggestion
        flash_attn_version = -1
        # will be set to a valid version if either flash attention v2, v3 or v4 is successfully imported
```

models/src/anemoi/models/layers/attention.py
```diff
@@ -349,38 +347,74 @@ def _init_rotary_embeddings(self, use_rotary_embeddings: bool, head_dim: int, fl
...
+            flash_attn_func = flash_attn_func_v3
+            self.use_flash_attn_v3 = True
+            self.use_flash_attn_v4 = False
         else:
+            # using flash attn v2
             LOGGER.info("Using flash attention v2")
-        return flash_attn, use_flash_attn_v3
+            flash_attn_func = flash_attn_func_v2
+            self.use_flash_attn_v3 = False
+            self.use_flash_attn_v4 = False
```
HCookie: A match statement makes more sense here, or just something else, I'm confused by this....

cathalobrien: Do you prefer this @HCookie ? 

I can't just import flash_attn.__version__ and branch off that, because you can have all 3 packages installed at the same time and flash_attn.__version__ returns the version of FA2. 

```
    def _import_flash_attn(self) -> tuple:
        """imports either flash attention v2, v3 or v4, based on what is installed. prioritising v4, then v3, then v2. if none are installed, raises an error.

        returns:
            flash attention function
        """
        # will be set to a valid version if either flash attention v2, v3 or v4 is successfully imported
        flash_attn_func = None

        self.use_flash_attn_v3 = False
        self.use_flash_attn_v4 = False

        e_v4 = None
        e_v3 = None
        e_v2 = None

        try:
            from flash_attn.cute import flash_attn_func

            LOGGER.info("Using flash attention v4")
            self.use_flash_attn_v4 = True
            return flash_attn_func
        except ImportError as e:
            e_v4 = e
            LOGGER.debug(f"Flash attention v4 not available: {e_v4}")

        try:
            from flash_attn_interface import flash_attn_func

            LOGGER.info("Using flash attention v3")
            self.use_flash_attn_v3 = True
            return flash_attn_func
        except ImportError as e:
            e_v3 = e
            LOGGER.debug(f"Flash attention v3 not available: {e_v3}")
        try:
            from flash_attn import flash_attn_func

            LOGGER.info("Using flash attention v2")
            return flash_attn_func
        except ImportError as e:
            e_v2 = e
            LOGGER.debug(f"Flash attention v2 not available: {e_v2}")

        raise ImportError(
            "Flash attention is not installed. Please install flash attention v4, v3 or v2 to use this attention implementation. "
            f"Attempted imports resulted in the following errors: "
            f"v4 import error: {e_v4} "
            f"v3 import error: {e_v3} "
            f"v2 import error: {e_v2} "
        )

```

HCookie: Sorry, what, they can all be installed at once ....

Issue #966, opened 2026-03-11
Remove graph dependency in train/tasks/base.py


JPXKQX:
### Is your feature request related to a problem? Please describe.

The training task currently depends on graph internals. This causes tight coupling across training, loss/scalers, diagnostics, and distributed setup, which is a problem for dynamic setups where we want to build the graph later (inside the model).

### Describe the solution you'd like

Refactor data flow in `tasks/` so responsibilities are cleanly separated:
- Move graph loading/creation into AnemoiModelInterface (pass graph_config, not graph_data).
- Let tasks read grid/mask/latlon metadata from datasets/datamodule (trainer.datamodule in lightning `setup(stage="fit")`???).
- Move `shard_shapes` computation into `BaseDDPStrategy.setup()`.
- Replace graph-attribute-based area-weight scalers with scalers that compute from dataset lat/lon (+ cutout mask for reweighting).
- Refactor output mask/diagnostic code to use latlons & cutout_mask from datasets/datamodule.

### Describe alternatives you've considered

- Keep current architecture and implement 2 levels of graph_creation (static vs dynamic): complexity increases over time and harder to maintain.
- Pass more constructor args into tasks (latlons, grid_size, cutout_mask, ...).

### Additional context

Expected outcomes:
- Cleaner separation of concerns and fewer torch-geometric leaks outside model internals.
- Better Lightning lifecycle alignment (especially distributed setup).
- Easier testing and extension of datasets, strategies, and losses.

Validation:
- Run training tests/integration paths.
- Test sharding. Confirm shard shapes match previous behaviour.
- Verify checkpoint supporting arrays still include expected masks/lat/lon fields.

### Organisation

ECMWF

PR #968, opened 2026-03-11
feat: automatic data nodes creation based on dataset entries

icedoom888:
## Description
This PR tackles automatric data nodes creation based on dataset created in the dataloader.

Reduces redundancy avoiding multiple specifications of the same datasets.
Realtes to #966 

----
📚 Documentation preview 📚: https://anemoi-training--968.org.readthedocs.build/en/968/

----
📚 Documentation preview 📚: https://anemoi-graphs--968.org.readthedocs.build/en/968/

----
📚 Documentation preview 📚: https://anemoi-models--968.org.readthedocs.build/en/968/


training/src/anemoi/training/train/train.py
```diff
@@ -29,6 +29,7 @@
 from pytorch_lightning.utilities.rank_zero import rank_zero_only
 from torch_geometric.data import HeteroData
 
+from anemoi.graphs.utils.config import integrate_data_nodes_in_config
```
MeraX, line 32:
```suggestion
from anemoi.models.utils.config import get_multiple_datasets_config
```
Is it possible to re-use the existent copy of this function?

graphs/src/anemoi/graphs/utils/config.py
```diff
@@ -0,0 +1,84 @@
...
+LOGGER = logging.getLogger(__name__)
+
+DEFAULT_DATASET_NAME = "data"
+DEFAULT_NODE_ATTR = {
+    "area_weight": {
+        "_target_": "anemoi.graphs.nodes.attributes.SphericalAreaWeights",
+        "norm": "unit-max",
+        "fill_value": 0,
+    }
+}
```
MeraX, line 24: This PR also introduces yet another place where defaults are stored (`DEFAULT_NODE_ATTR`). Are these defaults actually necessary or would it be possible to set these defaults simply as the default python argument values of `anemoi.graphs.nodes.AnemoiDatasetNodes`?

graphs/src/anemoi/graphs/utils/config.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    if "datasets" in config:
+        if isinstance(config, dict):
+            return config["datasets"]
+        return config.datasets
+
+    return OmegaConf.create({default_dataset_name: config})
+
+
+def integrate_data_nodes_in_config(config: DictConfig):
```
MeraX, line 39: I'm not sure how it's handled elsewhere in the code, but I would suggest noting here that `config` is actually modified. Thus I would also recommend considering returning nothing in this function.

training/src/anemoi/training/train/train.py
```diff
@@ -166,6 +167,9 @@ def graph_data(self) -> HeteroData:
         else:
             graph_filename = None
 
+        # Introduce Data Nodes in graph config
+        self.config = integrate_data_nodes_in_config(self.config)
```
MeraX, line 171: Are you manipulating the global `config` object here? Which version of the config is logged to MLflow?

Did you consider adding a flag to the configuration that explicitly enables the automatic generation of `AnemoiDatasetNodes` nodes? I think this would help catching config errors where the the user specification of custom note builders is dismissed due to a typo.

MeraX (commented):
Hi Alberto, thank you for this contribution. I think it helps simplifying the config. At the same time it makes the discoverability of alternative node builders from the configs alone harder, but I think that should be acceptable as long as the note builders are kept in the documentation.

I've put a few comments regarding the maintainability and user-friendliness, that I', happy to discuss.

graphs/src/anemoi/graphs/utils/config.py
MicheleCattaneo: I would agree, these global variables seem unecessary

training/src/anemoi/training/train/train.py
icedoom888: Ideally this was done to remove dependency of graphs from models!

graphs/src/anemoi/graphs/utils/config.py
icedoom888: At the time being "area_weight" is a necessary attribute for any data nodes, this isn't enforced in the code, and test will fail if not provided. 
I could try and find another workaround by enforcing creation somewhere else, but not sure where. Suggestions?

training/src/anemoi/training/train/train.py
icedoom888: Very good point @MeraX, will look into this!

PR #969, opened 2026-03-12
feat: network naming

icedoom888:
## Description
Introduces network naming: each netwrok component will be assigned a name. eg: encoder_era5, processor, decoder_era5..

This will be used in an upcoming PRs to introduce targeted model freezing and module loading:

- Freeze only specific networks.
   In a multi dataset setting you might want to freeze only certain parts of the full model, whilst training the rest.
   Example: freeze the global encoder and decoder trained with era5, and train only the LAM specific encoder and decoders and the common processor
  `submodules_to_freeze: [encoder_era5, decoder_era5]`
  
  ```
  def freeze_module_by_name(model, name):
    for module in model.modules():
        if getattr(module, "name", None) == name:
            for p in module.parameters():
                p.requires_grad = False
   ```

- Load only certain network weights from a given checkpoint.
  In a similar example you might want to load the encoder and decoder from a pretrained global model, but loading the processor from another model that was trained with global and LAM data already. 
   ```
   transfer_learning_dict: {
        dfr365473f43f: ['encoder_era5', 'decoder_era5']
        36zgrvg45g66: ['processor']
   }
   ```

### DIFF
--- a/models/src/anemoi/models/schemas/common_components.py
+++ b/models/src/anemoi/models/schemas/common_components.py
@@ -40,7 +40,7 @@ class TransformerModelComponent(PydanticBaseModel):
     "Number of attention heads. Default to 16."
     layer_kernels: Union[dict[str, dict], None] = Field(default_factory=dict)
     "Settings related to custom kernels for encoder processor and decoder blocks"
-    name: str = Field(default=None, example="Encoder/Decoder/Processor")
+    name: str | None = Field(default=None, example="Encoder/Decoder/Processor")
     "Name of the model"
 
 
@@ -61,7 +61,7 @@ class GNNModelComponent(BaseModel):
     "The number of extra hidden layers in MLP. Default to 0."
     layer_kernels: Union[dict[str, dict], None] = Field(default_factory=dict)
     "Settings related to custom kernels for encoder processor and decoder blocks"
-    name: str = Field(default=None, example="Encoder/Decoder/Processor")
+    name: str | None = Field(default=None, example="Encoder/Decoder/Processor")
     "Name of the model"
 
 
@@ -76,5 +76,5 @@ class PointWiseModelComponent(BaseModel):
     "Enable gradient checkpointing to reduce memory usage. Default to True."
     layer_kernels: Union[dict[str, dict], None] = Field(default_factory=dict)
     "Settings related to custom kernels for encoder processor and decoder blocks"
-    name: str = Field(default=None, example="Encoder/Decoder/Processor")
+    name: str | None = Field(default=None, example="Encoder/Decoder/Processor")
     "Name of the model"


ssmmnn11 (changes requested):
I am not keen on having names in these dataset agnostic parts of the model. Could this be solved by relying on self.encoder[dataset_name]  ... ?

models/src/anemoi/models/models/autoencoder.py
```diff
@@ -47,6 +47,7 @@ def _build_networks(self, model_config: DotDict) -> None:
 
             self.encoder[dataset_name] = instantiate(
                 model_config.model.encoder,
+                name=f"encoder_{dataset_name}",
```
MeraX, line 50: In principle, I think it's worth thinking about a way to reference individual components depending on their dataset. However, I would recommend not building any names by concatenating strings with a delimiter that could also be part of the dataset name itself.
I would prefer referencing the components by more like a path, e.g. a list `["encoder", "icon_dream"]`. Isn't there perhaps already an inventory dict with all encodes and dataset_names as keys?

ssmmnn11: https://docs.pytorch.org/tutorials/recipes/recipes/what_is_state_dict.html

Would be the recommended approach.

radiradev: can we use a module dict here e.g something like, then we still have the statedict but also can reference by dataset name or any othername
```python
class Net(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.encoders = nn.ModuleDict({
            "era5": nn.Conv2d(2, 8, kernel_size=3, padding=1), 
            "icon": nn.Conv2d(2, 8, kernel_size=3, padding=1)
        })
        self.decoders = nn.ModuleDict({
            "era5": nn.Conv2d(8, 2, kernel_size=3, padding=1),  
            "icon": nn.Conv2d(8, 2, kernel_size=3, padding=1)
        })

    def forward(self, datasets): 
        out = {}
        for dataset in datasets:
            x = self.encoders[dataset](datasets[dataset])
            x = self.decoders[dataset](x)
            out[dataset] = x
        return out
```

```python
print("Model's state_dict:")
for param_tensor in net.state_dict():
    print(param_tensor, "\t", net.state_dict()[param_tensor].size())
```
```
Model's state_dict:
encoders.era5.weight     torch.Size([8, 2, 3, 3])
encoders.era5.bias       torch.Size([8])
encoders.icon.weight     torch.Size([8, 2, 3, 3])
encoders.icon.bias       torch.Size([8])
decoders.era5.weight     torch.Size([2, 8, 3, 3])
decoders.era5.bias       torch.Size([2])
decoders.icon.weight     torch.Size([2, 8, 3, 3])
decoders.icon.bias       torch.Size([2])
```

JPXKQX: I would say that this is exactly the case in the main branch

JPXKQX: https://github.com/ecmwf/anemoi-core/blob/main/models/src/anemoi/models/models/encoder_processor_decoder.py#L37

icedoom888: @radiradev This is the current implementation I think!

radiradev: I guess there might be scenarios where you don't want the (e.g) encoder keys to match the dataset keys, but maybe not in the scope of this PR

PR #973, opened 2026-03-12
feat(models): remap supports keyword arguments + additional remaps: atanh, asinh, power, affine, displace_boundary_atoms

elkir:
## Description
This allows remap processors to pass parameters from the config to the transform functions/converters. Several remap converters are implemented: renormalised atanh, asinh, pure power transform, affine transform and a special transform that shifts boundary atomic values  away from the fixed bounds for easier modelling

#### `method_kwargs`
Many transforms are a family of functions with input parameters; this change allows us to specify these parameters from the config.

#### Extra transforms
For variables such as tcc, there is a lot of exact atomic mass at 0 and 1, with only a fraction of the values being in between, often clumped to the endpoints. This can be solved by an atanh transform distributing the values equally through the interval, together with the dispalce_boundary_atoms: shifting the boundaries away and then clamping them back to the boundary.

Similarly, for tp, this is usually highly skewed with a lot of mass at exactly 0. A power transform, together with a left boundary atom shift, obtains a region for modelling dry conditions and a reasonably Gaussian wet distribution.

See examples in [Preprocessing — Anemoi Models 0.0.0.post58 documentation](https://anemoi--973.org.readthedocs.build/projects/models/en/973/modules/preprocessing.html)

##  Additional notes ##
- This also cleans up old unused mappings from the deprecated multimapper, which are no longer useful

----
📚 Documentation preview 📚: https://anemoi-models--973.org.readthedocs.build/en/973/

----
📚 Documentation preview 📚: https://anemoi-training--973.org.readthedocs.build/en/973/

----
📚 Documentation preview 📚: https://anemoi-graphs--973.org.readthedocs.build/en/973/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
   - id: python-check-blanket-noqa # Check for # noqa: all
   - id: python-no-log-warn # Check for log.warn
 - repo: https://github.com/psf/black-pre-commit-mirror
-  rev: 26.1.0
+  rev: 26.3.1
   hooks:
   - id: black
     args: [--line-length=120]
@@ -41,7 +41,7 @@ repos:
     - --profile black
     - --project anemoi
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.15.4
+  rev: v0.15.9
   hooks:
   - id: ruff
     args:
@@ -59,12 +59,12 @@ repos:
   - id: docconvert
     args: ["numpy"]
 - repo: https://github.com/tox-dev/pyproject-fmt
-  rev: "v2.16.2"
+  rev: "v2.21.0"
   hooks:
   - id: pyproject-fmt
     args: ["--max-supported-python", "3.12"]
 - repo: https://github.com/jshwi/docsig # Check docstrings against function sig
-  rev: v0.79.0
+  rev: v0.82.2
   hooks:
   - id: docsig
     args:
--- a/graphs/pyproject.toml
+++ b/graphs/pyproject.toml
@@ -6,7 +6,6 @@
 # In applying this licence, ECMWF does not waive the privileges and immunities
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
-
 [build-system]
 requires = [ "setuptools>=60", "setuptools-scm>=8" ]
 
--- a/models/pyproject.toml
+++ b/models/pyproject.toml
@@ -6,7 +6,6 @@
 # In applying this licence, ECMWF does not waive the privileges and immunities
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
-
 [build-system]
 build-backend = "setuptools.build_meta"
 requires = [ "setuptools>=61", "setuptools-scm>=8" ]
--- a/training/pyproject.toml
+++ b/training/pyproject.toml
@@ -6,7 +6,6 @@
 # In applying this licence, ECMWF does not waive the privileges and immunities
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
-
 [build-system]
 requires = [ "setuptools>=60", "setuptools-scm>=8" ]
 
@@ -206,5 +205,5 @@ lint.ignore = [
   "TD003",
   "UP007",
 ]
-lint.flake8-import-conventions.extend-aliases."pytorch_lightning" = "pl"
+lint.flake8-import-conventions.extend-aliases.pytorch_lightning = "pl"
 lint.pydocstyle.convention = "numpy"


models/docs/modules/preprocessing.rst
```diff
@@ -25,6 +25,120 @@ the following classes:
    :no-undoc-members:
    :show-inheritance:
 
+**********
+Remapper
+**********
+
+The remapper module is used to do online transformations of the data using a set of predefined transforms and their inverses. This process is crucial for variables with pathological distributions, such as variables with a sharp peaks, long tails or other non-Gaussian shapes. It is especially important for diffusion models where the data distribution interacts with the noise distribution.
```
sahahner:
```suggestion
The remapper module is used to do in-place transformations of the data using a set of predefined transforms and their inverses. This process is crucial for variables with pathological distributions, such as variables with sharp peaks, long tails or other non-Gaussian shapes. It is especially important for diffusion models where the data distribution interacts with the noise distribution.
```

models/docs/modules/preprocessing.rst
```diff
@@ -25,6 +25,120 @@ the following classes:
...
+
+Example configuration:
+
+.. code:: yaml
+
+   data:
+      processors:
+         normalizer:
+           _target_: anemoi.models.preprocessing.multi_dataset_normalizer.TopNormalizer
+           _convert_: all
```
sahahner: Is the `_convert_: all` option necessary?

models/docs/modules/preprocessing.rst
```diff
@@ -25,6 +25,120 @@ the following classes:
...
+   Atanh remapper transform function examples.
+
+Example configuration:
+
+.. code:: yaml
+
+   data:
+      processors:
+         normalizer:
+           _target_: anemoi.models.preprocessing.multi_dataset_normalizer.TopNormalizer
```
sahahner: I don't think we have a `multi_dataset_normalizer` in anemoi-models at the moment?

models/src/anemoi/models/preprocessing/mappings.py
```diff
@@ -15,55 +17,224 @@ def noop(x):
...
+            ).all(), (
+                "input x must me greater or equal to 0, or use with clip_negative=True to clip negative values to 0"
+            )
+
+    # Apply transformation
     if lambd == 0:
-        return torch.log(x)
-    return (torch.pow(x, lambd) - 1) / lambd
+        return torch.log_(x)
+    return x.pow_(lambd).sub_(1.0).div_(lambd)
```
OpheliaMiralles, line 115: What is the difference between torch log_ and log?

elkir: All the underscore-terminated functions (`log_`,...) operate in-place on the tensor that was given, keeping fixed memory.

models/docs/modules/preprocessing.rst
sahahner: I have unresolved this conversation. Is the `_convert_: all` option necessary? We are not using this option for the other preprocessors, so I wold like to understand why/if it is necessary here.

As far as I understand not using `_convert_: all` is the recommended default.

models/docs/modules/preprocessing.rst
sahahner: There is no multi_dataset_normalizer.TopNormalizer in anemoi

models/docs/modules/preprocessing.rst
```diff
@@ -25,6 +25,120 @@ the following classes:
...
+           _convert_: all
+           config:
+             default: "mean-std"
+             max: ["tp","tcc"]
+         remapper1:
+           _target_: anemoi.models.preprocessing.remapper.Remapper
+           _convert_: all
```
sahahner: Is the _convert_: all option necessary for the new remapper configuration? 

We should avoid using _convert_: all because it strips away Hydra/OmegaConf features (like interpolation, type safety, and structured configs) and turns everything into plain Python objects. This makes configs less expressive and can hide bugs that Hydra would otherwise catch.

elkir: Sorry for resolving this comment prematurely. I thought I erased the `__convert__` option when resolving it, but it got lost in the commits somewhere.

The `__convert__` was inherited from a config we were working on, and after checking the documentation, it is indeed not necessary, especially in this form, where we're not storing the normalizer and other configs elsewhere but directly in the list of preprocessors.



Just out of curiosity: was it ever needed for split configs (where we're storing things in separate entries and referring to them through omegaconf), such as the one below, before? Or is this truly coming from some very outdated anemoi code?
```
data:
  normalizer:
    default: "mean-std"
    max: [tp]
  remapper:
    default: "none"
    boxcox-rescaled: [tp]
    method_kwargs: 
      boxcox-rescaled:
        lambd: 0.05
        symmetrical_bound: 1.0
  processors:
    normalizer:
      _target_: anemoi.models.preprocessing.multi_dataset_normalizer.TopNormalizer
      _convert_: all
      config: ${data.normalizer}
    remapper:
      _target_: anemoi.models.preprocessing.ds_remapper.TopRemapper
      _convert_: all
      config: ${data.remapper}
```

models/docs/modules/preprocessing.rst
elkir: I agree, see other comment. Hopefully, the current version has corrections to these now?

models/docs/modules/preprocessing.rst
elkir: roselved in another comment

models/docs/modules/preprocessing.rst
sahahner: Thank you for updating that. I define my processors configurations also in separate entries in the config files and do not use the convert entry.

models/src/anemoi/models/preprocessing/mappings.py
```diff
@@ -15,55 +17,224 @@ def noop(x):
...
+            "Power transform input x must satisfy x >= 0, or use with clip_negative=True to clip negative values to 0."
+        )
+    if tangent_linear_above_one:
+        lin_branch = x.clone().mul_(lambd).add_(1.0 - lambd)
+        pow_branch = x.pow_(lambd)
+        return torch.where(x > 1.0, lin_branch, pow_branch)
+    return x.pow_(lambd)
+
+
+def inverse_power_transform(x, lambd=0.33, tangent_linear_above_one=False):
```
Lun4m, line 158: This is a mirror of the comment I left on another PR since they have the same issue.

You are missing the `clip_negative` argument here.

It seems like in the way this is implemented you need the same arguments for both forward and backward transforms, but I think it'd make more sense to have a class for each mapper with forward and backward methods so you can "share" the arguments more explicitly?

This can always be left for a future refactor though.

sahahner: Thank you for raising this. I think we can indeed focus on that in another PR, maybe even the one that you mention :)

Lun4m: Unfortunately that PR is not targeting `main`, so I guess I'll open an issue when this one lands, if my suggestion makes sense

sahahner: okay, feel free to do that. that sounds good :)

Lun4m: The `clip_negative` argument is still missing though, so maybe let's wait for that before resolving? The training would outright crash without it, if you have it set in your config

Lun4m: @sahahner I'm sorry if my first comment was not super clear on the fact there was a bug, but it got merged in :melting_face:

sahahner: Hi, I saw that comment too late. Sorry about that. Could you please open an issue and tag @elkir as well? This should be a small change.

elkir: This has been resolved here https://github.com/ecmwf/anemoi-core/pull/1031

Apparently, some OpenClaw agent found this little bug

PR #975, opened 2026-03-16
fix(graphs,aicon): icon mesh vertex mask

fprill:
## Description

Bug-fix: Fixes cell-to-vertex indexing during coarsening of ICON meshes.

In the AICON encoder, edges are constructed between cell circumcenters (representing data points) and the three vertices of the enclosing hidden mesh triangle. However, a critical bug was introduced during recent refactorings of `icon_mesh.py`: Invalid edges were generated and edge shard partitions became empty due to incorrect indexing.

none (bug-fix)


JPXKQX (approved):
It looks good to me! Thanks Florian!

anaprietonem:
Hello @fprill @MeraX is this good to be merged? or are you looking to add/check anything else?

PR #979, opened 2026-03-16
[DRAFT]: suggestions for triton kernels

stadlmax:
## Description
- draft to discuss a few suggestions

- accelerate conversions between COO and CSR/CSC structures
- avoid calls to sort operations deterministic behavior is not desired via special Triton kernel

TBD

##  Additional notes ##
- makes custom kernels and custom operations compatible with torch.compile and AOT compilation workflows (`torch.library.custom_op` instead of `torch.autograd.Function`)
- minor suggestions here and there

### DIFF
--- a/models/src/anemoi/models/triton/gt.py
+++ b/models/src/anemoi/models/triton/gt.py
@@ -1,4 +1,5 @@
 # (C) Copyright 2025 Anemoi contributors.
+# (C) Copyright 2026 2026 NVIDIA CORPORATION & AFFILIATES.
 #
 # This software is licensed under the terms of the Apache Licence Version 2.0
 # which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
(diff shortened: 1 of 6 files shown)


models/src/anemoi/models/triton/gt.py
```diff
@@ -8,446 +8,484 @@
...
-    This is required to support non-square numbers of heads and/or channels.
-    Returns a mask for H, H*C and an offset for accessing into a 2D H*C matrix, ignoring padded values
-
-    masking apparently has a price, so if H and C are already powers of 2, nothing is returned
-    If H is already a power of 2 but C is not, a simpler H*C mask is returned
-
-    This function assumes a matrix layout of shape [H,C] for mask_H_C and H_C_off
-    """
-
-    # default mask (assume no padded values)
```
stadlmax, line 36: The default mask would be `None`, `True` AFAIK will be interpreted as a tile of size [1] and expanded accordingly. load/store operations still are faster as not every thread predicates these operations on its own. The difference should be marginal though. If inline-utility is preferred, I can also undo the change. For actual `None` masks, this helper doesn't work as Triton can't infer the type of a `None` value for this in-lined function.

models/src/anemoi/models/triton/gt.py
```diff
@@ -8,446 +8,484 @@
...
-        # tl.arange(H_pad, C_pad) doesnt work, because the arrays its offseting into aren't padded
-        # Therefore we make our own range, using unpadded major dimension (C)
-        H_C_off = tl.reshape(H_pad_off * C + C_pad_off, (H_pad * C_pad,))
-
-    return H_mask, H_C_mask, H_C_off
-
-
-@triton.jit
-def _gt_fwd(
+def _graphtransformer_fwd_kernel(
```
stadlmax, line 26: Names for Triton kernels are propagated. I might suggest something more informative than just `_gt_fwd` to keep profiles more informative.

models/src/anemoi/models/triton/gt.py
```diff
@@ -8,446 +8,484 @@
...
+        C_mask = (C_off < C)
+        H_C_mask = C_mask
+        H_PAD_VAL = None
+        H_C_PAD_VAL = 0.0
+    else:
+        H_mask = (H_off < H)
+        C_mask = (C_off < C)
+        H_C_mask = H_mask & C_mask
+        H_PAD_VAL = 0.0
+        H_C_PAD_VAL = 0.0
```
stadlmax, line 73: Explicit padding values likely don't hurt. IIRC, there is no guarantee across Triton versions that padding values are 0 and especially for reductions this can introduce sources of incorrectness.

models/src/anemoi/models/triton/gt.py
```diff
@@ -8,446 +8,484 @@
...
-    l_i = tl.zeros((H_pad,), dtype=tl.float32)  # sum of attention weights
-    m_i = tl.full((H_pad,), value=-float("inf"), dtype=tl.float32)  # running max for stability
-
-    # helpers to avoid repeated computations/indexing:
-    edge_ptr = E_ptr + neigh_start * H * C + H_C_off  # pointer to first edge_attr
-    e_idx = neigh_start  # first edge index
-    qk_scale: tl.constexpr = 1.0 / tl.sqrt(float(C))
+    dst_idx = tl.program_id(0).to(tl.int64)
+    head_idx = tl.program_id(1)
+    # option: batch_idx = tl.program_id(2)
```
stadlmax, line 44: Given custom kernels, it could be worth thinking about support for batching (i.e. multiple batches of data tensors for the same graph structure).

models/src/anemoi/models/triton/gt.py
```diff
@@ -8,446 +8,484 @@
...
+        qk_scale: tl.constexpr = (1.0 / tl.sqrt(float(C))) * 1.44269504089
+    else:
+        qk_scale: tl.constexpr = 1.0 / tl.sqrt(float(C))
 
     # for _ in tl.range(num_edges, warp_specialize=True):
-    for _ in range(num_edges):
-        e = tl.load(edge_ptr, mask=H_C_mask).to(tl.float32).reshape((H_pad, C_pad))
+    for eidx in range(neigh_start, neigh_end):
+        src_idx = tl.load(CSC_indices_ptr + eidx)
+        coo_idx = tl.load(CSC_map_to_coo_ptr + eidx)
```
stadlmax, line 97: indirect look-up in the main-loop to avoid explicit permutations of `edge_attr` for COO structures without introducing too much overhead for existing CSC structures

models/src/anemoi/models/triton/gt.py
```diff
@@ -8,446 +8,484 @@
...
         v_e = v + e
 
-        qk = tl.sum(q * k_e, axis=-1) * qk_scale  # Shape: [H]
+        qk = tl.sum(q * k_e, axis=-1, keep_dims=True) * qk_scale  # Shape: [H_PER_BLK, 1]
 
         m_ij = tl.maximum(m_i, qk)  # new running max
-        alpha_ij = tl.exp(qk - m_ij)  # attention weight for current edge
-        correction = tl.exp(m_i - m_ij)  # correction factor for previous accumulations
+        qk = qk - m_ij
+        if USE_FAST_EXP2:
```
stadlmax, line 115: EXP2 trick following upstream iterations of FA kernels. Especially on hardware with an increasing gap of memory bandwidth and performance of special function units, this could bring minor gains.

PR #981, opened 2026-03-17
test: add metadata check in training cycle tests

VeraChristina:
## Description
Adds a check to the training cycle tests (covering the main model classes) to check the expected structure of the metadata dictionary produced by the trainer.

Note that it is not a pydantic schema, and it is still evolving. The purpose is to track changes in the interface with anemoi-inference; see https://github.com/ecmwf/anemoi-core/issues/993. If we accidentally break the schema, this test will fail. If we want to make changes to the schema, this is a good point in time to check whether the change is compatible with anemoi-inference.

A first step to making the interface with anemoi-inference more explicit, without introducing the overhead of a pydantic schema.

### DIFF
--- a/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
+++ b/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
@@ -354,7 +354,7 @@ class CosineLatWeightedAttribute(BaseLatWeightedAttribute):
 
 
 class IsolatitudeAreaWeights(BaseLatWeightedAttribute):
-    """Latitude-weighted area weights for rectilinear grids.
+    r"""Latitude-weighted area weights for rectilinear grids.
 
     Attributes
     ----------
--- a/models/src/anemoi/models/layers/attention.py
+++ b/models/src/anemoi/models/layers/attention.py
@@ -57,6 +57,7 @@ class MultiHeadSelfAttention(nn.Module):
         num_heads: int,
         embed_dim: int,
         layer_kernels: DotDict,
+        attn_channels: Optional[int] = None,
         qkv_bias: bool = False,
         qk_norm: bool = False,
         is_causal: bool = False,
@@ -81,7 +82,10 @@ class MultiHeadSelfAttention(nn.Module):
         num_heads : int
             number of heads
         embed_dim : int
-            embedding dimension
+            Input and output embedding dimension
+        attn_channels : int, optional
+            Internal attention width used for q/k/v projections. If None,
+            defaults to embed_dim.
         qkv_bias : bool, optional
             bias for querys, keys and values, by default False
         qk_norm : bool, optional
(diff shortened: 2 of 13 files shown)


training/tests/integration/test_training_cycle.py
```diff
@@ -27,6 +27,72 @@
 
 LOGGER = logging.getLogger(__name__)
 
+# This schema defines the expected structure of the metadata dictionary produced by the trainer.
+# Note that it is not a pydantic schema, and it is still evolving.
+# The purpose is to track changes in the interface with anemoi-inference.
+# Before making changes to the schema, check whether the change is compatible with anemoi-inference.
+METADATA_SCHEMA = {
```
anaprietonem: shall this live in it's own file to make it easy to discover/track in the future?

training/tests/integration/test_training_cycle.py
```diff
@@ -27,6 +27,72 @@
...
+                "input_timesteps": None,
+                "ensemble": None,
+                "grid": None,
+            },
+        },
+    },
+}
+
+
+def assert_keys_exist(data: dict, schema: dict, path: str = "root") -> None:
```
anaprietonem, line 32: I appreciate and agree we might not want to include a Pydantic schema at the moment but would there a way to also check the typing of the entries in the schema to assert not just that key exist but also it has the right object on it?

training/tests/integration/test_training_cycle.py
```diff
@@ -27,6 +27,72 @@
...
+    "run_id": None,
+    "dataset": None,
+    "data_indices": None,
+    "provenance_training": None,
+    "timestamp": None,
+    "metadata_inference": {
+        "seed": None,
+        "run_id": None,
+        "dataset_names": list,  # list of datasets
+        "task": None,
```
anaprietonem: for task we could check the value returned is a valid one (ie a task supported in anemoi and not a 'random' one?

training/tests/integration/test_training_cycle.py
anaprietonem: To check my understanding -  this schema should always be matching what we have under https://github.com/ecmwf/anemoi-inference/blob/3be5d68baf661667a339d3292994910353a7cecc/src/anemoi/inference/metadata.py#L80 ?

VeraChristina: I have moved the schema to a separate file

VeraChristina: And I have clarified that it is a partial schema. We will need further changes in core and in inference to define a complete schema that we can validate here (i.e. check if it matches what inference expects in your reference). The idea of this PR is to safeguard the progress we are making towards a complete schema.

training/tests/integration/test_training_cycle.py
VeraChristina: That should be possible but I'd prefer to do it in a separate PR

training/tests/integration/test_training_cycle.py
VeraChristina: see above

anaprietonem (approved):
Thanks Vera for clarifying my questions. I think we should link in this PR the link the issue that captures the next step. Otherwise LGTM to be merged.

Issue #976, opened 2026-03-16
Expansion of config.graph.edges in mlflow not working


MeraX:
### What happened?

I noticed that all parameters configured under `config.graph.edges` are shown in mlflow as a list of dicts and lists instead of showing each parameter individually. This make the comparison of experiments with different graph configuration difficult. Note that configuration under `config.graph.nodes` is expanded fully as expected.

### What are the steps to reproduce the bug?

Train any (?) model and log with mlflow.

### Version

anemoi-models 0.13.0 , anemoi-training 0.10.0

### Platform (OS and architecture)

any

### Relevant log output

e.g. see https://mlflow.ecmwf.int/#/experiments/591/runs/3c7c2b78fa1041998cb6eb503525bde2 Search for parameter `config.graph.edges`. This is a randomly picked training log with probably other package version, but it shows the same issues as with the recent most version of training.

PR #982, opened 2026-03-17
fix: param expansion for mlflow

MeraX:
## Description
Parameters configured under `config.graph.edges` are shown in mlflow as a list of dicts and lists instead of showing each parameter individually. This make the comparison of experiments with different graph configuration difficult. Note that configuration under `config.graph.nodes` is expanded fully as expected.

With this PR, the expansion logic is expanded to Omegaconf and the mlflow-based logging no longer overrides the recursive extension by popping extended parameters before adding the remainer with `expanded_params.update(params)`.

https://github.com/ecmwf/anemoi-core/issues/976

##  Additional notes ##
Note that the number of parameters in mlflow slightly increase from 1202 to 1413 in my test training. This is because all lists of variables will be expanded now. See screenshot below (left column before this PR, right column after this PR)
Bildschirmfoto 2026-03-17 um 17 07 01
IMHO, this is the expected behaviour from https://github.com/ecmwf/anemoi-core/blob/7603aa31e213de525eedbe1034460ae97a659031/training/src/anemoi/training/diagnostics/mlflow/utils.py#L106-L112

### DIFF
--- a/.github/workflows/inactivity-bot.yml
+++ b/.github/workflows/inactivity-bot.yml
@@ -1,0 +1,40 @@
+name: Mark inactive issues and PRs
+
+on:
+  schedule:
+    - cron: "0 23 * * *"  # every day at 23pm on default(main) branch
+  workflow_dispatch: # Allows manual trigger
+
+permissions:
+  actions: write
+  contents: write # only for delete-branch option
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # Issue settings
+          days-before-issue-stale: 90 # (~3 months)
+          days-before-issue-close: 10
+          stale-issue-label: inactive
+          stale-issue-message: >
+            This issue has been marked as inactive due to inactivity.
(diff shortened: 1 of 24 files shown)


training/src/anemoi/training/diagnostics/mlflow/utils.py
```diff
@@ -12,6 +12,9 @@
 from collections import deque
 from typing import Any
 
+from omegaconf.dictconfig import DictConfig
```
JPXKQX: I think that both `DictConfig` and `ListConfig` could be imported directly from omegaconf

training/src/anemoi/training/diagnostics/mlflow/utils.py
```diff
@@ -42,79 +44,67 @@ def __repr__(self):
 
 
 def expand_iterables(
-    params: dict[str, Any],
+    params: Any,
```
anaprietonem, line 47: the type of params and return should be kept as before

MeraX: The function is not limited to `dict` (or `DictConfig`). It can process any type, but handles `dict[str, Any] | DictConfig[str, Any]` and `list | tuple | ListConfig` specially. 

However, the type hint and doc string were out of sync. I just fixed that.

PR #985, opened 2026-03-18
feat(models, training): Optimise spectral transform

samhatfield:
## Description
This PR implements two optimisations to spectral transforms:
1. reduces the number of transforms required to compute the spectral L2 loss from two to one
2. batches FFTs of the same length together

### Optimisation 1
The L2 spectral loss is currently computed like the following:
```python
pred_spectral = self._to_spectral_flat(pred)
target_spectral = self._to_spectral_flat(target)
diff = torch.abs(pred_spectral - target_spectral) ** 2
```
A spectral transform is a linear operation, so it's not necessary to transform each field individually. Instead you can just transform the residual itself, which reduces the number of calls to `_to_spectral_flat` from two to one:
```python
diff = torch.abs(self._to_spectral_flat(pred - target)) ** 2
```
This improves overall speed of training when spectral losses are used at N320 by about 5%.

### Optimisation 2
For classic reduced grids like N320, many latitudes have the same length. It makes sense to group these together into a batch and execute a batched FFT.

This provides a further ~20% of performance improvement for N320.

**Note that optimisation 2 does nothing for octahedral grids which have different length FFTs at each latitude.** There, [graphs are still the best solution](https://github.com/ecmwf/anemoi-core/pull/974).

https://github.com/ecmwf/anemoi-core/issues/599 https://github.com/ecmwf/anemoi-core/pull/974

##  Additional notes ##


OpheliaMiralles (dismissed):
Looks ok in theory. I'm not personally using the SpectralL2Loss but it shouldn't make any difference

training/src/anemoi/training/losses/spectral.py
```diff
@@ -105,6 +105,10 @@ def __init__(
         # Enforce loss to be calculated on full grids.
         self.supports_sharding = False
 
+        # Some transforms are proven to be linear, so we can compute the pred - target difference before taking the
+        # transform for those. This reduces the number of required transforms from two to one
+        self.diff_before_transform = False
```
sahahner, line 111: As we will remove the SpectralL2Loss, I think it is not necessary to introduce this change here. The flag is only used in the SpectralL2Loss.

samhatfield:
PR defunct. We are pursuing a CUDA-backend approach to optimise the spectral transforms.

PR #987, opened 2026-03-19
feat(models): add attn_dim to attenton computaton

ssmmnn11:
make the channel dimension in which the attention computation happens configurable.

attn_channels is by default equal to the embedding dimension but can be set smaller. This reduces memory usage in attention computation without decreasing the number of channels or the expressivity of the MLPs in the mappers and processor.

### DIFF
--- a/.github/workflows/inactivity-bot.yml
+++ b/.github/workflows/inactivity-bot.yml
@@ -5,13 +5,19 @@ on:
     - cron: "0 23 * * *"  # every day at 23pm on default(main) branch
   workflow_dispatch: # Allows manual trigger
 
+permissions:
+  actions: write
+  contents: write # only for delete-branch option
+  issues: write
+  pull-requests: write
+
 jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/stale@v9
         with:
-          repo-token: $
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
           # Issue settings
           days-before-issue-stale: 90 # (~3 months)


models/src/anemoi/models/layers/attention.py
```diff
@@ -102,16 +106,18 @@ def __init__(
...
+        if self.attn_dim <= 0:
+            raise ValueError(f"attn_dim must be > 0, got {self.attn_dim}")
+        if self.attn_dim % num_heads != 0:
+            raise ValueError(f"attn_dim ({self.attn_dim}) must be divisible by number of heads ({num_heads})")
 
         self.attention_implementation = attention_implementation
         self.use_alibi_slopes = use_alibi_slopes
 
         self.num_heads = num_heads
         self.embed_dim = embed_dim
```
anaprietonem: this could be removed right?

models/src/anemoi/models/layers/attention.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         num_heads: int,
         embed_dim: int,
```
anaprietonem, line 58: I assume we are keeping the `embed_dim` for backward compatibility but this could be updated or deprecated in the future? (we  could add a comment if so)

ssmmnn11: embed_dim is still the number of embedding channels and is used where we define the linear layers for the qkv, attn_dim - if set - defines a down projection for the attention computation.

models/src/anemoi/models/layers/attention.py
ssmmnn11: well spotted, I think we never used it. I removed it now

Issue #967, opened 2026-03-11
Loss scalers for spectral losses


sahahner:
### Is your feature request related to a problem? Please describe.

Spectral losses have been implemented in PR #788.
In a next step, we want to introduce loss scalers in the spectral dimension to be able to scale the different wave numbers.

### Describe alternatives you've considered

At the moment, we disable the loss scalers over the node dimension for spectral losses; therefore no scaling is applied in the spectral dimension, and the losses are added over the spectral dimension.

### Organisation

ECMWF

PR #992, opened 2026-03-19
feat(training): loss scalers for spectral losses and power spectrum loss

### LINKED ISSUE #725: Current loss reduction logic makes scaling over GRID dimension mandatory
In some situations, such as in the case of spectral losses (#678), the loss scaling over the GRID dimension has to be avoided as it becomes meaningless in the spectral space. However, if a user specifies a list of scalers that does not include a scaling over the GRID dimension, they will reach an error due to this:

https://github.com/ecmwf/anemoi-core/blob/dba42684794422e46b16c76f4c17a6db4859fd2c/training/src/anemoi/training/losses/base.py#L109-L114

it turns out this is related to how the reduction logic of the loss is implemented. For the GRID dimension, the values of the loss are first reduced via a sum, not a mean, and it is then assumed that the (weighted) average is implicitly computed later via the scaler. This is not ideal because in practice it makes the scaler over the GRID dimension mandatory, even when it's not needed. The current workaround is to specify a uniform scaling as suggested in the error message.

Ideally, we should use averaging directly in the reduction instead of delegating to the scaler.

sahahner:
This pull request introduces several important updates to the spectral loss and scaling framework. The main changes are the addition of new spectral dimension scalers, the depreciation of the `SpectralL2Loss`, introduction of a new `PowerSpectrumLoss`, and enhancements to ensure compatibility of scalers with spectral losses. Associated tests and imports are updated accordingly.

**Spectral Losses and Scalers:**
- Added new spectral dimension scalers: `SpectralDimensionScaler` and `LinearSpectralDimensionScaler`, providing uniform and linearly weighted scaling over spectral dimensions. 
- Introduced tests for the new spectral dimension scalers to verify correct behavior and formulas.

**Loss Function Updates:**
- Depercate `SpectralL2Loss`, as it is an expensive version of the MSE
- Introduce the `PowerSpectrumLoss`, which computes L2 loss on the power spectrum (energy per total wavenumber). All relevant imports, test cases, and references have been updated to use `PowerSpectrumLoss`. 

**Compatibility and Validation:**
- Added `_assert_spectral_scalers_compatible` to ensure that scalers applied on the grid dimension in spectral losses are correctly sized for the spectral domain, preventing mismatches between scaler and data shapes. 

## Description
Issue: https://github.com/ecmwf/anemoi-core/issues/967

Introduce scalers for Spectral L2 losses.

[](https://github.com/ecmwf/anemoi-core/issues/967)

----
📚 Documentation preview 📚: https://anemoi-training--992.org.readthedocs.build/en/992/

----
📚 Documentation preview 📚: https://anemoi-graphs--992.org.readthedocs.build/en/992/

----
📚 Documentation preview 📚: https://anemoi-models--992.org.readthedocs.build/en/992/


training/src/anemoi/training/losses/spectral.py
```diff
@@ -158,18 +209,20 @@ def forward(
...
-            pred_spectral = self._to_spectral_flat(pred)
-            target_spectral = self._to_spectral_flat(target)
-            diff = torch.abs(pred_spectral - target_spectral) ** 2
+        sc_pred = self.transform.forward(pred)
+        sc_target = self.transform.forward(target)
+        pred_amp = torch.sum(
+            sc_pred.real**2 + sc_pred.imag**2,
+            dim=-2,
+        )  # sum over order (M) dim to get power per wavenumber
+        target_amp = torch.sum(sc_target.real**2 + sc_target.imag**2, dim=-2)
```
frazane: This way of computing the power per wavenumber is valid for the SHT but not for the FFT. Would it make sense to move such operations on the transform classes themselves? This way, from inside the losses we can call the same methods but the implementation will be different (and correct) based on the transform.

sahahner: yes, good point. we can move this into the transform class and use an assert to check whether this is defined

sahahner:
Putting this into draft mode as main has moved on and people use this branch for experiments. Scalers and Power Spectrum Loss moved to this PR: https://github.com/ecmwf/anemoi-core/pull/1253

PR #1000, opened 2026-03-22
fix: allow control of multiprocessing context in the DataLoader

b8raoult:
## Description

We now use the [obstore](https://developmentseed.org/obstore/latest/) package to access S3, as it is also used by Zarr3. The package is mostly written in Rust, and does asynchronous I/O using the [Rust Tokio library](https://github.com/tokio-rs/tokio). Because of that, if one wants to train directly from S3, the Rust library will be running multiple threads, which are started at the beginning of the training, as the graph code enquires the latitudes/longitudes, amongst other calls. When the DataLoader forks its workers, the Rust threads are not inherited in the child processes and the training either hangs or segfault. This can be fixed by setting Python's [multiprocessing](https://docs.python.org/3/library/multiprocessing.html) mode to "spawn".

This change allows to user to control the multiprocessing mode from the hydra config.

Please note that more and more Python packages are using Rust.

PS: We can consider using "spawn" as the default.

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1000.org.readthedocs.build/en/1000/

----
📚 Documentation preview 📚: https://anemoi-graphs--1000.org.readthedocs.build/en/1000/

----
📚 Documentation preview 📚: https://anemoi-models--1000.org.readthedocs.build/en/1000/


training/src/anemoi/training/schemas/dataloader.py
```diff
@@ -136,3 +136,5 @@ class DataLoaderSchema(PydanticBaseModel):
     # TODO(Helen): Check that this equal or greater than the number of rollouts expected by callbacks ???
     read_group_size: PositiveInt = Field(example=None)
     "Number of GPUs per reader group. Defaults to number of GPUs (see BaseSchema validators)."
+    multiprocessing_context: str | None = Field(example=None)
```
HCookie:
```suggestion
    multiprocessing_context: str | None = Field(default=None, examples=[None, 'spawn', 'fork', 'forkserver'])
```
I would suggest adding a default

cathalobrien: good idea

PR #1027, opened 2026-04-06
fix(models): revisions triton gt kernel to improve stability

ssmmnn11:
Improve numerical stability of the Triton GT kernel. However, this results in a ~ 10% memory increase and a similar % in terms of performance degradation (pure kernel - less in a real training run). [Tests](https://github.com/ecmwf/anemoi-core/actions/runs/24045421978/job/70127518087).

blue: current triton kernel bf16
red: current triton kernel bf16, second try
green: resume blue curve with pyg compiled bf16
orange: pyg compiled bf16
violet: pyg compiled bf16, cast to float32
brown: revised triton kernel bf16

updated

test:

- make the attention logits very large -> softmax becomes very sharp
- every neighbor sends the same message -> output should no longer depend on the attention weights
- the gradients dq and dk should be zero or very close to zero
- check whether the backward pass still creates nonzero gradients

|dq| / |dk|

  - float32
      - GT before change:   0.000078 / 0.000703
      - GT now:                     0.000091 / 0.000454
  - bfloat16
      - GT before change:   1.468 / 18.760
      - GT now:                     0.321 /  3.615


cathalobrien:
Very nice work, 
to summarise this PR: it leaves the forward pass mostly unchanged, but saves the fwd output in fp32 and some softmax stats (`inv_l`). These values are reused in the backward pass which changes how `dS` is calculated. 

Is it possible to write a pytest which fails for main but passes for this branch?

cathalobrien:
I added a pytest, which passes on this branch but fails on main
```
_________ test_sharp_softmax_uniform_message_grads_zero[1000.0-dtype1] _________
models/src/anemoi/models/triton/test_gt_sharp_softmax.py:158: in test_sharp_softmax_uniform_message_grads_zero
    assert dk_norm / (dv_norm + 1e-12) < rel_tol, (
E   AssertionError: dk should be ~zero when all messages are identical, but |dk|/|dv| = 1.467430e-03 (scale=1000.0, dtype=torch.bfloat16). This indicates catastrophic cancellation in the backward pass.
E   assert (0.010635738261044025 / (7.247865676879883 + 1e-12)) < 0.001
________ test_sharp_softmax_uniform_message_grads_zero[10000.0-dtype1] _________
models/src/anemoi/models/triton/test_gt_sharp_softmax.py:158: in test_sharp_softmax_uniform_message_grads_zero
    assert dk_norm / (dv_norm + 1e-12) < rel_tol, (
E   AssertionError: dk should be ~zero when all messages are identical, but |dk|/|dv| = 1.467430e-03 (scale=10000.0, dtype=torch.bfloat16). This indicates catastrophic cancellation in the backward pass.
E   assert (0.010635738261044025 / (7.247865676879883 + 1e-12)) < 0.001
```

ssmmnn11:
see #1172

PR #1031, opened 2026-04-08
fix(models): add clip_negative parameter to inverse_power_transform

Bahtya:
## Problem

When `clip_negative: true` is set in the config for the power transform remapper, the inverse transform crashes with:

```
TypeError: inverse_power_transform() got an unexpected keyword argument 'clip_negative'
```

This happens because `power_transform` accepts `clip_negative` but `inverse_power_transform` does not. The Remapper passes the same `kwargs` dict to both forward and inverse functions.

## Fix

Add `clip_negative` parameter to `inverse_power_transform` for API compatibility. The inverse already clamps to non-negative values via `torch.clamp_(x, min=0.0)`, so no behavioral change is needed — just accept the parameter.

Fixes #1030


elkir (dismissed):
Yes, I remembered this issue for the Box-Cox, it slipped through for the power transform.

Looks good now

anaprietonem (approved):
Thanks for the contribution and for addressing reviewer's feedback.

PR #1045, opened 2026-04-15
fix!: loss NaN handling

pinnstorm:
## Description

Breaking out the fix for the nan-handling in the losses from PR #904

There is a long standing bug in torch.nanmean and torch.nansum meaning the backward pass will produce a NaN: [pytorch/pytorch#6718](https://github.com/pytorch/pytorch/issues/67180) (and other related issue)

The work around is to leave torch.mean/sum and just set the missing values to zero (further down in forward method) in both the predictions and targets so they have no impact on the loss.

Currently loss functions cannot handle nans, this is a fix.


training/src/anemoi/training/losses/base.py
```diff
@@ -72,8 +72,9 @@ def __init__(
 
         self.add_module("scaler", ScaleTensor())
 
-        self.avg_function = torch.nanmean if ignore_nans else torch.mean
-        self.sum_function = torch.nansum if ignore_nans else torch.sum
+        self.avg_function = torch.mean
+        self.sum_function = torch.sum
```
sahahner: Is the avg_function always torch.mean and the sum_function always torch.sum? Or are these fields possibly overwritten in a child class? Otherwise, we might not need this variable.

MeraX: My understanding is that these variables are not needed any more. I'm not aware of any child class overwriting these.

MeraX:
Hi @pinnstorm, thank you for breaking apart the other PR in independent pieces. In principle, I like your approach and solution. I could just deserve a little clean-up following Sarahs comment above. Just for averaging, I was wondering whether the averaging should be weighted by the inverse fraction of masked NaNs.

MeraX:
What is the status of this PR? I would very much welcome this fix in the main branch and next release (after a small revision, starting with addressing the bot comments.)

I would take the freedom to do a bit of this revision my self.

pinnstorm:
> What is the status of this PR? I would very much welcome this fix in the main branch and next release (after a small revision, starting with addressing the bot comments.)
> 
> I would take the freedom to do a bit of this revision my self.

Hi Marek 🙂! Sorry, yes I've been a bit slow on this after being pulled in a few other directions. I am very happy for you to contribute and do some of the revision if you have time for for this 🙏 , thanks very much!

MeraX:
I marked this as breaking, as we are removing the "public" API `avg_function` and `sum_function`.

MeraX:
Okay, I've merged the current master, added the new nan handling to spectral losses and loss wrappers, and extended the tests to make sure that there are no nans even in the gradients (I confirmed that the gradient test would have failed with the original code version).

MeraX (dismissed):
I would be glad to see this merged

training/src/anemoi/training/losses/base.py
```diff
@@ -214,6 +250,11 @@ def reduce(
             ),
         ).squeeze()
 
+        # Scale the result with the amount of masked nan-values.
+        # This adjusts the lost between steps with varying amounts of nans.
+        if nan_fraction > 0.0:
+            out /= 1 - nan_fraction
```
sahahner: I know we have talked about this already. 
I would not like to introduce a new scaling behaviour here by default. Since there seems to be interest in this functionality though, can we make it configurable in the config file, pelase?

training/src/anemoi/training/losses/spectral.py
```diff
@@ -147,13 +148,15 @@ def forward(
...
         group: ProcessGroup | None = None,
-        squash_mode: str = "avg",
+        squash_mode: Squash_mode = "avg",
         **kwargs,
     ) -> torch.Tensor:
         del kwargs  # unused
         is_sharded = grid_shard_slice is not None
         group = group if is_sharded else None
 
+        nan_fraction, pred, target = self.mask_nans(pred, target)
```
sahahner: For the spectral losses this functionality should not be enabled. The transform is not well defined with missing values. I would even go for an assert here, that makes the code break if ignore_nans is true for anything that has base class SpectralLoss (see suggested change above).

training/src/anemoi/training/losses/spectral.py
sahahner, line 98:
```suggestion
        assert not ignore_nans, (
            "Spectral losses cannot handle missing values; "
            "ignore_nans must be False"
        )
        BaseLoss.__init__(self, ignore_nans=ignore_nans)
        
```

sahahner:
Thank you @MeraX for putting extra work into this PR! I left a few comments already but overall this looks good. @pinnstorm could you possibly check whether Marek's refactor still works for your use case, please?

training/src/anemoi/training/losses/spectral.py
MeraX: I agree.

training/src/anemoi/training/losses/base.py
MeraX: My concern is that if we have varying amounts of nan values in each step, the loss would become jumpy, even if the model quality would be constant. But maybe, my approach is too simple when e.g. different weighting per variable are applied. So I would revert the scaling for now. I think this should not affect the gradients anyway.

training/tests/unit/losses/test_loss_function.py
```diff
@@ -589,6 +590,9 @@ def test_spectral_crps_with_target_without_ensemble_dim() -> None:
     assert out_total.numel() == 1, "squash=True should return scalar CRPS"
     assert torch.isfinite(out_total).all(), "Expected finite scalar loss with ignore_nans=True"
 
+    (grad,) = torch.autograd.grad(out_total, pred, retain_graph=True)
+    assert torch.isfinite(grad).all(), "Expected finite gradients"
+
 
 def test_spectral_crps_octahedral_irregular_grid_ignore_nans() -> None:
```
sahahner: To my understanding, ignore_nans should not be enabled for any spectral loss, including the spectral CRPS. I think these tests can be removed or adapted to check whether there is an assert that checks if ignore_nans is False

training/src/anemoi/training/losses/spectral.py
```diff
@@ -85,13 +86,15 @@ def __init__(
...
-            Whether to ignore NaNs in the loss computation.
+            Spectral losses cannot handle missing values;
+            ignore_nans must be False.
         scalers
             Kept for Hydra/config backwards compatibility. This module does not
             consume this argument directly (scaling is handled by BaseLoss).
         kwargs
             Additional arguments for the spectral transform.
         """
+        assert not ignore_nans, "Spectral losses cannot handle missing values; ignore_nans must be False"
```
MeraX, line 97: @OpheliaMiralles we are about to disable ignore_nan for spectral losses. However, you have once added a test for nans in spectral losses. Do you have a case with missing values when using spectral losses? How does that work with the current code? Or would you be fine with disabling nan-masking in spectral losses?

OpheliaMiralles: Might be a residual of code that I was using with InputOnlyImputer, where I applied a kind of mask_nans function before applying fft for precip. I am fine with disabling it because I am not using this imputer anymore, so please go ahead with this PR!

MeraX: Thank you. So we are just waiting for @pinnstorm confirmation before we can merge this PR.

MeraX:
I'm sorry for unintentionally deleting this branch. I just restored it.

training/src/anemoi/training/losses/base.py
sahahner: I see that you have removed the line entirely. When you have results on this, we can always include this later. :) Feel free to open an issue once you have results on this.

training/tests/unit/losses/test_loss_function.py
sahahner: Thank you for removing these leftover tests.

training/src/anemoi/training/losses/base.py
MeraX: yes. I think it is not so urgent. It's more important to take this PR over the finish line.

training/src/anemoi/training/losses/base.py
```diff
@@ -162,13 +189,16 @@ def reduce(
         Irrespective of `squash`, the output is reduced over the
         batch, ensemble and grid dimensions.
 
+        If self.ignore_nans is True, the loss is scaled by the
+         fraction of nan-values in the target.
```
pinnstorm: Do we need to remove this comment now?

training/src/anemoi/training/losses/base.py
```diff
@@ -28,6 +29,9 @@
 LOGGER = logging.getLogger(__name__)
 
 
+Squash_mode = Literal["avg", "sum"]
```
pinnstorm, line 32: Do we want this as a global or can we just have `Literal["avg", "sum"]` in the function definitions?

pinnstorm (commented):
Thanks a lot @MeraX for picking this up, looking good! I only had a couple of small comments/questions. I tested with the gridded-obs branch and this still works there, so all good. Thanks! 🙏

training/src/anemoi/training/losses/base.py
MeraX: good catch.

training/src/anemoi/training/losses/base.py
MeraX: Good question. I copied the style of the `CRPSBackend` in kcrps.py. I do prefer having the same type definition everywhere, i.e. using this single "global" object.

training/src/anemoi/training/losses/base.py
```diff
@@ -148,11 +146,40 @@ def scale(
...
+        target : torch.Tensor,
+            Target tensor
+
+        Returns
+        -------
+        torch.Tensor, torch.Tensor]
+            * 0-masked copy of ``pred`` if ``self.ignore_nans``, else ``pred``
+            * 0-masked copy of ``target`` if ``self.ignore_nans``, else ``target``
+        """
+        if self.ignore_nans:
```
sahahner, line 169: I went through the PR again and actually noticed that we are changing the logic of the ignore_nans flag:
**Before**: we changed the sum and avg functions, so NaNs in the prediction and target would be ignored.
**Now**: We check for NaNs only in the target! 
I propose to keep the logic from before, see the suggested change below.

training/src/anemoi/training/losses/base.py
```diff
@@ -148,11 +146,40 @@ def scale(
...
+            nan_mask = torch.isnan(target)
```
sahahner:
```suggestion
            nan_mask = torch.isnan(target+pred)
```

training/src/anemoi/training/losses/base.py
MeraX: @sahahner, I can confirm your finding. However, I assumed, that the model should not predict nans at all and especially not at positions that are finite in the target. So I though this change is an acceptable part of the fix. But I might have overseen conditions under which a model is right in outputting nans. Do you have an example?

sahahner: I see your argument from a training perspective. 
Nevertheless, we use these functions as validation metrics as well and then the positions of NaNs in prediction and target can be different.
I could also imagine, that we also run into the case where predicted fields/grid points are excluded from the loss by replacing them with NaNs (not defined, som dropout on variables).

We could make the separation clean and use two flags (ignore_nans_target, ignore_nans_predict)

MeraX: Thank you for your clarification. Then I would suggest staying with the original effect of nanmean/nansam is the best and I accepted your suggestion. I would leave thoughts about two flags or the introduction of any `assert torch.isfinite(X).all()` for a follow-up PR.

sahahner (approved):
When @pinnstorm is fine with closing the two open comments, I think this is ready to merge. Thank you very much for incorporating the changes @MeraX. Looking forward to using it.

training/src/anemoi/training/losses/spectral.py
pinnstorm: Good by me! See comment below

PR #1047, opened 2026-04-15
fix: update weight averaging config to not have duplicated commented out code. 

matschreiner:
Hi Jacob, here is what I meant in the ATS :) 

 - Add training/weight_averaging/ema.yaml and swa.yaml
  - Add - weight_averaging: null to defaults in all training configs                             
  - Remove inline weight_averaging: null and commented-out example configs                      
  - Can now be selected via CLI:
  training/weight_averaging=ema or                   
  training/weight_averaging=swa

### DIFF
--- a/training/src/anemoi/training/config/training/weight_averaging/ema.yaml
+++ b/training/src/anemoi/training/config/training/weight_averaging/ema.yaml
@@ -1,2 +1,3 @@
 _target_: pytorch_lightning.callbacks.EMAWeightAveraging
 decay: 0.999
+update_starting_at_step: ${training.lr.warmup}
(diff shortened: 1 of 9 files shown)


training/src/anemoi/training/config/training/weight_averaging/ema.yaml
```diff
@@ -0,0 +1,2 @@
+_target_: pytorch_lightning.callbacks.EMAWeightAveraging
+decay: 0.999
```
jakob-schloer, line 2:
```suggestion
decay: 0.999
update_starting_at_step: ${training.lr.warmup}
```

jakob-schloer: I believe this should be another default.

jakob-schloer (approved):
Good idea to move them to a seperate `weight_averaging` folder. This way we don't have that many duplicated lines.

PR #1048, opened 2026-04-15
feat(models): Learnable Residual Connections

### LINKED ISSUE #1043: Learnable residual connection

Autoregressive rollouts in Anemoi-based models suffer from error accumulation over long integration horizons. Each forecast step feeds back into the next, small systematic errors compound, eventually leads drift, blow-up, or distribution shift.

The standard residual formulation currently used, x(t+1) = x(t) + f_θ(x(t)). 

We could introduce a learnable scaling parameter α on the residual connection:
x(t+1) = α · x(t) + f_θ(x(t))
with α trainable and α < 1, errors in the state are contracted at each step rather than perfectly preserved, which should bound error growth during integration and improve long-rollout stability.

_No response_

### Additional context

_No response_

### Organisation

ECMWF

PortillaS-Predictia:
### Motivation:
The standard residual connection $x(t+1) = x(t) + f_{\alpha}(x(t))$ perfectly "preserves" errors in the state $x$ at each autoregressive update $x(t)\to x(t+1)$. During long rollouts, these errors accumulate and can destabilize forecasts. This PR introduces a minor change to lessen this effect.

PR #670 introduced _configurable residual connections_, expanding the (otherwise fixed) residual model. A natural enhancement is the introduction of _learnable residuals_ (as opposed to fixed operators). For instance, the `TruncatedConnection` may be enhanced as a _learnable low-pass filter_, in which the cutoff frequency for every variable is learned (taking advantage of the recently introduced SHTs #788). Likewise, the model does not need to be entirely `SkipConnection` or `NoResidual`, but a mixture of the two $x(t+1) = \theta\cdot x(t) + f_{\alpha}[x(t)]$ and the optimal behavior for each variable ($\theta$) is learned.

Originally, we drew inspiration from [Ornstein-Uhlenbeck SPs](https://en.wikipedia.org/wiki/Ornstein%E2%80%93Uhlenbeck_process):

$x(t+1) = (1 - \theta) \cdot x(t) + f_{\alpha}[x(t)]$.

With $\theta$ trainable and $0 < \theta < 1$, errors in the state are contracted at each step rather than preserved, which bounds error growth during integration.

An alternative explanation goes as follows - we build a _regularization potential_ $V[x(t)]$ that penalizes "undesirable" or "unphysical" states (that is, $V[x(t)]$ is high for "bad" $x$'s and low for "good" $x$'s), which we introduce in the model equation as $x(t+1) =x(t) + f_{\alpha}[x(t)] - \nabla V[x(t)]$. Now, notice the resemblance to the _gradient descent_ update rule $x\to x-\nabla V[x]$. That is, the field $- \nabla V[x(t)]$ "drifts" the state of the system towards the set of "desirable" states (low $V[x]$), away from the set of "undesirable" states (high $V[x]$).

For example, if we want to avoid "model explosion" in long rollouts, we would choose $V[x] = 0.5\theta x^2$ (that is, high for large $|x|$ and low for small $|x|$), so that $\nabla V[x] = \theta x$ and the model equation simplifies to  $x(t+1) =(1-\theta)x(t) + f_{\alpha}[x(t)]$, which would be our new residual connection (with the effect of "penalizing" states with large $|x|$). Furthermore, we may use $V[x] = 0.5\theta (x-\mu)^2$, so that the equilibrium point (that is, the state towards which the system "drifts") is some reference state $\mu$ (which, in turn, may depend on some external regressors, such as _insolation_, _co2_, etc.). Finally, we may introduce a _learnable low-pass filter_ that "penalizes" states with excess energy in the high frequencies (similar to the `TruncatedConnection` residual).

### Changes
New residual connection classes (`residual.py`):

- `ZeroConnection` — disables the residual entirely (`residual = 0`), forcing the model to forecast the full state.
- `ScalarOrnsteinConnection` — one learnable scalar α per prognostic variable. Simplest Ornstein variant with no spatial nor spectral structure.
- `SpectralOrnsteinConnection` — spatially-varying α(s) and bias μ(s) defined via low-order spherical harmonic coefficients. Optionally, supports a (learnable) spectral truncation of input fields (`truncate=True`) to remove high-frequency content from the residual connection. Supports both regular lat-lon and octahedral reduced grids.

Inverse SHT wrappers (`spectral_transforms.py`):

- Added `InverseRegularSHT` and `InverseOctahedralSHT` to complement the existing forward SHT classes, backed by the `InverseSphericalHarmonicTransform` in `spectral_helpers.py`.

### Example configuration
```
model:
  residual:
    _target_: anemoi.models.layers.residual.SpectralOrnsteinConnection
    lmax: 2
    grid: octahedral
    theta_init: 0.05
    truncate: true
    anti_aliasing: true
```

### Todos
- [x] Added tests for the residual connections `tests/layers/test_residual.py`-> All unit tests pass
- [x] Added brief documentation
- [x] Updated schemas
- [x] Debug training run with SpectralOrnsteinConnection + truncate: true on o96 using the deterministic and ensemble model
- [ ] Running test with sharding
- [ ] Performance tests
- [x] Running integration test (Integration test: https://github.com/ecmwf/anemoi-core/actions/runs/25367669967)
- [x] Opened an issue to for adapting the `SpectralOrnsteinConnection` for LAM #1080 (let's do a seperate PR on that because it is a bit harder for me to test)

#1043

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1048.org.readthedocs.build/en/1048/

----
📚 Documentation preview 📚: https://anemoi-graphs--1048.org.readthedocs.build/en/1048/

----
📚 Documentation preview 📚: https://anemoi-models--1048.org.readthedocs.build/en/1048/


samhatfield (dismissed):
I've only looked at spectral_transforms.py, but everything in there looks good to me 👍🏻

JPXKQX:
Hello, thank you very much for the PR, I think the PR is good to go. 

I only have one comment to add, and it is about the `ZeroConnection`. This was already proposed in a previous PR (when making the residual connection configurable), and the comments at that time were that this way is not optimal because it creates a full tensor in memory with no need. At that point, we agreed to keep it on the side until some scientific value had been demonstrated for a model or an implementation without the tensor creation is proposed.

[The integration tests](https://github.com/ecmwf/anemoi-core/actions/runs/25216790854) are passing.

PortillaS-Predictia:
Thanks Mario!

I only introduced the `ZeroConnection` for the sake of """completeness""", I never really use it. I'm fine with removing it.

models/src/anemoi/models/layers/residual.py
```diff
@@ -235,3 +242,304 @@ def _reshard(self, x, fn, shard_shapes=None, model_comm_group=None):
...
+        Variable names to exclude from spectral truncation (only used when
+        ``truncate=True``).
+    anti_aliasing : bool
+        If True (and ``truncate=True``), use anti-aliasing blending in the filter.
+    """
+
+    def __init__(
+        self,
+        lmax: int = 2,
+        grid: str = "legendre-gauss",
```
PortillaS-Predictia: I think this variable should now default to "regular", instead of "legendre-gauss". Thus, there would be two options - "regular" and "octahedral" (and "reduced", at some point, if we support it with an inverse transform in `spectral_transforms.py`).

models/src/anemoi/models/layers/residual.py
```diff
@@ -235,3 +242,304 @@ def _reshard(self, x, fn, shard_shapes=None, model_comm_group=None):
...
+        Maximum spherical harmonic degree for the theta/mu coefficients.
+    grid : str
+        Grid type: ``"legendre-gauss"`` for regular lat-lon, ``"octahedral"`` for
+        octahedral reduced grids.
+    theta_init : float
+        Initial value for theta.
+    theta_buff : float
+        Lower bound buffer for theta.
+    zmean_term : bool
+        Whether to include a zonal mean (mu) term.
```
PortillaS-Predictia: The "z" in "zmean_term" actually meant "zero", not "zonal". This variable's original purpose was different, then I changed it but never modified the name! We should drop the "z" in the name and the "zonal" in the description.

models/src/anemoi/models/layers/residual.py
```diff
@@ -235,3 +242,304 @@ def _reshard(self, x, fn, shard_shapes=None, model_comm_group=None):
...
+
+        self._internal_input_idx = list(data_indices.model.input.prognostic)
+        variables = data_indices.model.input.name_to_index
+        self._regressors_input_idx = [variables[f] for f in regressors]
+
+        self.nlat, self.nlon = _grid_shape_from_graph(graph, dataset_name)
+
+        sliced_stats = _slice_statistics_to_prognostic(statistics, data_indices)
+        theta = _ornstein_init_theta(theta_init, theta_buff, sliced_stats)
+        theta = np.sqrt(4 * np.pi) * np.log(theta / (1 - theta))
```
PortillaS-Predictia: This $\sqrt{4\pi}$ accounts for the SH's normalization, since the SHs were $\sqrt{4\pi}$-normalized in my code (thus, the inverse transform introduces a factor equal to $1/\sqrt{4\pi}$, so we need to "preemptively" account for it). But I believe this isn't the case anymore, @samhatfield?

models/src/anemoi/models/layers/residual.py
```diff
@@ -235,3 +242,304 @@ def _reshard(self, x, fn, shard_shapes=None, model_comm_group=None):
...
+    def _init_truncation(self, grid, lmax, theta_init, anti_aliasing, skip_truncate_variables, variables):
+        """Initialize spectral truncation parameters."""
+        if grid == "octahedral":
+            oct_lons = [20 + 4 * i for i in range(self.nlat // 2)]
+            oct_lons += list(reversed(oct_lons))
+            trunc = self.nlat - 1
+            self.x_fsht = SphericalHarmonicTransform(lons_per_lat=oct_lons, truncation=trunc)
+            self.x_isht = InverseSphericalHarmonicTransform(lons_per_lat=oct_lons, truncation=trunc)
+        else:
+            nlon = 2 * self.nlat
```
PortillaS-Predictia: Shouldn't we use `self.nlon` (which may or may not be equal to `2 * self.nlat`)? If we only want to support the `nlon = 2 * nlat` case, that's fine by me (I think the wrappers in `spectral_transforms.py` also asume `nlon = 2 * nlat`).

models/src/anemoi/models/layers/residual.py
jakob-schloer: That makes sense. Changed it to `use_mean`.

models/src/anemoi/models/layers/residual.py
jakob-schloer: Yes, you are right. We can use self.nlon here. Good catch.

models/src/anemoi/models/layers/residual.py
jakob-schloer: Very well spotted. You are right, different from your original inverse transform implementation the current code uses 1/(4π), so the compensating factor must be 4π. Can you confirm that @samhatfield?

models/src/anemoi/models/layers/residual.py
jakob-schloer: Yes, we also raise a `ValueError` now if another grid type is specified.

jakob-schloer:
> Hello, thank you very much for the PR, I think the PR is good to go.
> 
> I only have one comment to add, and it is about the `ZeroConnection`. This was already proposed in a previous PR (when making the residual connection configurable), and the comments at that time were that this way is not optimal because it creates a full tensor in memory with no need. At that point, we agreed to keep it on the side until some scientific value had been demonstrated for a model or an implementation without the tensor creation is proposed.
> 
> [The integration tests](https://github.com/ecmwf/anemoi-core/actions/runs/25216790854) are passing.

I've removed the `ZeroConnection` now. If we need it, it is easy to implement again.

models/src/anemoi/models/layers/residual.py
samhatfield: For the inverse transform, the Legendre polynomials used in the inverse transform _are_ normalised by $1/\sqrt{4\pi}$: https://github.com/ecmwf/anemoi-core/blob/main/models/src/anemoi/models/layers/spectral_helpers.py#L86-L87. Hence if you wanted to undo this, you should multiply by `np.sqrt(4 * np.pi)`.

Having said that, it's just a constant multiplier, so does it even matter?

jakob-schloer: I agree that it doesn't matter much but I believe we apply a normalisation of 1.0 in the direct transform, and 1/4pi in the inverse transform: ( https://github.com/ecmwf/anemoi-core/blob/main/models/src/anemoi/models/layers/spectral_helpers.py#L86-L87.)
```
norm_factor = np.sqrt(4 * np.pi)
norm_factor = 1.0 / norm_factor if inverse else norm_factor
vdm[0, 0, :] = norm_factor / np.sqrt(4 * np.pi)
```
So for the inverse we have: vdm[0,0] = (1/ sqrt(4π)) / sqrt(4π) = 1/(4π). 
I will leave the normalization as it is.

PortillaS-Predictia: Thanks for chiming in Sam!

I agree it's not very important. However, we are allowing the "user" to specify a `theta_init`, so we should respect that choice (or not allow it in the first place).

The "correction" constant should be $4\pi$ (see test below) (this isn't the SHs' "normalization", I explained myself poorly in my first comment, that's why Sam gave a different answer, see the test below to understand what I actually meant). I will modify it now.

Test,
```
from anemoi.models.layers.spectral_transforms import InverseOctahedralSHT
import torch
isht = InverseOctahedralSHT(90)
x = torch.zeros(45, 45, dtype=torch.complex64)
x[0, 0] = 1 + 0j  # 0th harmonic equal to 1
print(isht(x))  # 0.079577... which is 1/(4pi)
```

JPXKQX:
[Integration tests](https://github.com/ecmwf/anemoi-core/actions/runs/25367669967)

JPXKQX (approved):
Thank you for the PR, and congratulations to everyone involved!

PR #1050, opened 2026-04-17
fix: no learnable parameters for non-model nodes

### LINKED ISSUE #1049: Nodes for the `TruncatedConnection` don't pass assert.
### What happened?

When using `TruncatedConnection` , training fails with:

```
AssertionError("Number of trainable parameters for nodes group 'int' not provided in config.model.trainable_parameters.")
```

The user should not need to add the structural node to `config.model.trainable_parameters` at all — it is not a learnable node. Even if one adds it there as `int: 0`: the assert still stops it since the dictionary from the config is overwritten by `broadcast_config_keys`  in `models/models/base.py` which intentionally produces a new dict containing only the broadcasted model-node entries (`data` → dataset names, `hidden` → hidden node name) and drops all other nodes, like `int`. 

1. Set up a graph that includes an intermediate node group (e.g. `int`) used by `TruncatedConnection` as `truncation_nodes`.
2. Add the residual connection to the model config — with **no** entry for `int` in `trainable_parameters` (it is a structural node, not a learnable one):
   ```yaml
   model:
     residual:
       _target_: anemoi.models.layers.residual.TruncatedConnection
       data_nodes: data
       truncation_nodes: int
       edge_weight_attribute: gauss_weight
       row_normalize: true
     trainable_parameters:
       data: 0
       hidden: 0
       # int: 0 (adding this doesn't change the situation)
       data2hidden: 0
       hidden2data: 0
       hidden2hidden: 0
   ```
3. Run training. The job immediately crashes with the `AssertionError` above.

### Version

current main

irrelevant

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

rmi

dietervdb-meteo:
## Description
Fixes #1049 by setting number of trainable parameters for non-model nodes, like those to define a `TruncatedConnection`, to 0.


models/src/anemoi/models/layers/graph.py
```diff
@@ -72,10 +74,7 @@ def __init__(self, trainable_parameters: dict[str, int], graph_data: HeteroData)
         """Initialize NamedNodesAttributes."""
         super().__init__()
 
-        for nodes_name in graph_data.node_types:
-            assert (
-                nodes_name in trainable_parameters
-            ), f"Number of trainable parameters for nodes group '{nodes_name}' not provided in config.model.trainable_parameters."
+        trainable_parameters = defaultdict(int, trainable_parameters)
```
ssmmnn11, line 77: I think the only interaction is now that the hierarchical autoencoder will get 0 trainable parameters for some levels if not in config and this would have failed before. Though it should probably follow the design of the hierarchical encoder-processor-decoder. Is the hierarchical autoencoder still used? @icedoom888

ssmmnn11 (dismissed):
Thx! Would it be useful to also add a test?

dietervdb-meteo:
@ssmmnn11 , like putting an empty dictionary in https://github.com/ecmwf/anemoi-core/blob/2691d4b30ed7dc1ec5ac24109d48fa0132fa315f/models/tests/layers/test_graph.py#L131 ? or what would you like to test?

ssmmnn11 (approved):
LGTM! Many thanks!

PR #1051, opened 2026-04-19
docs(graphs): correct a diagram on how multiscale edges work

elkir:
## Description
Fix diagram which is in the docmuentation about how multiscale edges work.

The current diagram places the nodes to the centres at the faces, and then incorrectly assumes 3 connections at each refinement. This would be a dual to the actual graph which has the nodes at the triangle vertices, and therefore has 6 connections per refinement.

### Previous  (incorrect) image:
multi_scale_edges

### New corrected image:
option_1_recommended_y_at_2_1_small

##  Additional notes ##
image is in svg, I also have fairly small pngs which I can upload

----
📚 Documentation preview 📚: https://anemoi-training--1051.org.readthedocs.build/en/1051/

----
📚 Documentation preview 📚: https://anemoi-graphs--1051.org.readthedocs.build/en/1051/

----
📚 Documentation preview 📚: https://anemoi-models--1051.org.readthedocs.build/en/1051/


mchantry (approved):
Thanks for the contribution

PR #1052, opened 2026-04-20
feat: graph for skipped and multiscale suggestions

anaprietonem:
## Description
These are some suggestions for the implementation of the multi-scale and truncation matrices to be able to generated them on the fly, as an alternative way to the existing interface that allows to load the sparse matrices from disk.

**Note - this is still a WIP** - it will need to be rebased after the graph refactor. Currently the number of changes is quite large (github UI is also not doing the diff correctly). Main changes

- Explicit config definition to specify sparse graph generation in multiscale and truncation connections.
- Simplify truncation recipes
- Avoid graphFactory and projections folder - so the sparse matrices are generated where they are needed and reusing the nodes existing in the graph_data objects already used. Significantly reduce helpers and utility functions to resolve the configs.

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1052.org.readthedocs.build/en/1052/

----
📚 Documentation preview 📚: https://anemoi-graphs--1052.org.readthedocs.build/en/1052/

----
📚 Documentation preview 📚: https://anemoi-models--1052.org.readthedocs.build/en/1052/


training/src/anemoi/training/losses/multiscale.py
```diff
@@ -52,16 +64,62 @@ def __init__(
...
-            Filenames of the smoothing matrices (must preserve grid size)
-        autocast : bool
-            Whether to use automatic mixed precision for the projections
+            Deprecated.  Pass inside *multiscale_config* instead.
         """
-        super().__init__()
+        super().__init__(ignore_nans=ignore_nans)
 
-        self.smoothing_matrices = self._load_smoothing_matrices(loss_matrices_path, loss_matrices)
+        _has_matrices = loss_matrices and any(m is not None for m in loss_matrices)
```
ssmmnn11: here we will potentially still get {"loss_matrices_path" : "some/specified/path} in the config if the config has [None] for loss_matrices and then _load_smoothing_matrices() will expect a key and fail?

training/src/anemoi/training/schemas/base_schema.py
```diff
@@ -69,6 +69,51 @@ def expand_paths(config_system: Union[SystemSchema, DictConfig]) -> Union[System
     return config_system
 
 
+def _get(obj: Any, key: str) -> Any:
+    """Attribute-or-mapping access that works for both Pydantic models and plain dicts/DictConfigs."""
+    return getattr(obj, key, None) or (obj.get(key) if hasattr(obj, "get") else None)
+
+
+def _validate_multiscale_loss(training: Any) -> None:
```
ssmmnn11: will this work with configs like:

```
   training:
     training_loss:
       datasets:
         data:
           _target_: anemoi.training.losses.MultiscaleLossWrapper
           loss_matrices: [null]
           multiscale_config:
             num_scales: 4

```
or


```
   training:
     validation_metrics:
       datasets:
         data:
           multiscale:
             _target_: anemoi.training.losses.MultiscaleLossWrapper
             loss_matrices: [null]
             multiscale_config:
               num_scales: 4
```



```
   training:
     training_loss:
       datasets:
         data:
           _target_: anemoi.training.losses.CombinedLoss
           losses:
             - _target_: anemoi.training.losses.MultiscaleLossWrapper
               loss_matrices: [null]
               multiscale_config:
                 num_scales: 4
```

graphs/src/anemoi/graphs/builders.py
```diff
@@ -0,0 +1,193 @@
...
+        }
+    return smoothers
+
+
+def build_truncation_subgraph(
+    graph_data: HeteroData,
+    data_node_name: str,
+    truncation_config: Mapping | Any,
+) -> HeteroData:
+    """Add truncation nodes and edges to *graph_data* in-place.
```
ssmmnn11: Is this really in place now, or is it a new graph?

training/src/anemoi/training/losses/multiscale.py
```diff
@@ -26,17 +31,24 @@
 
 
 class MultiscaleLossWrapper(BaseLoss):
```
ssmmnn11, line 33: I think Schemas block this now from being inside a combinedloss, but we want this capability

ssmmnn11 (changes requested):
nice!

graphs/src/anemoi/graphs/builders.py
anaprietonem: I meant more that the original graph coordinates are reused. So I have updated the wording to make it more clear.

training/src/anemoi/training/losses/multiscale.py
anaprietonem: good catch, I updated the  logic now so that this should now be fixed

training/src/anemoi/training/schemas/base_schema.py
anaprietonem: Okey so the last one was indeed not working and the first two would have worked. I have fixed that now so that we have a multiscaleDisk and multiscaleOntheFly schemas and we check at runtime that either one of the options is defined (so case 1 and 2 fails). Regarding the CombinedLoss I fixes the schemas too, and added some tests here training/tests/unit/schemas/test_training_schemas.py.


Screenshot 2026-05-05 at 09 57 11
Screenshot 2026-05-05 at 09 57 11


I hope that's fixed but I have not looked at the CombinedLoss a lot, so if there is anything else I should check just let me know!

ssmmnn11 (approved):
great work!

HCookie:
This PR has added an undisclosed dependency on graphs, and must be fixed
https://github.com/ecmwf/anemoi-core/blob/1f2783c781edc6e000c579c11ca8cf17400c9079/models/src/anemoi/models/utils/config.py#L14

PR #1053, opened 2026-04-20
fix: to_refactor_branch

ssmmnn11:
some potential review findings


training/src/anemoi/training/diagnostics/callbacks/evaluation.py
```diff
@@ -70,6 +70,8 @@ def _eval(
             f"Set `task.validation_rollout` to at least {self.max_rollout}"
         )
 
+        # NOTE: This callback only works correctly when its configured rollout matches
```
JPXKQX, line 73: I think it could be lower than validation_rollout, too

Issue #1059, opened 2026-04-22
New losses for temporal downscaler


mc4117:
### Is your feature request related to a problem? Please describe.

For temporal consistency when using the temporal downscaler, we have been experimenting with using additional loss functions. We have seen large improvements in consistency especially for precipitation and so would strongly recommend using these loss functions when running the temporal downscaler and would like to implement these in anemoi-core. These include:

1. min across the window
2. max across the window
3. mean across the window 
4. diff across the window

### Organisation

ecmwf

PR #1060, opened 2026-04-22
test: add integration test for temporal downscaler ensemble

VeraChristina:
## Description
This PR adds a config template and an integration test for the temporal downscaler ensemble to ensure continued support of this use case. 

Once we add the custom losses for the temporal downscaler in https://github.com/ecmwf/anemoi-core/issues/1059, we'll update the template and the test.

### DIFF
--- a/training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
+++ b/training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
@@ -1,0 +1,49 @@
+defaults:
+- data: zarr
+- dataloader: native_grid
+- diagnostics: evaluation_ens
+- system: example
+- graph: encoder_decoder_only
+- model: transformer_ens
+- task: temporal_downscaler
+- training: ensemble
+- _self_
+
+config_validation: True
+
+### This file is for local experimentation.
+##  When you commit your changes, assign the new features and keywords
+##  to the correct defaults.
+# For example to change from default GPU count:
+# system:
+#   hardware:
+#     num_gpus_per_node: 1
+
+diagnostics:
+  plot:
+    callbacks: []
+  callbacks: []
+
+system:
+  input:
+    graph: graph_anemoi_new_${data.resolution}.pt
+    dataset: aifs-ea-an-oper-0001-mars-${data.resolution}-1979-2022-6h-v6
+    truncation: ${data.resolution}-o32-linear.mat.npz
+    truncation_inv: o32-${data.resolution}-linear.mat.npz
+    loss_matrices_path: null
+  hardware:
+    accelerator: auto
+    num_gpus_per_ensemble: 1
+    num_gpus_per_node: 1
+    num_nodes: 1
+    num_gpus_per_model: 1
+
+model:
+  num_channels: 128
+
+data:
+  resolution: o96
+
+training:
+  ensemble_size_per_device: 2
+  max_epochs: 1
--- a/training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
+++ b/training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
@@ -3,8 +3,8 @@ defaults:
 - dataloader: native_grid
 - diagnostics: evaluation_ens
 - system: example
-- graph: encoder_decoder_only
-- model: transformer_ens
+- graph: multi_scale
+- model: graphtransformer_ens
 - task: temporal_downscaler
 - training: ensemble
 - _self_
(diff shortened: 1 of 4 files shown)


training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -0,0 +1,49 @@
+defaults:
+- data: zarr
+- dataloader: native_grid
+- diagnostics: evaluation_ens
+- system: example
+- graph: encoder_decoder_only
+- model: transformer_ens
```
mc4117: could we change this to graphtransfomer_ens?

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -0,0 +1,49 @@
...
+diagnostics:
+  plot:
+    callbacks: []
+  callbacks: []
+
+system:
+  input:
+    graph: graph_anemoi_new_${data.resolution}.pt
+    dataset: aifs-ea-an-oper-0001-mars-${data.resolution}-1979-2022-6h-v6
+    truncation: ${data.resolution}-o32-linear.mat.npz
```
mc4117: I would remove the truncation matrices for now as we haven't tested this properly

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+model:
+  num_channels: 128
+
+data:
+  resolution: o96
+
+training:
+  ensemble_size_per_device: 2
+  max_epochs: 1
```
mc4117: could we do 2 epochs?

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+diagnostics:
+  plot:
+    callbacks: []
+  callbacks: []
+
+system:
+  input:
+    graph: graph_anemoi_new_${data.resolution}.pt
+    dataset: aifs-ea-an-oper-0001-mars-${data.resolution}-1979-2022-6h-v6
```
mc4117: I would put the default dataset as the 1h and then you have to overwrite the data.frequency too

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -0,0 +1,49 @@
...
+    num_nodes: 1
+    num_gpus_per_model: 1
+
+model:
+  num_channels: 128
+
+data:
+  resolution: o96
+
+training:
```
mc4117, line 42: can we overwrite the loss to the AlmostFairCRPS?

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
VeraChristina: yes, we could even remove this in the template and only overwrite in the test config?

training/tests/integration/config/test_temporal_downscaler_ensemble.yaml
```diff
@@ -0,0 +1,75 @@
...
+data:
+  frequency: 6h
+
+task:
+  _target_: anemoi.training.tasks.TemporalDownscaler
+  input_timestep: 12h
+  output_timestep: 6h
+  output_right_boundary: True
+
+training:
```
mc4117: I don't think we need to overwrite this here because I think we should just test the almost fair CRPS

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
mc4117: Yes that sounds good!

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
VeraChristina: you mean without the MultiScaleLossWrapper? The default set by `training: ensemble` is the `AlmostFairKernelCRPS` wrapped in the multi scale loss wrapper

mc4117: you're right sorry! actually we don't need to change anything here

VeraChristina: actually, maybe we should remove the `time_steps` scaler?

mc4117: I think that should be fine?

mc4117 (commented):
LGTM!

anaprietonem (approved):
LGTM from my side too!

PR #1061, opened 2026-04-22
feat(callbacks)!: add ens plot adapter and remove ensemble plot mixin

VeraChristina:
## Description

This PR adds an EnsemblePlotAdapter as a wrapper to the task specific plot adapters. It handles the additional ensemble dimension. This allows to use the basic plotting callbacks for ensemble trainings, and we can remove the ensemble mixin and most of the ensemble specific plotting callback classes.

Note that for ensemble models, the paths for `_target_` in the config under `diagnostics.plot.callbacks` change.

The PR keeps the EnsPlotSample as an additional callback because the behavior of this class differs from the overlayed PlotSample. (Currently, there also are two different sample plots for ensemble models, PlotSample and PlotEnsSample in `plot_ens.py`)
- PlotSample shows a single prediction (member 0) with an error map — useful to look at spatial accuracy of one realization.
- PlotEnsSample shows maps for truth, ensemble mean, ensemble mean error, ensemble spread, and the difference to the ensemble mean of all (or selected) ensemble members — useful for assessing spread and calibration.

Alternative: The PlotEnsSample is more relevant to evaluating ensemble models. We could deprecate PlotSample for ensemble models and make PlotEnsSample the default sample plot for ensemble models.

Test run here: https://github.com/ecmwf/anemoi-core/actions/runs/24789884281


anaprietonem:
@VeraChristina  I have update the title to include the ! (breaking change) as users will need to update the path in the plot_ens evaluation right? Could be good to add that to the description too

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -1012,6 +1014,101 @@ def _plot(
...
+            precip_and_related_fields,
+            colormaps,
+            per_sample,
+            every_n_batches,
+            dataset_names,
+            focus_area,
+            plotting_settings=plotting_settings,
+            **kwargs,
+        )
+        self.plot_members = members
```
anaprietonem, line 1108: where are `member` coming from? Is this something we could rather get out of the plot_adapter?

training/src/anemoi/training/train/methods/diffusion.py
```diff
@@ -61,6 +62,13 @@ def __init__(
 
         self.rho = config.model.model.diffusion.rho
 
+    @property
+    def plot_adapter(self) -> EnsemblePlotAdapter:
```
anaprietonem: Do I understand this is a bit of hack to handle diffusion?

anaprietonem (changes requested):
The PR looks good to me, nice contribution Vera. Mostly I would like to understand if the member can be provided by the plot_adapter and the part regarding diffusion/ensemble. I see that at the moment the plot_adapters are defining at the level of the 'tasks' while here we'd be interfering with the `training_method`. Is this showing that we might need to rethink this in the future?

training/src/anemoi/training/train/methods/diffusion.py
VeraChristina: I would say it's a second layer of adaptation. First we adapt to the task (temp downscaler, forecaster, etc.) and then we adapt to the method, i.e. for the ensemble type methods (crps, diffusion) we wrap the task-adapter in the ens-adapter. In this way we can for example get an adapter for an ensemble temporal downscaler -- by wrapping the temp-down-plot-adapter in an ens-adapter, which then handles both the ensemble and the temporal downscaling adaptations.

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: It's a plotting callback config entry for the PlotEnsSample. It configures which members to plot. So I don't think we can get it from the plot_adapter. But it looks like this got moved with merging in the callback context changes -- I'll double check where we should get this from now

mc4117: maybe I'm misunderstanding but is it not already in the ensembleplotadapter in this line https://github.com/ecmwf/anemoi-core/pull/1061/changes#diff-b4b0cc5dfa78da9544f41df97fa2dbd020f144bcff82a7a398893b66708905dfR146? I agree that it's something that should be set from the config

mc4117:
Really nice work! Would it be possible to get an example mlflow just to check what the new callbacks look like with the ensemble?

anaprietonem:
@ecmwf/anemoi_technical_subgroup while this PR includes a config breaking change, the changes lead to significant code simplification and ease of maintenance, and the deprecated key is now properly captured under `_DEPRECATED_TARGETS` on the schema Mixing, so anyone using old configs will get a proper error raised. That's why the PR is labelled as ATS approval not required.

anaprietonem (dismissed):
LGTM from my side, thanks Vera for clarifying the comments and for the nice contribution

anaprietonem (approved):

PR #1065, opened 2026-04-23
feat(graphs): Option to exclude self edges in KNN graphs

ssmmnn11:
Option to remove self edges in KNN graphs (same grid).

----
📚 Documentation preview 📚: https://anemoi-training--1065.org.readthedocs.build/en/1065/

----
📚 Documentation preview 📚: https://anemoi-graphs--1065.org.readthedocs.build/en/1065/

----
📚 Documentation preview 📚: https://anemoi-models--1065.org.readthedocs.build/en/1065/

### DIFF
--- a/graphs/docs/graphs/edges/knn.rst
+++ b/graphs/docs/graphs/edges/knn.rst
@@ -9,6 +9,10 @@ connections between two sets of nodes. Given two sets of nodes,
 (`source`, `target`), the KNN method connects all target nodes to their
 ``num_nearest_neighbours`` nearest source nodes.
 
+When ``source`` and ``target`` refer to the same node set, you can set
+``exclude_self_edges: true`` to drop self edges from the resulting
+graph.
+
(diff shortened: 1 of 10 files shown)


JPXKQX:
This could fit into a postprocessor. Do you have any concerns about including it?

ssmmnn11:
> This could fit into a postprocessor. Do you have any concerns about including it?

I think it is potentially slightly different: get KNN without self edges, vs remove all self edges from final graph. But I don't have a strong opinion.

ssmmnn11:
ok, now a post processor

JPXKQX:
I think it can be very useful and it doesn't have to be specific to the `KNNEdges` so users can use it with the `CutOffEdges` or other. I think it could be part of the base edge builder or a postprocessor.

JPXKQX (approved):
It looks good to me. Feel free to merge. Thanks Simon

PR #1066, opened 2026-04-23
fix(models): hierarchical with multiple datasets

JPXKQX:
## Description

This PR fixes the `forward()` pass for the hierarchical model. It is assumed that all datasets are encoded to the same hidden mesh and merged there before being encoded to other hidden meshes.

- Hierarchical Autoencoder doesn't work. To be fixed in another PR


anaprietonem (approved):
@JPXKQX changes LGTM, and since this functionality is broken in main I think it should be okey to merge it. 

In general it could be useful to have a chat @icedoom888 to discuss about this and the hierarchical autoencoder and what integration tests we can define to maintain this better

PR #1069, opened 2026-04-23
feat!: aggregate losses for temporal downscaler

### LINKED ISSUE #1059: New losses for temporal downscaler

mc4117:
## Description

For temporal consistency when using the temporal downscaler, we have been experimenting with using additional loss functions. We have seen large improvements in consistency especially for precipitation and so would strongly recommend using these loss functions when running the temporal downscaler and would like to implement these in anemoi-core. These include:

min across the window
max across the window
mean across the window
diff across the window

image

image

This PR is related to this issue #1059 

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1069.org.readthedocs.build/en/1069/

----
📚 Documentation preview 📚: https://anemoi-graphs--1069.org.readthedocs.build/en/1069/

----
📚 Documentation preview 📚: https://anemoi-models--1069.org.readthedocs.build/en/1069/


mc4117:
image
Example image from @Magnus-SI

mc4117:
Integration tests are now passing https://github.com/ecmwf/anemoi-core/actions/runs/24887996900

training/src/anemoi/training/config/training/ensemble.yaml
```diff
@@ -56,7 +56,11 @@ strategy:
 # don't enable this by default until it's been tested and proven beneficial
 loss_gradient_scaling: False
 
-
+# Optional time-aggregate loss added alongside the main training loss.
+# When set, the wrapper computes time-aggregated predictions (e.g. diff, mean)
+# and evaluates a loss on them, weighted by `weight`.
+# Only valid if the number of time steps is greater than 1.
+time_aggregate_loss: null
```
VeraChristina: I am wondering if we want to introduce a specifig config entry for the `time_aggregrate_loss` or if it could be configured using the `CombinedLoss` class instead? For example, to set up a time-aggregate loss in addition to a standard MSE, one would configure a CombinedLoss with two child losses, namely the standard MSE and the TimeAggregateLossWrapper (wrapping an MSE loss)?

training/src/anemoi/training/config/temporal_downscaler.yaml
```diff
@@ -23,3 +23,25 @@ diagnostics:
...
+        ignore_nans: False
+  time_aggregate_loss:
+    datasets:
+      data: # user-defined key in data
+        time_aggregation_types: [mean, max, min, diff]
+        weight: 1.0
+        loss_fn:
+          _target_: anemoi.training.losses.MSELoss
+          scalers: ['pressure_level', 'general_variable', 'node_weights']
+          ignore_nans: False
```
VeraChristina: Elaborating on the question above, this would be the config of the `AggregateLossWrapper` which would be one of the child losses of the `CombinedLoss`

VeraChristina (commented):
Thanks for adding this! One question regarding user interface and design: Do we want the time aggregate losses to be handled separately or could this be used via the CombinedLoss class (which provides functionality for weighting etc. and would mean all losses are handled via the same logic)? Perhaps the AggregateLossWrapper doesn't fit the interface that the CombinedLoss expects for child losses (in which case this suggestion wouldn't work, at least not out of the box) -- but I thought it might be worth considering

training/src/anemoi/training/config/training/ensemble.yaml
mc4117: I originally wanted to do it with the combined loss but then changed it in this commit (https://github.com/ecmwf/anemoi-core/pull/1069/changes/8a8a78890a2250530d9f425ecd64bef600db55ed#diff-0e2f9faff79204815efb0498d8c8aa3e6747abb5955dce3b1f6cd8d03ad6d72b) because it made the schema and inheritance from different configs very complicated. For example if you overwrote the combined loss over the ensemble training yaml, you would automatically have alpha and other almost fair properties in your combined loss breaking schema and also potentially leading to user errors

VeraChristina: Right, to avoid that we would need to always wrap in the combined loss (even when we only use one loss function, making the configs more complicated rather than less complicated)? I don't really like introducing a new loss part in the config and reimplementating the combination of losses for one specific case, when the combined loss wrapper already exists -- but I agree that we need working config overwrites and schemas

mc4117: the other option I was considering was making a training/loss config folder (similar to training scaler). But even then you would introduce four new files: single_loss; ensemble_loss; single_td_loss; ensemble_td_loss

VeraChristina: I would prefer the more verbose templates over having the changes in `train/methods/base.py` below to add the aggregate_losses

mc4117: I can revert to the old commit then and make the changes

training/src/anemoi/training/losses/base.py
```diff
@@ -339,6 +339,10 @@ def forward(
             Weighted loss
         """
         is_sharded = grid_shard_slice is not None
+        # When target has one fewer dimension than pred, insert the ensemble dim so
+        # broadcasting aligns (bs, t, 1, latlon, nvar) against (bs, t, ens, latlon, nvar).
+        if target.ndim == pred.ndim - 1:
+            target = target.unsqueeze(TensorDim.ENSEMBLE_DIM)
```
VeraChristina: does not seem related to the PR, does it fix an existing bug?

training/src/anemoi/training/losses/loss.py
```diff
@@ -28,6 +28,7 @@
 METRIC_RANGE_DTYPE = dict[str, list[int]]
 
 NESTED_LOSSES = ["anemoi.training.losses.MultiscaleLossWrapper"]
+WRAPPED_LOSSES = ["anemoi.training.losses.aggregate.TimeAggregateLossWrapper"]
```
VeraChristina, line 31: These seem to have the same role and logic below. Can we unify them?

VeraChristina (commented):
Thanks for making the changes! Two small things in the comments but otherwise LGTM. Could you add a small section in the docs for the temporal downscaler mentioning the purpose of these loss functions?

training/src/anemoi/training/losses/loss.py
VeraChristina: ah, that would break the configs, right?

mc4117: We could unify them. The problem is MultiscaleLoss uses `per_scale_loss` to define the function which doesn't really make sense for the TimeAggregateLoss. We could rename to `loss_fn` in both but this would be a breaking change

training/src/anemoi/training/losses/base.py
mc4117: I think it was needed for a previous version of the branch. I have reverted the change and will check if it causes problems

training/src/anemoi/training/losses/aggregate.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def __init__(
+        self,
+        time_aggregation_types: list[str],
+        loss_fn: BaseLoss,
+        ignore_nans: bool = False,
+    ) -> None:
+        super().__init__(ignore_nans=ignore_nans)
+        self.time_aggregation_types = time_aggregation_types
+        self.loss_fn = loss_fn
```
ssmmnn11: I think it would be best if TimeAggregateLossWrapper acts as a transparent wrapper around loss_fn. Currently the wrapper inherits from BaseLoss and has its own scalers but computation happens in loss_fn -> scaler updates may affect the wrapper, not the effective loss. So could we delegate/share wrapper metadata and methods -> scaler, add_scaler, update_scaler, has_scaler_for_dim, supports_sharding, needs_shard_layout_info, iter_leaf_losses.

This would make scaler inheritance/updates work when used inside CombinedLoss. And It should also make nested wrappers work correctly, like TimeAggregateLossWrapper(MultiscaleLossWrapper(...)) I think LossVariableMapper is a good example here or partially MultiscaleLossWrapper.

ssmmnn11 (changes requested):
Nice addition!

mc4117:
@ecmwf/anemoi_technical_subgroup I have updated the label to ATS approval not needed as this is not a breaking config change and does not change default config structure. If anyone disagrees please let me know!

training/src/anemoi/training/losses/aggregate.py
VeraChristina: @ssmmnn11 I have added a BaseLossWrapper class that delegates metadata and scaler methods to the inner loss, to make this more of a default for nested loss wrappers. It does not change the behaviour of the exisiting classes

ssmmnn11 (dismissed):
thank you for this nice contribution!

training/docs/modules/tasks.rst
```diff
@@ -155,6 +155,9 @@ Example: ``input_timestep="6H"``, ``output_timestep="3H"``,
 ``output_left_boundary=True`` produces output offsets
 ``[0H, 3H]`` and input offsets ``[0H, 6H]``.
 
+We strongly recommend using the time aggregate loss when training any
```
ssmmnn11: I would just say "This encourages the model  ...", not "We strongly recommend ...".

training/docs/modules/losses.rst
```diff
@@ -83,6 +83,63 @@ deterministic:
 
 .. _multiscale-loss-functions:
 
+***************************
+ Time Aggregate Loss Functions
+***************************
+
+A key challenge in temporal downscaling is **temporal consistency**: the
```
ssmmnn11: I would keep this technical and describe what it does - not so much interpretation, which can come in a technical report / paper.

training/src/anemoi/training/config/training/training_loss/ensemble_multiscale_aggregation.yaml
```diff
@@ -0,0 +1,30 @@
+datasets:
```
ssmmnn11, line 1: I think it would be useful to make the name more specific - something that say what is combined.

training/src/anemoi/training/config/training/training_loss/single_MSE_aggregation.yaml
```diff
@@ -0,0 +1,20 @@
+datasets:
```
ssmmnn11, line 1: I would recommend to make the name more specific - what is combined?

ssmmnn11 (commented):
some suggestions.

training/docs/modules/tasks.rst
mc4117: Or I could put something like `the default in anemoi is to use the time aggregate loss...`?

ssmmnn11: that makes sense!

training/src/anemoi/training/config/training/training_loss/ensemble_multiscale_aggregation.yaml
mc4117: I changed it to `composite` - is that clearer? The idea is just that for pydantic schema reasons if you want to inherit from the combined loss, you have to use this task

mc4117: changed to ensemble_multiscale_aggregation

training/src/anemoi/training/config/training/training_loss/single_MSE_aggregation.yaml
mc4117: changed to single_mse_aggregation

mc4117:
I confirm that the old code and the new code are equivalent 

train_afkcrps0 95_step-train_multi_dataset_loss_step

none
anaprietonem, line 1: what's is this file?

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
```diff
@@ -0,0 +1,141 @@
...
+import torch
+from pytorch_lightning.callbacks import Callback
+
+from anemoi.training.losses.base import BaseLoss
+from anemoi.training.utils.enums import TensorDim
+
+LOGGER = logging.getLogger(__name__)
+
+
+class PerTimestepMetrics(Callback):
```
anaprietonem, line 26: I might be missing something but where is this class used or instantiated?

training/src/anemoi/training/schemas/training.py
```diff
@@ -352,6 +368,33 @@ def check_no_deprecated_mixed_with_on_the_fly(self) -> Self:
...
+    time_aggregation_types: list[Literal["diff", "mean", "min", "max"]] = Field(min_length=1)
+    "Time aggregation operations to apply over the time dimension before computing the loss."
+    loss_fn: BaseLossSchema | CRPSSchema
+    "Inner loss function applied to each time-aggregated output."
+    scalers: list[str] | None = None
+    "Scalers to apply to the wrapped loss (delegated to inner loss_fn)."
+
+    @field_validator("loss_fn", mode="before")
+    @classmethod
+    def add_empty_scalers_to_inner(cls, v: Any) -> Any:
```
anaprietonem, line 387: why do we need this? since it's on the pydantic schema this would just be applied when pydantic is on so wondering if that's a problem or we've tested the config-validation off too already?

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
```diff
@@ -0,0 +1,141 @@
...
+                        continue
+
+                    for mkey, indices in val_metric_ranges.items():
+                        step_name = f"val_{metric_name}_metric/{dataset_name}/{mkey}/t_{t + 1}"
+
+                        metric_kwargs = {
+                            "scaler_indices": (..., indices),
+                            "grid_shard_slice": grid_shard_slice,
+                            "group": pl_module.model_comm_group,
+                        }
```
JPXKQX, line 132: what about the `pred_layout`, `target_layout` and `needs_shard_layout_info` arguments that appear in the base.py?

training/src/anemoi/training/losses/aggregate.py
```diff
@@ -0,0 +1,156 @@
+from __future__ import annotations
```
JPXKQX, line 10: the copyright/license is missing in this new file

training/docs/modules/losses.rst
```diff
@@ -91,6 +91,60 @@ deterministic:
...
+      -  Minimum over the output time window — penalises errors in
+         minimum values.
+
+   -  -  ``diff``
+      -  Consecutive step-to-step differences
+         (``pred[:, 1:] - pred[:, :-1]``) — penalises unrealistic
+         temporal transitions and discontinuities.
+
+The wrapper accumulates the specified loss function evaluated on each aggregation in
+turn and returns the sum. Because the ``time_steps`` scaler is
```
JPXKQX: does it return the sum? in aggregate.py, it seems it is computing the average:
```
        if self.time_aggregation_types:
            loss = loss / len(self.time_aggregation_types)
```

training/src/anemoi/training/losses/variable_mapper.py
```diff
@@ -364,9 +347,10 @@ def forward(
                 "without_scalers": without_scalers,
                 "grid_shard_slice": grid_shard_slice,
                 "group": group,
-                "squash_mode": squash_mode,
             },
```
anaprietonem, line 351: To confirm, you've checked this change is not having an impact with main defaults as all the implemented loss under the losses folder specify a squash_mode when calling the forward correct? (ie if one did not, then they would change from using avg to None)

training/src/anemoi/training/schemas/training.py
VeraChristina: I think I added this to prevent validation failure on the BaseLossSchema if the scalers are specified on the wrapper or handed down via the combined loss (as is intended) and therefore missing on the inner loss config. -- iirc, scalers were mandatory on the BaseLossSchema at the time. They now have an empty list as default. (yes, see change in line 266, same file)
If we want them to be mandatory (for cases without wrappers), we remove the default and keep the validator; if we are happy with the default empty list (i.e. no user warnings for forgetting to specify scalers), the validator is obsolete

training/docs/modules/losses.rst
mc4117: you're right, thanks for spotting this

none
mc4117: not sure. I have deleted it

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
mc4117: you're right it was there in my local version of configs but not in the github version. I've added it now

training/src/anemoi/training/losses/variable_mapper.py
mc4117: I think this change is not necessary actually so I have reverted

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
mc4117: added in and this callback is now tested in the integration test

training/src/anemoi/training/schemas/training.py
anaprietonem: Thanks Vera for clarifying what motivated the change, the follow up I have is then what's the expected behaviour when pydantic is off? are users required to pass the scalers:[] inside the inner losses and if not the code catches the error and raise it clearly?

VeraChristina: I have reverted the change on the base loss schema (i.e. scalers now required as before) and kept the validator to prevent validation errors for valid configs.

For pydantic off, the same default configs work (i.e. users do not need to pass `scalers: []` in the inner loss) because the code runs fine without scalers. The validator is only needed to prevent validation errors.

The main difference for pydantic on/off is the basic case (like just an MSE loss without wrapper) where there will be no warning for missing scalers if pydantic is off. This seems to be the intended difference for validation on/off, and this also is the case in main (so no change for existing use cases).

anaprietonem (approved):
thanks for addressing the comments and questions @mc4117  and @VeraChristina ! very nice feature! LGTM to be merged!

PR #1070, opened 2026-04-23
feat(training): flexible timestepping for forecaster

dietervdb-meteo:
## Description
When defining an hourly forecaster that has the same input and output horizon as the  'default' forecaster (in -6h, 0h, out 6h)  one is currently forced to set `multistep_input = 7`, `multistep_output = 6`, `timestep = 1`. This leads to 7 hourly input steps which has two important disadvantages: 1. increases memory load (in particular impactful for high spatial resolution models) 2. hourly analysis is not available for input in operational settings. This motivates a forecaster that can simply take -6h, 0h as input but predict hourly up to 6h in the future [with a single model forward]. Experiments with such a forecaster have shown it has slightly better error growth at late lead times compared to the one with 7 hourly input steps, be it at the cost of slightly less skill in the first 6 hours. This shows it is a valid (and possibly desirable) solution to the disadvantages listed above. The aim of this PR is to enable such a forecaster in anemoi.

To implement the forecaster described above we need to generalize beyond a configuration based on the number of input and output steps, together with a fixed timestep. Since under the hood the `BaseTask` already is defined in terms of (arbitrary) input and output offsets it is natural to go to a forecaster configuration based on offsets. 
To prevent the more general forecaster to interfere with the current simpler implementation we define it through a new class `OffsetForecaster`, with both this new task and the original `Forecaster` inheriting from `BaseForecaster`. 
What makes the forecasting tasks special is that they support rollout, and not every selection of offsets is compatible with rollout. For this reason we have carefully spelled out what we expect of the `OffsetForecaster` task (predictions from past to future, rollout up to any number of steps, no repeated forecasts), see `adr_003.md` added in this PR for details. This PR also puts into place a validation on instantiation that makes sure offsets as specified in the config lead to a valid forecaster.

Since for `OffsetForecaster` we are generalizing configurability, we consider it reasonable to do so arbitrarily widely inside the valid forecaster setup [see the adr for various exotic, but sensible, examples]. This however requires the rollout implementation to be slightly modified, from a `.roll()` based advancement to an index mapping based one (see the adr for motivation). Tests show this new implemtation is numerically identical and doesn't affect performance for cases where `OffsetForecaster` and `Forecaster` overlap.

See https://github.com/ecmwf/anemoi-core/discussions/941

----
📚 Documentation preview 📚: https://anemoi-training--1070.org.readthedocs.build/en/1070/

----
📚 Documentation preview 📚: https://anemoi-graphs--1070.org.readthedocs.build/en/1070/

----
📚 Documentation preview 📚: https://anemoi-models--1070.org.readthedocs.build/en/1070/


cathalobrien:
I ran a benchmark (n320->L6 global GT on rollout 6) comparing main with feat/flexible-timestepping . the speed and memory usage is essentially identical.
| branch                     | throughput (it/s) | peak alloc. men (GB) | 
|----------------------------|-------------------|----------------------|
| main                       | 0.91              | 62.7                 | 
| feat/flexible-timestepping | 0.90              | 62.7                 | 

Also looking at the memory snapshot and runtime traces, both looked essentially identical

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -74,32 +74,53 @@ class Forecaster(BaseTask):
...
                 "The following extra parameters were provided to %s but will be ignored: %s",
                 self.__class__.__name__,
                 kwargs,
             )
 
-        # Input: e.g. multistep_input=2, timestep=6H     ->  [-6H, 0H]
-        input_offsets = [-1 * i * self.timestep for i in range(multistep_input)]
-        # Outputs: e.g. multistep_output=1, timestep=6H  -> [[6H], [12H], [18H], ...] up to rollout.maximum
-        output_offsets = [(i + 1) * self.timestep for i in range(multistep_output)]
+        if multistep_input is not None or multistep_output is not None or timestep is not None:
```
mc4117: I think this part could be a bit cleaner as at the moment it's a bit difficult to follow? Also it seems to be missing the case where all the time specifications are none?

mc4117: If a user sets only multistep_input (forgetting multistep_output/timestep), I think this will still run and then multistep_output=None raises TypeError: 'NoneType' object cannot be interpreted as an integer.

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -74,32 +74,53 @@ class Forecaster(BaseTask):
...
+                "multistep_output, and timestep must not be provided."
+            )
+            input_offsets = sorted(frequency_to_timedelta(v) for v in input_offsets)
+            output_offsets = sorted(frequency_to_timedelta(v) for v in output_offsets)
+            rollout_shift = frequency_to_timedelta(rollout_shift)
+
         super().__init__(input_offsets=input_offsets, output_offsets=output_offsets)
+        self._rollout_shift = rollout_shift
+
+        if consistency_check:
```
mc4117: Maybe I'm missing something but i think if consistency_check = False then rollout_shift isn't properly set?

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: Agreed, if ATS decides to go for legacy support of old configs then we need to clean this up and make it airtight. If it is decided to break configs to new format then all of this can simply go. For now just put something simple in place that saved me changing all the default configs and related tests :)

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: It is set to `0H`, which means no shift at all. In some sense the only technically working case if one falls into some invalid setup, which is what this switch is for. The idea would be that only very advanced users would turn this off for training without rollout, or if they want to do something crazy like prediciting the past. In that case the user should set whatevr `rollout_shift` they want via the config. But if you have a good suggestion for what a proper setting would be I'm happy to implent it.

training/src/anemoi/training/tasks/forecaster.py
frazane: If we keep these arguments, are we nevertheless going to deprecate them?

mc4117: My preference would be to not deprecate but I think this should be a decision for ATS in a future PR

training/src/anemoi/training/schemas/tasks.py
```diff
@@ -46,6 +46,25 @@ class ForecasterSchema(BaseModel):
...
+class OffsetForecasterSchema(BaseModel):
+    """Configuration for the offset-based forecasting task."""
+
+    target_: Literal["anemoi.training.tasks.OffsetForecaster"] = Field(..., alias="_target_")
+    "Task class path for the offset-based forecasting task."
+    input_offsets: list[str] = Field(example=["-6H", "0H"])
+    "Input time offsets as duration strings."
+    output_offsets: list[str] = Field(example=["6H"])
+    "Output time offsets as duration strings."
+    rollout_shift: str = Field(default="0H", example="6H")
```
MicheleCattaneo:
```suggestion
    rollout_shift: str | None = Field(default=None, example="6H")
```
I find the default "0H" a bit misleading, it's a valid duration string, yet it is then just ignored and the shift is automatically set.

training/docs/adrs/adr-003.md
```diff
@@ -0,0 +1,348 @@
...
+
+### Validation scheme
+
+The conditions are checked at construction (`_validate_offsets` and
+`_validate_rollout_shift`), so an invalid forecaster is rejected early with an
+actionable `ValueError` rather than failing — or silently misbehaving — later during
+training or inference.
+
+- **Inferring `rollout_shift`.** If `rollout_shift` is omitted (defaults to
+  `"0H"`) it is inferred as the **largest** valid shift. If no valid shift exists,
```
MicheleCattaneo:
```suggestion
  `None`) it is inferred as the **largest** valid shift. If no valid shift exists,
```

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -259,4 +211,280 @@ def on_train_epoch_end(self, current_epoch: int) -> None:
...
+    Offsets are validated for consistency with rollout.
+    The default rollout shift is the maximum valid shift.
+    """
+
+    name: str = "offset-forecaster"
+
+    def __init__(
+        self,
+        input_offsets: list[str] | None = None,
+        output_offsets: list[str] | None = None,
```
MicheleCattaneo:
```suggestion
        input_offsets: list[str],
        output_offsets: list[str],
```
These can not be None. It would crash in the first lines of the constructor and it's not matching the schema.

MicheleCattaneo (commented):
Overall it looks good to me.
For now, just a minor comment about the offsets being optional (it does not match the schema and would crash).
Something that I think it's more important is the default rollout shift being `"0H"`.
I think if this is set and `consistency_check` is `False`, the model can silently not advance (because the default shift if a valid time string). Clearly a user mistake but I would avoid to put the option to make this mistake if we can. What do you think? For now I just suggested to put the default shift to None, to allow the automatic discovery of the best shift, but that would need some additional changes to handle the None type, e.g in the `OffsetForecaster`. If you agree that it should not be `"0H"`, I will look deeper into whats needed.
Also another question; what is the reason to have `consistency_check` turned off? Should it not always be done (and not be a constructor flag)?

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: Has become irrelevant as it was decided to go with separate classes.

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: Fixed, thanks.

training/src/anemoi/training/schemas/tasks.py
dietervdb-meteo: Set now as `"default"`, that should be clearer.

dietervdb-meteo:
@MicheleCattaneo , made consistency check mandatory.

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: Validation had become mandatory, so this eliminates this issue.

training/docs/modules/tasks.rst
```diff
@@ -135,6 +135,72 @@ Rollout behavior:
...
 
 
+******************
+ OffsetForecaster
+******************
+
+:class:`~anemoi.training.tasks.forecasting.OffsetForecaster` is an
+experimental generalization of
+:class:`~anemoi.training.tasks.forecasting.Forecaster` that declares the
+sample directly through time offsets, i.e. times relative to the start of the forecast:
```
mc4117:
```suggestion
sample directly through time offsets, i.e. times relative to the forecast initialisation time:
```

training/docs/modules/tasks.rst
```diff
@@ -135,6 +135,72 @@ Rollout behavior:
    ``multistep_input`` outputs are fed into the next rollout step.
 
 
+******************
+ OffsetForecaster
+******************
+
+:class:`~anemoi.training.tasks.forecasting.OffsetForecaster` is an
+experimental generalization of
```
mc4117:
```suggestion
experimental generalisation of
```

training/docs/modules/tasks.rst
```diff
@@ -135,6 +135,72 @@ Rollout behavior:
...
+:class:`~anemoi.training.tasks.forecasting.Forecaster` that declares the
+sample directly through time offsets, i.e. times relative to the start of the forecast:
+
+- ``input_offsets`` — input time offsets as frequency strings, e.g.
+  ``["-6H", "0H"]``.
+- ``output_offsets`` — target time offsets, e.g. ``["6H"]``.
```
mc4117:
```suggestion
- ``output_offsets`` — target time offsets as frequency strings, e.g. ``["6H"]``.
```

training/docs/modules/tasks.rst
```diff
@@ -135,6 +135,72 @@ Rollout behavior:
...
+
+- ``rollout`` / ``validation_rollout`` — as for
+  :class:`~anemoi.training.tasks.forecasting.Forecaster`.
+
```
mc4117, line 153: I would put a heading here like 
"Examples"

training/docs/modules/tasks.rst
```diff
@@ -135,6 +135,72 @@ Rollout behavior:
...
+The standard forecaster (two input steps, one output step, ``6H``
+timestep) written with offsets:
+
+.. code:: yaml
+
+  task:
+    _target_: anemoi.training.tasks.OffsetForecaster
+    input_offsets: ["-6H", "0H"]
+    output_offsets: ["6H"]
```
mc4117, line 165: could you add rollout here too?

training/src/anemoi/training/config/task/offset_forecaster.yaml
```diff
@@ -0,0 +1,14 @@
+_target_: anemoi.training.tasks.OffsetForecaster
+input_offsets: ["-6H", "0H"]
+output_offsets: ["6H"]
+
+# Rollout configuration for autoregressive forecasting
+rollout:
```
mc4117, line 6: but shouldn't this now be defining the rollout shift?

training/src/anemoi/training/schemas/tasks.py
```diff
@@ -46,6 +46,23 @@ class ForecasterSchema(BaseModel):
     "Number of rollouts to use for validation. If unset, validation uses the training rollout."
 
 
+class OffsetForecasterSchema(BaseModel):
+    """Configuration for the offset-based forecasting task."""
+
+    target_: Literal["anemoi.training.tasks.OffsetForecaster"] = Field(..., alias="_target_")
+    "Task class path for the offset-based forecasting task."
+    input_offsets: list[str] = Field(example=["-6H", "0H"])
```
mc4117, line 54: I would add a minimum length for the input and output offsets

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -59,47 +59,39 @@ def load_state_dict(self, state: dict) -> None:
...
 
-class Forecaster(BaseTask):
-    """Forecasting task implementation.
-
-    Builds input and output offsets from ``multistep_input``,
-    ``multistep_output`` and a ``timestep`` string (e.g. ``"6H"``).
+class BaseForecaster(BaseTask):
+    """Abstract forecasting task implementation.
 
+    Actual implementations should inherit from this class.
```
mc4117, line 65: I'm not sure I would add this line as it's not there in the basetask for example?

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -259,4 +211,285 @@ def on_train_epoch_end(self, current_epoch: int) -> None:
...
+    def _compute_advance_map(self) -> dict[str, list[tuple[int, int]]]:
+        """Pre-compute index mappings for input advancement during a rollout step."""
+        out_to_idx = {o: j for j, o in enumerate(self._output_offsets)}
+        in_to_idx = {i: j for j, i in enumerate(self._input_offsets)}
+        advance_map = {"inin": [], "outin": []}
+        for new_idx, in_offset in enumerate(self._input_offsets):
+            shifted_in = in_offset + self._rollout_shift
+            if shifted_in in out_to_idx:
+                advance_map["outin"].append((out_to_idx[shifted_in], new_idx))
+            else:
```
mc4117, line 356: what if shifted_in is not in in_to_idx either? would be good to have an assert here I think

training/tests/unit/tasks/test_forecaster.py
```diff
@@ -372,3 +373,216 @@ def test_rollout_advance_input_reapplies_boundary_truth_and_refreshes_forcing()
     torch.testing.assert_close(updated[0, -1, 0, :, 0], torch.tensor([10.0, 200.0]))
     # forcing variable should be refreshed from batch for both grid points
     torch.testing.assert_close(updated[0, -1, 0, :, 1], torch.tensor([1000.0, 2000.0]))
+
+
+# ── OffsetForecaster: equivalence with Forecaster on a regular grid ────────────
```
mc4117, line 378: it would be great to have a test where the offsets are irregular e.g.  ([-4H, -2H, -1H, 0H] → ["1H", "2H"])

training/docs/modules/tasks.rst
MicheleCattaneo: Is anything "hardcoding" hours at the moment? Assuming a 10minute gridded dataset, would something not work?

mc4117: you're right it doesn't need to be hours

training/docs/modules/tasks.rst
```diff
@@ -135,6 +135,72 @@ Rollout behavior:
...
+ OffsetForecaster
+******************
+
+:class:`~anemoi.training.tasks.forecasting.OffsetForecaster` is an
+experimental generalization of
+:class:`~anemoi.training.tasks.forecasting.Forecaster` that declares the
+sample directly through time offsets, i.e. times relative to the start of the forecast:
+
+- ``input_offsets`` — input time offsets as frequency strings, e.g.
+  ``["-6H", "0H"]``.
```
mc4117, line 148: I would put an example here of a different frequency other than hours

training/docs/modules/tasks.rst
dietervdb-meteo: Indeed, there is even a unit test using minutes.

training/src/anemoi/training/config/task/offset_forecaster.yaml
dietervdb-meteo: I think it shouldn't. I believe the 'rollout configuration' and the key `rollout` is a bit misleading here actually, it is specifically about rollout training, while e.g. `rollout_shift` is a feature off the model/task that is also of relevance in inference, independent of the type of rollout training is performed [the `rollout_shift`, `step_shift` earlier, was a property of the forecaster task class, not the rollout configuration class.] 

I think it could be better to rename the `rollout` key to `rollout_curriculum` or something, since it is about how one wants to schedule the rollout training. @JPXKQX , what do you think? [I believe it was you that set up a separate rollout configuration class for the forecaster?] 

Still, setting `rollout_shift` explicitly through the config will be something that will be used only in a very niche set of cases by very few users, so we might also just keep things as they are for now. But that's just my opinion, happy to hear if people see things differently.

training/src/anemoi/training/schemas/tasks.py
dietervdb-meteo: Ok, done. [also the validation during `__init__` would fail btw]

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: Fine, removed.

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: This is actually guaranteed by the validation of `rollout_shift`. Are you worried that is not robust enough?

training/tests/unit/tasks/test_forecaster.py
dietervdb-meteo: done

training/docs/modules/tasks.rst
dietervdb-meteo: done

training/docs/modules/tasks.rst
dietervdb-meteo: done

training/docs/modules/tasks.rst
dietervdb-meteo: done

dietervdb-meteo:
[Integration tests](https://github.com/ecmwf/anemoi-core/actions/runs/31477144526) pass. We can add the `run-expensive-tests` label all the way at the end.

training/src/anemoi/training/config/task/offset_forecaster.yaml
mc4117: Ok I'm happy to leave this as it is for now and then we can maybe revisit the naming of rollout in a future PR

training/docs/modules/tasks.rst
```diff
@@ -135,6 +135,83 @@ Rollout behavior:
...
+    rollout:
+      start: 1
+      epoch_increment: 1
+      maximum: 6
+
+The rollout shift
+=================
+
+The model can forecast arbitrarily far into the future by rollout: it is
+applied repeatedly, each step feeding its own outputs (and inputs) back as the inputs of
```
mc4117, line 195:
```suggestion
applied repeatedly, each step feeding its own outputs back as the inputs of
```

training/src/anemoi/training/tasks/forecaster.py
mc4117: Ah I hadn't spotted that it was guaranteed there. Thanks for clarifying!

training/docs/modules/tasks.rst
dietervdb-meteo: Wouldn't we like to be precise? Typically part of the input of a forecast step is reused in the next step upon rollout. (e.g. -6 0 -> 6, 0 6 -> 12 etc)

ssmmnn11 (dismissed):
Great work! 

Minor suggestions in:

https://github.com/ecmwf/anemoi-core/pull/1355

dietervdb-meteo:
Thanks @ssmmnn11, adopted those suggestions.

mc4117 (approved):
LGTM! Thanks for all your hard work. Once tests pass I'm happy to merge

PR #1074, opened 2026-04-24
feat: Support for multiple dataset data-loader with different frequencies

OpheliaMiralles:
## Description
This PR introduces flexibility in the forecaster for sparse data loading. Usecase is mainly linked to the use of several dataset with varying frequencies, especially if the frequencies differ a lot (e.g. 6h and 5min).

The sparse rollout behavior lives in the `Forecaster` task, and the data loading pieces were adjusted so sparse mixed-frequency windows are supported. If frequencies of the datasets align, handling of indices should be similar as before. If frequencies differ then the `DataLoader` should select exact corresponding offsets.

I am open to alternative design choices.

In practice, it means you should be able to use a config with e.g. 
```
config_validation: false

task:
  _target_: anemoi.training.tasks.Forecaster
  multistep_input: 12 
  multistep_output: 12
  timestep: 5m
  rollout:
    start: 1
    epoch_increment: 0
    maximum: 1
  validation_rollout: 1

data:
  frequency: 5m
  datasets:
    nwp:
      forcing: [forcing_var]
      diagnostic: []
      target: []
    radar:
      forcing: [forcing_var]
      diagnostic: []
      target: []

dataloader:
  training:
    datasets:
      nwp:
        dataset_config:
          dataset: /path/to/nwp.zarr
          frequency: 6h
          drop: []
        start: 2020-01-01 00:00:00
        end: 2020-01-02 00:00:00
      radar:
        dataset_config:
          dataset: /path/to/radar.zarr
          frequency: 5m
        start: 2020-01-01 00:00:00
        end: 2020-01-02 00:00:00
  validation:
    datasets:
      nwp:
        dataset_config:
          dataset: /path/to/nwp.zarr
          frequency: 6h
        start: 2020-01-02 00:00:00
        end: 2020-01-03 00:00:00
      radar:
        dataset_config:
          dataset: /path/to/radar.zarr
          frequency: 5m
        start: 2020-01-02 00:00:00
        end: 2020-01-03 00:00:00
  test:
    datasets: {}
```
In the case where multiple data sources are used with different time frequencies, we want to be able to not overload the memory by interpolating the frequency of the coarser (in time) dataset to the higher-resolution one. This PR allows to pass in some desired time steps (e.g. last available NWP timestep and last hour of radar data) and predict whichever time step too (e.g. the next two hours of rain). This is a draft but will be continuously updated. 

##  Additional notes ##
# What Changed

- updated `anemoi.training.tasks.Forecaster`
- added `training/config/task/sparse_forecaster.yaml`
- updated datamodule / relative time index handling 
- preserved per-dataset sparse timing metadata for mixed-frequency loading
- added focused unit/integration coverage for sparse forecaster and sparse metadata handling

# Testing

 `.venv/bin/python -m pytest training/tests/unit/tasks/test_forecaster.py training/tests/unit/data/test_datamodule.py training/tests/unit/data/test_multidataset.py training/tests/integration/test_sparse_dataset_time_indices.py -q`


OpheliaMiralles:
@copilot

dietervdb-meteo:
This is a nice and interesting use case @OpheliaMiralles ! I think it would be good to refactor the offset logic a bit. Currently there are two kinds of offsets used with quite some overlap, the ones of the `BaseTask` which get created via the `multistep` entries [but the actual timedeltas never get used], and then the dataset specific ones via the `dataset_time_indices`. To me this hints at two refactors. 1. make per dataset `BaseTask` offsets 2. Define a `BaseForecaster` that works with pure offsets and without the `multistep` attributes. Then have separate subclasses `Forecaster` (which works with multistep) and `SparseForecaster` which works with dataset specific offsets (currently the `dataset_time_indices`) and doesn't require `multistep` anymore. These two refactors fit quite well with how things are set up in #1070.

Another point of overlap with #1070 is the generalization of rollout. We should set up a meeting and discuss!

JPXKQX:
Hello Ophelia, Dieter, I leave here my 2 cents, following also some of the discussions during the task decoupling! I think there are three functional requirements for this task:
- Per-dataset configuration. This is not specific to the forecaster, as other users could benefit from it in later use cases. The base class could handle this with minimal changes by converting `relative_date_indices` from a `list[int]` to a `dict[str, list[int]]`. The tasks now don't depend on anything else. They are created directly from the config files, but there is a dependency on other parts of the code, such as the `TrainingModule` or the `DataModule`. The main open question is how to handle the per-dataset configuration. Does this have to be explicitly defined for all datasets, and if so, how and when should we check the datasets mentioned in the task and the dataloader? Should we introduce a dependency of the task on the datamodule? This would be somehow a circular dependency, as the task is used to create the data readers.
- Non-uniform steps (in the forecaster). I think this could be minor, as the underlying `BaseTask` supports any offsets. The question here is whether to support different `Forecaster` classes for different levels of customisation (`SimpleForecaster`, `PerDatasetForecaster`, `NonUniformForecaster`, ...), to support only a general class or something else.
- Rollout customisation. Less urgent, but it will be important when we start combining datasets with different frequencies

OpheliaMiralles:
Right now:
- this PR adds mixed-frequency handling within the existing `Forecaster` path. On main, Forecaster defines one global input/output window from `multistep_input`, `multistep_output,` and `timestep`, and that same temporal structure is effectively assumed for every dataset. In this PR, the `datamodule` computes per-dataset relative input and target indices on the shared model time grid and exposes them through the metadata; the task then uses those indices to select the appropriate dataset-specific windows at runtime. It allows datasets with different native cadences to participate in the same training sample, even when a coarser dataset contributes fewer frames within the same model window.
- the rollout scheme itself is not generalised as part of this PR: the current `Forecaster`  still uses a single shared model timestep and advances rollout in consecutive blocks on that grid. So mixed native frequencies are supported here, but fully custom per-step rollout schedules are more in scope for #1070.

I chose to make the existing `Forecaster` more flexible rather than creating new tasks based on @ssmmnn11 comment on slack, but this is debattable. I would of course be happy to discuss the overlap with #1070 and/or support for an eventual `FlexibleSparseForecaster`. I agree that if we want to generalise this properly, we should agree on something consistent.

training/tests/integration/config/test_mixed_frequency_multidatasets.yaml
```diff
@@ -0,0 +1,61 @@
...
+    nordic_radar:
+      forcing: ["forcing_var"]
+      diagnostic: []
+      target: []
+
+task:
+  _target_: anemoi.training.tasks.Forecaster
+  multistep_input: 1
+  multistep_output: 1
+  timestep: "5m"
```
mc4117: I was just wondering given that one dataset doesn't have 5 mins, what does this timestep here mean? Does it take 5 mins for the datasets which have 5 mins and the smallest timestep for the datasets that don't?

training/tests/unit/schemas/test_task_schemas.py
```diff
@@ -0,0 +1,38 @@
...
+        "_target_": "anemoi.training.tasks.Forecaster",
+        "multistep_input": 1,
+        "multistep_output": 1,
+        "timestep": "5m",
+        "rollout": {"start": 1, "epoch_increment": 0, "maximum": 1},
+        "validation_rollout": 1,
+    }
+
+
+def test_forecaster_schema_accepts_standard_forecaster_config() -> None:
```
mc4117: I think this is being tested somewhere else already?

training/tests/integration/test_sparse_dataset_time_indices.py
```diff
@@ -0,0 +1,92 @@
...
+
+def test_sparse_forecaster_metadata_keeps_per_dataset_sparse_windows(mocker: MockFixture) -> None:
+    cfg = load_sparse_multidatasets_config()
+    fake_readers = {
+        "meps_source": FakeDatasetReader(
+            dataset_name="meps_source",
+            frequency="1h",
+            start="2020-01-01T00:00",
+            stop="2020-01-03T00:00",
+            step_minutes=60,
```
mc4117: I didn't see step_minutes in the integration test? was just wondering how this differs from the frequency?

training/tests/integration/config/test_sparse_multidatasets.yaml
```diff
@@ -0,0 +1,61 @@
+# Sparse multi-dataset integration config used by metadata-oriented tests.
```
mc4117: I think there should be a config file in the main set of configs like the multi.yaml to show people how to do multiple timesteps and then the integration test can just use this config with some small overrides?

training/src/anemoi/training/data/data_reader.py
```diff
@@ -258,6 +285,169 @@ def tree(self, prefix: str = "") -> Tree:
         return tree
 
 
+class RelativeTimeReader:
```
mc4117, line 489: I don't see where this RelativeTimeReader is used, but I'm probably just missing it. Would you mind pointing me to the bit of the code where it's called? Would this now be the default reader?

mc4117:
Thanks for this! 
The logic looks good and I've made a couple of early comments mainly for my understanding. One thing is that I noticed that the commit history of this branch is a bit tangled with commits that are already on main. I don't know if it's possible to do a clean rebase so only the commits of this PR are part of the history?

models/src/anemoi/models/models/base.py
```diff
@@ -129,7 +134,9 @@ def _calculate_shapes_and_indices(self, data_indices: dict) -> None:
             self.output_dim[dataset_name] = self._calculate_output_dim(dataset_name)
 
     def _calculate_input_dim(self, dataset_name: str) -> int:
-        return self.n_step_input * self.num_input_channels[dataset_name] + self.node_attributes.attr_ndims[dataset_name]
+        return (
+            self.n_step_input[dataset_name] if isinstance(self.n_step_input, dict) else self.n_step_input
```
mc4117: you could extract this
  `self.n_step_input[dataset_name] if isinstance(self.n_step_input, dict) else self.n_step_input`
into a helper method on the base model (e.g. `self._get_n_step_input(dataset_name)`) to make it more robust to code changes later

training/src/anemoi/training/data/data_reader.py
```diff
@@ -258,6 +285,169 @@ def tree(self, prefix: str = "") -> Tree:
...
+                dates_ns[lower_indices[valid_lower]] - missing_dates_ns[valid_lower],
+            )
+            upper_distances[valid_upper] = np.abs(
+                dates_ns[upper_indices[valid_upper]] - missing_dates_ns[valid_upper],
+            )
+
+            use_upper = upper_distances < lower_distances
+            nearest_indices = np.where(use_upper, upper_indices, lower_indices)
+            nearest_distances = np.minimum(lower_distances, upper_distances)
+            tolerance_ns = max(1, (frequency_to_seconds(self.reader.frequency) * 1_000_000_000) // 2)
```
mc4117, line 597: I would maybe log a warning if this fallback is triggered as people might accidentally silently pick the wrong analysis cycle for datasets with large native frequency gaps. You could also make the tolerance configurable?

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -125,17 +212,115 @@ def get_offsets(self, mode: str | None = None) -> list[datetime.timedelta]:
...
+                msg = "`dataset_name` is required for sparse rollout input updates."
+                raise ValueError(msg)
+
+            requested_output_relative_times = self._requested_output_relative_times(
+                dataset_name,
+                rollout_step=rollout_step,
+            )
+            step_specs = []
+            for relative_time in [
+                int(relative_time + (rollout_step + 1) * self.num_output_timesteps)
```
mc4117, line 202: Could you add an assertion or bounds check here to confirm this doesn't index past the batch boundary?

training/tests/integration/config/test_mixed_frequency_multidatasets.yaml
OpheliaMiralles: Yes it's mostly the "model" timestep, e.g. when we rollout for time t, t+1 means t+(1 x timestep) in this setup. If there is only 1 one-hour input in the (multistep_input x timestep) window then only 1 frame gets loaded

training/src/anemoi/training/data/data_reader.py
OpheliaMiralles: multidataset as the default sample reader indeed

OpheliaMiralles:
> Thanks for this! The logic looks good and I've made a couple of early comments mainly for my understanding. One thing is that I noticed that the commit history of this branch is a bit tangled with commits that are already on main. I don't know if it's possible to do a clean rebase so only the commits of this PR are part of the history?

I'll try, indeed right now the history is very messy

models/src/anemoi/models/interface/__init__.py
```diff
@@ -59,7 +59,7 @@ def __init__(
         self,
         *,
         config: DictConfig,
-        n_step_input: int,
+        n_step_input: int | dict[str, int],
```
dietervdb-meteo, line 62: Why assymetry between input and output? Also for output we can allow per dataset specification?

models/src/anemoi/models/models/base.py
```diff
@@ -44,7 +44,7 @@ def __init__(
         model_config: DictConfig,
         data_indices: dict,
         statistics: dict,
-        n_step_input: int,
+        n_step_input: int | dict[str, int],
         n_step_output: int,
```
dietervdb-meteo: idem: lets allow per dataset for output as well.

models/src/anemoi/models/interface/__init__.py
dietervdb-meteo: Also, would it be safer to always force it to be a dict. And simply give the dataset the name `data` for a single dataset model? [I believe this is already implemented this way in various other places?]

training/src/anemoi/training/tasks/base.py
```diff
@@ -123,6 +177,77 @@ def get_batch_output_indices(self, **kwargs) -> list[int]:
...
+        if not candidate_times:
+            msg = (
+                f"Dataset '{dataset_name}' has no forcing/boundary time at or before relative time "
+                f"{relative_time}. Available times: {available_times}"
+            )
+            raise ValueError(msg)
+        sampled_time = candidate_times[-1]
+
+        LOGGER.info(
+            "Sparse rollout dataset=%s requested_time=%s sampled_time=%s",
```
dietervdb-meteo: Anything about rollout should not be part of the base task. It is forecaster specific.

dietervdb-meteo (changes requested):
I will need to look at the data handling part in more detail. A lot of stuff has been added there... Here are some first comments.

training/tests/integration/test_mixed_frequency_dataset_time_indices.py
VeraChristina, line 1: Can we move this test and the config file above to the unit tests? They mock data and only seem to test if time windows are stored in the metadata as expected

training/src/anemoi/training/tasks/base.py
```diff
@@ -202,9 +362,56 @@ def fill_metadata(self, md_dict: dict) -> None:
             "timestep": timestep,  # backwards compatibility with inference
         }
 
-        dataset_names = md_dict["metadata_inference"]["dataset_names"]
         for dataset_name in dataset_names:
-            md_dict["metadata_inference"][dataset_name]["timesteps"] = timesteps
```
VeraChristina, line 220: Maybe I am missing something but I don't think this should be removed. The purpose of the method `fill_metadata` is to populate the md_dict with metadata to be stored in the checkpoint for use in inference. Can you please restore the metadata we store in the checkpoint and move additional timestep handling to a different method?

VeraChristina: Also, if this PR needs changes in the structure of the metadata under `metadata_inference`, please check regarding implications in inference. Ideally, keep them backwards compatible but that's not always possible.

VeraChristina (commented):
A couple of comments below. The main concern is regarding metadata stored in the checkpoint for inference

OpheliaMiralles:
@copilot+claude-opus-4.6 resolve the merge conflicts in this pull request

OpheliaMiralles:
Can someone review this please I am tired of merging main and the hundred conflicts, this is only support for different frequencies in a multi dataset setting

PR #1075, opened 2026-04-27
fix: Changed FFT2D low-pass filter from enabled to disabled by default

evenmn:
## Description
We have observed that the FFT2D low-pass filter implemented in Anemoi for the FFT2D reduces the quality of the generated fields. Disabling the low-pass filter improves the fine-scale structures of the fields, in particular visible for precipitation and wind. The default setting should therefore be running without the filter.

In the future, we can discuss if it's useful to include the filter code at all, but for now, let's keep it.

##  Additional notes ##


havardhhaugen:
Looks good, I will bring this up at ATS just to flag the change since it gives us significantly better results.

havardhhaugen (approved):

anaprietonem:
@havardhhaugen did you address this at ATS last week? shall we update the label and merge it?

OpheliaMiralles (approved):

PR #1077, opened 2026-04-29
fix: poetry issue

NatalieZelenka:
## Description
Closes #1023

Removes a dependency to a extra from the main dependencies list, which will allow poetry to work for some use cases. But it does make it slightly less clear which dependencies are related to mlflow.

#1023

##  Additional notes ##
Through discussion with @gmertes and @anaprietonem, we decided not to move the mlflow dependencies to be optional dependencies since this would change the behaviour for existing users. 


anaprietonem (approved):
LGTM and integration test passed fine, feel free to merge Natalie, and thanks for fixing it!

PR #1078, opened 2026-04-29
feat(graphs): Add option to disable node scaling by degree

matschreiner:
# Add option to disable node scaling by degree
                                                                    
In the 3D graph visualization, nodes are scaled proportionally to
their number of incoming connections. For dense graphs this helps 
identify high-degree nodes, but it can obscure the spatial
distribution of the node set itself. For example near the poles.                              
                                                      
This PR adds a "Scale nodes by degree" checkbox (checked by       
default) to the Global Controls panel. When unchecked, all nodes
are rendered at uniform size regardless of their degree. The      
change takes effect immediately without reloading.  

Changes:                                                          
- interactive_3d.html.jinja: new checkbox in the controls panel;
computeNodeScale returns uniform scale when the checkbox is off;  
onchange listener triggers an immediate re-render.  

image

image

image

### DIFF
--- a/graphs/src/anemoi/graphs/plotting/interactive_3d.html.jinja
+++ b/graphs/src/anemoi/graphs/plotting/interactive_3d.html.jinja
@@ -183,6 +183,13 @@ nor does it submit to any jurisdiction. #}
                     <input id="scaleFactor" type="number" min="0.5" max="3.0" step="0.1" value="1.0" />
                 </div>
             </div>
+
+            <div class="control-row">
+                <span>Scale nodes by degree</span>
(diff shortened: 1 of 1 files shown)


frazane:
Fully support this, especially when working in high-res the scaling by degree is a pain. I wonder if we could make it a tunable factor instead of a on/off button (or both)? Just brainstorming, I am also happy with the current solution.

matschreiner:
@frazane , ok if you are happy with this, then maybe we can implement the tunable factor in a future PR and merge this as is?

frazane (approved):

PR #1081, opened 2026-04-29
feat(callbacks)!: remove RolloutEvalEns

VeraChristina:
## Description
This PR removes the RolloutEvalEns callback, which apart from dead code was verbatim the same as its base class RolloutEval.
If we don't want the breaking change in the configs, we could introduce an alias.

### DIFF
--- a/training/tests/unit/diagnostics/test_callbacks.py
+++ b/training/tests/unit/diagnostics/test_callbacks.py
@@ -20,7 +20,6 @@ import yaml
 from anemoi.training.diagnostics.callbacks import _get_progress_bar_callback
 from anemoi.training.diagnostics.callbacks import get_callbacks
 from anemoi.training.diagnostics.callbacks.evaluation import RolloutEval
-from anemoi.training.diagnostics.callbacks.evaluation import RolloutEvalEns
 
 NUM_FIXED_CALLBACKS = 3  # ParentUUIDCallback, CheckVariableOrder, RegisterMigrations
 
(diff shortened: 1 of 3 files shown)


mchantry:
@VeraChristina is it clear for how long it has been dead code?
I don't think we need an alias for something that has been dead.

VeraChristina:
@mchantry I meant: it's verbatim the same as the base class except for a few snippets of dead code. The RolloutEvalEns has been working (it's just literally the same as RolloutEval). -- I agree though, that we probably don't need an alias.

training/tests/unit/diagnostics/test_callbacks.py
anaprietonem: Two cents, shall we also have a test where we check RolloutEval with a batch that has more than one ensemble member? to check the 'ensemble side' is properly covered?

anaprietonem:
The PR LGTM, @jakob-schloer I believe you might have used that callback before or use it at the moment, so please if you could also have a quick look, that would be great!

training/tests/unit/diagnostics/test_callbacks.py
VeraChristina: good point! Have parametrized the test to test batch shapes with n_ens=1 and n_ens=3

jakob-schloer:
> The PR LGTM, @jakob-schloer I believe you might have used that callback before or use it at the moment, so please if you could also have a quick look, that would be great!

Looks good. Thanks for cleaning up dead code @VeraChristina.

anaprietonem (approved):
@VeraChristina thanks for the nice work. Feel free to merge

Issue #1083, opened 2026-04-30
Pydantic schema used by anemoi-training config validate is missing the log_hyperparams field in MlflowSchema


rockydevnet:
### What happened?

The Pydantic schema used by anemoi-training config validate is missing the log_hyperparams field in MlflowSchema such that if you try to set log_hyperparams: False, config validation will fail.

### What are the steps to reproduce the bug?

With the mlflow logger enabled using the `anemoi.training.diagnostics.mlflow.azureml.AnemoiAzureMLflowLogger`, and `log_hyperparams: False`, when you run training (`anemoi-training train  --config-name config`), or config validation ('anemoi-training config validate --config-name config'), the validation of the AnemoiAzureMLflowLogger configuration fails.

For example, this is the `mlflow.yaml` configuration I am using:

```yml
mlflow:
  enabled: True
  _target_: anemoi.training.diagnostics.mlflow.azureml.AnemoiAzureMLflowLogger
  offline: False
  authentication: True
  tracking_uri: null
  experiment_name: 'hackathon-test'
  project_name: 'Anemoi'
  system: False
  terminal: False
  run_name: null # If set to null, the run name will be the a random UUID
  on_resume_create_child: True
  expand_hyperparams: # Which keys in hyperparams to expand
    - config
  log_hyperparams: False
  http_max_retries: 3
  max_params_length: 2000
  save_dir: ${system.output.logs.mlflow}
  log_model: False
  identity: "default"
interval: 100 # passed to trainer.log_every_n_steps
```

If you set `config_validation: False`, `anemoi-training train  --config-name config` runs successfully, indicating that this is a validation issue.

### Version

anemoi-training[azure]==0.11.0

### Platform (OS and architecture)

ubuntu:22.04

### Relevant log output

```shell
Error executing job with overrides: ['system.input.dataset=/mnt/azureml/cr/j/6a69a705598648eebaa13ccde26edf97/cap/data-capability/wd/INPUT_data_dir/output_dataset.zarr', 'system.input.graph=/mnt/azureml/cr/j/6a69a705598648eebaa13ccde26edf97/cap/data-capability/wd/output_dir/graph/graph.pt', 'system.output.root=/mnt/azureml/cr/j/6a69a705598648eebaa13ccde26edf97/cap/data-capability/wd/output_dir']
Traceback (most recent call last):
  File "/opt/conda/envs/core/lib/python3.11/site-packages/anemoi/utils/cli.py", line 264, in cli_main
[...]

PR #1084, opened 2026-04-30
fix: add missing parameters for logging hyperparameters in MlflowSchema

rockydevnet:
## Description
Fix bug #1083: The Pydantic schema: `MlflowSchema` is missing two parameters: `log_hyperparams` and `prefix`. These are available and work at runtime, but because they are missing from the schema, config validation incorrectly fails indicating that they are illegal extra inputs.

Bug fix for #1083

#1083

##  Additional notes ##
I have added tests for both implementations of the `BaseAnemoiMLflowLogger`, i.e. for `AnemoiMLflowLogger` and for `AnemoiAzureMLflowLogger` to assert that the constructor parameters all match the schema so that we will catch any other future missing parameters.

### DIFF
--- a/training/src/anemoi/training/config/diagnostics/log/mlflow.yaml
+++ b/training/src/anemoi/training/config/diagnostics/log/mlflow.yaml
@@ -9,6 +9,8 @@ mlflow:
   system: False
   terminal: True
   run_name: null # If set to null, the run name will be the a random UUID
+  prefix: '' # Prefix for metric keys logged to MLflow
+  log_hyperparams: True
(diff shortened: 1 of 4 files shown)


HCookie (changes requested):
Hi,
Thanks for fixing this.
Can you add some entries into the config file for `mlflow` to showcase these options? Once we do that, then I can will merge this

HCookie (approved):

HCookie:
Thank you for addressing my comments, it is now merged. Thanks for your contribution

PR #1085, opened 2026-05-01
feat(models,training): dictionarify encoder/decoder/residual and extract into Hydra defaults

HCookie:
## Summary

> [!WARNING]
> Still to be tested and models run

This PR makes encoder, decoder, and residual configs dictionary-based (keyed by dataset name) in the model code, and extracts the duplicated config blocks into reusable Hydra defaults.

### Dictionarify encoder/decoder/residual (models)

- Encoder, decoder, and residual are now stored as `torch.nn.ModuleDict` keyed by dataset name, allowing per-dataset configuration
- Adds `_get_nested_configuration()` to `BaseGraphModel` which resolves either a flat config (single `_target_`) or a nested dict of per-dataset configs
- Updates all model classes (`AnemoiModelEncProcDec`, diffusion, ensemble, hierarchical, autoencoder) to use the new nested lookup
- Moves encoder/decoder/processor schemas into their own modules
- Updates the GLU/Transformer assertion in `train.py` to iterate over potentially multiple encoder/decoder configs

### Extract configs into Hydra defaults (training)

- Extracts duplicated encoder, decoder, and residual blocks from all model YAML files into reusable configs under `model/encoder/`, `model/decoder/`, and `model/residual/`
- All 11 model configs now reference these via Hydra `defaults:` list instead of inlining the full block
- Where configs differ, separate named defaults preserve the exact resolved config

### Validation command

- Adds `--open` and `--viewer` flags to `anemoi-training config validate` to open the resolved config in an editor

## New default configs

**Encoder:** gnn, gt16, gt16_qknorm, transformer, pointwise
**Decoder:** gnn, gt16, gt16_qknorm, gt16_qknorm_zeroinit, transformer, pointwise
**Residual:** skip, truncated

## Model config mapping

| Model config | Encoder | Decoder | Residual |
|---|---|---|---|
| graphtransformer | gt16 | gt16 | skip |
| gnn | gnn | gnn | skip |
| transformer | gt16 | gt16 | skip |
| graphtransformer_ens | gt16 | gt16 | skip |
| transformer_ens | gt16 | gt16 | skip |
| graphtransformer_diffusion | gt16_qknorm | gt16_qknorm | skip |
| graphtransformer_diffusiontend | gt16_qknorm | gt16_qknorm | skip |
| transformer_diffusion | gt16_qknorm | gt16_qknorm_zeroinit | skip |
| transformer_diffusiontend | gt16_qknorm | gt16_qknorm_zeroinit | skip |
| transformer_transformermapper | transformer | transformer | skip |
| point_wise | pointwise | pointwise | skip |

----
📚 Documentation preview 📚: https://anemoi-training--1085.org.readthedocs.build/en/1085/

----
📚 Documentation preview 📚: https://anemoi-graphs--1085.org.readthedocs.build/en/1085/

----
📚 Documentation preview 📚: https://anemoi-models--1085.org.readthedocs.build/en/1085/

### DIFF
--- a/models/src/anemoi/models/models/hierarchical.py
+++ b/models/src/anemoi/models/models/hierarchical.py
@@ -37,16 +37,18 @@ class AnemoiModelEncProcDecHierarchical(AnemoiModelEncProcDec):
         self.encoder_graph_provider = nn.ModuleDict()
         self.encoder = torch.nn.ModuleDict()
         for dataset_name in self.dataset_names:
+            sub_encoder_config = self._get_nested_configuration(model_config.model.encoder, dataset_name)
+
             self.encoder_graph_provider[dataset_name] = create_graph_provider(
                 graph=self._graph_data[(dataset_name, "to", self._graph_name_hidden[0])],
-                edge_attributes=model_config.model.encoder.get("sub_graph_edge_attributes"),
+                edge_attributes=sub_encoder_config.get("sub_graph_edge_attributes"),
                 src_size=self.node_attributes.num_nodes[dataset_name],
                 dst_size=self.node_attributes.num_nodes[self._graph_name_hidden[0]],
-                trainable_size=model_config.model.encoder.get("trainable_size", 0),
+                trainable_size=sub_encoder_config.get("trainable_size", 0),
             )
 
             self.encoder[dataset_name] = instantiate(
-                model_config.model.encoder,
+                sub_encoder_config,
                 _recursive_=False,  # Avoids instantiation of layer_kernels here
                 in_channels_src=self.input_dim[dataset_name],
                 in_channels_dst=self.input_dim_latent,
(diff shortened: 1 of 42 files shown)


JPXKQX:
Hi Harrison,
Thanks for opening this pull request. There are a few things that would be good to consider when refactoring this part of the configuration. 
- I think it would be useful to add an `optional` field for both the encoder and the decoder. This should initially be set to `False`, but could be set to `True` if we want the model to run with or without this. Currently, we are not checking this, so if a model is trained with multiple encoders and we then pass only one during inference, the model will produce a forecast.
- Secondly, with observations, we aim to support encoders/decoders that take one or multiple datasets. In that case, we would break the 1:1 relationship between datasets and encoders/decoders.
One possible proposal for the unrolled config would be:
```
encoder:
   0: # or another name
       datasets: [ "data" ] # only 1 dataset is supported for now
       optional: False
       mapper:
           ...
```
What's your view on these defaults (encoder@encoder.data) with a structure like this?

HCookie:
@JPXKQX I feel as if we run the risk of nesting things too far deep if we do that. 
I've proposed some changes in #1092 regarding your questions and would be keen for your review

models/src/anemoi/models/models/hierarchical.py
```diff
@@ -123,16 +125,18 @@ def _build_networks(self, model_config):
             src_nodes_name = self._graph_name_hidden[i]
             dst_nodes_name = self._graph_name_hidden[i + 1]
 
+            sub_downscale_config = self._get_nested_configuration(model_config.model.encoder, src_nodes_name)
```
ssmmnn11, line 128: does this work? I think _get_nested_configuration needs dataset name not src_nodes_name.

training/src/anemoi/training/config/model/residual/truncated.yaml
```diff
@@ -0,0 +1,2 @@
+_target_: anemoi.models.layers.residual.TruncatedConnection
+# TODO: Add defaults here
```
ssmmnn11, line 2: please add defaults here :-)

training/src/anemoi/training/train/train.py
```diff
@@ -261,19 +261,34 @@ def _validate_transfer_learning_datasets(
     @cached_property
     def model(self) -> pl.LightningModule:
         """Provide the model instance."""
+        encoder_config = (
+            [self.config.model.encoder]
+            if "_target_" in self.config.model.encoder
```
ssmmnn11, line 266: should we have a helper here? -> not all callers need to know that _target_ means shared config?

ssmmnn11 (changes requested):
nice!

PR #1086, opened 2026-05-01
feat(models,residual): new drop argument for residual connections

JPXKQX:
## Description

This pull request adds support for selectively excluding specific prognostic variables from the residual (skip) connection in Anemoi models. Users can now specify a list of variable names to drop from the skip connection using a new `drop` argument in both configuration files and model schemas. The changes also update configuration files to expose this new option.

----
📚 Documentation preview 📚: https://anemoi-training--1086.org.readthedocs.build/en/1086/

----
📚 Documentation preview 📚: https://anemoi-graphs--1086.org.readthedocs.build/en/1086/

----
📚 Documentation preview 📚: https://anemoi-models--1086.org.readthedocs.build/en/1086/


models/src/anemoi/models/layers/residual.py
```diff
@@ -44,8 +64,16 @@ def forward(
         """
         pass
 
+    def _drop_variables(self, x: torch.Tensor) -> torch.Tensor:
+        """Zero out specified prognostic variables in the input tensor."""
+        if len(self.drop_indices):
+            x[..., self.drop_indices] = 0.0  # Zero out the prognostic variables specified in drop list
```
ssmmnn11: depending on how steps looks, this x is a view or a copy. If it is a view, then you mutate x here and this breaks the model. 

```
  def _drop_variables(self, x: torch.Tensor) -> torch.Tensor:
      if not self.drop_indices:
          return x

      x = x.clone()
      x[..., self.drop_indices] = 0.0
      return x
```

models/src/anemoi/models/layers/residual.py
```diff
@@ -17,18 +18,37 @@
...
+    def __init__(
+        self,
+        drop: list[str] = [],
+        data_indices: IndexCollection | None = None,
+    ) -> None:
         super().__init__()
+        self.drop_names = drop
+
+        model_data_indices = data_indices.model.input
+        assert all(
```
ssmmnn11: breaks if data_indices is None as advertised by the type hint.

models/src/anemoi/models/layers/residual.py
```diff
@@ -17,18 +18,37 @@
...
+            v in model_data_indices.includes for v in self.drop_names
```
ssmmnn11: I think this includes forcing + prognostic so accepts dropping forcing variables?

ssmmnn11 (changes requested):
thanks for the contribution!

anaprietonem:
To clarify, discussing with @JPXKQX the changes here are still a WIP. @NRaoult, Nina if you'd have some time could you test this could and report if you find the feature useful for your work in snow?

models/src/anemoi/models/layers/residual.py
JPXKQX: Fixed!

models/src/anemoi/models/layers/residual.py
JPXKQX: Fixed!

models/src/anemoi/models/layers/residual.py
JPXKQX: Fixed!

PR #1087, opened 2026-05-02
feat: projections for spectral losses

frazane:
This PR adds optional sparse projection (or regridding) as a pre-processing step to all spectral losses, enabling their use when for instance the native data grid doesn't map cleanly to the transform's expected layout.
  
  - training/losses/spectral.py: Add `subgrid` and `projection_config` parameters to the `SpectralLoss` base class; apply selection of a subgrid (e.g. as `output_mask` for LAM, or as an optional slice) and optional sparse projection before any spectral transform (FFT2D, DCT2D, SHT variants) via a new `_prepare_spatial` step.
  - models/layers/spectral_transforms.py: Remove `nodes_slice` from FFT2D (now handled upstream as `subgrid`); extract `prepare_for_fft` helper from forward.
  - models/layers/graph_provider.py: Add `ProjectionGraphProvider.from_config` factory supporting three construction modes: file-based matrix, existing graph edges, or on-the-fly KNN to a target grid; add `normalize_projection_edges_name` helper.
  - graphs/builders.py: Add `build_node_to_node_projection_subgraph` to build a KNN subgraph between arbitrary source/target node pairs.
  - training/losses/multiscale.py: Replace inline `_apply_projector` with the shared `SparseProjector.apply_with_provider`.
  - training/schemas/training.py: Schema updates for the new loss parameters.
  - Tests: New unit tests for all spectral losses with projection, and for graph provider construction modes.

---
Note: originally this PR was only targeting the FFT2D transform but was later adapted to provide a more general projection mechanism for spectral losses, following suggestions from @anaprietonem and @ssmmnn11 (thank you both!). Some additional refactoring was made to other parts of the codebase.

---

----
📚 Documentation preview 📚: https://anemoi-training--1087.org.readthedocs.build/en/1087/

----
📚 Documentation preview 📚: https://anemoi-graphs--1087.org.readthedocs.build/en/1087/

----
📚 Documentation preview 📚: https://anemoi-models--1087.org.readthedocs.build/en/1087/


training/tests/unit/losses/test_loss_function.py
```diff
@@ -518,6 +519,41 @@ def test_spectral_crps_fft_and_dct() -> None:
...
+    # Create a mock sparse matrix (e.g., identity)
+    sparse_mat = eye(grid, format="csr")
+
+    # Mock scipy.sparse.load_npz to return our sparse matrix
+    mocker.patch("scipy.sparse.load_npz", return_value=sparse_mat)
+
+    loss = get_loss_function(
+        DictConfig(
+            {
+                "_target_": "anemoi.training.losses.spectral.SpectralCRPSLoss",
```
anaprietonem, line 591: according to the docs of SpectralCRPSLoss 

```
    Works with:
      - FFT2D
      - DCT2D
      - Reduced SHT
      - Octahedral SHT
    """
``` 

but the changes for the projection matrix are just defined for the FFT2D, so bet because of the kwargs of the SpectraltTransform this won't fail, but it should properly be raised or cought that projection matrix won't affect the other cases or have you thought about moving this functionality to the base spectral transform?

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -79,6 +87,8 @@ def __init__(
         patch_padding: bool
             If True, allow non-divisible `(y_dim, x_dim)` by zero-padding on the
             bottom/right edges before patch extraction.
+        projection_matrix: str | None
```
anaprietonem: Francesco, right now for the ResidualTruncation and the Multiscale Loss which also take this sparse matrices there is the option to pass a on-the-fly config as an alternative to loading this matrices from this disk. See for example the docs for multiscale https://github.com/ecmwf/anemoi-core/blob/a5eeaef1615906974f684e39197f17be7fe9afb4/training/docs/user-guide/models.rst#L209. it would be nice to do the same for this case where you can either pass a npz file or build the projection matrix on the fly,

training/tests/unit/losses/test_loss_function.py
```diff
@@ -518,6 +519,41 @@ def test_spectral_crps_fft_and_dct() -> None:
         assert out_total.numel() == 1, f"{transform}: scalar CRPS expected"
 
 
+def test_spectral_crps_fft2d_projection(mocker: MockerFixture) -> None:
```
anaprietonem, line 575: Could we have also a test with the nodes_slice with the projection matrix?

models/src/anemoi/models/layers/spectral_transforms.py
anaprietonem: Similarly since in both cases you're taking the ProjectionGraphProvider and the SparseProjector, could it be worth abstracting this to its own class than then get's used across the different loss functions?

training/tests/unit/losses/test_loss_function.py
frazane: > have you thought about moving this functionality to the base spectral transform?

I have but my understanding is that it's not a necessity to project the data on a different grid in order to use the other transforms. On the other hand, if one has a non-regular grid they _have to_ reproject in order to use the FFT2D. Or am I missing something? 

That's why I wanted to keep the scope of this PR narrow, but if you think this is the right time and it would benefit other use cases to implement a more general solution, I am happy to do it. What do you think?

models/src/anemoi/models/layers/spectral_transforms.py
frazane: I think the goal here is to project / "regrid" to a predetermined grid so I don't see the use for computing it on the fly (or do it for multiple scales). Could you explain? Were you envisioning applying the spectral losses on different levels of smoothing? Again, as above, this would change the scope of the PR (which is for now about regridding in order to use FFT2D).

OpheliaMiralles (approved):
Looks good to me, and I think the scope defined in the PR is clear enough. I also don't see an immediate need to generalise it to all spectral losses or to other losses

models/src/anemoi/models/layers/spectral_transforms.py
ssmmnn11: I think the point is that the matrix is then build from the graph config and all the information to train the model is carried via the config instead of having a file you need to bring along.

frazane: My point is that in many cases people bring regridding matrices computed from external software. So we don't need to build the matrix from the graph config at all. No?

ssmmnn11: some ideas here #1108

frazane: Thanks @ssmmnn11! I was also working on reusing the `projection_config`, but you were a lot faster. Also, I was still implementing it at the FFT2D transform level rather than for all spectral losses - your solution is more general. Happy to test your branch and merge, let's keep in touch!

frazane:
Update: the projection and the node slice do not currently support sharding on the GRID dimension. I am looking into this now and will push something as soon as I can get it to work on a small test run.

ssmmnn11:
> Update: the projection and the node slice do not currently support sharding on the GRID dimension. I am looking into this now and will push something as soon as I can get it to work on a small test run.

did we have it in #1108?

frazane:
> > Update: the projection and the node slice do not currently support sharding on the GRID dimension. I am looking into this now and will push something as soon as I can get it to work on a small test run.
> 
> did we not have it in #1108?

Yes, but I am stupid. :)

ssmmnn11:
> > > Update: the projection and the node slice do not currently support sharding on the GRID dimension. I am looking into this now and will push something as soon as I can get it to work on a small test run.
> > 
> > 
> > did we not have it in #1108?
> 
> Yes, but I am stupid. :)

😆 I am pretty sure that is not true.

frazane:
@ssmmnn11 the upside of not implementing the sharding inside the loss is that I might have found a more fundamental bug...I noticed that even though my loss had a `supports_sharding: False` attribute, somehow the tensors were arriving sharded anyway to the forward pass.

Turns out in `ensemble.py` we override `compute_dataset_loss_metrics` but never call `_prepare_tensors_for_loss` https://github.com/ecmwf/anemoi-core/blob/390ce34c1fbf89bf37cc8733cc500fe2bc7237ee/training/src/anemoi/training/train/methods/ensemble.py#L171

so we never actually reach the logic from https://github.com/ecmwf/anemoi-core/blob/390ce34c1fbf89bf37cc8733cc500fe2bc7237ee/training/src/anemoi/training/train/methods/base.py#L576-L582

that gathers tensors if the loss does not support sharding. I pushed a fix in cc4f37f. With this, tensors are automatically gathered just before the loss is computed, and the projection works.

ssmmnn11:
Very good. We just need to carefully test that this still works with all different combinations (batch sharded / ensemble dim / losses that support sharding) etc.

frazane:
@ssmmnn11 tested a few configurations end-to-end:

* deterministic model, mse loss with model sharding: OK
* deterministic model, spectral l2 loss with projection and without model sharding: OK
* deterministic model, spectral l2 loss with projection and with model sharding: OK
* ensemble model, spectral crps with projection and with model sharding: OK
* ensemble model, combined loss (crps + spectral crps with projection) with model sharding: there was a bug, fixed by a7317d8 and now OK

frazane:
Integration tests OK https://github.com/ecmwf/anemoi-core/actions/runs/27025868259

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -38,3 +38,24 @@ def forward(self, x: torch.Tensor, projection_matrix: torch.Tensor) -> torch.Ten
             for i in range(x.shape[0]):
                 out.append(torch.sparse.mm(projection_matrix, x[i, ...]))
         return torch.stack(out)
+
+    def apply_with_provider(self, batch: torch.Tensor, provider: object) -> torch.Tensor:
```
ssmmnn11: Can we think of a better name here? the logic is usually 

provider.get_edges -> sparse_projection

training/src/anemoi/training/losses/spectral.py
```diff
@@ -106,6 +128,15 @@ def __init__(
         # Enforce loss to be calculated on full grids.
         self.supports_sharding = False
 
+        self.nodes_slice = slice(*(nodes_slice or (0, None)))
```
ssmmnn11: will it work with sharding?

training/src/anemoi/training/losses/spectral.py
```diff
@@ -126,8 +157,21 @@ def __init__(
             msg = f"Unknown transform type: {transform}"
             raise ValueError(msg)
 
+    def _apply_nodes_slice(self, x: torch.Tensor) -> torch.Tensor:
+        # Slice the grid dim as a view, avoiding an explicit index-tensor allocation.
+        index = [slice(None)] * x.ndim
+        index[TensorDim.GRID] = self.nodes_slice
+        return x[tuple(index)]
+
+    def _prepare_spatial(self, x: torch.Tensor) -> torch.Tensor:
```
ssmmnn11: can we think of a better name?

training/src/anemoi/training/losses/spectral.py
ssmmnn11: here _prepare_spatial() is not called, will this cause issues with nodes_slice, projection_config?

training/src/anemoi/training/train/methods/base.py
```diff
@@ -1002,9 +1005,12 @@ def calculate_val_metrics(
                 if target_layout is not None:
                     metric_kwargs["target_layout"] = target_layout
                 if getattr(metric, "needs_shard_layout_info", False):
+                    # grid_shard_sizes must stay consistent with grid_shard_slice: if the tensors
+                    # were gathered to the full grid (grid_shard_slice is None), the metric must be
+                    # told it is not sharded, otherwise it would re-shard an already-full tensor.
                     metric_kwargs.update(
                         grid_dim=self.grid_dim,
-                        grid_shard_sizes=self.grid_shard_sizes[dataset_name],
+                        grid_shard_sizes=self.grid_shard_sizes[dataset_name] if grid_shard_slice is not None else None,
```
ssmmnn11, line 1028: I think we also need to check PerTimestepMetrics, training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -628,3 +649,94 @@ def get_edges(
...
+        except ImportError:
+            pass
+        if not isinstance(config, dict):
+            config = dict(config)
+        if not config:
+            return None
+
+        has_matrix = "matrix_path" in config and config["matrix_path"] is not None
+        has_edges = "edges_name" in config and config["edges_name"] is not None
+
```
ssmmnn11, line 680: row_normalize is dropped when using a file?

ssmmnn11 (dismissed):
Nice! Thanks for your contribution! 

Some suggestions:
- could you say if we still want to support sharding, or if this should be a follow up
- some docstrings are a bit awkward ("This is abc, without def and no xyz"), maybe we can clean them up a bit to make them more human friendly - just state what it does :-)

training/src/anemoi/training/losses/spectral.py
frazane: It will not. But since we don't support sharding for spectral losses, I guess it's okay?

training/src/anemoi/training/losses/spectral.py
frazane: Good spot, forgot to add it when merging main.

frazane:
> could you say if we still want to support sharding, or if this should be a follow up

Do you mean to support sharding for the projections or for the spectral losses in general? In any case, I would prefer to leave this to a follow up, though I am not even sure _how_ we would do it.

models/src/anemoi/models/layers/sparse_projector.py
frazane: What if we refactor the code so the provider is owned by the projector, instead of passing it to it here?

models/src/anemoi/models/layers/graph_provider.py
frazane: I assumed when someone provides a regridding matrix it's not needed (we don't) but this might not always be true, thanks for pointing out

training/src/anemoi/training/train/methods/base.py
frazane: Should be addressed by dcbee5e but I will run some end-to-end configs with some additional validation metrics to check

frazane: Can confirm it works!

dietervdb-meteo:
@frazane , I changed `nodes_slice` to `subgrid` in the PR description per our discussion.

ssmmnn11:
> > could you say if we still want to support sharding, or if this should be a follow up
> 
> Do you mean to support sharding for the projections or for the spectral losses in general? In any case, I would prefer to leave this to a follow up, though I am not even sure _how_ we would do it.

Yes, this was already implemented in the suggestions I provided. But no worries. The best path forward is probably that this goes in without sharding support, and then we implement the sharding once this is in.

frazane:
> The best path forward is probably that this goes in without sharding support, and then we implement the sharding once this is in.

Agreed. It would be great for us to have this merged rather soon because without this we cannot use spectral losses at all.

frazane:
Results from some test runs below, both after 5 epochs. The first is with pointwise CRPS + SpectralCRPS computed on a cartesian domain after the projection, the second one is with pointwise CPRPS only. 

image

image

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -2,39 +2,30 @@
...
 class SparseProjector(torch.nn.Module):
-    """Applies sparse projection matrix to input tensors.
+    """Applies a sparse projection matrix to input tensors.
 
-    Stateless projector that receives matrix as input to forward().
+    Stateless: the matrix is passed to :meth:`forward`, not stored.
     """
 
     def __init__(self, autocast: bool = False) -> None:
-        """Initialize SparseProjector.
```
anaprietonem, line 20: minor comment but why removing the docstrings here?

graphs/src/anemoi/graphs/builders.py
```diff
@@ -86,6 +86,65 @@ def _expand_smoother_config(cfg: dict | Any) -> dict[str, dict]:
...
+    num_nearest_neighbours = config["num_nearest_neighbours"]
+    sigma = config["sigma"]
+
+    subgraph = HeteroData()
+    subgraph[source_node_name].x = graph_data[source_node_name].x
+    subgraph[source_node_name].num_nodes = graph_data[source_node_name].num_nodes
+
+    graph_config: dict = {
+        "nodes": {},
+        "edges": [_knn_edge_cfg(source_node_name, target_node_name, num_nearest_neighbours, sigma)],
```
anaprietonem, line 125: We can update this at later stage, but just curious and related to what we were discussing with Bastien, do you expect this to require more flexibility in the future to use other types of edge builders?

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
```diff
@@ -88,7 +88,16 @@ def _eval_per_timestep(
         for dataset_name, y_pred in y_preds.items():
             y = y_targets[dataset_name]
             n_timesteps = y.shape[TensorDim.TIME]
-            grid_shard_slice = pl_module.grid_shard_slice.get(dataset_name)
+
+            # Gather the grid up front when any loss/metric does not support sharding, so non-sharding
```
mc4117, line 92: was just wondering if you tested this without this modification? I'm a bit confused as to why this is still needed?

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -74,6 +74,27 @@ def create_graph_provider(
         return NoOpGraphProvider()
 
 
+def normalize_projection_edges_name(
```
anaprietonem, line 77: why do we need this? Since we don't have to do that in the sparse projector https://github.com/ecmwf/anemoi-core/blob/main/graphs/src/anemoi/graphs/export.py wondering if we can keep both aligned, but I might be missing something

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
frazane: Yes, it broke my code with a shape mismatch. The tensors were being passed still sharded to a metric that does not support sharding.

models/src/anemoi/models/layers/graph_provider.py
frazane: It was in https://github.com/ecmwf/anemoi-core/pull/1108/changes/400890291f6d167b60b9f2d032241a31c6c67a83. I'll have a look.

graphs/src/anemoi/graphs/builders.py
frazane: Good question! I think here since the goal is just "map the nodes from one partition to another" (I didn't really have truncation / smoothing, etc. in mind) KNN is a natural choice, but I agree that it could be nice to extend this. If that's okay for you, I'd keep this for later. Generally speaking, I think we could come up with a coherent abstraction for "projections using sparse matrices": right now we have different flavours and implementations of the same thing, especially in how we specify the configs. It would be great if we could have a dedicated `projection` key, with a fixed schema (or more than one, resolved based on some discriminative key), that can be reused across all configs (residual connection truncation, multi-scale losses, regridding before losses or plotting, etc.).

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
frazane: Could you point me to where _without this change_ the code gathers tensors before a metric that does not support sharding? Then we will be confused together! (I am also looking 👀 )

models/src/anemoi/models/layers/graph_provider.py
frazane: Will simplify and only allow the conventional `(src, "to", dst)` shape!

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
frazane: Aha, I _think_ what's happening is that normally we go through [`compute_dataset_loss_metrics`](https://github.com/ecmwf/anemoi-core/blob/1a99ae09313f3324608ab4918e0f88505c789e08/training/src/anemoi/training/train/methods/base.py#L692), which calls `_prepare_tensors_for_loss`, and then `_compute_metrics` which calls `calculate_val_metrics`. 

With your changes we call `calculate_val_metrics` directly so skip gathering the tensors, which is why this diff fixes it.

models/src/anemoi/models/layers/graph_provider.py
anaprietonem: yes that sounds good to me! thanks Francesco

training/docs/modules/losses.rst
```diff
@@ -255,6 +255,97 @@ Supported transforms include:
    ``use_graphed_rfft=True`` in the section of the config file corresponding to your spectral loss. This can provide
    significant speedups, but may not be supported on all devices and can have higher memory usage.
 
+.. note::
+
+   Before the transform is applied the grid can be subset to a subgrid by setting the optional ``subgrid`` argument.
```
anaprietonem, line 260: Could we have below in the docs one example of configuration that includes the subgrid either as a tuple or output_mask?

training/src/anemoi/training/losses/spectral.py
```diff
@@ -126,12 +154,27 @@ def __init__(
             msg = f"Unknown transform type: {transform}"
             raise ValueError(msg)
 
+    def _select_subgrid(self, x: torch.Tensor) -> torch.Tensor:
+        # Obtain a subgrid by slicing the grid dim as a view, avoiding an explicit index-tensor allocation.
+        index = [slice(None)] * x.ndim
+        index[TensorDim.GRID] = self.subgrid
```
anaprietonem, line 169: I think it be useful for debugging purposes to have some logger  statements to show the shape of x after applying the subgridding and the reprojection

training/src/anemoi/training/losses/spectral.py
```diff
@@ -126,12 +154,27 @@ def __init__(
...
     def _to_spectral_flat(self, x: torch.Tensor) -> torch.Tensor:
-        """Transform to spectral domain and flatten spectral dimensions."""
-        x_spec = self.transform.forward(x)
-        # flatten only transformed spatial/spectral dims into one "mode" axis
-        spatial_start_dim = x.ndim - 2
-        return x_spec.flatten(start_dim=spatial_start_dim, end_dim=-2)
+        """Transform to spectral domain and flatten the transformed dims into one "mode" axis."""
+        x_spec = self._to_spectral(x)
+        # the transform splits the single grid dim into two spectral dims; flatten them back to one
+        return x_spec.flatten(start_dim=x_spec.ndim - 3, end_dim=-2)
```
anaprietonem, line 188: why has this changes from x_spec.ndim-2 to x_spec.ndim - 3?

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
mc4117: Ok I had a bit of a closer look in to this again.

The PerTimestep Metrics callback calls `calculate_val_metrics` directly, bypassing `compute_dataset_loss_metrics` which would be overkill so without this change the tensor gathering step would be skipped for non-sharding metrics like the spectral losses. 

So to summarise LGTM!

training/src/anemoi/training/schemas/training.py
```diff
@@ -436,11 +436,73 @@ class HuberLossSchema(BaseLossSchema):
...
+    edge_weight_attribute: str | None = None
+    src_node_weight_attribute: str | None = None
+    row_normalize: bool = False
+
+    @model_validator(mode="after")
+    def check_mode(self) -> Self:
+        has_matrix = self.matrix_path is not None
+        has_edges = self.edges_name is not None
+        has_target_grid = self.num_nearest_neighbours is not None
+
```
anaprietonem, line 475: Those checks are still implemented at the code level, to raise errors when pydantic is off, correct?

training/docs/modules/losses.rst
```diff
@@ -255,6 +255,97 @@ Supported transforms include:
...
+
+   # Option A: reuse an existing graph edge set
+   training_loss:
+     datasets:
+       your_dataset_name:
+         _target_: anemoi.training.losses.spectral.SpectralL2Loss  # any spectral loss
+         transform: fft2d
+         x_dim: 256
+         y_dim: 128
+         projection_config:
```
anaprietonem, line 340: Just to confirm, both projection_config and subgrid are not breaking config changes and the preference is to not update the base configs, as both needs to be defined for regional models based on the high-res domain, correct?

training/docs/modules/losses.rst
anaprietonem: And a minor thing, but subgrid shouldn't be allowed for the SHL transforms as I believe in those we need to compute the spectra for the whole domain. Is that the case? @sahahner or @samhatfield can confirm

anaprietonem (commented):
Left some comments, thanks a lot and nice work @frazane! Happy to review later or feel free to ping me if anything is not clear

training/docs/modules/losses.rst
sahahner: Yes, `subgrid` should not be allowed for spherical harmonic transforms.

frazane: Good point, will fix it. Again similarly to what I mentioned in another comment, these options to manipulate data (subsetting, regridding/projecting, etc.) should ideally be consolidated into proper abstractions at some point. I don't love the way it's implemented here but for now it will do I guess.

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
anaprietonem: Just to confirm, Francesco the error you were seeing is because you were trying to train a temporal downscaler with this? cause my understanding is the per_time_step metrics is a callback we don't need to configure otherwise

training/docs/modules/losses.rst
anaprietonem: Yes I get your point about abstracting this better, but I think I would try to finish this, and then we can have a chat about standardising this across the code base? At the moment to me the interface is good enough as in all cases (expecting callbacks that I think it's a special case) we define a config where we accept both on the fly and from disk definitions

training/src/anemoi/training/diagnostics/callbacks/per_timestep_metrics.py
frazane: I was not actually training a temporal downscaler, but I was told to test all possible combinations so I set up a quick end-to-end dummy training for that as well.

training/src/anemoi/training/losses/spectral.py
frazane: It changes from `x.ndim - 2` to `x_spec.ndim - 3`. They are the same thing because the transform adds an extra dimension, I just found it cleaner to use the `x_spec` tensor rather than the original `x` to keep in mind that the transform adds the extra dim.

training/src/anemoi/training/schemas/training.py
frazane: There are some checks here that are also informative in case of errors: https://github.com/ecmwf/anemoi-core/pull/1087/changes#diff-16d6e840532df39257e09f7cbee58ba5142f80d230f0d2f1c03aa23ee5f7ee9bR646

training/docs/modules/losses.rst
frazane: Can confirm, they are keyword-only args defaulting to None. When neither is set, `_select_and_project` is a no-op.

anaprietonem:
Thanks for addressing the comments @frazane! the PR LGTM and I am happy to approve but I noticed the IT are failing https://github.com/ecmwf/anemoi-core/actions/runs/28225648972/job/83616858618. The error though is one that should be fixed in main, so if you can update the branch and re-trigger them to ensure they pass, then I can approve and merge!

anaprietonem (approved):
Thanks for addressing the comments Francesco and checking that the IT was running fine! LGTM

PR #1088, opened 2026-05-04
fix(graphs): fix node selection in interactive 3D graph viewer

matschreiner:
#  Filter node selection highlights to visible edge sets                                        
                                                                                               
  When clicking a node in the 3D graph viewer, the previous behaviour highlighted all connected
   nodes across every edge set, regardless of which edge sets were toggled on in the Edges     
  panel. This made it impossible to isolate connections for a specific edge set — for example,
  selecting only data_to_hidden would still highlight nodes connected via hidden_to_hidden or  
  any other edge set.                                 

  The fix skips hidden edge sets when computing the connected-node set on click. Now the       
  highlighted neighbours reflect only the edge sets currently visible in the UI.
                                                                                               
  Changes:                                            
  - interactive_3d.html.jinja: selectNode now checks edgeSetMeshes[i].visible before including
  an edge set's connections.                                                                   
   

Before: 

After:

### DIFF
--- a/graphs/src/anemoi/graphs/plotting/interactive_3d.html.jinja
+++ b/graphs/src/anemoi/graphs/plotting/interactive_3d.html.jinja
@@ -706,8 +706,9 @@ nor does it submit to any jurisdiction. #}
                 selectedNode = nodeName;
                 connectedNodes = new Set();
 
-                // Compute adjacency on the fly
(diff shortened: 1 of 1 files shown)


frazane (approved):
Thanks for the fix!

PR #964, opened 2026-03-10
refactor(models, training): shard_shapes

japols:
## Description
This PR simplifies the sharding metadata used to track tensor shapes across ranks. 

Previously, we stored the full tensor shape for each rank (e.g. list[list[int]]). This required layers to manually track reshapes and dimension changes whenever tensors were transformed, which made the sharding logic fragile and tightly coupled to tensor layouts.

This refactor introduces `ShardSizes = Union[list[int], None]`, representing the per-rank shard sizes along *only* the sharded dimension. Layers now propagate this information through a bundled `GraphShardInfo` / `BipartiteGraphInfo` which tracks shard metadata for both nodes and edges. The full shape expansion only happens at the level of the communication primitive where shapes are assumed equal for non-sharded dimensions across ranks.

Also refactor the all-to-all primitives for head/channel <-> grid sharding into a single common all-to-all primitive.

##  Additional notes ##
For now I've tested sharding for a global model across combinations of:
* ensemble/single
* keep_batch_sharded True/False
* transformer/graphtransformer
* head/edge sharding

please feel free to also test your favourite use case.

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

----
📚 Documentation preview 📚: https://anemoi-training--964.org.readthedocs.build/en/964/

----
📚 Documentation preview 📚: https://anemoi-graphs--964.org.readthedocs.build/en/964/

----
📚 Documentation preview 📚: https://anemoi-models--964.org.readthedocs.build/en/964/

PR #1099, opened 2026-05-07
fix: Sharding for learnable residual connections. 

jakob-schloer:
## Description

Add model-sharding support for `SpectralOrnsteinConnection` so it works correctly when `num_gpus_per_model > 1`.

`SpectralOrnsteinConnection` was incompatible with model sharding (grid split across multiple GPUs). Two issues existed:

1. The inverse SHT produces a spatial weight field with the full grid shape, but when sharded, `x` only contains the local grid shard. Broadcasting fails on the mismatched grid dimension.
2. When `truncate=True`, the forward SHT requires all latitude rings to produce correct spectral coefficients. Operating on a grid shard yields incorrect results silently.

##  Additional notes ##
The `gather_tensor` in `_apply_truncation` introduces an all-gather collective per forward pass which would become a problem for very large grids.

## Todos
- [x] Waiting for merging the `shard_shapes` refactor #964 
- [ ] Testing with deterministic model
- [ ] Testing with ensemble model
- [ ] Integration test


jakob-schloer:
Thanks @japols adapting the sharding to the most recent logic. I've just tested the sharding running two identical models once with 1 GPU/Model vs 2 GPU/Model. 
The training loss is identical see [mlflow](https://mlflow.ecmwf.int/#/metric?runs=[%223782ab9660ad4c5da35aeeb4bd53561a%22,%225bf50a965c1447e1b449ad1811dedf27%22]&metric=%22rollout%22&experiments=[%22802%22]&plot_metric_keys=%5B%22train_multi_dataset_loss_step%22%5D&plot_layout=%7B%7D&x_axis=step&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=[]&last_linear_y_axis_range=[]) with the 2 gpm (2.85 it/s) is faster than the 1 gpm (2.44 it/s).

PR #1102, opened 2026-05-07
refactor(models): graph partitioning

japols:
## Description
Refactor graph partitioning logic for sharding and chunking into a unified `GraphPartition` abstraction that exploits dst-sorted edges for fast subgraph operations via slicing.

Previously, edge sharding across GPUs and chunking within a GPU used separate code paths with expensive `bipartite_subgraph` calls. This PR consolidates both into a single `GraphPartition` dataclass that precomputes dst/edge splits and materialises subgraphs via simple slice operations.

This lays the groundwork for extending edge sharding to processors (non-bipartite graphs) and for more efficient halo exchange communication schemes.

* faster subgraph operations via slicing instead of index_select
* clearer graph sharding/chunking code in mapper.py
* no sorting overhead when using Triton GT

##  Additional notes ##


japols:
Benchmark tests [pass](https://github.com/ecmwf/anemoi-core/actions/runs/23009063665/job/66814102554) with ~10-20% throughput improvements and ~10% reduced peak memory usage.

models/src/anemoi/models/distributed/graph.py
```diff
@@ -20,6 +20,13 @@
 from anemoi.models.distributed.shapes import get_shard_sizes
 
 
+def model_is_distributed(
```
cathalobrien: This could be moved to distributed/utils.py ?

models/src/anemoi/models/distributed/khop_edges.py
```diff
@@ -226,3 +261,297 @@ def drop_unconnected_src_nodes(
...
+        return slice(start, end)
+
+    def _get_dst_range(self, partition_id: int) -> slice:
+        start, end = get_partition_range(self.dst_splits, partition_id)
+        return slice(start, end)
+
+    def _relabel_dst_nodes(self, edge_index: Adj, partition_id: int) -> None:
+        """Relabel dst indices from global to partition-local (in-place).
+
+        Caller must ensure edge_index is a clone if the original must be preserved.
```
cathalobrien: could you make this explicit by adding an in_place parameter?
e.g.
```
def _relabel_dst_nodes(self, edge_index: Adj, partition_id: int, in_place) -> Tuple[None,edge_index]:
```

models/src/anemoi/models/distributed/khop_edges.py
```diff
@@ -226,3 +261,297 @@ def drop_unconnected_src_nodes(
...
+    edge_splits: list[int]
+
+    def materialise(
+        self,
+        partition_id: int,
+        x: PairTensor,
+        edge_attr: Tensor,
+        edge_index: Adj,
+        cond: Optional[PairTensor] = None,
+    ) -> tuple[PairTensor, Tensor, Adj, Tensor, Optional[PairTensor]]:
```
cathalobrien, line 85:
```
tuple[PairTensor, Tensor, Adj, Tensor, Optional[PairTensor]]
```
would it be worthwhile to represent this as a BipartiteGraphShard object, the list of objects is getting quite long

models/src/anemoi/models/distributed/khop_edges.py
```diff
@@ -226,3 +261,297 @@ def drop_unconnected_src_nodes(
...
+    shard_info : BipartiteGraphShardInfo
+        Current shard metadata.
+    model_comm_group : ProcessGroup, optional
+        Model communication group.
+    cond : tuple[Tensor, Tensor], optional
+        Conditioning tensors (cond_src, cond_dst).
+
+    Returns
+    -------
+    tuple[PairTensor, Tensor, Adj, BipartiteGraphShardInfo, Optional[PairTensor]]
```
cathalobrien, line 350: Could this return a BipartiteGraphShard object instead of a list of objects? If someone wanted to test something, or reuse these functions in a new codebase, it is some work to line up the list of objects of different types correctly

models/src/anemoi/models/layers/block.py
```diff
@@ -644,6 +648,7 @@ def apply_gt(
         edges: Tensor,
         edge_index: Adj,
         size: Union[int, tuple[int, int]],
+        edges_are_dst_sorted: bool = False,
```
cathalobrien: Rather then adding more arguments, could edges, edge_index, size and edges_are_dst_sorted be combined into a subgraph object?

models/src/anemoi/models/layers/block.py
```diff
@@ -653,10 +658,17 @@ def apply_gt(
...
+            assert is_edge_index_dst_sorted(edge_index), (
+                "edges_are_dst_sorted=True but edge_index is not sorted by destination node. "
+                "This indicates a bug in edge sorting/sharding."
+            )
+
         if self.graph_attention_backend == "triton":
-            csc, perm, reverse = edge_index_to_csc(edge_index, num_nodes=conv_size, reverse=True)
-            edges_csc = edges.index_select(0, perm)
-            args_conv = (edges_csc, csc, reverse)
+            csc, _, reverse = edge_index_to_csc(
```
cathalobrien: great to see the index select vanish :)

should we edit edge_index_to_csc to not return perm anymore? now that we don't need it for the index_select?

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -88,7 +89,7 @@ def get_edges(
         dst_coords: Optional[Tensor] = None,
         model_comm_group: Optional[ProcessGroup] = None,
         shard_edges: bool = True,
-    ) -> Union[tuple[Tensor, Adj, Optional[ShardSizes]], Tensor]:
+    ) -> Union[tuple[Tensor, Adj, Optional[ShardSizes], bool], Tensor]:
```
cathalobrien: similar comment as earlier, but could we replace this list of results with an Edges object?

models/src/anemoi/models/distributed/khop_edges.py
```diff
@@ -226,3 +261,297 @@ def drop_unconnected_src_nodes(
...
+        Caller must ensure edge_index is a clone if the original must be preserved.
+        """
+        dst_offset = get_partition_range(self.dst_splits, partition_id)[0]
+        edge_index[1] -= dst_offset
+
+
+def _drop_unconnected_src_nodes(x_src: Tensor, edge_index: Adj) -> tuple[Tensor, Adj, Tensor]:
+    """Drop src nodes with no edges and relabel src indices to be contiguous.
+
+    Modifies edge_index[0] in-place. Caller must clone edge_index beforehand
```
cathalobrien: Rather then a comment which might be missed, we could make the in_place nature explicit by adding an optional parameter

```
_drop_unconnected_src_nodes(x_src: Tensor, edge_index: Adj, in_place: bool = false) -> tuple[Tensor, Adj, Tensor]:
```

models/src/anemoi/models/distributed/khop_edges.py
```diff
@@ -226,3 +261,297 @@ def drop_unconnected_src_nodes(
...
+        """
+        x_src, x_dst = x
+
+        # slice edges and dst nodes for this partition
+        edge_range = self._get_edge_range(partition_id)
+        edge_attr_subset = edge_attr[edge_range]
+        edge_index_subset = edge_index[:, edge_range].clone()  # clone to avoid in-place corruption
+
+        dst_range = self._get_dst_range(partition_id)
+        x_dst_subset = x_dst[dst_range]
```
cathalobrien, line 118: should we also clone for these subsets? or why is clone() only needed for edge_index_subset ?

models/src/anemoi/models/distributed/khop_edges.py
```diff
@@ -226,3 +261,297 @@ def drop_unconnected_src_nodes(
...
+    updated_shard_info = BipartiteGraphShardInfo(
+        src_nodes=shard_info.src_nodes,
+        dst_nodes=partition.dst_splits,
+        edges=partition.edge_splits,
+    )
+
+    return (x_src_local, x_dst), edge_attr, edge_index, updated_shard_info, cond_local
+
+
+def build_graph_partition(edge_index: Adj, num_parts: int, num_nodes: tuple[int, int]) -> GraphPartition:
```
cathalobrien: should we add some tests that demonstrate the building of graph partitions, dropping unconnected source nodes and converting a given graph to a local subset? would also be nice documentation of how to use these features.

models/src/anemoi/models/triton/utils.py
```diff
@@ -11,18 +11,36 @@
...
 
 import torch
+from torch import Tensor
 from torch_geometric.typing import Adj
 from torch_geometric.utils import index_sort
 from torch_geometric.utils.sparse import index2ptr
 
 
-def edge_index_to_csc(edge_index: Adj, num_nodes: Optional[Tuple[int, int]] = None, reverse: bool = True):
+def sort_edge_index_by_dst(edge_index: Adj, max_value: int = None) -> Tuple[Adj, Tensor]:
```
cathalobrien: This function is used in layers/graph_provider.py and distributed/khop_edges.py 
should we move this function somewhere outside triton/?  distributed/utils.py or somewhere else

models/src/anemoi/models/triton/utils.py
```diff
@@ -11,18 +11,36 @@
...
+    """Sort edge indices by destination node."""
+    _, perm = index_sort(edge_index[1], max_value=max_value, stable=True)
+    return edge_index[:, perm], perm
+
+
+def is_edge_index_dst_sorted(edge_index: Adj) -> bool:
```
cathalobrien: same as above, this function is just used in block.py. can we move it somewhere outside triton/?

models/src/anemoi/models/triton/utils.py
```diff
@@ -11,18 +11,36 @@
...
+def edge_index_to_csc(
+    edge_index: Adj, num_nodes: Optional[Tuple[int, int]] = None, reverse: bool = True, assume_sorted: bool = False
+):
     """Convert edge indices to CSC format, optionally also building reverse (CSR-like) metadata.
 
     Args:
         edge_index (LongTensor): [2, num_edges] edge indices (src, dst).
         num_nodes (Tuple[int, int], optional): (num_src, num_dst).
         reverse (bool): If True, also build CSR-like info for per-source iteration.
+        assume_sorted (bool): If True, assume the edge indices are already sorted by dst nodes.
```
cathalobrien: I would change assume_sorted to sorted

models/src/anemoi/models/triton/utils.py
```diff
@@ -11,18 +11,36 @@
...
+
+def is_edge_index_dst_sorted(edge_index: Adj) -> bool:
+    """Check whether edge_index is sorted by destination node (edge_index[1])."""
+    dst = edge_index[1]
+    if dst.numel() <= 1:
+        return True
+    return bool(torch.all(dst[1:] >= dst[:-1]).item())
+
+
+def edge_index_to_csc(
```
cathalobrien, line 25: Now that we don't have the expensive index select here, we could add this into the forward function of the triton gt and remove the apply_gt function

cathalobrien (commented):
great work! nice performance improvements now that index_select is removed and good foundation for what comes next ;)

Now we are passing edges, edge_indices, num_nodes and edges_are_dist_sources around as parameters, and returning a list of these from functions. Could it be worthwhile to abstract these as a Subgraph object? so that people implementing a new model don't have to do so much plumbing.

Also some standalone tests for the new graph partition functions would be nice to see as documentation for how to use them.

models/src/anemoi/models/layers/block.py
ssmmnn11: should this be like this?

```
  csc, perm, reverse = edge_index_to_csc(
      edge_index,
      num_nodes=conv_size,
      reverse=True,
      assume_sorted=edges_are_dst_sorted,
  )

  if perm is not None:
      edges = edges.index_select(0, perm)
```

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -27,6 +27,7 @@
 from anemoi.models.distributed.khop_edges import shard_edges_1hop
 from anemoi.models.distributed.shapes import ShardSizes
 from anemoi.models.layers.graph import TrainableTensor
+from anemoi.models.triton.utils import sort_edge_index_by_dst
```
ssmmnn11: probably this should not live in triton utils anymore? is there a better place for it?

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -360,15 +364,20 @@ def _get_edges_impl(
...
+    ) -> tuple[Tensor, Adj, Optional[ShardSizes], bool]:
         """Implementation of get_edges, separated for checkpointing."""
         # Build graph from coordinates
         edge_attr, edge_index = self.build_graph(src_coords, dst_coords)
+        edge_index, perm = sort_edge_index_by_dst(edge_index, max_value=dst_coords.shape[0])
+        edge_attr = edge_attr[perm]
 
         if shard_edges:
-            return shard_edges_1hop(edge_attr, edge_index, src_coords.shape[0], dst_coords.shape[0], model_comm_group)
+            edge_attr, edge_index, edge_shard_sizes = shard_edges_1hop(
```
ssmmnn11, line 390: shard_edges_1hop sorts again? Though not a real issue ...

ssmmnn11 (dismissed):
Very nice! 

Just check the trainable edges please / a migration path for these.

models/src/anemoi/models/layers/graph_provider.py
japols: you're right, I changed shard_edges_1hop to use the new codepath instead

models/src/anemoi/models/distributed/khop_edges.py
japols: I agree this would be cleaner, maybe for a future PR

models/src/anemoi/models/distributed/khop_edges.py
japols: same as above, pretty standard pyg-like output but could be cleaned up

models/src/anemoi/models/layers/block.py
japols: same as above

models/src/anemoi/models/layers/block.py
japols: I would still keep it for the case where a user provides an unsorted edge_index

models/src/anemoi/models/layers/graph_provider.py
japols: same as above

models/src/anemoi/models/distributed/khop_edges.py
japols: good point, changed it

models/src/anemoi/models/distributed/khop_edges.py
japols: in this case we don't need to as we're only returning a slice, not modifying in-place via relabeling

models/src/anemoi/models/distributed/khop_edges.py
japols: good point, added some tests

models/src/anemoi/models/triton/utils.py
japols: I would still keep it as an explicit to_input_format conversion, see above

japols:
> Very nice!
> 
> Just check the trainable edges please / a migration path for these.

I added a migration that handles the edge-attribute permutations if necessary. It is applied by default in the transfer-learning path and in `on_load_checkpoint` in the lightning module as it would otherwise fail silently.

cathalobrien (approved):

Issue #919, opened 2026-02-19
Multi-domain: A dynamic way of training across domains and resolutions


einrone:
### Issue:
Currently, anemoi-core does not support dynamically switching between different global datasets, regional datasets or experiment with different resolutions during the same training without introducing additional encoders and decoders.

Each training run assumes a fixed spatial configuration (same domain and resolution), which makes it difficult to train a model with mixed datasets where samples come from different regions and resolutions during each training iteration steps. 

### without multi-domain:

1. Introducing additional transfer learning steps when fine-tuning to a new domain.
2. Transfer learning can cause catastrophic forgetting, where learned features of previous steps may be forgotten.
3. When training with multiple encoders, the similarities between datasets are not optimally used.
4. Harder to achieve cross-domain or cross-resolution generalization as the model is not able to continuously see different regional (or global at different resolution) samples
5. Harder to get a good prediction on a unseen domain i.e domain(s) that is not contained in the training dataset (model has not generalized)

### Quick summary of the solution:
multi-domain is a concept that allows the user to train a model on different datasets simultaneously, without introducing additional model components, e.g encoders and decoders. Compared to multi-dataset PR, the model architecture is a single encoder-processor-decoder structure that can be supported with different back bones. This is done by removing all graph and grid dependency in the model architecture i.e not initializing the model (during creation) with a graph. Tested with both deterministic and ensemble models (implemented in a fork on the DestinE repository).  Down below there is an illustration of the model trains, see gif:

<img src="https://github.com/user-attachments/assets/54790f1c-2092-4a95-bc43-7b66cc6027f7"
     width="500"
     height="300">

During training the dataloader holds different anemoi dataset objects which are labeled. For each training iteration the sample is randomly fetched and provides an output containing the data (shape = [time, Ensemble, Variable, Grid]) and its label, which is ties the data and its corresponding graph.  An image of the output from the dataloader is provided

<img width="500" height="271" alt="Image" src="https://github.com/user-attachments/assets/bb18614e-8ac0-449b-b6ad-3e96d11491ca" />

Additional details can be provided in a document or in a meeting. 

### What needs to be changed/modified/implemented: 
Each of these tasks can be a separate PR or combined PRs. This can be of course be discussed

**anemoi-models**:
- [ ] layers/mapper.py (remove all grid dependencies)
- [ ] layers/block.py (remove all grid dependencies)
- [ ] layers/processor.py (remove all grid dependencies)
[...]

PR #1107, opened 2026-05-08
feat: Multi-domain, Introducing dataloader functionality

einrone:
## Description

This PR includes dataloader functionality that is being is used for multi-domain training. See
[Issue 919](https://github.com/ecmwf/anemoi-core/issues/919) for more in depth explanation.

This PR may introduce a breaking change, where we have introduced a baseclass for multi-domain dataloader and multi-dataset

For more in-depth description see [Issue 919](https://github.com/ecmwf/anemoi-core/issues/919l)

##  Additional notes ##

Notice that this PR does include a test for multi-domain dataloader, but this is a small PR out of several PRs to enable multi-domain.Thus regarding this, we do not fully enable multi-domain yet, so the full solution i.e training cannot yet be tested.

----
📚 Documentation preview 📚: https://anemoi-training--1107.org.readthedocs.build/en/1107/

----
📚 Documentation preview 📚: https://anemoi-graphs--1107.org.readthedocs.build/en/1107/

----
📚 Documentation preview 📚: https://anemoi-models--1107.org.readthedocs.build/en/1107/


einrone:
On ATS it was agreed no configuration will be provided in this PR, hence removing config changes and instantiation in datamodule.

training/tests/unit/data/test_multidomain.py
```diff
@@ -0,0 +1,72 @@
...
+from pytest_mock import MockFixture
+
+from anemoi.training.data.multidomain import MultiDomainDataset
+
+
+class TestMultiDomain:
+    """Test MultiDomainDataset instantiation and properties."""
+
+    @pytest.fixture
+    def multi_domain(self, mocker: MockFixture) -> MultiDomainDataset:
```
anaprietonem, line 22: wouldn't it more representative to have 2 datasets with different frequencies,dates and resolution for the unit-tests of multi-domain?

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,198 @@
...
+            worker_id : int
+                The ID of the current worker (0-indexed).
+
+        Returns
+        -------
+            None
+        """
+        self.worker_id = worker_id
+
+        for dataset in self.dataset_names:
```
anaprietonem, line 134: why do you have to loop over the datasets for the worker init in the multidomain dataset? We could at least have this inner loop as base_initialisation that get's reused for multi-ds and multi-domain

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,198 @@
...
+            LOGGER.info(
+                ("Worker %d (%s, pid %d, base_seed %d, sanity rnd %f)"),
+                worker_id,
+                self.label,
+                os.getpid(),
+                base_seed,
+                sanity_rnd,
+            )
+
+    def get_sample(self, domain_name: str, index: int) -> dict[str, torch.Tensor]:
```
anaprietonem, line 168: @JPXKQX  you had some suggestions/ideas here to abstract this better with a sampler?

training/src/anemoi/training/data/anemoidataset.py
```diff
@@ -0,0 +1,265 @@
...
+from torch.utils.data import IterableDataset
+
+from anemoi.models.distributed.balanced_partition import get_balanced_partition_sizes
+from anemoi.models.distributed.balanced_partition import get_partition_range
+from anemoi.training.data.data_reader import BaseAnemoiReader
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AnemoiDataset(IterableDataset, ABC):
```
anaprietonem, line 32: While discussing with @VeraChristina yesterday, anemoi-datasets now include units. Specially for multi-domain where we load many datasets in the same Datasets we should have a check that goes over the data_readers and inspect the variable_metadata for the units to ensure those are the same

anaprietonem: Similarly how do you that at the moment for the variable names in multi-domain @Sooph-knmi @einrone ?

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,198 @@
...
+from anemoi.training.utils.time_indices import normalize_time_indices
+from anemoi.training.utils.time_indices import offset_time_indices
+from training.src.anemoi.training.data.anemoidataset import AnemoiDataset
+
+LOGGER = logging.getLogger(__name__)
+
+# it may be that multidomain is not a good name for this, but it is what it is for now
+
+
+class MultiDomainDataset(AnemoiDataset):
```
anaprietonem, line 33: Can you add some documentation to explain what's a Multi-domain dataset?  We could extend https://anemoi.readthedocs.io/projects/training/en/latest/modules/data.html#multi-dataset

anaprietonem: Can a multi-domain dataset combine NativeGridDatasets and TrajectoryDatasets? or should we have some assert that the data_readers in multi-domain do not specify any trajectory?

anaprietonem (commented):
Left a few comments, thanks for the nice work @einrone and @Sooph-knmi

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def __iter__(self) -> Generator[dict[str, torch.Tensor], None, None]:
+        """Return an iterator that yields a tuple torch.Tensor and its corresponding domain name.
+
+        Returns
+        -------
+        tuple[torch.Tensor, str]
+            A tuple containing the tensor sample and its corresponding domain name
+        """
+        if self.shuffle:
```
anaprietonem, line 196: so in this case the shuffling is applied per dataset_reader since each can have it's own valid_date_indices, is that right?

training/tests/unit/data/test_multidomain.py
einrone: Yes that is true. However would you also like this test with an additional test which includes your suggestions?

training/src/anemoi/training/data/multidomain.py
einrone: In the multi-domain dataloader you want to have control over how much data is partitioned per worker. That is the worker needs to have correct start end of time-indices of a given domain (we dont want to move into a time index that belongs to another domain) but also correct shard size for the given domain.  If that makes sense?

einrone: In the v1 implementation we did not implement this type of solution but it was quite unstable on different HPC, hence this solution fixed it.

training/src/anemoi/training/data/anemoidataset.py
einrone: Yes, we do have a check for this. This check has been implemented in the refactored code in the DestinE repository but in the datamodule. If I understood you correctly, do you want me to include a check in the constructor or as a "private" method within the dataloader class?

training/tests/unit/data/test_multidomain.py
Sooph-knmi: Usually with multi-domain we would not have different frequencies (this is more something for multi-in/multi-out). But we do have different resolutions and dates. I can adapt the unit test to make it a bit more different.

training/src/anemoi/training/data/multidomain.py
einrone: I mean in theory yes, I did implement this feature in the DestinE repository but it was an older version of the trajectory dataset. At that time this functionality was built into the single dataset class. However, I have not tested this functionality with multi-domain, have you @Sooph-knmi ?

training/src/anemoi/training/data/multidomain.py
einrone: When shuffle=True, indices (dataset_reader for a given dataset) are shuffled independently for each dataset/domain first. Each domain is shuffled indices are then sliced to the worker’s assigned chunk. After that, the per-domain chunks are flattened into a single list of (domain, index) pairs. So this list includes all domains, because we want to sample domains randomly. This combined list is shuffled again, so samples from different domains are mixed before iteration. Finally, samples are yielded by looking up each (domain, index) pair.

einrone:
> Left a few comments, thanks for the nice work @einrone and @Sooph-knmi

Hi, thank you for the reply and walking through the code. Really appreciate that, thanks. Let me know if there something which are a bit unclear or something that needs to be changed in order to fit into the current codebase.

training/src/anemoi/training/data/multidomain.py
Sooph-knmi: Yes, but also, we need to have the same dataset across workers to avoid timeouts. So we shuffle each data reader shard per dataset before combining them, which enforces this property.

training/src/anemoi/training/data/multidomain.py
Sooph-knmi: Hi @anaprietonem, as far as I can tell, TrajectoryDatasets are also not supported by multi-dataset? I am not so sure about how Anemoi deals with TrajectoryDatasets but maybe for now it is not a requirement to support it? In that case we can add an assertion as you mention.

Sooph-knmi: I took a quick look, it seems to be supported in the usable_indices selection. So I see no reasons why it would not be able to combine both datasets, as long as the correct indices are selected the samples can be shuffled and the correct sample can be passed to the model.

training/src/anemoi/training/data/multidomain.py
anaprietonem: Thanks for the explanation, could we then try to at least reduce the code duplication and have the inner loop call a utility function from the base dataset?

training/src/anemoi/training/data/multidomain.py
anaprietonem: Thanks for checking Sophie, indeed I am not trying to push for this to be supported mostly trying to ask if it makes sense to do it, and if not then suggest the code have an assert to prevent that.

training/src/anemoi/training/data/anemoidataset.py
anaprietonem: Yes I'd go for a private method in the dataloader class. If useful this the work @VeraChristina  did for units https://github.com/ecmwf/anemoi-core/pull/1140/changes#diff-593510609c04c672c6651eff017a6c3ae8993ec36973d852e6ec4c52377d4081 so we should reuse something similar to https://github.com/ecmwf/anemoi-core/blob/a161913b3dd0bf23843a7b77feb875ef3f637597/training/src/anemoi/training/utils/variables_metadata.py

training/src/anemoi/training/data/multidomain.py
anaprietonem: I think it would be good to document this better.

einrone:
Resolved conflict with main and ran `test_multidomain.py` (2/2 passed 3 warnings) and `test_multidataset.py`  (5/5 passed 3 warnings). Warnings comes from deprecation warnings from anemoi regarding torch geometric and torch.jit.script

training/src/anemoi/training/data/multidomain.py
einrone: Do you want that documentation in anemoi.readthedocs or briefly explained in a docstring?

einrone:
@anaprietonem I added unit check in the dataloader and also created additional test, could you have a look if this is the implementation you were aiming for.

anaprietonem:
> @anaprietonem I added unit check in the dataloader and also created additional test, could you have a look if this is the implementation you were aiming for.

Yes thanks a lot Aram for adding this! I think the only things missing area:
- Updating the docs
- Adding the assert/error catching if gridded and trajectory datasets are used
- Sampling mechanism we discussed with @JPXKQX  last week! 

feel to ping me for review again when those are added or if you have any questions!

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,238 @@
...
+        }
+        # Normalize the date indices to use slices where possible, which can improve downstream indexing performance.
+        self.relative_date_indices = {
+            name: normalize_time_indices(indices) for name, indices in relative_date_indices.items()
+        }
+        LOGGER.info("valid date indices: %s", self.valid_date_indices)
+        self.n_samples_per_worker = {}  # overwrite base to empty dict
+        self.chunk_index_range = {}  # overwrite base to empty dict
+
+    def _check_datasets_units(self) -> None:
```
VeraChristina, line 80: This method doesn't seem to be called anywhere, unless I am missing something? Also, we noticed in other contexts that it can be useful to have an option to disable the checks, see e.g. https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/train/train.py#L337. We have added config options under the key `check_variables_compatibility` in several places in the training config (for the different checks), using the schema `CheckVariablesCompatibilitySchema`.

training/src/anemoi/training/data/datamodule.py
```diff
@@ -136,6 +137,19 @@ def _get_dataset(
...
+                strategy,
+                data_readers=data_readers,
+                relative_date_indices=relative_date_indices,
+                shuffle=shuffle,
+                label=label,
+                epoch=self.epoch,
+                rollout=len(tuple(self.task.steps(label))),
+                **dataset_options,
+            )
+
```
dietervdb-meteo: Why not absorb this in the above, and make MultiDataset default instead of None? Or even break configs and demand for strategy to be set to `_target_.... MultiDataset` in the default case.

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,305 @@
...
+        self.valid_date_indices = {
+            name: np.arange(len(anchors), dtype=np.int64) for name, anchors in self.anchors.items()
+        }
+        # Normalize the date indices to use slices where possible, which can improve downstream indexing performance.
+        self.relative_date_indices = {
+            name: normalize_time_indices(indices) for name, indices in relative_date_indices.items()
+        }
+
+    @cached_property
+    def frequency(self) -> dict[str, datetime.timedelta]:
```
dietervdb-meteo: Don't really see why this can't go and we simply inherit the `def frequency` from parent class?

training/src/anemoi/training/data/multidataset.py
```diff
@@ -86,23 +86,23 @@ def __init__(
         if single_seq and multi_seq:
             msg = (
                 "Currently mixing single-sequence datasets (global time axis) with "
-                "Trajectory datasets (init x step axes) in the same MultiDataset is unsupported. "
+                f"Trajectory datasets (init x step axes) in the same {self.__class__.__name__} is unsupported. "
```
dietervdb-meteo:
```suggestion
                f"Trajectory datasets (init x step axes) in the same IterableDataset is unsupported. "
```

training/src/anemoi/training/data/multidataset.py
```diff
@@ -86,23 +86,23 @@ def __init__(
...
                 f"Single-sequence: {single_seq}. Trajectory: {multi_seq}. "
             )
             raise ValueError(msg)
 
-        # Compute valid (sequence, position) anchors and a flat index over them
-        # that the shuffle/shard logic operates on.
+        self._set_date_indices(relative_date_indices)
```
dietervdb-meteo, line 103: when looking at the diff here, seems like many comments in the original code have dissappeared. Can they be added back, at the appropriate place?

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,305 @@
...
+        super().__init__(
+            data_readers=data_readers,
+            relative_date_indices=relative_date_indices,
+            shuffle=shuffle,
+            label=label,
+            epoch=epoch,
+            rollout=rollout,
+        )
+        self._check_datasets_units(**dict(check_variables_compatibility or {}))
+        LOGGER.info("valid date indices: %s", self.valid_date_indices)
```
dietervdb-meteo: Maybe this log message can be moved to the parent class? And maybe to debug? is this something we want to see in a typical log?

training/src/anemoi/training/data/multidomain.py
dietervdb-meteo: Completely agree with Ana here, we can define a utility function 'set_samples_per_worker` in the parent class, and only override that. The rest of `per_worker_init` can then be directly inherited.

training/src/anemoi/training/data/multidomain.py
```diff
@@ -0,0 +1,305 @@
...
+        self.n_samples_per_worker = {}
```
dietervdb-meteo: both of these can be moved to `per_worker_inits` (or even the sub-utility, see comment there), which is where these things get initialized anyways.

dietervdb-meteo (changes requested):
Already some comments on streamlining the two DataSet classes even further.

dietervdb-meteo:
PS I also unresolved a comment by Ana above, as I think she's right, and it wasn't fully taken into account.

dietervdb-meteo:
I removed the breaking label, since we currently have fallbacks that make the old configs work. Personally I would be in favor of removing this extra code and make the PR (config) breaking. Lets discuss.

einrone:
Quick rockachat or an actual meeting :) ?

Issue #1100, opened 2026-05-07
rollout training doesn't store the correct rollout step


JPXKQX:
### What happened?

I think there is a problem with the 'rollout' step when resuming a job using the `run_id`. When I resume a job at the third epoch, it should use rollout step 3, but it starts with rollout 2 again. See the image of the mlflow metric, rollout.

<img width="984" height="298" alt="Image" src="https://github.com/user-attachments/assets/f0875142-c72b-44ec-87cf-9bb10809d73d" />

### What are the steps to reproduce the bug?

Resume a rollout training with `run_id`.

### Version

main

### Organisation

ECMWF

frithjofschick:
Hi Mario, I proposed a fix in [this PR draft](https://github.com/ecmwf/anemoi-core/pull/1109). The code changes got bigger than I would have liked for a first potential contribution, but I tried to write a clear description of the proposed changes.

I just saw that the status of this issue is in `To be triaged` ... please let me know if you are accepting contributions in this format and apologies if I didn't follow established triage processes.

frithjofschick:
The fix was merged - I believe this issue can be closed :)

PR #1109, opened 2026-05-08
fix(training): persist rollout step across job resume

frithjofschick:
## Description
This change enables the `Forecaster` task class to persist rollout curriculum state across job resume.

Specifically, this change:
- adds `state_dict`/`load_state_dict` methods to `RolloutConfig` to own its own serialisation
- wires them through a new `training_runtime_state_dict`/`load_training_runtime_state_dict` interface on `BaseTask`, intended for any mutable training runtime state that must survive a checkpoint/resume cycle
- hooks into `on_save_checkpoint`/`on_load_checkpoint` on `BaseTrainingModule` to persist this state at the training checkpoint level

The rollout state is kept out of the model's `state_dict` so that it is invisible to inference code (e.g. `anemoi-inference`).

Old checkpoints that predate this change have no `task_state` key, but `checkpoint.get("task_state", {})` handles this gracefully by passing an empty dict to `load_training_runtime_state_dict`, which leaves the rollout step at `start`.

`RolloutConfig.step` was reset to `start` on every resume because it is mutable runtime state that was never written to or read from checkpoints.

[Issue 1100](https://github.com/ecmwf/anemoi-core/issues/1100)

## Additional notes


training/src/anemoi/training/train/methods/base.py
```diff
@@ -452,6 +452,15 @@ def _update_checkpoint_state_dict_for_load(self, checkpoint: dict[str, Any]) ->
             if full_key.startswith(processor_prefixes):
                 state_dict[full_key] = value
 
+    def get_extra_state(self) -> dict:
+        if hasattr(self, "task"):
+            return self.task.extra_state_dict()
+        return {}
+
+    def set_extra_state(self, state: dict) -> None:
+        if hasattr(self, "task"):
```
frithjofschick: Forgot to comment on this specific if clause - it looks a bit awkward. I introduced this check, because `set_extra_state` can be called during `load_state_dict`, which in some tests is invoked on a `DummyTrainingModule` that was constructed via `__new__`, bypassing `__init__ `and therefore never setting `self.task`.

The cleaner fix would be to initialise `self.task = None` in `BaseTrainingModule.__init__` before the real assignment and change the guard to `if self.task is not None`. That makes the invariant explicit in the constructor rather than hiding the assumption in a `hasattr` call. However, I didn't want to mess with default assignments in a core module in my first PR :) Happy to make that change if you'd like me to though.

frithjofschick: On second thought, one should keep `task` as a required dependency of `BaseTrainingModule`, but instead give the test stubs some sort of minimal task. Then this if clause can be removed entirely.

JPXKQX:
Hi @frithjofschick, welcome and thank you for your first contribution to anemoi! 

The PR looks good overall, and the test coverage looks great.

My only concern is about storing the rollout step in the model's state_dict`. The rollout step is training runtime state, not model state. For example, `anemoi-inference` will only care about the model itself and not the rollout step.

I think the right place for this is the Lightning checkpoint level, using `on_save_checkpoint`/`on_load_checkpoint` on `BaseTrainingModule`.

What do you think?

frithjofschick:
Hi @JPXKQX, thanks for your review!

Agreed - handling the training runtime state on Lightning level is more sensible.

I adjusted the PR to use `on_save_checkpoint`/`on_load_checkpoint` for keeping track of the `rollout` variable and simplified the tests so that I could remove the `if hasattr(self, "task")` guards in `on_save_checkpoint`/`on_load_checkpoint` of `BaseTrainingModule`

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -42,6 +42,14 @@ def increase(self) -> None:
             self.step += 1
             LOGGER.info("Rollout window length has been increased to %d.", self.step)
 
+    def state_dict(self) -> dict:
+        """Return serialisable state."""
+        return {"step": self.step}
```
JPXKQX: could we add here all the other `RolloutConfig` arguments?

JPXKQX:
Thank you for making the changes so quickly! 

Unfortunately, I am unable to run integration tests from an external branch, but I would be happy to approve them if you could run a check locally:  
- Run a rollout training. Stop it after a few epochs (fewer than the `rollout.maximum`) and then resume it. Then check in MLFlow that the logged `rollout` makes sense.

frithjofschick:
Thank you! I'm going to address your comment and run the integration test this evening :)

training/src/anemoi/training/tasks/forecaster.py
frithjofschick: I would consider the other parameters of `RolloutConfig` config parameters that should be read from the YAML defining the training at construction time on every run (`start`, `epoch_increment` and `maximum`). If we add them to the state dictionary here, then we might end up with two sources of truth for them and would have to add validation to alert us to discrepancies IMO. Would you mind if I therefore leave this as is for now?

frithjofschick:
I ran a small training job [1] for two epochs, resumed for another two epochs and checked that the logged `rollout` makes sense. It is properly persisted across the job resume:
image
image

Two observations/changes that I made during testing:
- PyTorch-Lightning was calling `on_train_epoch_end` once during the resume/restore phase, before any epoch actually trains, which caused one extra increment of `rollout`. Therefore, I had to add tracking for the last epoch that was processed, so the same epoch can't trigger two increments.
- The `Forecaster` module was logging `rollout` with `on_step=True`, but PyTorch-Lightning is using `log_every_n_steps` (default is `n=50`) and so `rollout` was missing in MLFlow in my small test. I changed the `Forecaster` to log `rollout`  with `on_epoch=True` instead, since it is constant within the epoch.

[1] I used this config:
```
defaults:
  - data: zarr
  - dataloader: native_grid
  - diagnostics: evaluation
  - system: example
  - graph: multi_scale
  - model: gnn
  - task: forecaster
  - training: single
  - _self_

config_validation: True

data:
  resolution: o48

system:
  hardware:
    num_gpus_per_node: 1
    accelerator: cpu
  output:
    root: /Users/frithjof/Desktop/anemoi-core/training/output
  input:
    dataset: /Users/frithjof/Desktop/anemoi-core/training/era5-o48-2020-2021-6h-v1.zarr
    graph: graph-o48.pt

dataloader:
  training:
    datasets:
      data:
        dataset_config:
          dataset: ${system.input.dataset}
          frequency: ${data.frequency}
          drop: []
        start: 2020
        end: 2021
  limit_batches:
    training: 5
    validation: 1

model:
  num_channels: 32
  num_layers: 2

task:
  rollout:
    start: 1
    epoch_increment: 1
    maximum: 20

training:
  max_epochs: 4

diagnostics:
  log:
    mlflow:
      enabled: True
      offline: True
      experiment_name: "anemoi-rollout-test"
      project_name: "anemoi-local-integration-tests"
      run_name: "rollout-cpu-run-v2"
      tracking_uri: "file:///Users/frithjof/Desktop/anemoi-core/training/output/mlruns"
```

JPXKQX:
Hi, I have some news following discussions with other colleagues. I thought this was introduced with some of the latest developments, but apparently, this behaviour was already known and is caused by an undesirable feature of our dataloader. 

I still think this is the way to go, but we will need to make some changes to the data loaders before merging this PR. Sorry for not raising this earlier. It was great to have your contribution here. 

Here are some more details on the issue if you are interested: Currently, the data loader loads the same batch for all epochs. This means that, in epoch 0, it loads a batch size of 14 (needed for rollout of 12 with two input steps), even though only three (2+1) are needed for that epoch. This is an important performance bottleneck.  As a result, users employ three or four different rollout configurations to resume training.
- start = 0; maximum = 4.
- start=4, maximum=8.
- start=8, maximum=12.
This is instead of starting with 1 and resuming the training. The idea is to include the rollout in the checkpoint, but first we will need to fix the Dataloader behaviour to avoid causing any performance degradation.

HCookie:
Btw, I'll be looking at some of the dataloader issues we have with rollout in the near future. So we will be able to unblock this PR with that work.

HCookie:
I've opened that work I discussed here https://github.com/ecmwf/anemoi-core/pull/1156

mc4117:
> I ran a small training job [1] for two epochs, resumed for another two epochs and checked that the logged `rollout` makes sense. It is properly persisted across the job resume: image image
> 
> Two observations/changes that I made during testing:
> 
> * PyTorch-Lightning was calling `on_train_epoch_end` once during the resume/restore phase, before any epoch actually trains, which caused one extra increment of `rollout`. Therefore, I had to add tracking for the last epoch that was processed, so the same epoch can't trigger two increments.
> * The `Forecaster` module was logging `rollout` with `on_step=True`, but PyTorch-Lightning is using `log_every_n_steps` (default is `n=50`) and so `rollout` was missing in MLFlow in my small test. I changed the `Forecaster` to log `rollout`  with `on_epoch=True` instead, since it is constant within the epoch.
> 
> [1] I used this config:
> 
> ```
> defaults:
>   - data: zarr
>   - dataloader: native_grid
>   - diagnostics: evaluation
>   - system: example
>   - graph: multi_scale
>   - model: gnn
>   - task: forecaster
>   - training: single
>   - _self_
> 
> config_validation: True
> 
> data:
>   resolution: o48
> 
> system:
>   hardware:
>     num_gpus_per_node: 1
>     accelerator: cpu
>   output:
>     root: /Users/frithjof/Desktop/anemoi-core/training/output
>   input:
>     dataset: /Users/frithjof/Desktop/anemoi-core/training/era5-o48-2020-2021-6h-v1.zarr
>     graph: graph-o48.pt
> 
> dataloader:
>   training:
>     datasets:
>       data:
>         dataset_config:
>           dataset: ${system.input.dataset}
>           frequency: ${data.frequency}
>           drop: []
>         start: 2020
>         end: 2021
>   limit_batches:
>     training: 5
>     validation: 1
> 
> model:
>   num_channels: 32
>   num_layers: 2
> 
> task:
>   rollout:
>     start: 1
>     epoch_increment: 1
>     maximum: 20
> 
> training:
>   max_epochs: 4
> 
> diagnostics:
>   log:
>     mlflow:
>       enabled: True
>       offline: True
>       experiment_name: "anemoi-rollout-test"
>       project_name: "anemoi-local-integration-tests"
>       run_name: "rollout-cpu-run-v2"
>       tracking_uri: "file:///Users/frithjof/Desktop/anemoi-core/training/output/mlruns"
> ```

Hi Sorry for being a bit late to this! Nice work! I was just wondering, what config did you use when you resumed the run?  Was it the same as the first run?

I was also wondering if you could explain a bit more about how you deal with the problem about pytorch running on_train_epoch_end before restarting. I've also noticed this and it's quite annoying so it would be great if you had a fix!

frithjofschick:
> Hi Sorry for being a bit late to this! Nice work! I was just wondering, what config did you use when you resumed the run? Was it the same as the first run?
> 
> I was also wondering if you could explain a bit more about how you deal with the problem about pytorch running on_train_epoch_end before restarting. I've also noticed this and it's quite annoying so it would be great if you had a fix!

Hi Mariana, thanks for taking a look!

On the config: Yes, identical to the first run. The only thing I changed was `training.max_epochs` (from 2 to 4).

The local integration test was:

1. Set `training.max_epochs` to 2, start training and note the run id `X`.
1. Set `training.max_epochs` to 4, resume with `training.run_id=X`.
1. Confirm in MLflow that the rollout metric continues from where it left off instead of restarting.

On the `on_train_epoch_end ordering`: The epoch-end hook fires again for the epoch that was already completed before the checkpoint was written (weird behaviour, agreed!). I made the increment idempotent by storing `last_increased_epoch` in the rollout state and only incrementing when `current_epoch != last_increased_epoch`. This way the replayed `on_train_epoch_end` becomes a no-op and the rollout stays correct across the training resume boundary.

frithjofschick:
Hi @HCookie @JPXKQX :) looks like the performance bottleneck with the data loader caused by the rollout length [was resolved here](https://github.com/ecmwf/anemoi-core/pull/1169). Is there anything else blocking this PR?

anaprietonem:
Hey @frithjofschick, sorry for the delay! PR looks good, but we would like to ensure the integration tests are passing, Since the PR is coming from a fork we can't them easily. So could you run those and show the output results here please? You can see an example of how we execute them on our CI here https://github.com/ecmwf/anemoi-core/blob/main/.github/workflows/integration-tests-hpc.yml. 
- the integration test requires 1 gpu  can be run doing `python3 -m pytest -v training/tests/integration --slow`
- and for the performance benchmarking you need 2 gpus and then run `python -m pytest -v training/tests/integration/test_benchmark.py --slow --multigpu`

frithjofschick:
Hey @anaprietonem ,

Thanks for the guidance! I managed to successfully run the integration tests, but not the performance benchmarking due to missing data (the benchmarks expect some hardcoded zarr files to be available in the local file system).

## Integration tests

The **integration tests** I ran on a lambda.ai instance with 8 v100s and the following environment (dev install from this branch):
```
platform linux -- Python 3.12.9, pytest-9.1.1, pluggy-1.6.0 -- /home/ubuntu/anemoi-core/.venv/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /home/ubuntu/anemoi-core/training
configfile: pytest.ini
plugins: hydra-core-1.3.3, anyio-4.14.0, skip-slow-1.1.0, asyncio-1.4.0, hypothesis-6.155.7, mock-3.15.1, typeguard-4.5.2
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
```

I had to run a few more compute heavy integration tests separately, because I ran into issues with memory allocations otherwise. So I first ran
```
pytest --slow -v -s   "training/tests/integration/test_training_cycle.py::test_training_cycle_edm_transport[transport_edm_diffusion_tendency]" "training/tests/integration/test_training_cycle.py::test_training_cycle_imerg_target"  "training/tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion]" "training/tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion_tendency]" "training/tests/integration/test_training_cycle.py::test_training_cycle_temporal_downscaler_ensemble" "training/tests/integration/test_training_cycle.py::test_evaluator"
```
Output extract:
```
=================================================================== 6 passed, 650 warnings in 911.73s (0:15:11) ====================================================================
```

Then I ran the remaining integration tests
```
cd training
pytest tests/integration --slow -v -s \
  --deselect "tests/integration/test_training_cycle.py::test_training_cycle_edm_transport[transport_edm_diffusion_tendency]" \
  --deselect "tests/integration/test_training_cycle.py::test_training_cycle_imerg_target" \
  --deselect "tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion]" \
  --deselect "tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion_tendency]" \
  --deselect "tests/integration/test_training_cycle.py::test_training_cycle_temporal_downscaler_ensemble" \
  --deselect "tests/integration/test_training_cycle.py::test_evaluator"
```
Output extract:
```
 ===================================================== 44 passed, 11 skipped, 6 deselected, 1518 warnings in 1138.99s (0:18:58) =====================================================
```
 
## Performance benchmarking

Regarding the **performance benchmarking**: These, I had more trouble with, because it seems like the benchmarks rely on data with hardcoded paths that are available locally on your HPC. For example, when running
```
pytest "training/tests/integration/test_benchmark.py::test_benchmark_dataloader[graphtransformer]" --slow --multigpu -x
```
the benchmarking fails, because it expects `/home/mlx/ai-ml/datasets/aifs-ea-an-oper-0001-mars-n320-1979-2023-6h-v8.zarr` to be available in the attached filesystem.
```
============================================================================= short test summary info ==============================================================================
FAILED training/tests/integration/test_benchmark.py::test_benchmark_dataloader[graphtransformer] - zarr.errors.PathNotFoundError: nothing found at path '/home/mlx/ai-ml/datasets/aifs-ea-an-oper-0001-mars-n320-1979-2023-6h-v8.zarr'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
========================================================================== 1 failed, 5 warnings in 6.98s ===========================================================================
```

Are there publicly available .zarr files that can be used in place of these hardcoded datasets for the performance benchmarking?

anaprietonem:
> Hey @anaprietonem ,
> 
> Thanks for the guidance! I managed to successfully run the integration tests, but not the performance benchmarking due to missing data (the benchmarks expect some hardcoded zarr files to be available in the local file system).
> 
> ## Integration tests
> The **integration tests** I ran on a lambda.ai instance with 8 v100s and the following environment (dev install from this branch):
> 
> ```
> platform linux -- Python 3.12.9, pytest-9.1.1, pluggy-1.6.0 -- /home/ubuntu/anemoi-core/.venv/bin/python3
> cachedir: .pytest_cache
> hypothesis profile 'default'
> rootdir: /home/ubuntu/anemoi-core/training
> configfile: pytest.ini
> plugins: hydra-core-1.3.3, anyio-4.14.0, skip-slow-1.1.0, asyncio-1.4.0, hypothesis-6.155.7, mock-3.15.1, typeguard-4.5.2
> asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
> ```
> 
> I had to run a few more compute heavy integration tests separately, because I ran into issues with memory allocations otherwise. So I first ran
> 
> ```
> pytest --slow -v -s   "training/tests/integration/test_training_cycle.py::test_training_cycle_edm_transport[transport_edm_diffusion_tendency]" "training/tests/integration/test_training_cycle.py::test_training_cycle_imerg_target"  "training/tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion]" "training/tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion_tendency]" "training/tests/integration/test_training_cycle.py::test_training_cycle_temporal_downscaler_ensemble" "training/tests/integration/test_training_cycle.py::test_evaluator"
> ```
> 
> Output extract:
> 
> ```
> =================================================================== 6 passed, 650 warnings in 911.73s (0:15:11) ====================================================================
> ```
> 
> Then I ran the remaining integration tests
> 
> ```
> cd training
> pytest tests/integration --slow -v -s \
>   --deselect "tests/integration/test_training_cycle.py::test_training_cycle_edm_transport[transport_edm_diffusion_tendency]" \
>   --deselect "tests/integration/test_training_cycle.py::test_training_cycle_imerg_target" \
>   --deselect "tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion]" \
>   --deselect "tests/integration/test_training_cycle.py::test_training_cycle_multidatasets_edm_transport[edm_diffusion_tendency]" \
>   --deselect "tests/integration/test_training_cycle.py::test_training_cycle_temporal_downscaler_ensemble" \
>   --deselect "tests/integration/test_training_cycle.py::test_evaluator"
> ```
> 
> Output extract:
> 
> ```
>  ===================================================== 44 passed, 11 skipped, 6 deselected, 1518 warnings in 1138.99s (0:18:58) =====================================================
> ```
> 
> ## Performance benchmarking
> Regarding the **performance benchmarking**: These, I had more trouble with, because it seems like the benchmarks rely on data with hardcoded paths that are available locally on your HPC. For example, when running
> 
> ```
> pytest "training/tests/integration/test_benchmark.py::test_benchmark_dataloader[graphtransformer]" --slow --multigpu -x
> ```
> 
> the benchmarking fails, because it expects `/home/mlx/ai-ml/datasets/aifs-ea-an-oper-0001-mars-n320-1979-2023-6h-v8.zarr` to be available in the attached filesystem.
> 
> ```
> ============================================================================= short test summary info ==============================================================================
> FAILED training/tests/integration/test_benchmark.py::test_benchmark_dataloader[graphtransformer] - zarr.errors.PathNotFoundError: nothing found at path '/home/mlx/ai-ml/datasets/aifs-ea-an-oper-0001-mars-n320-1979-2023-6h-v8.zarr'
> !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
> ========================================================================== 1 failed, 5 warnings in 6.98s ===========================================================================
> ```
> 
> Are there publicly available .zarr files that can be used in place of these hardcoded datasets for the performance benchmarking?

Great thanks for checking the Integration parts, indeed for the benchmarking I had missed that. I will  try check those tomorrow taking a copy from this and get back to you

frithjofschick:
Thank you! Btw I forgot to mention: After pulling my fork of anemoi-core and checking out this branch I had to fetch the git tags manually, otherwise pip will fallback to non-existing tags and fail to install the packages as done in the testing workflows:
```
git fetch https://github.com/ecmwf/anemoi-core.git 'refs/tags/*:refs/tags/*'
```

anaprietonem (approved):
thanks @frithjofschick for the contribution and for all the reviewers for the feedback and questions! The tests are passing https://github.com/ecmwf/anemoi-core/actions/runs/28007587545/job/83133927849 (note the error in the PB tests is known, and not related to this PR)

MicheleCattaneo:
Sorry for commenting on a closed PR. I however noticed that one of the conditions to increase the rollout is still
```
current_epoch % self.epoch_increment == 0
```
For cases with `epoch_increment>1` the rollout is incremented at the end of epoch 0 (because `0 % self.epoch_increment == 0` is true) rather than after `epoch_increment` epochs.

With the default value `self._last_increased_epoch: int = -1`, I think a simple fix would be to set the condition to 
```
current_epoch % self.epoch_increment == self.epoch_increment - 1
```
to get a more intuitive behaviour for the first `epoch_increment` epochs. 
What do you think?

PR #1112, opened 2026-05-11
feat: export JSON schema

b8raoult:
## Description

Add the command `anemoi-training schema` to export the JSON schema of a training config. Useful for validation tools or IDE such as vscode, that will use it to autocomplete and highlight errors.

The command prints the full JSON Schema (Pydantic model_json_schema()) for the training config's Pydantic model tree. This is useful for IDE autocompletion/validation of YAML configs, generating docs, or external tooling that wants to validate configs without running the CLI's own validate command.
Another use case where this can be useful is for agentic development as the output of this command can be used to augment prompts.

##  Additional notes ##


frazane:
Perhaps better as a subcommand of `anemoi-training config`?

anaprietonem:
> Perhaps better as a subcommand of `anemoi-training config`?

Hey @frazane good suggestion, agreed with it. I had a go, could you review it and see if that's what you have in mind?

frazane (commented):
Hi @anaprietonem, I think this makes sense!

On a more general note: how do we plan to use the schema? I am assuming it cannot be used in combination with structured configs, but only with "unrolled" ones. Is that so? Perhaps a few lines explaining how these schemas are intended to be used (on a high level) would help give them more visibility so more people will experiment with them.

aaron-hopkinson (approved):

anaprietonem:
> Hi @anaprietonem, I think this makes sense!
> 
> On a more general note: how do we plan to use the schema? I am assuming it cannot be used in combination with structured configs, but only with "unrolled" ones. Is that so? Perhaps a few lines explaining how these schemas are intended to be used (on a high level) would help give them more visibility so more people will experiment with them.

We discussed this over a meeting and updated the PR description to provide users/developers guidance about this!

PR #1113, opened 2026-05-12
fix(training): Add anemoi weight averaging classes in order to work with Imputers and Scalers

jakob-schloer:
## Description

For weight averaging, a copy of the model is created at the beginning of training and its parameters are updated/averaged over the course of training. The pytorch lightning classes for weight averaging pair parameters and buffers positionally via `zip`, which fails in Anemoi when used together with:

1. Imputers (`ConstantImputer` and friends) register scratch buffers (`nan_locations`, `loss_mask_training`) that get reassigned with new shapes on the first forward pass. The averaged model's deep-copy still holds the original `shape=(0,)` tensor, so the per-batch `update_parameters` of the PL `AverageModel` call crashes with a shape mismatch.

2. Updating loss scalers (`NaNMaskScaler`, etc.) call `ScaleTensor.update_scaler` every batch — this pops and re-registers the scaler buffer, shuffling the buffer order in the live model relative to the averaged snapshot. With matching shapes the positional `zip` silently mis-pairs tensors; with mismatched shapes it crashes.

## Changes

- `anemoi.training.diagnostics.callbacks.weight_averaging.AveragedModel` — name-based matching, shape filtering, and a float/non-float split (non-float buffers are sync'd from source rather than averaged).
- `WeightAveraging` — overrides `setup`, `_swap_models`, `_copy_average_to_current` so all transfer paths between live and averaged model use name-based matching.
- `EMAWeightAveraging` and `SWAWeightAveraging` — thin subclasses mirroring `pytorch_lightning.callbacks.EMAWeightAveraging` (and the
  analogous SWA configuration).
- `_get_weight_averaging_callback` still instantiates whatever the user targets, but logs a warning if a stock `pytorch_lightning.callbacks.*` class is configured, since those will crash when used in combination with imputers or updating scalers.

## Usage

```yaml
weight_averaging:
   _target_: anemoi.training.diagnostics.callbacks.weight_averaging.EMAWeightAveraging
   decay: 0.999
```
## Notes
Since we still allow stock `pytorch_lightning.callbacks.*` classes, this is not a breaking change

----
📚 Documentation preview 📚: https://anemoi-training--1113.org.readthedocs.build/en/1113/

----
📚 Documentation preview 📚: https://anemoi-graphs--1113.org.readthedocs.build/en/1113/

----
📚 Documentation preview 📚: https://anemoi-models--1113.org.readthedocs.build/en/1113/


training/src/anemoi/training/diagnostics/callbacks/weight_averaging.py
```diff
@@ -7,81 +7,245 @@
...
+            b_avg_det = b_avg.detach()
+            if self.use_buffers and b_avg.is_floating_point():
+                avg_self.append(b_avg_det)
+                avg_model.append(b_model_)
+                if self.n_averaged == 0:
+                    b_avg_det.copy_(b_model_)
+            else:
+                sync_pairs.append((b_avg_det, b_model_))
+
+        return avg_self, avg_model, sync_pairs
```
HCookie: For convenience, I'd suggest making this return type a dataclass

training/src/anemoi/training/diagnostics/callbacks/weight_averaging.py
```diff
@@ -7,81 +7,245 @@
...
-
-
-def _safe_copy_average_to_current(self: Any, pl_module: Any) -> None:
-    """Copy averaged buffers to the current model.
-
-    Same as ``_safe_swap_models`` but for the copy performed at the end of training.
+
+    def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: str) -> None:  # noqa: ARG002
+        if stage != "fit":
+            return
```
HCookie, line 133: Does this effectively check that it's not be called in sanity checking?
Seems a bit fragile, as the attribute this sets is needed later

training/src/anemoi/training/diagnostics/callbacks/weight_averaging.py
```diff
@@ -7,81 +7,245 @@
...
 from pytorch_lightning.callbacks import Callback
+from pytorch_lightning.callbacks import WeightAveraging as _PLWeightAveraging
+from pytorch_lightning.utilities.model_helpers import is_overridden
+from pytorch_lightning.utilities.rank_zero import rank_zero_warn
+from torch.optim.swa_utils import AveragedModel as _TorchAveragedModel
+from torch.optim.swa_utils import get_ema_avg_fn
+from torch.optim.swa_utils import get_swa_avg_fn
+from torch.optim.swa_utils import get_swa_multi_avg_fn
+from torch.utils._foreach_utils import _get_foreach_kernels_supported_devices
+from torch.utils._foreach_utils import _group_tensors_by_device_and_dtype
```
HCookie: I'm slightly uncomfortable on the use of these "private" methods.

anaprietonem:
@jakob-schloer I am marking this as draft, as from what I understand, it might be better to check if there are ways we can adapt the dynamic scalers and imputers so that we don't have to hack the EMA callback so much. And @ssmmnn11 is looking at preparing a branch with some suggestions.

training/src/anemoi/training/diagnostics/callbacks/weight_averaging.py
jakob-schloer: I will introduce a NamedTuple class to handle the arguments more cleanly.

training/src/anemoi/training/diagnostics/callbacks/weight_averaging.py
jakob-schloer: The setup hook only runs once per Trainer entrypoint, with stage being either "fit" | "validate" | "test" | "predict". We only build the average model for stage=='fit' — validate/test/predict have nothing to average. 

The validation sanity check runs inside fit, after setup(stage="fit") has already executed. So by the time sanity checking happens, the  average model is already built. So this is not about sanity-checking at all. 

The check also simply mimics the upstream PL implementation:
https://github.com/Lightning-AI/pytorch-lightning/blob/fe6b1cc4e80ae0396e2e404c16e6b6968ad5437e/src/lightning/pytorch/callbacks/weight_averaging.py#L145

training/src/anemoi/training/diagnostics/callbacks/weight_averaging.py
jakob-schloer: Yes, I was simply using the same logic as the upstream PL scripts, but we can get rid of them. At the moment, both private functions are only used by the PL standard SWA model using the "foreach" parameter update, i.e. `multi_avg_fn`. As far as I understand, this is simply an optimized parameter update of the `avg_fn`. We can drop the `multi_avg_fn` because we don't use it in the custom implementations.

HCookie (approved):
Thanks for addressing my comments, looks good to me

PR #1129, opened 2026-05-20
feat: Abstract asynchronous plotting executor

### LINKED ISSUE #1127: Abstract asynchronous plotting executor

As describer in https://github.com/ecmwf/anemoi-core/issues/769, we want to abstract the asyncronous executor so that we can have. a base class and then extend that for the asynchronous part. 

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

anaprietonem:
## Description
Existing implementation mixed async infrastructure (thread, loop, executor) directly into BasePlotCallback, making it hard to reason about, test independently, or swap execution strategies. The new design follows the strategy pattern, the callback holds an executor and calls schedule; the executor owns all concurrency details.

Summary of the changes:

- Introduces BasePlotExecutor as an abstract base class defining a unified schedule(fn, trainer, *args, **kwargs) / shutdown() interface for plot execution strategies.
- Adds SyncPlotExecutor — a trivial implementation that calls the plot function immediately on the calling thread.
- Adds AsyncPlotExecutor — migrates the existing background-thread + asyncio event loop logic out of BasePlotCallback and into a self-contained class, making async execution an implementation detail rather than scattered state on the callback.
- Simplifies BasePlotCallback: removes the _plot / _async_plot dual-path, start_event_loop, submit_plot, and _plot_with_error_catching methods. The plot() method now always delegates to self._executor.schedule(...), with sync or async behavior determined by which executor is instantiated at construction. teardown() is reduced to a single self._executor.shutdown() call.
- Error handling (os._exit(1) on unhandled exceptions) is centralised in BasePlotExecutor._run, so both executors benefit from it consistently.
- Adds a new unit test file (test_plot_executors.py) to check sync execution correctness, error handling for both executors, async scheduling, argument forwarding, and shutdown behaviour.

https://github.com/ecmwf/anemoi-core/issues/1127

[](https://github.com/ecmwf/anemoi-core/issues/1127)

##  Additional notes ##


anaprietonem:
Integration tests https://github.com/ecmwf/anemoi-core/actions/runs/26149743218

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -108,19 +187,16 @@ def __init__(
 
         init_plot_settings()
 
-        self.plot = self._plot
-        self._executor = None
         self._error: BaseException = None
```
VeraChristina: this doesn't seem to be used anywhere

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -249,34 +301,15 @@ def _plot(
...
-        # run_in_executor doesn't support keyword arguments,
-        await loop.run_in_executor(
-            self._executor,
-            self._plot_with_error_catching,
-            trainer,
-            args,
-            kwargs,
-        )  # because loop.run_in_executor expects positional arguments, not keyword arguments
-
     @rank_zero_only
```
VeraChristina, line 312: Do we need the rank zero guard on both `plot` and `_plot`?

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -77,6 +77,85 @@ def from_plot_config(cls, plot_cfg: DictConfig, save_basedir: str | Path | None)
...
+    Runs a single-threaded executor backed by a dedicated asyncio event loop,
+    allowing plot functions to be submitted from the main thread without blocking it.
+    """
+
+    def __init__(self) -> None:
+        self._executor = ThreadPoolExecutor(max_workers=1)
+        self.loop: asyncio.AbstractEventLoop | None = None
+        self._loop_ready = threading.Event()
+        self.loop_thread = threading.Thread(target=self._start_event_loop, daemon=True)
+        self.loop_thread.start()
```
VeraChristina: Do we want `loop` and `loop_thread` to be public?

VeraChristina (commented):
Thanks for making this part of the code more readable by extracting the sync/async execution logic! A couple of minor comments below.

When I run the global integration test with `asynchronous: True` I get a couple of warnings "Task was destroyed but it is pending!" -- However, all plots are present and look fine.

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: good catch, I removed it from _plot!

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: indeed I have removed it now

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: yep fair, updated now too!

anaprietonem:
Thanks for the review @VeraChristina, I have addressed the comments and improved the shutting down system so hopefully the messages for task being destroyed should now be gone or at least minimised. I tested locally with 2 gpus and results looked fine. I am running IT one more time here https://github.com/ecmwf/anemoi-core/actions/runs/26275717140

VeraChristina (approved):
Thanks for addressing the feedback, LGTM!

PR #1130, opened 2026-05-20
fix(training): gather sharded batches if loss does not support sharding

frazane:
This PR makes sure that in the ensemble training the batches are actually gathered before computing the loss, if the loss does not support sharding.

---


japols:
good spot, thanks for the fix!

have you tried testing this with keep_batch_sharded=False yet? happy to approve after that.

frazane:
@japols yes, it runs with `keep_batch_sharded=False` without problems.

japols (approved):
LGTM

PR #1131, opened 2026-05-20
fix(training): allow setting blosc num threads via config

cathalobrien:
## Description
This PR adds an optional entry to the **dataloader** config which sets how many threads is used by blosc to decompress Zarr chunks
```
blosc_num_threads: PositiveInt | None = Field(default=None)
"Sets how many threads used when decompressing Zarr chunks. If None, the thread count will be determined by Blosc."
```
The default value is None, meaning the thread count will be determined by blosc. In my tests this is 8 threads (numcodecs v0.15.1 on AG).

In theory, changing the number of threads could impact performance. But in my tests there were no performance difference between 1 and 64 threads while training an n320 model. However, this could vary depending on the CPU node and the number of dataloader processes available.

Multiple users have reported this error on arm architectures like Grace-Hopper systems. This error is non-deterministic 

The fix for this error is to set **blosc_num_threads=1** 
```
  File "/nfs/dh2_02_perm_a/ecm1032/venvs/myvenv_ag_test/.venv/lib/python3.12/site-packages/anemoi/datasets/data/stores.py", line 230, in __getitem__
    return self.data[n]
           ~~~~~~~~~^^^
  File "/nfs/dh2_02_perm_a/ecm1032/venvs/myvenv_ag_test/.venv/lib/python3.12/site-packages/zarr/core.py", line 796, in __getitem__
    result = self.get_orthogonal_selection(pure_selection, fields=fields)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nfs/dh2_02_perm_a/ecm1032/venvs/myvenv_ag_test/.venv/lib/python3.12/site-packages/zarr/core.py", line 1078, in get_orthogonal_selection
    return self._get_selection(indexer=indexer, out=out, fields=fields)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nfs/dh2_02_perm_a/ecm1032/venvs/myvenv_ag_test/.venv/lib/python3.12/site-packages/zarr/core.py", line 1341, in _get_selection
    self._chunk_getitems(
  File "/nfs/dh2_02_perm_a/ecm1032/venvs/myvenv_ag_test/.venv/lib/python3.12/site-packages/zarr/core.py", line 2186, in _chunk_getitems
    self._process_chunk(
  File "/nfs/dh2_02_perm_a/ecm1032/venvs/myvenv_ag_test/.venv/lib/python3.12/site-packages/zarr/core.py", line 2099, in _process_chunk
    chunk = self._decode_chunk(cdata)
            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nfs/dh2_02_perm_a/ecm1032/venvs/myvenv_ag_test/.venv/lib/python3.12/site-packages/zarr/core.py", line 2355, in _decode_chunk
    chunk = self._compressor.decode(cdata)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "numcodecs/blosc.pyx", line 585, in numcodecs.blosc.Blosc.decode
  File "numcodecs/blosc.pyx", line 414, in numcodecs.blosc.decompress
RuntimeError: error during blosc decompression: 0

```

----
📚 Documentation preview 📚: https://anemoi-training--1131.org.readthedocs.build/en/1131/

----
📚 Documentation preview 📚: https://anemoi-graphs--1131.org.readthedocs.build/en/1131/

----
📚 Documentation preview 📚: https://anemoi-models--1131.org.readthedocs.build/en/1131/

### DIFF
--- a/training/src/anemoi/training/utils/worker_init.py
+++ b/training/src/anemoi/training/utils/worker_init.py
@@ -14,12 +14,10 @@ from numcodecs.blosc import get_nthreads
 from numcodecs.blosc import set_nthreads
 from torch.utils.data import get_worker_info
 
-from anemoi.training.schemas.base_schema import BaseSchema
-
 LOGGER = logging.getLogger(__name__)
 
 
-def worker_init_func(worker_id: int, dataloader_config: BaseSchema) -> None:
+def worker_init_func(worker_id: int, blosc_num_threads: int | None = None) -> None:
     """Configures each dataset worker process.
 
     Calls per_worker_init() on each dataset object.
@@ -28,8 +26,9 @@ def worker_init_func(worker_id: int, dataloader_config: BaseSchema) -> None:
     ----------
     worker_id : int
         Worker ID
-    dataloader_config : BaseSchema
-        Dataloader configuration
+    blosc_num_threads : int, optional
+        Number of threads to use for Blosc decompression, by default None.
+        If None, the number of threads will be determined by Blosc.
 
     Raises
     ------
@@ -38,7 +37,6 @@ def worker_init_func(worker_id: int, dataloader_config: BaseSchema) -> None:
 
     """
     # read the dataloader config to set the number of threads blosc should use when decompressing zarr chunks.
-    blosc_num_threads = dataloader_config.get("blosc_num_threads", None)
     if blosc_num_threads is not None:
         set_nthreads(blosc_num_threads)
         LOGGER.info("Set number of threads used by Blosc to %d.", blosc_num_threads)
--- a/training/src/anemoi/training/utils/worker_init.py
+++ b/training/src/anemoi/training/utils/worker_init.py
@@ -10,7 +10,8 @@
 
 import logging
 
-from numcodecs.blosc import set_nthreads, get_nthreads
+from numcodecs.blosc import get_nthreads
+from numcodecs.blosc import set_nthreads
 from torch.utils.data import get_worker_info
 
 from anemoi.training.schemas.base_schema import BaseSchema
--- a/training/src/anemoi/training/utils/worker_init.py
+++ b/training/src/anemoi/training/utils/worker_init.py
@@ -10,7 +10,7 @@
 
 import logging
 
-from numcodecs.blosc import set_nthreads as blosc_set_nthreads
+from numcodecs.blosc import set_nthreads, get_nthreads
 from torch.utils.data import get_worker_info
 
 from anemoi.training.schemas.base_schema import BaseSchema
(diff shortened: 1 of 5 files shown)


training/src/anemoi/training/utils/worker_init.py
```diff
@@ -10,12 +10,16 @@
...
+from numcodecs.blosc import set_nthreads
 from torch.utils.data import get_worker_info
 
+from anemoi.training.schemas.base_schema import BaseSchema
+
 LOGGER = logging.getLogger(__name__)
 
 
-def worker_init_func(worker_id: int) -> None:
+def worker_init_func(worker_id: int, dataloader_config: BaseSchema) -> None:
```
MeraX: Hi @cathalobrien thanks for this PR. Did you consider just passing blosc_num_threads here? This would disentangle the worker from the rather complex and intransparent config object.

anaprietonem: I agree with Marek, it's a nice contribution but I think it'd cleaner to pass the blosc_num_threads to the worker_init_function rather than the whole dataloader_config.

training/src/anemoi/training/utils/worker_init.py
```diff
@@ -24,13 +28,23 @@ def worker_init_func(worker_id: int) -> None:
...
+        Dataloader configuration
 
     Raises
     ------
     RuntimeError
         If worker_info is None
 
     """
+    # read the dataloader config to set the number of threads blosc should use when decompressing zarr chunks.
+    blosc_num_threads = dataloader_config.get("blosc_num_threads", None)
```
anaprietonem: Could we add a line to the docs about what value should users/developers set for this? (ie when is it useful to leave the default and when is better to overwrite it?)

training/src/anemoi/training/utils/worker_init.py
```diff
@@ -24,13 +28,23 @@ def worker_init_func(worker_id: int) -> None:
...
+    if blosc_num_threads is not None:
+        set_nthreads(blosc_num_threads)
```
anaprietonem, line 41: what's is this doing? is it defining an ENV variables that is later used by torch in the per_worker_init set up?

anaprietonem: Would it be feasible to add a unit test to showcase how this parameter behaves?

anaprietonem (commented):
Thanks Cathal, a few comments! nice work!

anaprietonem:
From the error stack trace, the error is also being triggered in zarr in anemoi-datasets, so have you consider if the fix should be applied there? or you see benefit from being able to configure it at training time?

training/src/anemoi/training/utils/worker_init.py
cathalobrien: sure, done

training/src/anemoi/training/utils/worker_init.py
cathalobrien: it calls [this](https://numcodecs.readthedocs.io/en/v0.15.1/compression/blosc.html#numcodecs.blosc.set_nthreads) function, which changes how many threads numcodecs uses internally. [Internally](https://github.com/Blosc/c-blosc/blob/c8072c21dcce8daf6dd19a0bbae8c53458ad68cc/blosc/blosc.c#L1958) it (re)initialises blosc with the new thread count 

regarding a unit test, sure I'll add one

training/src/anemoi/training/utils/worker_init.py
cathalobrien: good idea, done!
https://anemoi--1131.org.readthedocs.build/projects/training/en/1131/troubleshooting.html#error-during-blosc-decompression

PR #1136, opened 2026-05-21
fix(training): enable coupling of validation rollout to training rollout

ssmmnn11:
currently validation rollout is independent from training rollout which means the validation loss is not directly comparable to the training loss. This pull request makes the validation rollout setting optional, if not set, the training rollout is used in validation.

### DIFF
--- a/training/src/anemoi/training/tasks/forecaster.py
+++ b/training/src/anemoi/training/tasks/forecaster.py
@@ -63,7 +63,7 @@ class Forecaster(BaseTask):
         multistep_output: int,
         timestep: str,
         rollout: dict | None = None,
-        validation_rollout: int = 1,
+        validation_rollout: int | None = None,
         **kwargs,
     ) -> None:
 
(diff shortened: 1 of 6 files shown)


anaprietonem:
@sahahner @jakob-schloer it would be good if you can have a look at this.  As Simon mentioned above it makes it more robust to ensure the validation rollout used the settings defined in the training mode but would be good you can confirm this doesn't affect the RolloutEval callback for which the validation_rollout parameter is used

sahahner:
Thanks Simon, makes sense that the validation metrices should always track at least the current rollout used in training. I thought actually, that this is what the current validation is doing? 
I see a tracked per-variable metric for the current rollout step (for single models, using the default metric tracking, no additional callback), for example here:
Screenshot 2026-05-22 at 14 27 11

What validation loss/metric is this effecting?

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -114,20 +116,24 @@ def get_offsets(self, mode: str | None = None) -> list[datetime.timedelta]:
         if mode == "training":
             rollout_step = self.rollout.maximum
         elif mode == "validation":
-            rollout_step = self.validation_rollout
+            rollout_step = self.rollout.maximum if self.validation_rollout is None else self.validation_rollout
         else:
             LOGGER.debug(
                 "Unknown mode '%s' for %s.get_offsets(), defaulting to training rollout.",
```
jakob-schloer:
```suggestion
                "Unknown mode '%s' for %s.get_offsets(), defaulting to maximum of training and validation rollout.",
```

jakob-schloer (dismissed):
Looks good to me. I ran a couple of test combinations with the RolloutEvalCallbacks which worked fine.

HCookie:
For awareness, I am fixing the rollout full loading in this PR. https://github.com/ecmwf/anemoi-core/pull/1156
It might have some fun conflicts.

jakob-schloer (approved):
LGTM! Thanks for your work @ssmmnn11.

PR #1118, opened 2026-05-15
feat(graphs): Enable multi-level lam resolution in stretched grid with cutout mask #1115

jasonjunweilyu:
## Description

This PR enables stretched-grid graph to have multi-level lam resolutions. A function is added to create multi stretched-grid tri nodes. Stretched grid node builder and multi-scale edges are updated for this feature. Tests are added for the stretched grid node builder and updated for multi-scale edges.

## What problem does this change solve?

New feature.

## What issue or task does this change relate to?

#1115 

##  Additional notes ##

See example multi-resolution lam region test config, log and plots in the comment section below.
For single-resolution lam stretched-grid, this PR is backward compatible, so no change to the previous config. 

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

PR #1137, opened 2026-05-21
feat(graphs): add MutualKNNEdges edge builder

frazane:
Adds a `MutualKNNEdges` edge builder: an edge is kept only if the target is among the source node's nearest targets and the source is among the target node's nearest sources — the intersection of `KNNEdges` and `ReversedKNNEdges`.

Across a resolution transition (e.g. the global/regional boundary of stretched grid), `KNNEdges`/`CutOffEdges` give lopsided, one-sided coupling. `MutualKNNEdges` tapers connectivity symmetrically and bounds node degree on
both sides. `num_nearest_neighbours` and `reversed_num_nearest_neighbours` may differ to control fan-in/fan-out independently; with `source_mask_attr_name` / `target_mask_attr_name` it confines edges to the boundary.

Note that this was developed in tandem with AI coding agents.

### Summary of changes
- `intersect_edges` helper in `utils.py`
- `MutualKNNEdges` in `edges/builders/knn.py` (torch-cluster + scikit-learn paths)
- `MutualKNNEdgeSchema` added to the recipe schema union
- "Mutual K-Nearest Neighbours" section in `docs/graphs/edges/knn.rst`
-  Unit, schema, and end-to-end tests.

### Examples

This is what we would get as cross-boundary edges when using something like cutoff-based edges:

```yaml
_target_: anemoi.graphs.edges.CutOffEdges
    cutoff_distance_km: 200
    max_num_neighbours: 22
    source_mask_attr_name: coarse_mask
    target_mask_attr_name: dense_mask
```
image

And this is with the mutual KNN edges, which gives a more homogeneous connectivity for the nodes at the boundaries:

```yaml
_target_: anemoi.graphs.edges.MutualKNNEdges
    num_nearest_neighbours: 6
    source_mask_attr_name: coarse_mask
    target_mask_attr_name: dense_mask
```
image

---

----
📚 Documentation preview 📚: https://anemoi-training--1137.org.readthedocs.build/en/1137/

----
📚 Documentation preview 📚: https://anemoi-graphs--1137.org.readthedocs.build/en/1137/

----
📚 Documentation preview 📚: https://anemoi-models--1137.org.readthedocs.build/en/1137/


JPXKQX:
Hi Francesco, I think this looks really cool, thanks! The approach we are currently following for contributions to new features is to demonstrate some benefit in terms of computational cost, skill, or other factors. Some other PRs are already waiting to showcase this, such as #1118 and #1086.
I think this could homogeneise the connectivity of those nodes in the boundary and have a positive impact, but it would be good to see some plots of this. Thanks again for your contributions

frazane:
> Hi Francesco, I think this looks really cool, thanks! The approach we are currently following for contributions to new features is to demonstrate some benefit in terms of computational cost, skill, or other factors. Some other PRs are already waiting to showcase this, such as #1118 and #1086. I think this could homogeneise the connectivity of those nodes in the boundary and have a positive impact, but it would be good to see some plots of this. Thanks again for your contributions

Hi Mario, I'll try to come up with some plots. However, I would like to point out that this is the possibly the most natural way to handle connectivity across regional boundaries, and I see no reason not to implement it. Adding a new KNN variant also enables others to explore other graph designs: it's not only about regional modelling. Additionally, please consider that while I agree with 

> we are currently following for contributions to new features is to demonstrate some benefit in terms of computational cost, skill, or other factors

making this a hard requirement is a bit unreasonable. I would argue it should be on a case-by-case basis, weighing added complexity, size of the PR (note that here ~300 of the total LOC changes is tests and docs, and there are zero changes to the existing code), and how hard it is to demonstrate the benefit. As you might be aware, training regional models is quite the effort. One of the main benefits I would expect from this feature is a slightly improved handling of dynamic processes (e.g. fronts) across the boundaries*. However, running an empirical comparison for such a small, natural change, is not a luxury that we can afford, both in terms of time and compute. So I will try to focus on other aspects like model throughput and peak memory and hopefully we will see some benefits there.

*despite the wide perception field of the processor, we think the model might still be relying on closest connections a lot because we see some artifacts at the boundaries, especially where the connectivity is sparse

frazane:
Perhaps this PR was framed the wrong way? I am not even sure this is a "new feature", but more of a convenience / quality of life change for us.

We already anyway invest quite some time in finding the optimal graph connectivity at the boundaries to avoid orphan nodes while having homogeneous degrees. For now we've done it by iterating and tweaking numbers for hours (our graph generation takes several minutes), and it's still sub-optimal. With this PR you can save time and get a better result.

mpvginde:
> We already anyway invest quite some time in finding the optimal graph connectivity at the boundaries to avoid orphan nodes while having homogeneous degrees. For now we've done it by iterating and tweaking numbers for hours (our graph generation takes several minutes), and it's still sub-optimal. With this PR you can save time and get a better result.

I second this. This PR gives us much more direct control over boundary graph-connectivity.

dietervdb-meteo (commented):
Nice functionality and implementation, tested it in our in house setup and seems to work smoothly. 
Two comments:
1. couldn't we save some lines of code and improve maintainability by subclassing `MutualKNNEdges` off of `KNNEdges` (or the other way around?) 
2. It could worth changing the default stretched grid graph config to use this more refined setup using the MutualKNNEdges. It will increase visibility and also automatically have it pass integration tests as well as benchmarking. It would extend the scope of this PR a little bit since it would also touch training...

frazane:
Hi @dietervdb-meteo, glad it ran smoothly in your setup!

1. The only shared code would be the positive-int check in `__init__`; both `_compute_*` methods run a forward and a reversed KNN and intersect, so they'd be fully overridden with no reuse. Subclassing would save a couple of lines but the relationship is ambiguous: mutual KNN is not KNN, it's the intersection of two KKNs (ReversedKNNEdges can on the other hand subclass KNNEdges because it's just the same thing in reverse).

2. Agreed it'd help visibility and test coverage, but I think setting a new functionality as default is a bit risky. I would be in favour of doing it in a follow up, after more people from the community have worked with it for a while.

dietervdb-meteo (approved):
Ok, makes sense to postpone a possible change in the default config.

I see no objection to merging apart from the tests being fixed.

PR #1140, opened 2026-05-26
feat: units check

VeraChristina:
## Description
This PR adds unit checks when resuming training runs or when transferring weights from a checkpoint. The aim is to ensure consistent units in datasets across different stages of training.
- It gathers variables_metadata from the checkpoint after loading
- It compares the units found in the checkpoint to the metadata of the datasets used in the current run, using functionality from anemoi-transform. For each dataset that is both in the checkpoint and in the current run:
   - If either metadata is missing (old or new), a warning is logged and the run proceeds
   - If both exist and there is a mismatch between old and new, a ValueError is raised

At the moment we do not check for unit compatibility between the datasets used in different steps of training when resuming a run or during transfer learning. This can lead to unnoticed inconsistencies between datasets.

I have tested the PR with a dummy checkpoint, trained on era5 and transfer learning/resuming with era6 (https://anemoi.ecmwf.int/datasets/aifs-e6-an-oper-5503-mars-o96-2016-2017-6h-v0). As expected, a value error is raised for tp:
```tp: Units are not compatible: m vs kg m**-2```

This PR addresses step 1 of https://github.com/ecmwf/anemoi-core/issues/1134

##  Additional notes ##
- Since the two paths (resuming versus transfer learning) load the old checkpoint at different points, we implement the unit checks at suitable points in the two paths after the checkpoints are loaded.

### CODE REVIEW

File: training/src/anemoi/training/data/datamodule.py [hunk 1]
```diff
@@ -200,3 +200,8 @@ def fill_metadata(self, metadata: dict) -> None:
                 "diagnostic": [input_index_to_name[int(index)] for index in input_data_indices["diagnostic"]],
             }
             metadata["metadata_inference"][dataset_name]["variable_types"] = variable_types
+
+            dataset_metadata = datasets_config.get(dataset_name, {})
+            variables_metadata = dataset_metadata.get("variables_metadata")
```
anaprietonem: is this just units or more things?

VeraChristina: This is potentially more things, like mars request etc.; the compatibility check [here](https://github.com/ecmwf/anemoi-transform/blob/43edb1e67eba33965af23b7c55412246a9541525/src/anemoi/transform/variables/__init__.py) also checks for other things, e.g. surface vs pressure vs model level if that information is available.

We could only store the units in the metadata_inference if we prefer, @gmertes any thoughts?

gmertes: Adding the full variables_metadata to the metadata_inference is a good idea 👍 Makes it explicit that this information is used in inference.

training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -20,6 +20,63 @@
...
+                "Skipping unit compatibility check for this dataset.",
+                dataset_name,
+            )
+            continue
+
+        ckpt_vars = {name: Variable.from_dict(name, data) for name, data in ckpt_var_meta.items()}
+        ds_vars = {name: Variable.from_dict(name, data) for name, data in ds_var_meta.items()}
+
+        try:
+            Variable.check_compatibility(ckpt_vars, ds_vars)
```
gmertes, line 89: We need a version bump for anemoi-transform: check_compatibility only exists in >=0.3.0

training/src/anemoi/training/utils/checkpoint.py
```diff
@@ -125,6 +125,16 @@ def transfer_learning_loading(model: torch.nn.Module, ckpt_path: Path | str) ->
...
+    # Extract variables_metadata for unit compatibility check
+    metadata = checkpoint.get("hyper_parameters", {}).get("metadata", {}).get("metadata_inference", {})
+    ckpt_variables_metadata = {}
+    for dataset_name in model._ckpt_model_name_to_index:
+        ds_inference = metadata.get(dataset_name, {})
+        vm = ds_inference.get("variables_metadata")
+        if vm is not None:
+            ckpt_variables_metadata[dataset_name] = vm
+    model._ckpt_variables_metadata = ckpt_variables_metadata or None
+
```
gmertes, line 134: This looks almost identical to the snippet in the BaseTrainingModule (except for model vs self). It's small enough that I think a copy-paste is ok, but it risks them getting out of sync. Do you think it's worth putting this in a utility function? This utils/checkpoint.py file seems like a good place for it.

gmertes (commented):
Very nice 👌 

Do I see it correctly that it currently checks all variables in the dataset, but doesn't account for any variables that are dropped and unused by the model? Should we limit to only checking variables that are used? (idk if that's hard to do, or if that's what we want so feel free to ignore)

training/src/anemoi/training/utils/variables_metadata.py
VeraChristina: good catch, done!

training/src/anemoi/training/utils/checkpoint.py
VeraChristina: Good point, that's why I put the check into a utility function, have added this one to utils/variables_metadata.py (to avoid circular imports in checkpoint.py)

anaprietonem (dismissed):
Thanks Vera for addressing the feedback, this LGTM to me. Could be good to add the check from the integration tests as a comment to show those are passing.

VeraChristina:
Added a small check to the dataset test to make sure we also drop variables_metadata for dropped variables (re Gert's comment above). 
And here are the integration tests: https://github.com/ecmwf/anemoi-core/actions/runs/26631766651

anaprietonem (dismissed):

VeraChristina:
@gmertes I have removed the variables_metadata from metadata_inference as discussed, and updated the rest accordingly

gmertes (approved):

Issue #1142, opened 2026-05-27
Restrict spectral loss to `output_mask`


dietervdb-meteo:
### What happened?

# Problem
Loss for LAM models should only be computed over the area defined by the `output_mask`. For spatial losses this is automatically taken care of through the `GraphNodeAttributeScaler` (see [here](https://github.com/ecmwf/anemoi-core/blob/f0743b78893f04bff18057b68ca81e17f3c777b4/training/src/anemoi/training/losses/scalers/node_attributes.py#L64)). For spectral losses such an implementation is lacking.

# Proposed solution
1. allow arbitrary `kwargs` in losses
2. pass `output_mask` as kw-argument to the generic loss call
3. have base spectral loss pick up `output_mask`, use it to crop `x` to the regional domain before the spectral transform is applied

# Comment
A more comprehensive and less ad hoc/per loss treatment of `output_mask` would be a nicer long term solution, but this will be easier to implement after some refactoring related to observations will be implemented, and is thus currently postponed. Hence the rather minimal proposed solution above.

### What are the steps to reproduce the bug?

Have a look at the current implementation.

### Version

current main

### Platform (OS and architecture)

irrelevant

### Relevant log output

```shell

```

### Organisation

RMI

dietervdb-meteo:
Fix implemented via #1087 .

PR #1143, opened 2026-05-27
fix: spectral losses for lam

dietervdb-meteo:
## Description
Makes `SpectralLoss` respect `output_mask`, i.e. for lam models loss should be computed only on regional domain. 

Spectral transform, and then spectral loss, is [automatically] computed only over region determined by `output_mask`, as was already the case for spatial losses.

#1142

##  Additional notes ##
Additional arguments for SpectralLosses, such as e.g. `x_dim` and `y_dim` as set in the config should be wrt the region selected by the `output_mask`, not wrt the full data grid.

Currently draft PR. Tests will be added once some sanity checks have been performed.

### DIFF
--- a/training/src/anemoi/training/losses/base.py
+++ b/training/src/anemoi/training/losses/base.py
@@ -47,6 +47,7 @@ class BaseLoss(nn.Module, ABC):
     def __init__(
         self,
         ignore_nans: bool = False,
+        **kwargs,
     ) -> None:
         """Node- and feature_weighted Loss.
 
@@ -70,6 +71,7 @@ class BaseLoss(nn.Module, ABC):
             Allow nans in the loss and apply methods ignoring nans for measuring the loss, by default False
 
         """
+        del kwargs
         super().__init__()
 
         self.add_module("scaler", ScaleTensor())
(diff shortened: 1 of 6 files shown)


ssmmnn11:
would it be an option to use a projection matrix for that:

#1087 

just do a 1-to-1 mapping for the points you want to keep?

dietervdb-meteo:
Thanks @ssmmnn11  for the heads up. Indeed looks like it can be rewired to use some of that functionality. At first sight  through `node_slice` rather than through a sparse mapping though, i.e. it seems that all one needs to do is to translate the lam `output_mask` to the appropriate `node_slice`. Will have a look.

ssmmnn11:
why not via a sparse mapping?

dietervdb-meteo:
If you need to get part of a tensor, isn't slicing more straightforward than a projection? At least conceptually. Would the projection be able to handle non-contiguous cases? (but so far `output_mask` is `True` on a contiguous set of indices in all use cases I'm aware of). We also want to change the dimension size, i.e. not fill with a bunch of zeros, would the sparse projection do that?

frazane:
@dietervdb-meteo indeed you should be good using the slice if it already gives you a set of points that can be reshaped to a 2-dimensional regular grid. The projection part in #1087 is only necessary when the data is unstructured and needs a regridding/projection before being reshaped (for instance in the case of the ICON regional grid) so it can be passed to the 2D transforms.

Note that the slice option is basically made ad-hoc for a "cutout tensor". It's true that it won't work in the non-contiguous case - we would need an explicit index there.

dietervdb-meteo:
I did a refactor of this using the #1087 functionality, see #1174. I set it up as a PR to #1087. The discussion can continue there. Will close this PR.

PR #1145, opened 2026-05-27
fix(training): load model on CPU to avoid GPU OOMs

elkir:
## Description
When using an existing checkpoint through the `load_weights_only=true`, `transfer_learning=false`, the current behaviour uses implicit `map_location=None`, which is meant to reload the model onto the device it was saved from.

But we load the checkpoint back to the GPU ( where `rank:0` saved it) only to rewrite the fresh model's parameters on the CPU, and that is then moved back to the GPU when `Trainer.fit()` is called

This explicitly loads the checkpoint on the `"cpu"` to begin with.

Some additional testing on NVIDIA GPUs might be needed to confirm there are no negative effects.

This was revealed to us when we encountered OOM errors when fork-reloading model weights from a model sharded across 8 GPUs, which was previously trained and fitting the memory fine.

The OOM might be an AMD-specific issue, as at the OOM time during the load of the 13GB checkpoint, is caused by a clash with 56 GB/64GB total is held outside PyTorch's caching allocator (HIP runtime, RCCL??). This fix solves this OOM.


cathalobrien (dismissed):
Hey @elkir, thanks for the contribution.

The suggestion seems reasonable. I extended the same change to transfer_learning_loading() to stay consistent, and added some comments explaining the rationale.

I tested this branch by running the `training/tests/unit/train/test_checkpoint_loading.py` tests on a node with an Nvidia GPU. I modified the tests so that the original models were created on a GPU and loaded onto a CPU. the tests passed fine

cathalobrien (approved):

japols (approved):

PR #1141, opened 2026-05-27
feat(training): Initial spectral AMSE implementation

sahahner:
This pull request introduces a new loss function to the spectral losses module and updates the module's public API accordingly. The most significant change is the addition of the `SpectralAMSELoss` class, which implements the AMSE loss based on a recent research paper. The changes are as follows:

**New Loss Function Implementation:**

* Added the `SpectralAMSELoss` class to `spectral.py`, implementing the AMSE loss in the spectral domain as described by Subich et al. (arXiv:2501.19374, 2025). This loss computes amplitude and coherence differences between predicted and target spectra, and is compatible with multiple spectral transforms.

**API and Import Updates:**

* Added `SpectralAMSELoss` to the list of imports in `losses/__init__.py`, making it available for use throughout the package.
* Included `"SpectralAMSELoss"` in the `__all__` list in `losses/__init__.py`, ensuring it is part of the module's public API.

##  Additional notes ##
AMSE was introduced in Subich et al, 2025.
@jameschappell has thankfully provided their implementation and we are incorporating it in anemoi.

see https://github.com/ecmwf/anemoi-core/issues/1147

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

PR #1148, opened 2026-05-28
feat(training): Initial spectral AMSE implementation (#1141)

### LINKED ISSUE #1147: Introduce Adjusted Mean Squared Error

_No response_

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

sahahner:
This pull request introduces a new loss function to the spectral losses module and updates the module's public API accordingly. The most significant change is the addition of the `SpectralAMSELoss` class, which implements the AMSE loss based on a recent research paper. The changes are as follows:

**New Loss Function Implementation:**

* Added the `SpectralAMSELoss` class to `spectral.py`, implementing the AMSE loss in the spectral domain as described by Subich et al. (arXiv:2501.19374, 2025). This loss computes amplitude and coherence differences between predicted and target spectra, and is compatible with multiple spectral transforms.

**API and Import Updates:**

* Added `SpectralAMSELoss` to the list of imports in `losses/__init__.py`, making it available for use throughout the package.
* Included `"SpectralAMSELoss"` in the `__all__` list in `losses/__init__.py`, ensuring it is part of the module's public API.

##  Additional notes ##
AMSE was introduced in Subich et al, 2025.
@jameschappell has thankfully provided their implementation and we are incorporating it in anemoi.

see https://github.com/ecmwf/anemoi-core/issues/1147

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License
Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

##  Additional notes ##


sahahner:
Reagrding this comment from @anaprietonem 

https://github.com/ecmwf/anemoi-core/pull/1141#issuecomment-4553956892

I am happy to run a test with the AMSE for a global model.

training/src/anemoi/training/losses/spectral.py
```diff
@@ -130,6 +130,99 @@ def _to_spectral_flat(self, x: torch.Tensor) -> torch.Tensor:
...
+        grid_shard_slice: slice | None = None,
+        group: ProcessGroup | None = None,
+        squash_mode: str = "avg",
+        **kwargs,
+    ) -> torch.Tensor:
+        del kwargs  # unused
+        is_sharded = grid_shard_slice is not None
+        group = group if is_sharded else None
+
+        with torch.amp.autocast(device_type="cuda", enabled=False):
```
frazane:
```suggestion
        with torch.amp.autocast(device_type=pred.device.type, enabled=False):
```

training/src/anemoi/training/losses/spectral.py
```diff
@@ -130,6 +130,99 @@ def _to_spectral_flat(self, x: torch.Tensor) -> torch.Tensor:
...
+      and :math:`M` is the zonal wavenumber, consistent with the original paper.
+      The sum over :math:`M` gives per-total-wavenumber power spectra.
+    - ``fft2d`` / ``dct2d``: :math:`L` is the meridional frequency index and
+      :math:`M` is the zonal frequency index. The sum over :math:`M` gives
+      per-meridional-wavenumber power spectra. This is mathematically valid but
+      no longer corresponds to the spherical-harmonic-motivated formulation of
+      the paper. For ``dct2d``, coefficients are real-valued; computing coherence
+      per coefficient before summing would degenerate to :math:`\pm 1`, but the
+      :math:`M`-sum here avoids this by computing coherence as a cosine similarity
+      over the zonal-frequency profile at each :math:`L`.
```
frazane: As mentioned in #992, I think in general the following rule should apply: the loss should know about the formula of the metric, the transform should know about what its coefficients mean. It's also a good rule to easily enforce losses and transforms being two orthogonal concepts, which was in our original goals. 

As far as I understand, all terms in the AMSE formula (PSD pred, PSD target, cross) can be computed by either a SHT or a FFT/DCT. Maybe I am missing it, but I am not seeing what makes the AMSE only valid in the spherical harmonics case.

frazane (commented):
Hi Sara, thanks for working on this! My comment is pretty much the same as with the other PR about separating loss-specific and transform-specific logic.

training/src/anemoi/training/losses/spectral.py
sahahner: I was at the moment changing this. Now the AMSE fails if the used transform contains no implementation of the PSD. Could you possibly provide one for the transform that you tend to use, please?

sahahner: Indeed the AMSE could be calculated but I agree with @jameschappell here that is does something else. Would you be up to testing it on your regional setup? If not possible no worries. I think then I would argue to exclude the AMSE for 2d transforms for now.

frazane: Nice, I saw your change, I think it goes in the right direction. Except maybe

```python
cross = torch.real(pred_spec * torch.conj(target_spec)).sum(dim=-2)
```

is also still hardcoded for the SHT coefficients layout. I suppose we could have a `cross_spectral_density` or something along these lines. Having said that, computing the PSD in the FFT/DCT case could be a bit more involved and we can leave it for subsequent PRs, but at least we already have the right structure in place! Thanks for making the change.

frazane: > it does something else

In what way? I guess the most important question here is whether both implementations will preserve the "idea" for the AMSE and will provide the same incentives while training a model. Mathematically there are differences for sure, but will the signal given to the model take the same aspects into consideration (addressing the MSE smoothing problem)?

sahahner: FFT2D/DCT2D do not have an isotropic “total wavenumber” like SHT. In their case, AMSE is computed per meridional frequency band (summing over zonal frequencies), not per total wavenumber. So in that sense, yes, the AMSE signal is preserved and I guess that should still counter MSE smoothing, but in an anisotropic (grid-aligned) way. I would be interested in seeing some results and how the forecasted fields look like.

sahahner: > cross_spectral_density

I have added that function now to the SHT class.

frazane: I suppose the only thing remaining is updating the docstring, perhaps also with a TODO for implementing the methods for the FFT/DCT

sahahner: I have changed the docstrings. Feel free to have a look.

sahahner:
The AMSE implementation works as intended. Thank you very much @jameschappell for providing this nice work.
Results on a global o96 model show that the mean power of an AMSE-trained model is closer to a physics-based model in comparison to an MSE-trained model, while the MSE of medium-range forecasts (AMSE-trained) is degraded by about 5% when compared to an MSE-trained model.
spectra_10v levelist _240_2606_AMSE_June_23

frazane (approved):
Very nice, spectral losses for the win!

jameschappell:
Thanks @sahahner for pushing this over the line while I was away! Just wanted to quickly comment on the discussion regarding the FFT/DCT version of this to say that I have tested this within our FastNet codebase and also see improved spectral performance within a regional domain on a regular lat/lon grid when using the AMSE combined with a 2D FFT. While the exact meaning of the terms in the AMSE differ slightly due to the different transform used (as Sara mentioned above), this approach is still effective in improving the spectral consistency of the model so we may want to make this available too (in a future PR).

PR #1151, opened 2026-05-28
feat: custom discrete cosine transform

dietervdb-meteo:
## Description
EXPERIMENTAL

The current DCT implementation, which relies on `torch-dct`, has an issue which might be quite specific to RMI: 

1. we work with CERRA, which has a 1069x1069 grid, 
2. we work on LUMI which has AMD hardware.

These two combine in a destructive way, since 1069 is a large prime number which does not go well with the `HipFFT` backend on Rocm. 

```
RuntimeError: cuFFT error: HIPFFT_EXEC_FAILED
```

For this reason we implement a custom made 2d DCT into anemoi. It currently avoids FFT completely and works via explicit matrix multiplication, so that it will also work for grids with relatively large prime number sizes. Note that the default `torch_dct.dct_2d` does not seem a very optimized implementation, so it could be the custom implementation here improves on it in general.

This is currentlty a draft PR, performance and sanity still needs to be checked.

Since we want to test this in our particular setup (CERRA based LAM) this is PR towards #1143 rather than main.   

Example use in config:
```yaml
training_loss:
  datasets:
    data:
      scalers: ['*']
      loss_weights: [1.0, 0.1]
      losses:
        - _target_: anemoi.training.losses.CRPS
          alpha: 0.95
        - _target_: anemoi.training.losses.spectral.SpectralCRPSLoss
          scalers: ['pressure_level', 'general_variable', 'nan_mask_weights']
          alpha: 0.95
          transform: 'dct2d'
          apply_filter: False
          x_dim: 1069
          y_dim: 1069
          dct_backend: custom_dct   #new
          dct_backend_kwargs:       #new
            norm: unnormalized      #new
```

### DIFF
--- a/models/src/anemoi/models/layers/custom_dct.py
+++ b/models/src/anemoi/models/layers/custom_dct.py
@@ -1,0 +1,271 @@
+# (C) Copyright 2025 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+from __future__ import annotations
+
+import logging
+import math
+from typing import Callable
+from typing import Literal
+
+import torch
+import torch.nn as nn
+from torch import Tensor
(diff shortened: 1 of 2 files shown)


dietervdb-meteo:
We tested this, though it works technically it seems quite a bit slower than FFT without an increase in skill/spectrum. So we retire this for now.

Issue #769, opened 2025-12-19
Plotting Refactor – Exploration & Requirements Gathering


VeraChristina:
### Goal
Improve maintainability, clarity, and usability of plotting callbacks across the codebase.
Scope of this Issue
This issue is for exploration and documentation only.
Implementation will be split into separate follow-up tickets once we understand the scope.
 
### Areas to Investigate
1. Code Structure & Fragile Assumptions
•	Identify tight coupling and brittle logic within existing plotting components
•	Note areas where assumptions break with changes in training, or with new plots/use cases
•	Can we support plotting from logged outputs instead of recomputing in the callbacks
2. Sharding / Distributed Training Compatibility
•	Understand how current plots behave under sharded or distributed operation
•	Identify pain points
3. Lightning Callback Integration
•	Review how callbacks interact with PyTorch Lightning internals
4. Multi-Dataset Plotting
•	Evaluate how multiple datasets are handled
•	Define requirements for effective handling of multiple datasets trainings
 
### Plot Functionality & Tools
5. Replace / Integrate Map-Based Plotting
•	Investigate using earthkit-plots for map-based outputs
•	Explore integration opportunities with Datashader where applicable
6. Plot Audit
•	Identify existing plots that may be obsolete or rarely used, e.g. long rollout plots?
•	List potential new plots needed, based on user experience or research workflows
7. Plot Performance Bottlenecks
•	Review slow plotting components, e.g. PlotSpectrum
8. CLI / Workflow Separation
•	Investigate running plots outside training:
•	CLI subcommand for validation visualisation / metric reporting
9. Asynchronous Computation
•	Revisit current setup for async plotting
 
### Outcomes Expected From This Issue
•	Documented overview of current plotting challenges
•	Concrete list of follow-up tasks + tickets
•	Prioritization suggestions (cleanup, replacement, retirement)

anaprietonem:
Update on steps and scope of changes. 
Standardise callback configuration and runtime architecture to:

- Improve maintainability
- Enable multi-dataset flexibility
- Support future extensibility (e.g., Earthkit backend)
- Preserve current behavior during migration
- Proposed Configuration Model

**1. Runtime Settings** 

Introduce a dedicated settings section for runtime-only behavior:
```
settings:
  schedule:
    batch: 750
    epoch: 5

  backend: datashader
  projection_kind: equirectangular
  asynchronous: true

  sample_idx: 0
  per_sample: 6

[...]

PR #1153, opened 2026-05-28
feat: callback evaluate cli

anaprietonem:
## Description

Adds anemoi-training evaluate, a standalone command to run one full validation pass on a trained checkpoint without resuming training
- AnemoiEvaluator extends AnemoiTrainer, replacing trainer.fit() with trainer.validate()
- Checkpointing and weight-averaging callbacks are automatically disabled regardless of config
- Raises RuntimeError immediately if no checkpoint is configured (run_id, fork_run_id, or warm_start), preventing silent evaluation on random weights
- Registers .anemoi-training-evaluate hidden script entry point in pyproject.toml 
- Adds user guide page (docs/user-guide/evaluation.rst) covering checkpoint loading, DDP behaviour, config selection, and an example override config

Note this PR is kept minimal to simply introduce this command. In the future or separate PR where we will work on abstracting and reorganising the plotting code to an evaluate module, and consider if this code can be decoupled from pytorch-lightning to reuse it for inference.

Includes an evaluate command to allow for quick evaluation and decouple it from training.

[
](https://github.com/ecmwf/anemoi-core/issues/769?issue=ecmwf%7Canemoi-core%7C1152)

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1153.org.readthedocs.build/en/1153/

----
📚 Documentation preview 📚: https://anemoi-graphs--1153.org.readthedocs.build/en/1153/

----
📚 Documentation preview 📚: https://anemoi-models--1153.org.readthedocs.build/en/1153/


VeraChristina (commented):
Thanks for adding this! I have added an integration test for the evaluator on top of this branch here: https://github.com/ecmwf/anemoi-core/pull/1157

It required a small fix, and I also see a different error when running with `load_weights_only=False` -- might be due to the test setup, though. Happy to discuss here or on the PR to this branch

training/src/anemoi/training/commands/evaluate.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class EvaluateBase(Command, ABC):
+    accept_unknown_args = True
+
+    @staticmethod
+    def add_arguments(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
+        return parser
+
+    def _merge_sysargv(self, args: argparse.Namespace) -> str:
```
VeraChristina: Does this differ from what we do for the training command? Can we get this method from a shared base class?

training/src/anemoi/training/commands/evaluate.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # this will turn "/env/bin/anemoi-training evaluate" into "/env/bin/.anemoi-training-evaluate"
+        # the dot at the beginning is intentional to not interfere with autocomplete
+        modified_sysargv = argv.with_name(f".{argv.name}-{args.command}")
+
+        if hasattr(args, "subcommand"):
+            modified_sysargv += f"-{args.subcommand}"
+        return str(modified_sysargv)
+
+    def prepare_sysargv(self, args: argparse.Namespace, unknown_args: list[str] | None = None) -> None:
```
VeraChristina: as for _merge_sysargv, does this differ from the training command?

training/src/anemoi/training/commands/evaluate.py
```diff
@@ -0,0 +1,93 @@
...
+
+class Evaluate(EvaluateBase):
+    """Commands to evaluate Anemoi models."""
+
+    def run(self, args: argparse.Namespace, unknown_args: list[str] | None = None) -> None:
+        self.prepare_sysargv(args, unknown_args)
+
+        LOGGER.info("Running anemoi evaluation command with overrides: %s", sys.argv[1:])
+        main()
+
```
VeraChristina: Do we expect other subclasses to be added? Otherwise, we might not need the ABC with abstract method here.

VeraChristina (dismissed):
Two minor comments below (feel free to ignore), the PR looks good to me! 

In the future we could think about whether we want to make it easier to run this on, say, every kth saved checkpoint in one run_dir -- so that it becomes even easier to use as a replacement for plotting during training runs.

training/src/anemoi/training/commands/evaluate.py
anaprietonem: Fair point, I simplified it now!

VeraChristina (approved):
Thanks for addressing the feedback, LGTM!

PR #1154, opened 2026-05-28
fix: ensemble training shape clean-up

ssmmnn11:
- Makes CRPS consistent.
 - Keeps ensemble shape for targets: [batch, time, 1, grid, variable].
 - Lets the CRPS loss remove the singleton ensemble dimension when computing the loss.
 - Fixes spectral CRPS so it works with the same target shape.
 - Runs the spectral CRPS transform outside autocast.


training/src/anemoi/training/losses/kcrps.py
```diff
@@ -72,6 +72,17 @@ def _validate_arguments(alpha: float, backend: CRPSBackend) -> None:
             msg = f"Unknown CRPS backend {backend!r}. Expected one of: 'naive', 'stable'."
             raise ValueError(msg)
 
+    def mask_nans(self, pred: torch.Tensor, target: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+        """Mask CRPS inputs while preserving each tensor's ensemble size."""
+        if self.ignore_nans:
```
sahahner: I stumbled over this check on self.ignore_nans in the mask_nans function. Should we move the "if self.ignore_nans" check outside of the function and only call it only if self.ignore_nans==True?

training/src/anemoi/training/losses/spectral.py
```diff
@@ -324,17 +324,20 @@ def forward(
...
+            tgt_spec = einops.rearrange(tgt_spec, "b t 1 m v -> b t v m 1")
+            return self._kernel_crps(pred_spec, tgt_spec, alpha=self.alpha)
 
-        pred_spec = einops.rearrange(pred_spec, "b t e m v -> b t v m e")  # ensemble dim last for preds
-        tgt_spec = einops.rearrange(tgt_spec, "... m v -> (...) v m")  # remove ensemble dim for targets
         if self.no_autocast:
-            with torch.amp.autocast(device_type="cuda", enabled=False):
-                crps = self._kernel_crps(pred_spec, tgt_spec)
+            with torch.amp.autocast(device_type=pred.device.type, enabled=False):
+                crps = compute_spectral_crps()
```
sahahner: Even though this works perfectly fine the way it is written, I think it would be cleaner and easier to read to pass `pred` and `target` here.

ssmmnn11: good point. moved to context / nullcontext instead

training/src/anemoi/training/losses/kcrps.py
ssmmnn11: Yes we should. thx!

sahahner:
Thank you for these changes. 
Aligning the target and prediction shapes for the ensemble models is great.
Running the spectral CRPS transform outside autocast also looks good to me. This does change the default behaviour of the spectralCRPS loss as far as I see.
Did you notice any effect on runtime?

training/src/anemoi/training/losses/kcrps.py
```diff
@@ -72,6 +72,15 @@ def _validate_arguments(alpha: float, backend: CRPSBackend) -> None:
             msg = f"Unknown CRPS backend {backend!r}. Expected one of: 'naive', 'stable'."
             raise ValueError(msg)
 
+    def mask_nans(self, pred: torch.Tensor, target: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
+        """Mask CRPS inputs while preserving each tensor's ensemble size."""
+        target_nan_mask = torch.isnan(target).any(dim=2, keepdim=True)
+        pred_nan_mask = torch.isnan(pred).any(dim=2, keepdim=True)
```
frazane: Is there a convention on whether we hardcode dimension indices or we use e.g. `TensorDim.ENSEMBLE_DIM`?

training/src/anemoi/training/losses/kcrps.py
```diff
@@ -160,13 +171,14 @@ def forward(
     ) -> torch.Tensor:
         is_sharded = grid_shard_slice is not None
 
-        y_pred, y_target = self.mask_nans(y_pred, y_target)
+        if self.ignore_nans:
+            y_pred, y_target = self.mask_nans(y_pred, y_target)
```
frazane, line 176: If we are anyway dropping the whole sample (setting everything to zero) when any ensemble member or the target is NaN, isn't it cheaper to let NaNs propagate through the CRPS computation and mask them only once later? 

(maybe out of scope for this PR and can be ignored, but just wanted to flag this)

sahahner (dismissed):
Looks good to me. Thank you, Simon.

sahahner (approved):

Issue #1134, opened 2026-05-20
Units in training


VeraChristina:
### Is your feature request related to a problem? Please describe.

Variables may have different units in different datasets. In training this can lead to issues with finetuning. We need to handle the case where a dataset provided fo finetuning has different units than the dataset that was used in pre-training. 

### Describe the solution you'd like

PR https://github.com/ecmwf/anemoi-transform/pull/258 adds 'units' to the variable class and functionality to check for compatibility of units. We want to use this information to better handle units in training.

Steps:
- [x] At the begin of a fine-tuning run, use the functionality from transforms to check for compatibility of the units specified in the metadata of the checkpoint (stored as part of the dataset metadata) and the units of variables in the dataset provided. Throw an error for incompatible units. https://github.com/ecmwf/anemoi-core/pull/1140
- [x] Make sure we provide unit information to inference. -- Already passed as part of the dataset metadata.
- [x] Check units of target versus output variables for compatibility.
- [ ] Consider implementing on-the-fly conversion of units.
- [ ] Optional: Make units to be used by a checkpoint configurable via the config for the first training run. Fallback would be to use the units of the dataset provided in the first training run.

PR #1155, opened 2026-05-28
feat: add unit check for target variables

VeraChristina:
## Description
This PR adds a unit check for target variables, i.e. we check whether the units of the targets and of the corresponding output variables are compatible.

It also adds optional entries to the training configs to configure the variable checks in line with https://github.com/ecmwf/anemoi-transform/pull/303.

See https://github.com/ecmwf/anemoi-core/issues/1134 for more context

----
📚 Documentation preview 📚: https://anemoi-training--1155.org.readthedocs.build/en/1155/

----
📚 Documentation preview 📚: https://anemoi-graphs--1155.org.readthedocs.build/en/1155/

----
📚 Documentation preview 📚: https://anemoi-models--1155.org.readthedocs.build/en/1155/


training/src/anemoi/training/utils/variables_metadata.py
```diff
@@ -92,6 +92,59 @@ def check_variables_metadata_compatibility(
...
+                target_var,
+                pred_var,
+                target_var,
+            )
+            continue
+
+        pred_variable = Variable.from_dict(pred_var, variables_metadata[pred_var])
+        target_variable = Variable.from_dict(target_var, variables_metadata[target_var])
+
+        compatible, reason = pred_variable.compatible(target_variable, return_reason=True)
```
gabrieloks: I generally agree that we should check the units. However, it's worth noting that, especially during the early stages of research, the units may not accurately reflect reality, or the researcher may forget to update them.

A good example is the IMERG dataset. The units in the original dataset are mm/hr. Some preprocessing is required to convert it into an AIFS-friendly Zarr dataset, and during that process the metadata may not be updated accordingly. As a result, the user could encounter an error due to this check, even though the dataset itself is valid.

I would suggest adding a flag such as:

units_check: True (default)

This would make users aware of potential unit inconsistencies while still allowing them to use the dataset when necessary.

VeraChristina: Just to be sure I understand, we have datasets where all of the following apply?
- Their metadata contains unit information. (If one of the datasets does not have units, this check throws a warning but doesn't error.)
- The variables have been scaled to the correct magnitude / transformed via a filter to be compatible with AIFS outputs.
- The unit information in the metadata is not correct, so that when comparing the unit metadata this check would fail. That is, the filter updated the variables but not the metadata.

gabrieloks: This happens here for example:

https://anemoi.ecmwf.int/datasets/aifs-ea-an-oper-0001-mars-n320-1979-2024-6h-v1-for-single-v2?tab=statistics

units of tp is [m]

https://anemoi.ecmwf.int/datasets/nasa-imerg-grib-n320-1998-2024-6h-v1?tab=statistics

units of tp 

kg m-2 s-1

And I've used the IMERG dataset for training but we forgot to update the metadata during the dataset creation phase

VeraChristina: I see, thanks for the context! Generally, I'd have a preference to patch the metadata of the dataset (or to delete incorrect metadata) rather than adding an option to disable checks -- but I agree that we need a feasible way of using research datasets. @b8raoult would that be easy to do e.g. for the IMERG dataset above?

gabrieloks (approved):

PR #1159, opened 2026-06-02
fix(training)!: parameter freezing

JPXKQX:
## Description

This PR fixes a current bug with model freezing. If a specific encoder (e.g. `encoder.data`) is specified, no parameters are frozen, but still the logs say "parameters successfully frozen".

- A test had been added.

Currently, we should specify "data" and this would freeze "encoder.data.*" and "decoder.data.*", but it could be the case of freezing other intermediate layers of the processor if they are named "data" in this case. We propose a more explicit method of defining the layers to be frozen, which requires the entire path within the model to be specified.

This PR changes how the submodules to freeze are specified. Therefore, it's marked as breaking.

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those, please refer to https://anemoi.readthedocs.io/en/latest/***

----
📚 Documentation preview 📚: https://anemoi-training--1159.org.readthedocs.build/en/1159/

----
📚 Documentation preview 📚: https://anemoi-graphs--1159.org.readthedocs.build/en/1159/

----
📚 Documentation preview 📚: https://anemoi-models--1159.org.readthedocs.build/en/1159/


training/src/anemoi/training/utils/checkpoint.py
```diff
@@ -143,16 +143,19 @@ def freeze_submodule_by_name(module: nn.Module, target_name: str) -> None:
...
             for param in child.parameters():
+                LOGGER.info("Freezing parameter: %s", param.shape)
                 param.requires_grad = False
+        elif target_name.startswith(name + "."):
+            new_target = target_name.replace(name + ".", "", 1)
+            freeze_submodule_by_name(child, new_target)
         else:
-            # Recursively search within children
-            freeze_submodule_by_name(child, target_name)
+            LOGGER.info("Skipping submodule: %s", name)
```
VeraChristina: This might be a confusing log statement if multiple submodules_to_freeze are specified. Perhaps include the target_name for easier debugging?

training/src/anemoi/training/utils/checkpoint.py
```diff
@@ -143,16 +143,19 @@ def freeze_submodule_by_name(module: nn.Module, target_name: str) -> None:
...
     """
     for name, child in module.named_children():
         # If this is the target submodule, freeze its parameters
         if name == target_name:
             for param in child.parameters():
+                LOGGER.info("Freezing parameter: %s", param.shape)
                 param.requires_grad = False
+        elif target_name.startswith(name + "."):
+            new_target = target_name.replace(name + ".", "", 1)
+            freeze_submodule_by_name(child, new_target)
```
VeraChristina: This requires to specify the full submodule path for each submodule to freeze, is this clear from the docs?

VeraChristina (dismissed):
Thanks for fixing this! Just two minor comments below

training/src/anemoi/training/utils/checkpoint.py
```diff
@@ -143,16 +143,19 @@ def freeze_submodule_by_name(module: nn.Module, target_name: str) -> None:
...
         else:
-            # Recursively search within children
```
mc4117, line 168: was this recursive statement not working before or are you saying that this is a better way to implement it?

JPXKQX: This didn't work for any complex string, e.g. 'encoder.data', because the child would always be the name of a submodule (without any '.').

training/src/anemoi/training/utils/checkpoint.py
JPXKQX: I have updated the logs. I have set this to `debug`, and am now logging whether or not the target submodule is found.

training/src/anemoi/training/utils/checkpoint.py
JPXKQX: Documentation has been updated

JPXKQX:
Integration tests running [here](https://github.com/ecmwf/anemoi-core/actions/runs/27285522635)

mc4117 (approved):
LGTM! Thanks for finding the bug

Issue #595, opened 2025-10-09
Caution - Using dropout with model sharding - Revisit random seed generation


anaprietonem:
### Is your feature request related to a problem? Please describe.

As raised by @ssmmnn11, current implementation for random seed generation implies that every rank in the model group gets the same random seed. While each rank gets different inputs,  if dropout is activated the elements dropped would the "same" which can lead to unexpected behaviour or patterns in the model between ranks.

While we still want to preserve a way to generate a synced random seed across ranks to generate spatial correlated random numbers for sharding, there could be a secondary random stream that allow to have random seeds different across ranks for layers  like dropout.

Ebraheem1:
Hi,

By diving into the codebase to think about the root cause(s) behind this issue, I think it's also related to #498. I'd like to take this on. I have an ongoing analysis that points to different spots. I'd articulate them on the way so this comment will be edited with additional info whenever I stumble upon more root causes.

### First potential reason [Dataloader workers are seeded with a rank-agnostic base seed]

In file `training/src/anemoi/training/data/multidataset.py`, I see two issues:

- All workers across all processes/GPUs get the same seed. `get_base_seed()` returns the same value regardless of which process or worker is calling it. Thus, stochastic data augmentation will provide identical outcomes across different processes. This is definitely sub-optimal in terms of learning effectiveness.
- Workers within the same GPU are not differentiated from each other. `worker_id` is captured (`self.worker_id = worker_id`) but never mixed into the seed.

In my eyes, the correct seed for worker `w` on global rank `r ` should be a function of `(base_seed, r, w)`, e.g., a hash or arithmetic combination.

ssmmnn11:
Thank you for you comment @Ebraheem1. We have a common base seed to sync model and ensemble groups. The way to tackle this is to use different random streams for operations that require to be synchronised between certain ranks (data loading, model init) and operations where we don't want this kind of synchronisation. 

Regarding the dataloader: all workers shuffle the global sample index, and then each worker picks its chunk.

Ebraheem1:
Thanks for mentioning this dataloader chunk thing @ssmmnn11 . You are definitely right here.

The only thing I can point to now is:

In `training/src/anemoi/training/distributed/strategy.py`, `seed_rnd` derives a single seed from `(model_comm_group_id + 1) * base_seed` and applies it via `pl.seed_everything` on every rank:
```
def seed_rnd(model_comm_group_id: int, global_rank: int) -> None:
[...]

PR #1160, opened 2026-06-02
feat: introduce rank-local and synced source of randomness

ssmmnn11:
Introduce separate model RNG streams for distributed training. The default pytorch RNG is now seeded independently per rank, so dropout can use the rank-local random source. A synced pytorch RNG stream is available through a context manager for operations that must stay identical across ranks, such as model initialization and selected noise sampling operations.


ssmmnn11:
#595

models/src/anemoi/models/layers/ensemble.py
```diff
@@ -181,7 +182,8 @@ def forward(
             self.noise_channels,
         )
```
VeraChristina, line 192: Can we make sure that this is the same across all ranks, specifically for grid_size if self.noise_graph_provider is None? I.e. prevent callers from passing shard-local sizes which would desyncronize the synced stream. E.g. take the full grid size and shard inside, or add a check?

models/src/anemoi/models/layers/ensemble.py
VeraChristina, line 218: Do we need maybe_checkpoint here?

training/src/anemoi/training/train/train.py
```diff
@@ -348,7 +350,9 @@ def model(self) -> pl.LightningModule:
 
         training_method_cfg = self.config.training.method
         training_method_cls = get_class(training_method_cfg._target_)
-        model = instantiate_with_runtime_kwargs(training_method_cfg, **kwargs)  # Task -> pl.LightningModule
+        seed_torch_rng_sources(self.initial_seed, self.strategy.global_rank, reset_synced=True)
```
VeraChristina, line 370: This resets the groups to the stream derived from sync_group_id=0. Elsewhere, we set sync_group_id=model_comm_group_id. Do we need to restore this after model init?

models/src/anemoi/models/distributed/random.py
```diff
@@ -0,0 +1,295 @@
...
+
+    new_cuda_devices = _cuda_devices_in_state(current_state).difference(
+        _cuda_devices_in_state(default_state),
+    )
+    if new_cuda_devices:
+        msg = (
+            "CUDA was initialized inside use_synced_torch_rng(). "
+            "Initialize CUDA before entering the synced random context."
+        )
+        raise RuntimeError(msg)
```
VeraChristina, line 295: This raises after modifying the _synced_state. This seems fine for training runs because they crash here, but not sure how it interacts with tests.

models/tests/layers/processor/test_pointwise_processor.py
```diff
@@ -71,15 +71,71 @@ def test_pointwisemlp_processor_with_sharding_dropout_forward(pointwisemlp_proce
...
+        x,
+        batch_size,
+        shard_info,
+        model_comm_group=fake_model_comm_group,
+    )
+
+    assert output.shape == x.shape
+
+
+def test_pointwisemlp_processor_checkpointed_dropout_preserves_rng(pointwisemlp_processor_init):
```
VeraChristina, line 86: do we need a seed_torch_rng_sources here?

VeraChristina (commented):
Thanks for adding this! I think the separate saved RNG state acting as synced stream generally works well. A couple of comments below mostly regarding maintainability, like making sure future contributions don't desyncronise the synced steam.

One small comment along those lines regarding RandomFourierEmbeddigns in diffusion.py (which wasn't modified so I can't comment there directly): I would prefer the torch.randn call in the init to be explicitly wrapped in use_synced_torch_rng and not just implicitly being called in the model construction in train.py (which is wrapped in the context) -- just to be more explicit, it works as it is.

More generally, I am wondering if we can make the two states of the stream more explicit for future contributors, e.g. having `with_synced_torch_rng` and `with_local_torch_rng`. (This doesn't enforce the correct context but it might help against forgetting about the context.) Maybe even wrap torch.randn calls in `synced_randn` and `local_randn` which assert they are in the correct context?

Similarly, perhaps it's worth thinking about adding guards against rank-specific calls to toch.randn within synced contexts? E.g. an optional check for shape consistency (broadcast shape from rank 0 and assert all ranks agree) inside synced_randn controlled by a debug flag? Then we could add an integration test that runs in debug mode. (Maybe this is too cautious but I imagine these type of issues would be hard to debug.)

PR #1161, opened 2026-06-02
feat: flex-attention backend

cathalobrien:
This PR adds a [flex attention](https://pytorch.org/blog/flexattention/) backend to the attention layers.

Flex attention has been available in pytorch since ~2.6 but it wasn't stable for our use case. Now it seems more stable (I tested aifs-single v2 with torch 2.11 and it worked fine).

The specific use case here was that a user running on CPU installed the triton cpu build and wanted to use flex attention rather than SDPA.

I have also enabled support for the [flash attention v4 backend of flex attention](https://pytorch.org/blog/flexattention-flashattention-4-fast-and-flexible/) by checking if flash attention v4 is installed and passing the right kernel options to flex attention so it can use it. 

Performance is slightly slower then flash attention v2 (0.34 it/s vs 0.38 it/s), but much faster then sdpa. 

I added a correctness test which compares against sdpa, and a small benchmark test which compares all 3 attention kernels standalone.
```
anemoi-core/models/tests/integration/attention/test_benchmark_attention.py::test_attention_backend_benchmark[forward-global] Attention benchmark: window_size=None, dtype=torch.float16, shape=(1, 16, 40320, 32)
  flex: 36.25 ms
  sdpa: 27.32 ms
  flash: 21.16 ms
PASSED
anemoi-core/models/tests/integration/attention/test_benchmark_attention.py::test_attention_backend_benchmark[forward-sliding_window] Attention benchmark: window_size=1120, dtype=torch.float16, shape=(1, 16, 40320, 32)
  flex: 4.88 ms
  sdpa: 112.95 ms
  flash: 1.95 ms
PASSED
anemoi-core/models/tests/integration/attention/test_benchmark_attention.py::test_attention_backend_benchmark[fowrward & backward-global] Attention benchmark: window_size=None, dtype=torch.float16, shape=(1, 16, 40320, 32)
  flex: 28.18 ms
  sdpa: 26.16 ms
  flash: 21.11 ms
PASSED
anemoi-core/models/tests/integration/attention/test_benchmark_attention.py::test_attention_backend_benchmark[fowrward & backward-sliding_window] Attention benchmark: window_size=1120, dtype=torch.float16, shape=(1, 16, 40320, 32)
  flex: 5.04 ms
  sdpa: 114.19 ms
  flash: 1.96 ms
PASSED
```


models/src/anemoi/models/layers/attention.py
```diff
@@ -359,6 +362,101 @@ def forward(
...
+        self.block_mask = None
+        self._block_mask_key = None
+        self._compile = True
+        self._use_flash4_backend = False
+
+        if not kwargs.get("use_triton_backend", False):
+
+            # Try import flash attention v4
+            # if this is avilable it can be used as a backend for flex attention which gives approx 2x performance
+            # One reason to use flex attention with the flash attewntion v4 backend, ratehr then using flash attention v4 directly, is
```
HCookie:
```suggestion
            # One reason to use flex attention with the flash attention v4 backend, rather then using flash attention v4 directly, is
```

models/src/anemoi/models/layers/attention.py
```diff
@@ -359,6 +362,101 @@ def forward(
...
+            # flex attentions support for custom block masks.
+            # if flash attention is not available then the trion backend will be used for flex attention
```
HCookie:
```suggestion
            # if flash attention is not available then the triton backend will be used for flex attention
```

HCookie (dismissed):
Awesome work, looks good to me

models/src/anemoi/models/layers/attention.py
```diff
@@ -359,6 +371,114 @@ def forward(
...
+            block_mask_kwargs = {}
+            compiled_block_mask = torch.compile(
+                self.create_block_mask
+            )  # REQUIRED, otherwise entire seq_len^2 array will be materialised
+            block_mask = compiled_block_mask(mask_mod, *mask_shape, query.device, **block_mask_kwargs)
+
+        if self._compile:
+            flex_attn = torch.compile(
+                partial(self.attention, block_mask=block_mask, kernel_options=self._kernel_options), dynamic=False
+            )
```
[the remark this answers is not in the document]
cathalobrien: I dont think so buddy

cathalobrien:
n320 config runs with torch 2.11, but fails with torch 2.10
```
RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling `cublasGemmEx( handle, opa, opb, m, n, k, alpha_ptr, a, CUDA_R_16F, lda, b, CUDA_R_16F, ldb, beta_ptr, c, std::is_same_v<C_Dtype, float> ? CUDA_R_32F : CUDA_R_16F, ldc, compute_type, CUBLAS_GEMM_DEFAULT_TENSOR_OP)`
```
I will add a min version check.
currently this PR will have to wait then until we increase the max versions of pytorch

PR #1169, opened 2026-06-05
feat(training): load only data required for current rollout length

ssmmnn11:
This branch changes rollout training so the dataloader follows the current rollout length instead of always using the maximum configured rollout.

- Training dataloader offsets use the current rollout length.
- Validation without validation_rollout also uses the current training rollout window.
- Train/validation/test datasets are updated when the epoch changes.
- Dataloader worker persistence depends on whether epoch_increment > 0
- Worker seeds include the epoch.

@cathalobrien @HCookie @japols 

----
📚 Documentation preview 📚: https://anemoi-training--1169.org.readthedocs.build/en/1169/

----
📚 Documentation preview 📚: https://anemoi-graphs--1169.org.readthedocs.build/en/1169/

----
📚 Documentation preview 📚: https://anemoi-models--1169.org.readthedocs.build/en/1169/

### DIFF
--- a/training/docs/user-guide/performance-optimisation.rst
+++ b/training/docs/user-guide/performance-optimisation.rst
@@ -277,10 +277,8 @@ their recommended settings
 
 .. note::
 
-   Longer rollout increases the CPU memory required by the dataloaders.
-   It can be beneficial to break rollout runs into multiple runs (e.g.
-   rollout 1->6 and rollout 7->12) and tune the number of workers for
-   both runs accordingly.
+   Dataloader CPU memory can increase as the active rollout length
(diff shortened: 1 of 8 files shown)


cathalobrien (dismissed):
Nice work.

Can you please edit this file `anemoi-core/training/docs/user-guide/performance-optimisation.rst` to remove the note about splitting up rollout training to save CPU memory

```rst
.. note::

   Longer rollout increases the CPU memory required by the dataloaders.
   It can be beneficial to break rollout runs into multiple runs (e.g.
   rollout 1->6 and rollout 7->12) and tune the number of workers for
   both runs accordingly.
```

cathalobrien (approved):
thanks!

PR #1133, opened 2026-05-20
feat!: Forecast dataset dataloader

mc4117:
## Description

This PR implements a new dataloader for reading 5-dimensional trajectory/forecast datasets in the Anemoi training framework. The new dataloader enables training on forecast-structured data where each sample represents a complete forecast run rather than a sliding window over a continuous timeseries.

## What problem does this change solve?

The existing NativeGridDataset dataloader was designed for 4D gridded datasets with shape [(dates, variables, ensembles, cells)], using sliding windows over a continuous timeseries. However, forecast datasets have a fundamentally different structure with shape (base_dates, variables, ensembles, steps, cells) where:
- Each base date represents an independent forecast initialization
- The forecast step dimension represents lead times within each forecast
- Samples should never cross forecast initialization boundaries

This PR introduces TrajectoryDataset to handle this structure properly and removes backwards compatibility for 4D datasets

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/contributor_license_agreement.md)

PR #1170, opened 2026-06-05
feat: dataloader for forecast dataset 2

floriankrb:
Another attempt to use trajectories datasets, https://github.com/ecmwf/anemoi-core/pull/1133

_AI generated: Claude wrote most of it with appropriate prompts._

# What problem does this change solve?

  Training on forecast/trajectory datasets — where data has an explicit lead-step axis on top of the initialisation axis  (5-D: base_dates × variables × ensembles × steps × cells). This is needed by the temporal interpolation task and any  setup that must sample windows inside a single forecast without crossing initialisation boundaries.

#  Why the previous approach is limited

  #1133 makes the 5-D data fit the existing pipeline by collapsing the two real axes into one. It flattens (init, step) into a virtual 1-D time series of length num_inits × forecast_steps and fabricates a dates array (init + step·freq),then spends effort recovering it. Those dates are synthetic and collide whenever forecasts overlap in valid time. (also it  duplicate some of the anemoi-datasets code.

#  What this PR does instead

anemoi-datasets already models this correctly. A zarr with layout: trajectories opens through open_dataset() into a TrajectoriesZarr that exposes base_dates, steps, base_frequency, step_frequency, statistics, and step/base subsetting. No fabricated dates.

*Design* — a 2-D sample anchor (sequence, position) replaces the flat int index + trajectory_ids:

  - A sample is addressed by (sequence, position); relative time offsets are applied within one sequence, so a sample can never span two sequences — by construction, not by filtering.
  - Analysis (NativeGridDataset): a single sequence, position = time index. Anchor (0, t) is exactly the old flat index t → fully backwards-compatible.
  - Forecast (ForecastDataset): one sequence per initialisation (base date), position = forecast step. frequency =  step_frequency. A sampling mode controls how anchor positions are drawn per forecast — non_overlapping (default) or all.

  The rest of the pipeline is unchanged in spirit: a reader still turns (anchor + offsets) into a (T, E, G, V) tensor,  MultiDataset enumerates / shuffles / shards anchors, and the task slices the T axis. Tasks, losses and the model are untouched — anemoi-core/models needs no changes.

  This also makes non-overlapping training first-class: each forecast is an independent sequence, so "one window per init" is just a sampling mode rather than a side effect of trajectory-id bookkeeping.

*Running it* : `anemoi-training train --config-name  temporal_downscaler_forecast`. Not tested yet, waiting for `aifs-od-fc-oper-0001-mars-o96-2016-2025-6h-1h-v5` to build. But it could be run on smaller test datasets.

### DIFF
--- a/training/src/anemoi/training/config/dataloader/forecast_native_grid.yaml
+++ b/training/src/anemoi/training/config/dataloader/forecast_native_grid.yaml
@@ -21,15 +21,6 @@ batch_size:
   validation: 4
   test: 4
 
-# ============
-# Default effective batch_size for training is 16
-# For the o96 resolution, default per-gpu batch_size is 2 (8 gpus required)
-# The global lr is calculated as:
-# global_lr = local_lr * num_gpus_per_node * num_nodes / gpus_per_model
-# Assuming a constant effective batch_size, any change in the per_gpu batch_size
-# should come with a rescaling of the local_lr to keep a constant global_lr
-# ============
-
 # runs only N training batches [N = integer | null]
 # if null then we run through all the batches
 limit_batches:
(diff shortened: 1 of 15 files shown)


mc4117:
Closed due to duplicate work

Issue #1171, opened 2026-06-08
PerTimestepMetrics reimplements a separate validation loop


ssmmnn11:
`PerTimestepMetrics` repeats a lot of code from computation of the validation metrics:

  - input and target loading
  - model forward pass
  - ensemble gather
  - post-processing
  - metric loop
  - metric kwargs setup
...

This makes it easy for the callback to get out of sync with the rest of the code. For example, fixes to metric kwargs in `BaseTrainingModule.calculate_val_metrics()` also need to be copied here.

A clean-up is required so PerTimestepMetrics only handles the per-timestep slicing and naming

mc4117:
Closed by PR #1173

PR #1173, opened 2026-06-08
fix: simplify per timestep metrics

mc4117:
Clean up per timestep callback 

## Description
Related to issue #1171

- Remove duplicate model forward pass, ensemble gather, and post-processing
- Read y_pred from validation_step outputs (index 2) instead of re-running the model
- Get targets from task.get_targets + _collapse_ens_dim (cheap batch indexing)
- Delegate metric computation, post-processing, and kwargs setup entirely to calculate_val_metrics, so any future fixes there apply automatically
- Update tests to use the new interface (outputs tuple, mocked calculate_val_metrics)


frazane:
Could we align on how we handle metrics that don't support sharding between this PR and #1087? See https://github.com/ecmwf/anemoi-core/pull/1087/changes/dcbee5e325416eba160432a4cbcf3f6a0c67778d.

mc4117:
Hi @frazane I'm not sure I understand your comment? These metrics inherit directly from the computed validation output so are not affected by sharding

mc4117:
Discussed offline with @frazane and we agree that this PR should go first followed by PR #1087

VeraChristina:
Thanks for simplifying this and removing duplicate computations! The PR looks good to me. Do we have a test run somewhere / is the callback covered by the temporal downscaler integration test? (If not, it might be worth adding it.)

anaprietonem:
Agree with @VeraChristina, code looks much cleaner than before so thanks @mc4117. I was wondering (and fair if it's out of the scope for now) but if we really need this. In the sense that what you're doing here is getting the timestep dimension over y and then looping over that and computing the metrics. So could we not abstract the logging we do in the base method to account for this? https://github.com/ecmwf/anemoi-core/blob/e4faaaed083d581365f95ed171407377344af6c6/training/src/anemoi/training/train/methods/base.py
(motivation is then you'd wouldn't to worry about configuring the callback or not)

mc4117:
> Thanks for simplifying this and removing duplicate computations! The PR looks good to me. Do we have a test run somewhere / is the callback covered by the temporal downscaler integration test? (If not, it might be worth adding it.)

I checked and it is indeed in the temporal dowscaler integration test

mc4117:
> Agree with @VeraChristina, code looks much cleaner than before so thanks @mc4117. I was wondering (and fair if it's out of the scope for now) but if we really need this. In the sense that what you're doing here is getting the timestep dimension over y and then looping over that and computing the metrics. So could we not abstract the logging we do in the base method to account for this? https://github.com/ecmwf/anemoi-core/blob/e4faaaed083d581365f95ed171407377344af6c6/training/src/anemoi/training/train/methods/base.py (motivation is then you'd wouldn't to worry about configuring the callback or not)

I wanted to keep this feature as optional because you might not always want breakdown by hourly. Could we do this if it moved to base.py?

if we want configurability, you'd need some flag to switch it on/off so in that case I don't see much benefit compared to the callback.

anaprietonem (dismissed):
Thanks for clarifying my question Mariana, if you've checked that this is being tested in the integration test then good from my side.

mc4117:
this is an example mlflow https://mlflow.ecmwf.int/#/experiments/54/runs/3a223a13c7bc4810821881119ca8ef4f

VeraChristina (dismissed):
Thanks for fixing the tests, LGTM!

frazane (approved):
LGTM!

PR #1174, opened 2026-06-08
fix: spectral loss for lam

dietervdb-meteo:
## Description
Refactor of #1143. For LAM models the spectral loss should be computed only over the regional domain defined by the `output_mask`. As such we allow the `subgrid` argument of spectral losses to be set to `output_mask`, which under the hood will be replaced by a slice representation of the actual output mask object.

Example config:
```yaml
training_loss:
     datasets:
       your_dataset_name:
         _target_: anemoi.training.losses.spectral.SpectralCRPSLoss
         transform: fft2d
         subgrid: "output_mask"
         x_dim: 1069
         y_dim: 1069
```

##  Additional notes ##
Note that the above, for current most typical use cases, is equivalent to
```yaml
training_loss:
     datasets:
       your_dataset_name:
         _target_: anemoi.training.losses.spectral.SpectralCRPSLoss
         transform: fft2d
         subgrid: [0, 1142761]
         x_dim: 1069
         y_dim: 1069
```
which would work without the minor code changes introduced in this PR. Still, having the 'automatic' implementation as introduced by this PR has the following dvantages:
1. users will not need to explicitly know how their `output_mask` is sliced out of the data grid.
2. a change of output mask will remain restricted to a single part of the config, that will then automatically propagate through

----
📚 Documentation preview 📚: https://anemoi-training--1174.org.readthedocs.build/en/1174/

----
📚 Documentation preview 📚: https://anemoi-graphs--1174.org.readthedocs.build/en/1174/

----
📚 Documentation preview 📚: https://anemoi-models--1174.org.readthedocs.build/en/1174/

### DIFF
--- a/training/docs/modules/losses.rst
+++ b/training/docs/modules/losses.rst
@@ -241,12 +241,13 @@ Supported transforms include:
   native grid of ERA5 such as N320.
 * ``OctahedralSHT``: Spherical harmonic transform (SHT) on the octahedral reduced Gaussian grid.
 
-.. note::
+.. notes::
 
-   SHT-based transforms expect a flattened reduced-grid ordering:
+   1. SHT-based transforms expect a flattened reduced-grid ordering:
    ``[batch, ensemble, grid_points, variables]`` and return spectral coefficients with
    shape ``[batch, ensemble, l, m, variables]`` where ``l = truncation + 1``.
 
+<<<<<<< Updated upstream
 .. note::
 
    ``ReducedSHT`` and ``OctahedralSHT`` both perform a spherical harmonic transform on a reduced Gaussian grid.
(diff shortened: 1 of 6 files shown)


dietervdb-meteo:
Hmm, the changes to the argument of the spectral loss break some tests. Will fix that after @frazane or someone else confirms they agree with the general idea of this PR.

frazane:
Nit: maybe we could use a more general name than `nodes_slice`, since this PR would also allow referring to index arrays such as boolean masks? Something like `nodes_idx` or `nodes_subset` that then resolves to a slice for a list/tuple and to a supporting array if it's a string. Later we could also support specifying a mask from a `.npy` file or similar. 

Apart from that I am okay with the general idea, could you add some docs?

dietervdb-meteo:
Docs and test added.

frazane (approved):
Looks good!

PR #1181, opened 2026-06-11
fix(training)!: make user-specified search path take precedence and simplify Hydra search paths hierarchy

### LINKED ISSUE #1093: User-provided `--config-path` does not take precendence in config search
## Issue description

When users specify the directory containing their training configs, e.g.

```
anemoi-training train --config-path /path/to/configs --config-name my_config
```

the explicitly provided path does not actually take precedence, and training can run with entirely different configs than the user expects.

This happens because the search-path plugin prepends the current working directory (and the ANEMOI_CONFIG_PATH and anemoi home ~/.config/anemoi/training directories) ahead of Hydra's primary path:

https://github.com/ecmwf/anemoi-core/blob/c1d83955eda1d75001d09f2b686fb0a388218ab6/training/src/hydra_plugins/anemoi_searchpath/anemoi_searchpath_plugin.py#L62-L70

Since the path passed with --config-path enters through Hydra's @hydra.main and is not prepended ahead of these, the prepended CWD wins.

For reference, this is the rationale behind the current design (thanks @gmertes for the explanation):

> The design of this predates anemoi, so it's quite ancient :smile: The original rationale is as follows:
>
> * You have a set of base config files in your home folder under ~/.config/anemoi/training/config . This contains things like the hardware and system configuration that you set once and then doesn't change. There is another plugin that looks for configs in this folder, but it has a lower priority than the cwd one.
> * Then for each training run you have another folder somewhere else, that contains the config for that training run (let's say ~/my-new-model) This contains all the config entries that do change: model, data, training etc.
> * Then when you want to train, you cd ~/my-new-model && anemoi-training train ... . The plugin will load that training-specific config from ~/my-new-model, and overlay it on top of the "base" configs from your ~/.config/anemoi/training/config

## Expected behavior

The path explicitly provided by the user should take precedence over anything else.

## Proposed solution

The plan is to switch the prepends to appends (reversing their order) so the --config-path path keeps top priority, and at the same time simplify the discovery surface by removing the anemoi-home and ANEMOI_CONFIG_PATH search paths, which appear unused and mainly cause confusion about where a config value came from.

That leaves three search paths, in decreasing priority: 
1. explicitly provided `--config-path`
2. current working directory
3. Packaged configs (pkg://anemoi.training/config)

The packaged-configs path is untouched - it ships the default config groups referenced in config.yaml's defaults list, so it stays.

frazane:
## Description
Makes the user-provided `--config-path` take precedence in the config search, and simplifies the Hydra search path to a small, predictable hierarchy.

The search-path plugin *prepended* the current working directory (and `ANEMOI_CONFIG_PATH` and the `~/.config/anemoi/training` home dir) ahead of Hydra's primary `@hydra.main` path. Because `--config-path` enters through that primary path, the prepended CWD shadowed it — so `anemoi-training train --config-path /my/configs ...` silently ran with different configs than the user asked for.

This change:
- **appends** the current working directory instead of prepending it, so an explicit `--config-path` keeps top priority;
- **removes** the anemoi-home (`~/.config/anemoi/training`) and `ANEMOI_CONFIG_PATH` search paths, which were a frequent source of "where did this config value come from?" confusion and are effectively unused;
- leaves the packaged-defaults path (`pkg://anemoi.training/config`) **unchanged** as the lowest-priority fallback that ships the default config groups in `config.yaml`'s `defaults:` list.

Resulting search path, in decreasing priority: `--config-path` → current working directory → packaged defaults.

The now-orphaned `anemoi-training config training-home` subcommand (which copied configs into the no-longer-searched home dir) is removed; use `anemoi-training config generate` instead. The user-guide docs are updated to match.

**Breaking changes:**
- Configs under `~/.config/anemoi/training` and `ANEMOI_CONFIG_PATH` are no longer auto-discovered — pass `--config-path` or use the working directory.
- The `config training-home` subcommand is removed.

Closes #1093.

## Additional notes
Note that this was developed in tandem with AI coding agents.

---

----
📚 Documentation preview 📚: https://anemoi-training--1181.org.readthedocs.build/en/1181/

----
📚 Documentation preview 📚: https://anemoi-graphs--1181.org.readthedocs.build/en/1181/

----
📚 Documentation preview 📚: https://anemoi-models--1181.org.readthedocs.build/en/1181/


JesperDramsch:
Thanks for digging into this, @anaprietonem asked me to look at the PR, as I built this back in the day. The search-path precedence genuinely is under-documented (sorry!), and pinning down the expected behaviour and writing it down is useful regardless. I think there are two separate things bundled here, though, and I'd treat them differently.

The precedence fix — making `--config-path` win — is a reasonable call. Worth flagging that it's a design choice more than a clear-cut bug, though: I've had a user expect the exact opposite back in the day (local/CWD overriding `--config-path`), so whichever direction we pick, most of the value is in documenting the behaviour so people stop guessing. Happy to see that part land.

The removal of `ANEMOI_CONFIG_PATH` and `~/.config/anemoi/training` is the part I'd like to slow down on. That's institutional infrastructure, and pulling it out is a bigger decision than the precedence fix actually requires. If it turns out I'm the only one relying on it (it was really only ever shown in slides back then I'm afraid) then I'm genuinely not precious about keeping it. My concern is the failure mode: removing a search path doesn't raise anything, it just silently falls back to the packaged defaults. So anyone who *was* relying on it would train with a different config and get no signal that anything changed, and possibly train multi-day models without ever noticing why their model degraded / changed.

So I'd say: keep the precedence fix and the documentation, but treat the search-path removal as its own discussion, and either way document it loudly (changelog breaking-change note + a migration line) so it can't fail silently.

One gentle, general note, since the description mentions this was developed with coding agents: I've fairly consistently seen agents lean toward removing code they don't have full context on... it's a known failure mode worth extra care around, especially for deletions and breaking changes. Not a knock on this PR specifically, just a flag to keep a close eye on the “remove” parts of agent-assisted diffs. I've had success in writing in the instructions (e.g. AGENTS.md or CLAUDE.md) that this is not a greenfield project we're contributing to.

Happy to help document the precedence behaviour or sort out the search-path ordering if that's useful.

frazane:
Hello @JesperDramsch, thanks for the comments. Just to clarify, I worked on these changes _after_ receiving feedback from ECMWF colleagues, and going in this direction of retaining only the `--config-path` and CWD options is what I was advised to do. Additionally, I reached out to the community about whether anyone was using the ANEMOI_CONFIG_PATH or the `~/.config/anemoi/training` options, and this does not appear to be the case. So as far as the design choices are concerned, I believe it's safe to keep them as is and we could focus on improving the implementation. If you push back on this, I propose we let the ATS decide. What do you think?

JesperDramsch:
Thanks @frazane, the community check plus the steer by my colleagues is plenty for me. No need for the ATS explicitly as they will have to review a breaking change like this anyways, especially regarding config discovery.

One small suggestion, a runtime signal rather than just a doc one: if `ANEMOI_CONFIG_PATH` is set in the environment but no longer read, emit a one-off deprecation warning at startup for a release or two. That reaches the people who actually had it set, who won't see it in the release notes before rerunning a pipeline, and turns a silent fallback to packaged defaults into a visible "heads up, this is ignored now" rather than a config swap they only notice when a multi-day model comes out wrong. But I don't know how deprecation warnings are exactly handled at the moment. It's cheap to add, but no worries I'm just sad it never got use as it was received quite positively when I presented it back then. But I won't stand in the way of course, as it's just bloat in that case.

gmertes (commented):
The implementation LGTM. When we release this we could write an extra paragraph at the top of the Github release notes to highlight this change for extra visibility.

anaprietonem (approved):
thanks to the reviewers, and @frazane for looking at this. LGTM too!

PR #1185, opened 2026-06-12
fix(models): fix test file discovery and add missing test coverage

ChrisBoettner:
## Description

I added two small test fixes in `anemoi-models`.

**1. `balanced_partition.py` was never collected by pytest**

`models/tests/distributed/balanced_partition.py` was missing the `test_`
prefix required by pytest's default `python_files = test_*.py` rule. There is no
override in the `pytest.ini`, so all parametrised tests for
`get_balanced_partition_sizes` and `get_partition_range` are silently skipped. I
assume this was not intentional.

Two additional tests are added:

- `test_get_balanced_partition_range` — the helper called by
  `multidataset.py` to partition date indices across dataloader workers
  had no test coverage.
- `test_get_partition_range_invalid_partition_id` — checks the
  `ValueError` raised on an out-of-range partition id.

**2. Stale TODO in `test_attention.py`**

The init test contained `# TODO: Make assert in MHSA to check if
embed_dim is divisible by num_heads`. That check already exists at
`attention.py:120` (`attn_channels % num_heads != 0 -> ValueError`).
The comment was removed and a parametrised test is added that checks
the error path for both `MultiHeadSelfAttention` and
`MultiHeadCrossAttention`.

No existing issue.

## Additional notes

All 14 new tests pass. Pre-commit hooks (black, isort, ruff, docsig)
pass clean on both changed files.

By opening this pull request, I affirm that I agree to the Contributor License Agreement.

### DIFF
--- a/models/tests/distributed/test_balanced_partition.py
+++ b/models/tests/distributed/test_balanced_partition.py
@@ -9,6 +9,7 @@
 
 import pytest
 
+from anemoi.models.distributed.balanced_partition import get_balanced_partition_range
 from anemoi.models.distributed.balanced_partition import get_balanced_partition_sizes
 from anemoi.models.distributed.balanced_partition import get_partition_range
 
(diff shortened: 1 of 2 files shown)


mc4117:
LGTM! Thanks for the contribution. I have run the tests and the tests pass

cathalobrien (approved):

PR #1187, opened 2026-06-12
fix: fix integration tests with torch >= 2.11

cathalobrien:
## Description
integration tests were failing with torch 2.11 and greater. 
specifically the edm_diffusion_transport tests which run at the end of the test suite. 
the breaking change was related to how torch compile reuses compiled code *across* tests. If you ran the failing tests standalone, they would pass. they only fail when run at the end of the test suite, when many other things have been compiled.

the solution is to reset torch compile (specially the frontend, Dynamo) after each test. 
```
@pytest.fixture(autouse=True)
def _reset_dynamo() -> None:
    """Reset torch compile state before and after each test.

    This prevents 'recomputed metadata doesn't match checkpointed value' errors observed with torch 2.12.
    """
    import torch._dynamo

    torch._dynamo.reset()
    yield
    torch._dynamo.reset()
```
PyTorch [appears](https://github.com/pytorch/pytorch/blob/7d28bf8e2bf24163375b6904056caffa5a58242d/test/dynamo/test_compile.py#L27) to do the same before their compilation related tests . Other fixes like changing torch compile settings were tried but didn't help.


cathalobrien:
[integration tests pass](https://github.com/ecmwf/anemoi-core/actions/runs/27416015213/job/81028928386)

mc4117 (approved):
LGTM!

PR #1189, opened 2026-06-12
feat(training,models): add radial spectral density to cartesian transforms and enable AMSE

### LINKED ISSUE #1188: Support the spectral AMSE loss with the 2D transforms (FFT2D, DCT2D), not only the SHT family
## Context

`SpectralAMSELoss` was added in #1148. It is currently usable only with the spherical-harmonic transforms (`RegularSHT`, `ReducedSHT`, `OctahedralSHT`): its `__init__` asserts that the configured transform provides `power_spectral_density` and `cross_spectral_density` methods (`training/src/anemoi/training/losses/spectral.py`), and those two methods are defined only on the `SHT` base class (`models/src/anemoi/models/layers/spectral_transforms.py`). `FFT2D` and `DCT2D` do not implement them, so configuring the AMSE loss with `transform: fft2d` or `transform: dct2d` trips the assertion at construction time.

We deliberately left the 2D-transform variant out of #1148 to keep that PR focused, and agreed to follow it up separately.

## What is needed

Implement `power_spectral_density` and `cross_spectral_density` on `FFT2D` and `DCT2D` so the AMSE formulation has a per-band amplitude and cross-spectrum to work with, mirroring the SHT implementations.

For the SHT these sum over the order `M` to produce a per-degree-`L` quantity. The 2D transforms produce a `(ky, kx)` spectrum rather than an `(L, M)` one, so the analogue is to bin the 2D spectrum into radial-wavenumber bands and sum within each band — i.e. define what plays the role of `L` for the 2D case. The AMSE `forward` (`SpectralAMSELoss`) should then work unchanged on the resulting per-band tensors.

## Notes

This is useful for regional / limited-area models on regular lat/lon grids, where the SHT does not apply. As noted in the #1148 discussion, AMSE combined with a 2D FFT has been tested in the FastNet codebase and improves spectral consistency on a regional regular-grid domain; the individual AMSE terms have a slightly different physical meaning under a 2D transform than under the SHT, but the loss remains effective.

Scope: add the two methods to `FFT2D`/`DCT2D`, unit tests for them and for `SpectralAMSELoss` instantiated with each 2D transform, and a short docs note that AMSE now supports the 2D transforms.

frazane:
## Description

Adds per-band power and cross spectral density to the Cartesian 2D transforms `FFT2D` and `DCT2D`, and uses them to enable the spectral AMSE loss (`SpectralAMSELoss`) on those transforms. Previously only the spherical-harmonic transforms carried a spectral density, so AMSE was SHT-only. Follow-up to #1148. 

New feature, in two parts.

First, the core addition: a radially-binned spectral density for the 2D transforms. `power_spectral_density` and `cross_spectral_density` existed only on the `SHT` base class, where they sum over the order `M` to give a per-degree-`L` quantity. The Cartesian transforms produce a `(ky, kx)` plane instead, so the analogue implemented here bins that plane into integer radial-wavenumber bands `L = round(sqrt(ky**2 + kx**2))` and sums within each band. The result is a per-`L` power spectrum (and cross-spectrum) for `FFT2D` and `DCT2D`, useful on its own for any spectral diagnostic or loss on regular and limited-area grids, where the SHT does not apply.

Second, the immediate consumer: `SpectralAMSELoss` asserts that its transform exposes `power_spectral_density`/`cross_spectral_density`, so it raised at construction for `transform: fft2d` / `transform: dct2d`. With the methods above in place, AMSE now works with both. The SHT path is untouched; this is not a breaking change.

Implementation notes:

- A shared `RadialSpectralDensity` mixin provides both methods for `FFT2D` and `DCT2D`. The per-band reduction uses `torch.real(c * conj(c))` (and `torch.real(a * conj(b))` for the cross-spectrum), which is correct for both complex (FFT) and real (DCT) coefficients. The `SHT` implementation cannot be reused directly because it reads `.imag`, which raises on the real DCT coefficients.
- Patch-wise `FFT2D` (`patch_size` set) is rejected by `SpectralAMSELoss`, since it produces a per-patch spectral tensor with no single `(ky, kx)` plane to bin.

Closes #1188.

##  Additional notes ##

Under a 2D transform the individual AMSE terms have a slightly different physical meaning than under the SHT, but the formulation is unchanged. 

CC @sahahner and @jameschappell (cannot add you as reviewer but your feedback would be welcome!). 

---

----
📚 Documentation preview 📚: https://anemoi-training--1189.org.readthedocs.build/en/1189/

----
📚 Documentation preview 📚: https://anemoi-graphs--1189.org.readthedocs.build/en/1189/

----
📚 Documentation preview 📚: https://anemoi-models--1189.org.readthedocs.build/en/1189/


models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -127,6 +178,8 @@ def __init__(
                 patch_y, patch_x = self.patch_size
                 self.filter = self.lowpass_filter(patch_x, patch_y)
 
+        self._init_radial_bands()
```
sahahner: What is the overhead of always initialising the radial bands even though power_spectral_density or cross_spectral_density are not required?

frazane: I haven't measured, but it's a good point that we don't always need them. Perhaps a cached property is better?

sahahner: yes, something in that direction sounds good

models/tests/layers/test_spectral_density.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def _make(transform: str, x_dim: int, y_dim: int):
+    if transform == "fft2d":
+        return FFT2D(x_dim=x_dim, y_dim=y_dim)
+    pytest.importorskip("torch_dct")
+    return DCT2D(x_dim=x_dim, y_dim=y_dim)
+
+
+def test_radial_band_index_values() -> None:
```
sahahner: How much of the spectral density is already tested in the loss function test `test_amse_2d_transforms`? 
There are quite a few new tests here, and I wonder whether there is a risk of duplicated testing.

How general are the power density tests and could they be applied to the SHT version as well?

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -46,7 +46,58 @@ def forward(
         """
 
 
-class FFT2D(SpectralTransform):
+class RadialSpectralDensity:
```
sahahner: I took a different approach for introducing a base class for the PSD functionality for the SHTs and introduced a base class for all SHT where the definition of the PSD align.
`class SHT(SpectralTransform)`

Your approach diverges from that (not meaning it is wrong), but it would be nice if this aligns. It seems that all 2D transforms are based on `RadialSpectralDensity`, so why not make it a base class for 2D transforms that inherits from the more general `SpectralTransform`?

training/docs/modules/losses.rst
```diff
@@ -300,6 +300,28 @@ Use this for global models on the reduced Gaussian grid (only N320 supported so
 Truncation is by default set to 319 for n320 grids, but can be set to a higher or lower value in the config file.
 This truncation parameter defines how many wave numbers are included in the spectral representation.
 
+Spectral AMSE
+-------------
+
```
sahahner, line 414: Thank you for including documentation on the AMSE!

models/src/anemoi/models/layers/spectral_transforms.py
frazane: Done in e6db5deed

models/src/anemoi/models/layers/spectral_transforms.py
frazane: This makes sense, thanks! We are now more aligned, see 61e89fb1d

models/tests/layers/test_spectral_density.py
frazane: There was indeed some duplication. More generic tests are now parameterized over `["fft2d", "dct2d", "sht"]`, see 0b32ab4da!

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -46,7 +47,69 @@ def forward(
...
+        elif self._radial_wavenumber_kind == "index":
+            ky = torch.arange(self.y_dim, dtype=torch.float32)
+            kx = torch.arange(self.x_dim, dtype=torch.float32)
+        else:
+            raise ValueError(f"Unknown radial wavenumber kind: {self._radial_wavenumber_kind}")
+
+        ky_grid, kx_grid = torch.meshgrid(ky, kx, indexing="ij")
+        band = torch.sqrt(ky_grid**2 + kx_grid**2).round().long()
+        # flattened (ky*kx,) bucket id per spectral coefficient
+        return band.reshape(-1), int(band.max().item()) + 1
```
sahahner: I think before this was a buffer. Is there a specific reason this is now not a buffer anymore? Can this always be on CPU?

models/tests/layers/test_spectral_density.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    # FFT2D 4x3: |fftfreq*N| -> ky=[0,1,2,1] (y_dim=4), kx=[0,1,1] (x_dim=3)
+    # band[i,j] = round(sqrt(ky[i]**2 + kx[j]**2)), flattened row-major (y outer, x inner).
+    t = FFT2D(x_dim=3, y_dim=4)
+    expected = torch.tensor([0, 1, 1, 1, 1, 1, 2, 2, 2, 1, 1, 1])
+    assert torch.equal(t.radial_band_index, expected)
+    assert t.n_radial_bands == 3
+
+
+def test_radial_bands_are_built_lazily() -> None:
```
sahahner: This test does not seem necessary as it checks whether `functools.cached_property` works.

models/tests/layers/test_spectral_density.py
```diff
@@ -0,0 +1,120 @@
...
+    else:  # DCT-II cosine mode
+        field = torch.cos(torch.pi * (yy + 0.5) * ky0 / y_dim) * torch.cos(torch.pi * (xx + 0.5) * kx0 / x_dim)
+
+    spec = t.forward(field.reshape(1, 1, 1, y_dim * x_dim, 1))
+    psd = t.power_spectral_density(spec)[0, 0, 0, :, 0]
+    expected = round((ky0**2 + kx0**2) ** 0.5)
+    assert (psd[expected] / psd.sum()).item() > 0.99
+
+
+def test_dct_real_coeffs_do_not_use_imag() -> None:
```
sahahner: There are other tests that already check this so I think it can also be removed.

models/tests/layers/test_spectral_density.py
sahahner: I still find the number of tests rather high and found a few that seem to be duplicating tests. Feel free to have a look at my other comments.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -46,7 +47,69 @@ def forward(
...
+        return self._sum_over_radial_bands(torch.real(spectral_coeffs * torch.conj(spectral_coeffs)))
+
+    def cross_spectral_density(self, spectral_coeffs_a: torch.Tensor, spectral_coeffs_b: torch.Tensor) -> torch.Tensor:
+        """Return per-band cross-spectral density: sum of ``Re[a conj(b)]`` within each band."""
+        return self._sum_over_radial_bands(torch.real(spectral_coeffs_a * torch.conj(spectral_coeffs_b)))
+
+    def _sum_over_radial_bands(self, per_mode: torch.Tensor) -> torch.Tensor:
+        """Collapse the two spectral dims ``[..., ky, kx, v] -> [..., L, v]``."""
+        index, n_radial_bands = self._radial_bands
+        index = index.to(per_mode.device)
```
sahahner: I found the .to(device) here that should probably be avoided. Could you give this another look please?

frazane: Could you be more specific? We were registering a buffer in the initialization before, then we removed it to avoid initializing the radial bands if not needed. But this requires us at some point to call `to()`, no?

sahahner: it is possible to write the fields to buffer if the function is called?

sahahner:
Thank you, @frazane, for this nice work and for enabling more spectral loss functions for the 2D spectral transforms. 
Did you try these loss functions out on an example and could share the result here?

frazane:
> Thank you, @frazane, for this nice work and for enabling more spectral loss functions for the 2D spectral transforms. Did you try these loss functions out on an example and could share the result here?

Hi @sahahner, thanks for the helpful review, definitely improved the quality of the PR. I am unable to do an end-to-end test with actual data because I need #1087 to be merged in order to use spectral losses at all. If we want to see results I would wait for that one and then I'll post something here!

models/src/anemoi/models/layers/spectral_transforms.py
frazane: We could define a lazy, empty buffer at initialization and fill it when the call the cached property?

frazane: See last commit, this way we get both: the radial band index is a buffer, and we avoid calling `to` at every step.

sahahner: Yes, I brought our solutions together and now it looks fine but would be still good to test it during training.

frazane: Thanks for the help @sahahner! Will post again on this PR when I run some end-to-end tests successfully.

guiletj:
Hi @frazane, 
I merged `PR#1087` and `PR#1189` on a fork of anemoi-core to use `SpectralAMSELoss` on the LAM part of a stretched grid model, here is the link to the fork (https://github.com/guiletj/anemoi-core/tree/feat/train-stretched-with-spectral-loss).
However, when I start a training, a bug about devices appears (`RuntimeError: Expected all tensors to be on the same device, but got index is on cpu, different from other tensors on cuda:0 (when checking argument in method wrapper_CUDA_index_add_)`).
To solve it, I applied some changes to the code in the function `_ensure_radial_bands`, here is the commit (https://github.com/guiletj/anemoi-core/commit/6724f1c4b05a86f47453f941f994d47929882d34).
Do you have other ideas to solve this problem or some comments ?

frazane:
Thanks a lot @guiletj for reporting this problem and proposing a fix! @sahahner I think this relates a bit to the changes we were discussing last week. Could you have a look?

frazane:
@guiletj I could reproduce your error. I pushed a fix in 5a1b4c5. It keeps the lazy build (except for ky, kx which are cheap) but anchors it to the buffer system instead of placing devices by hand, which is something that we generally try to avoid.

Could you confirm it's resolved on your side? Also, out of curiosity, can you confirm the loss is behaving as expected?

models/tests/layers/test_spectral_density.py
```diff
@@ -0,0 +1,73 @@
...
+
+    total = torch.real(a * torch.conj(a)).flatten(-3, -2).sum(dim=-2)
+    torch.testing.assert_close(psd_a.sum(dim=-2), total)
+    torch.testing.assert_close(t.cross_spectral_density(a, a), psd_a)
+
+    cross = t.cross_spectral_density(a, b)
+    assert torch.all(cross**2 <= psd_a * psd_b * (1 + 1e-9) + 1e-9)
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="needs CUDA to exercise cross-device buffers")
```
sahahner: I see the motivation behind the test, but in general, I think if we use the buffers, we do not need to include this test here. At other places where we use the buffer (pre/postprocessors), we also do not add additional tests for the functionality of the buffer.

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -48,7 +48,74 @@ def forward(
...
+    """Base class for the 2D Cartesian spectral transforms (:class:`FFT2D`, :class:`DCT2D`).
+
+    Defines the per-radial-band power and cross spectra the 2D transforms share, binning
+    their ``(ky, kx)`` spectral plane into integer radial-wavenumber bands.
+    """
+
+    # "fft": signed FFT wavenumbers (|k| in 0..N//2); "index": cosine indices 0..N-1.
+    _radial_wavenumber_kind: str = "fft"
+
+    def _register_radial_geometry(self) -> None:
```
sahahner: Can we move the functionality of `_register_radial_geometry` into `_ensure_radial_bands`? That would allow us to avoid calling `_register_radial_geometry` in the basic 2D FFT transform. 
As far as I understand we never need `kx` and `ky` without calling `_ensure_radial_bands`.

sahahner:
Thank you, @guiletj, for raising the bug. Thanks, @frazane, for removing the manual passing around of the device, which can be fragile, for example, when using multiple GPUs.
I made a few minor comments.

If one of you could share their experience with using the AMSE loss function, that would be great!
Thank you :)

models/src/anemoi/models/layers/spectral_transforms.py
frazane: > As far as I understand we never need kx and ky without calling _ensure_radial_bands.

That's true, however I cannot seem to make it work without a manual device placement. I don't think we can easily get both fully lazy behavior and no manual device placement. Open for suggestions!

guiletj:
Hi @frazane and @sahahner,

Thank you for your fix. I tried to make a small training with it this morning and it went fine, so the fix solve the bug on my side too. The branch I used for my tests is here : https://github.com/guiletj/anemoi-core/tree/feat/train-stretched-with-spectral-loss.

You seem to be interested by having experience with the AMSE loss function so here is a link to an experiment with `SpectralAMSELoss` : https://mlflow.ecmwf.int/#/experiments/726/runs/43b88a6b580a425caae24080d82fce64/model-metrics.

In addition, about the performance of the `SpectralAMSELoss`, it decreases with the steps but we observe huge values during the training of this experiment. Do you have the same order of magnitudes for the loss ?

frazane:
> In addition, about the performance of the `SpectralAMSELoss`, it decreases with the steps but we observe huge values during the training of this experiment. Do you have the same order of magnitudes for the loss ?

Hi @guiletj, yes the magnitude is expected and it's due to the loss being sum-reduced over the GRID dimension instead of mean-reduced. For pointwise losses this works because each point is automatically scaled by the area weight (per the default configuration scalers), however this is not done for spectral losses since grid weighting becomes meaningless in spectral space. @sahahner is working on addressing this, see https://github.com/ecmwf/anemoi-core/pull/992.

models/src/anemoi/models/layers/spectral_transforms.py
sahahner: I see the issue now and propose a refactor (see commit). _radial_band_index is now built eagerly and registered as a non-persistent buffer. Since only SpectralAMSELoss needs the buffer, I moved the _register_radial_bands() call into SpectralAMSELoss.__init__ (and made it idempotent), so plain FFT2D/DCT2D usage pays nothing.

sahahner:
I have proposed a refactor that does not need any manual device placement and makes use of the buffer. To avoid the unnecessary creation of variables in the buffer, I have moved that to the loss functions (at the moment only AMSE) that actually use the functionality.

Would you mind testing that, please?

frazane:
@sahahner thanks, your refactor makes sense, and I can confirm it works on my end!

models/src/anemoi/models/layers/spectral_transforms.py
```diff
@@ -48,7 +48,66 @@ def forward(
...
-class FFT2D(SpectralTransform):
+class Cartesian2DTransform(SpectralTransform):
+    """Base class for the 2D Cartesian spectral transforms (:class:`FFT2D`, :class:`DCT2D`).
+
+    Defines the per-radial-band power and cross spectra the 2D transforms share, binning
+    their ``(ky, kx)`` spectral plane into integer radial-wavenumber bands.
+    """
+
+    # "fft": signed FFT wavenumbers (|k| in 0..N//2); "index": cosine indices 0..N-1.
+    _radial_wavenumber_kind: str = "fft"
```
sahahner, line 59: Should this possibly be moved to the FFT2D transform?

sahahner:
Thank you for confirming that @frazane. I think from the code perspective this is ready to merge, but I would like to confirm that the effect of the AMSE on the PSD  when using it with 2D transform is as intended.

guiletj:
Hi @frazane and @sahahner,

I need to correct my previous feedback about the changes made to the code.
This morning, I realized that I hadn't tested the right code (the one with your corrections). So, I tested it and it returns an error :

>     File "/home/guiletj/perm/venvs/recipes_stage_juliette-b7ade5def8/lib64/python3.11/site-packages/anemoi/models/layers/spectral_[transforms.py](http://transforms.py/)", line 70, in _register_radial_bands
>     ky = (torch.fft.fftfreq(self.y_dim, device=device) * self.y_dim).abs()
>     ^^^^^^
>     NameError: name 'device' is not defined.

With @ThomasRieutord, we propose a new fix which works on our side, here is the link to the commit : https://github.com/guiletj/anemoi-core/commit/e377b68bedb34a0e43c3c5b8d731b16cc533f62e. This fix use the function `get_distributed_device()` from `anemoi.graphs.utils` in `spectral_transforms.py`.

Thank you !

sahahner:
> Hi @frazane and @sahahner,
> 
> I need to correct my previous feedback about the changes made to the code. This morning, I realized that I hadn't tested the right code (the one with your corrections). So, I tested it and it returns an error :
> 
> > ```
> > File "/home/guiletj/perm/venvs/recipes_stage_juliette-b7ade5def8/lib64/python3.11/site-packages/anemoi/models/layers/spectral_[transforms.py](http://transforms.py/)", line 70, in _register_radial_bands
> > ky = (torch.fft.fftfreq(self.y_dim, device=device) * self.y_dim).abs()
> > ^^^^^^
> > NameError: name 'device' is not defined.
> > ```
> 
> With @ThomasRieutord, we propose a new fix which works on our side, here is the link to the commit : [guiletj@e377b68](https://github.com/guiletj/anemoi-core/commit/e377b68bedb34a0e43c3c5b8d731b16cc533f62e). This fix use the function `get_distributed_device()` from `anemoi.graphs.utils` in `spectral_transforms.py`.
> 
> Thank you !

Hi @guiletj, thank you for testing! Could it be that you are not on the newest version of the code?

I think the line of code should look like this?
`ky = (torch.fft.fftfreq(self.y_dim) * self.y_dim).abs()`
https://github.com/ecmwf/anemoi-core/pull/1189/changes#diff-dc6a21152fbed1d04e44f2d73e11a5db7193cb66caa70b48e783ec528767bea3R70

guiletj:
> > Hi @frazane and @sahahner,
> > I need to correct my previous feedback about the changes made to the code. This morning, I realized that I hadn't tested the right code (the one with your corrections). So, I tested it and it returns an error :
> > > ```
> > > File "/home/guiletj/perm/venvs/recipes_stage_juliette-b7ade5def8/lib64/python3.11/site-packages/anemoi/models/layers/spectral_[transforms.py](http://transforms.py/)", line 70, in _register_radial_bands
> > > ky = (torch.fft.fftfreq(self.y_dim, device=device) * self.y_dim).abs()
> > > ^^^^^^
> > > NameError: name 'device' is not defined.
> > > ```
> > 
> > 
> > With @ThomasRieutord, we propose a new fix which works on our side, here is the link to the commit : [guiletj@e377b68](https://github.com/guiletj/anemoi-core/commit/e377b68bedb34a0e43c3c5b8d731b16cc533f62e). This fix use the function `get_distributed_device()` from `anemoi.graphs.utils` in `spectral_transforms.py`.
> > Thank you !
> 
> Hi @guiletj, thank you for testing! Could it be that you are not on the newest version of the code?
> 
> I think the line of code should look like this? `ky = (torch.fft.fftfreq(self.y_dim) * self.y_dim).abs()` https://github.com/ecmwf/anemoi-core/pull/1189/changes#diff-dc6a21152fbed1d04e44f2d73e11a5db7193cb66caa70b48e783ec528767bea3R70

I updated the code with the newest version and it works. 
So everything is good !
Thank you

Issue #690, opened 2025-11-20
Increase in node joggle required when computing Voronoi region in Planar Area Weights when dealing with 1km datasets.


icedoom888:
### What happened?

Line 179 in: 
v = Voronoi(extended_points, qhull_options="QJ Pp")

Fails with error:
`
scipy.spatial._qhull. QhullError: QH6229 qhull input error: 51 attempts to construct a convex hull with joggled input.      Increase joggle above 'QJ0.064' or modify qh_JOGGLE... parameters in user_r.h While executing: | qhull v pp qj Options selected for Qhull 2020.2.r 
2020/08/31: run-id 2039614516 voronoi Pprecision-ignore Qbbound-last-qj Pgood _run 51 QJoggle 0.064 _joggle-seed 1305198295 _max-width 6.4 Error-roundoff 8.8e-15 Width-outside 1.7e-14 _wide-facet 5.2e-14 _maxoutside 2.7E-12 Last point added to hull was p302409.
`

This Qhull error usually means thepoint cloud is nearly degenerate, and Qhull can't build a convex hull or Voronoi diagram without applying more "joggle" (random perturbation).

This happens when:

- Points are almost coplanar, collinear, or otherwise low-dimensional.
- The dataset contains duplicate or extremely close points.
- The dataset is massive (hundreds of thousands of points).
- Numerical precision issues (very large or very tiny coordinates).

Increasing joggle factor to:
v = Voronoi(extended_points, qhull_options="QJ0.2 Pp")
Fixes the issue.

### What are the steps to reproduce the bug?

Create graph with 1km datasets and ICON grids.

### Version

graphs v0.6.2

### Platform (OS and architecture)

Linux balfrinew-ln002 5.14.21-150500.55.65_13.0.73-cray_shasta_c #1 SMP Fri Jun 14 21:25:32 UTC 2024 (d052318) x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell

```

### Organisation

Meteoswiss

PR #1191, opened 2026-06-13
perf(graphs): vectorize PlanarAreaWeights with shoelace area

frazane:
## Description

`PlanarAreaWeights` (and `MaskedPlanarAreaWeights`, which inherits it) computed each node's planar Voronoi-cell area in a Python loop that constructed one `scipy.spatial.ConvexHull` per node. This replaces that loop with a vectorised shoelace-formula computation over all Voronoi regions at once, with an exact `ConvexHull` fallback for the rare degenerate regions where the shoelace formula would be wrong. The public behaviour of both attributes is unchanged.

Performance, not a bugfix or feature. The per-node `ConvexHull` loop is the dominant cost when generating stretched-grid / LAM graphs that use planar area weights.

Measured on a stretched lam=10 graph (cutout of a 1 km regional dataset + a global N320 one; 1.69M data nodes), comparing `main` and this branch on the *same* Voronoi tessellation so only the area extraction differs:

| planar-area computation, 1.69M nodes | `main` | this branch |
| --- | --- | --- |
| `PlanarAreaWeights.compute_area_weights` (area step) | 108.3 s | 1.8 s |

About 60x faster on the part this PR changes. The surrounding `Voronoi` build (~60 s) and everything else are untouched; this area step was the single biggest cost in data-node generation for such graphs.

The output is numerically equivalent to the previous `ConvexHull` result: across all 1.69M cells the maximum relative difference is 1.1e-08 (float32-epsilon level), and the degenerate cells are matched exactly by the fallback. It is not bit-identical — the shoelace and `ConvexHull` float computations differ at ~1e-11, so a small fraction of cells land on the far side of a float32 rounding boundary. This is below the precision of the stored (float32) weights.

No tracking issue — found while profiling `anemoi-graphs create` on stretched-grid recipes.

## Additional notes

How it works:

- For each Voronoi region the cell area is the shoelace formula `1/2 |sum_i (x_i*y_{i+1} - x_{i+1}*y_i)|`, evaluated for all regions at once via flat indexing and `np.add.reduceat`. SciPy returns 2-D region vertices in boundary order, so for a convex cell this equals `ConvexHull(...).volume`.
- Voronoi cells are convex, but qhull's input joggling (`QJ`, needed for numerically hard grids) can return degenerate regions carrying interior or near-collinear vertices, where a plain shoelace under-counts the area. Such regions are detected (a convex polygon's consecutive edge-turn cross products all share one sign, so a region with both signs is non-convex) and recomputed exactly with `ConvexHull`, matching the previous behaviour. On the lam=10 graph above this affected 5 of 1.69M regions; detection false positives are harmless, as a clean cell simply takes the exact path.

Tests: a clean-equivalence test (areas match the per-node `ConvexHull` reference on a clustered patch resembling a regional cutout) and a degenerate-fallback test (injects an interior vertex into a real Voronoi region and asserts the exact hull area is recovered — verified to fail by ~18% without the fallback). The full `graphs` test suite passes, aside from pre-existing failures that only need the optional `h3` dependency.

---


graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -158,6 +158,69 @@ def _get_boundary_ring(self, points: np.ndarray, resolution: float) -> np.ndarra
...
+
+        # Flag non-convex regions: a convex polygon's consecutive edge-turn cross products
+        # all share one sign, so a region with both signs has an interior/degenerate vertex.
+        edge_x, edge_y = x[nxt] - x, y[nxt] - y
+        turn = edge_x * edge_y[nxt] - edge_y * edge_x[nxt]
+        has_pos = np.add.reduceat((turn > 0).astype(np.intp), starts) > 0
+        has_neg = np.add.reduceat((turn < 0).astype(np.intp), starts) > 0
+        flagged = np.flatnonzero(has_pos & has_neg)
+
+        # Exact ConvexHull fallback for the rare degenerate regions (matches the old code).
```
anaprietonem, line 217: why do you still need this? what are are ` rare degenerate regions`?

anaprietonem:
Hey @frazane nice one! Out of curiosity, how were you profiling the graph creation?

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
frazane: Sometimes the data (as in our case, see https://github.com/ecmwf/anemoi-core/issues/690) can be numerically hard for Qhull to compute Voronoi regions, so we need to set a large joggle to be able to do it. When we do, some regions are "degenerate" in the sense that they have near-collinear, near-duplicate vertices and things like this. The old way of computing the areas automatically handles these cases, so we use it as a fallback since the shoelace algorithm does not.

frazane:
> Hey @frazane nice one! Out of curiosity, how were you profiling the graph creation?

Hi @anaprietonem nothing fancy, just `time.perf_counter` on a high level, and `cProfile` + [snakeviz](https://jiffyclub.github.io/snakeviz/) to go into details, but the former was actually more than enough!

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -158,6 +158,69 @@ def _get_boundary_ring(self, points: np.ndarray, resolution: float) -> np.ndarra
...
         return np.concatenate([expanded_hull, np.vstack(boundary_points)])
 
+    @staticmethod
+    def _voronoi_region_areas(v: Voronoi, n_points: int) -> np.ndarray:
+        """Polygon area of the first ``n_points`` Voronoi regions, vectorised.
+
+        Uses the shoelace formula instead of a per-node :class:`scipy.spatial.ConvexHull`.
+        SciPy returns each 2-D region's vertices in boundary order, so for a convex cell
+        this matches ``ConvexHull(...).volume`` to float-rounding level. Heavy joggling
+        (qhull ``QJ``) can yield degenerate, non-convex regions where shoelace
```
anaprietonem, line 168: but aren't we using "QJ Pp" for the Voronoi computation?

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -158,6 +158,69 @@ def _get_boundary_ring(self, points: np.ndarray, resolution: float) -> np.ndarra
...
+        cross = x * y[nxt] - x[nxt] * y
+        areas = 0.5 * np.abs(np.add.reduceat(cross, starts))
+
+        # Flag non-convex regions: a convex polygon's consecutive edge-turn cross products
+        # all share one sign, so a region with both signs has an interior/degenerate vertex.
+        edge_x, edge_y = x[nxt] - x, y[nxt] - y
+        turn = edge_x * edge_y[nxt] - edge_y * edge_x[nxt]
+        has_pos = np.add.reduceat((turn > 0).astype(np.intp), starts) > 0
+        has_neg = np.add.reduceat((turn < 0).astype(np.intp), starts) > 0
+        flagged = np.flatnonzero(has_pos & has_neg)
```
anaprietonem, line 215: I think it could worth adding a logger here to indicate how many of the points are being flagged

anaprietonem (commented):
From my side the PR is fine. I left some minor comments for clarification. Still I think it could be great if before merging we can get confirmation from the LAM and SG side that the output for the recipe construction is the same!

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
frazane: For now yes, but currently we have a hardcoded fix with the joggle in our development branches. We plan to make it a configurable option (or perhaps we could have a machanism to auto-tune the joggle) in an upcoming PR. See https://github.com/ecmwf/anemoi-core/issues/690.

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
frazane: Good idea!

frazane:
> From my side the PR is fine. I left some minor comments for clarification. Still I think it could be great if before merging we can get confirmation from the LAM and SG side that the output for the recipe construction is the same!

I can confirm with SG on our side results are byte-identical. I agree it would be nice to get an independent confirmation though.

anaprietonem:
@dietervdb-meteo @mpvginde  @MeraX  since you have LAM set ups, would you have some time to run the code on this PR and confirm the performance improvements for the graph recipe please?

@havardhhaugen, if you could do a check on the SG side it would also be mcuh appreciated!

MeraX:
Thank you @anaprietonem for pinging me. I can't test the AICON embedded grid with the current code, but I'm expecting benefit from the PR, so I would approve merging it.

MeraX (approved):

mpvginde:
> @dietervdb-meteo @mpvginde @MeraX since you have LAM set ups, would you have some time to run the code on this PR and confirm the performance improvements for the graph recipe please?
> 
> @havardhhaugen, if you could do a check on the SG side it would also be mcuh appreciated!

Hi @anaprietonem, for LAM we only calculate the loss over the high-res region, so no averaging of the loss over boundary and high-res. In addition, each grid cell in the high-res part represents an equal area in the Lambert Conformal projection so we typically use UniformWeights in our graph-workflows. 

But I will give it a go to see if things work as expected.

mpvginde (approved):
Checked for the RMI-LAM graph (with PlanarAreaWeights). 
Graph creation takes: 
- main branch: ~ 248 sec
- this branch: ~ 130 sec

Area weight statistics are identical.

Issue #1135, opened 2026-05-21
Reduce duplicate recomputations in plotting callbacks


VeraChristina:
### Is your feature request related to a problem? Please describe.

- PlotLoss recomputes loss inside callback
- Other callbacks reprocess batch data

### Describe the solution you'd like

Extend the plot adapter to handle the shared computations, once for all plotting callbacks.
- Move BasePlotAdditionalMetrics.process() (and perhaps PlotLoss' loss recomputation) into adapter methods
- Wrap results in a cached plot payload held on the base callback class, computed lazily per-batch
- The contract with the trainer stays unchanged
This eliminates duplicate computation across callbacks, keeps the trainer decoupled, and leverages the adapter pattern already in place.

Note that this increases the responsibilities of the plot adapter. Rethinking the coupling of the adapter to the task hierarchy is not in the scope of this issue.

### Describe alternatives you've considered

One alternative is for the trainer's predict_step to return a richer payload. Upsides: no redundant loss/normalisation calls, callbacks become pure consumers. (would need to do this lazily only on plotting batches)
Downsides: This would increase the couplign of the trainer to plotting concerns; payload might grow over time when new plots are added.

PR #1195, opened 2026-06-15
feat: shared processing in plot callbacks

VeraChristina:
## Description
Reduce the duplication of post-processing in the plotting callbacks by introducing a plot playload that all callbacks fetch from the plot adapters and that caches post-processed tensors on first access.

https://github.com/ecmwf/anemoi-core/issues/1135

##  Additional notes ##


training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -903,22 +886,54 @@ def process(
             self.latlons = {}
 
         if dataset_name not in self.latlons:
-            self.latlons[dataset_name] = pl_module.model.model._graph_data[dataset_name].x.detach()
-            self.latlons[dataset_name] = np.rad2deg(self.latlons[dataset_name].cpu().numpy())
+            # Fallback: extract latlons if not already populated by prepare_payload
+            self.latlons[dataset_name] = np.rad2deg(
```
anaprietonem: why is this needed? shouldn't we check if that's on the payload? * okey I do see above that self.latlons is actually coming from the payload but still not clear when the Fallback is hit

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -903,22 +886,54 @@ def process(
...
+                tuple(
+                    pl_module.plot_adapter.select_members(
+                        denormed_output[step_idx, self.sample_idx : self.sample_idx + 1],
+                        members,
+                    )
+                    for step_idx in range(denormed_output.shape[0])
+                ),
+            )
+        else:
+            # Fallback for callers without a payload (e.g. direct process() calls)
```
anaprietonem: similarly when is this Fallback hit?

anaprietonem: If we do need the Fallback, it would good  to refactor the below to some utility function to avoid the hardcoding of very similar code between the payload computation and the fallback

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -903,22 +886,54 @@ def process(
...
+                        self.post_processors[dataset_name](x[dataset_name][:, ...].detach().cpu(), in_place=False)[
+                            self.sample_idx : self.sample_idx + 1
+                        ],
+                        members,
+                    )
+                    for x in outputs.predictions
+                ),
+            )
 
+        output_tensor = pl_module.plot_adapter.prepare_plot_output_tensor(output_tensor)
```
anaprietonem: this is equivalent to the function below for process_output_tensor so why do we need that function still? can't we get rid of it?

anaprietonem (commented):
Nice work Vera! left a few comments to clarify

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: good point, it's only hit when prepare_payload wasn't called before process (which doesn't happen in the current use of the class). I'll remove the fallback and make this clear in the docstring

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: as above

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: the method process_output_tensor was added in the recent merge of main. As far as I understand it's used to process auxiliary outputs (from plot_kwargs) not present in the payload. There is some overlap with the code here, but process_output_tensor denormalises outputs, while here we use already denormalised outputs -- I could extract the common part into a helper function but that's only three lines of code.

training/src/anemoi/training/diagnostics/callbacks/plot.py
```diff
@@ -365,47 +365,26 @@ def on_validation_batch_end(
...
+            self.post_processors = payload.post_processors
+            self.latlons = payload.latlons
+            self._payload = payload
 
             plot_kwargs = self._plot_kwargs_from_output(pl_module, output)
             output = TrainingStepOutput(
                 loss=output.loss,
                 metrics=output.metrics,
-                predictions=gathered_predictions,
+                predictions=payload.outputs[1],
```
anaprietonem: could this be cleaner if we do something like?

```

@dataclass
class GatheredOutputs:
    loss: torch.Tensor
    predictions: list[dict[str, torch.Tensor]]

gathered = GatheredOutputs(
    loss=output.loss,
    predictions=[
        {
            dataset_name: pl_module.allgather_batch(dataset_pred, dataset_name)
            for dataset_name, dataset_pred in pred.items()
        }
        for pred in preds
    ],
)

self._cached_payload = PlotPayload(
    batch_idx=batch_idx,
    batch=gathered_batch,
    outputs=gathered,  # ← now a named object
    post_processors=post_processors,
    latlons=latlons,
    feature_indices=feature_indices,
)
```

Though looking at this a bit more, I am getting confused why we have the loss in the output payload payload.outputs[0] but in TrainingStepOutput we use output.loss?

training/src/anemoi/training/diagnostics/callbacks/plot.py
anaprietonem: Yes I want to check better the need for process_output_tensor but I think that's out of the scope for this PR, so a helper function will be fine.

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: good catch, it's a reference to the legacy tuple format of outputs (which was replaced with TrainingStepOutput), I had missed this after merging, have removed now

training/src/anemoi/training/diagnostics/callbacks/plot.py
VeraChristina: done!

anaprietonem (dismissed):
Thanks for addressing the comments Vera, nice work! LGTM to be merged

anaprietonem:
Closing the PR, due to the issues with memory management. Feature implemented under https://github.com/ecmwf/anemoi-core/commit/2a72fd81e9d78351100ed3aa783f427308d5d68c

PR #998, opened 2026-03-20
feat: loading orchestration layer in checkpoint system refactor

JesperDramsch:
**Part of the checkpoint system refactor (#248).** Implements the Loading Orchestration Layer — four strategies for applying checkpoint data to models, built on Phase 1 infrastructure (PR #501).

### What this PR does

Adds four loading strategies that read from `CheckpointContext.checkpoint_data` (populated by source stages) and load weights/state into models:

| Strategy | Use case | Key behavior |
|----------|----------|-------------|
| `WeightsOnlyLoader` | Load pretrained weights, discard training state | `extract_state_dict()` → `load_state_dict(strict=False)` |
| `ColdStartLoader` | Pretrained weights + explicit state reset | Inherits from `WeightsOnlyLoader`, resets epoch/step to 0 |
| `TransferLearningLoader` | Flexible loading with shape-mismatch handling | Non-mutating filter, `skip_mismatched`, metadata tracking |
| `WarmStartLoader` | Full training resume | Restores model + optimizer + scheduler + epoch + step |

Supporting infrastructure:

- `TrainingState` dataclass — structured training state with `from_checkpoint()` / `to_dict()`
- `LoadingStrategy` base class with shared `_apply_anemoi_metadata()` helper
- State dict utilities: `filter_state_dict()`, `match_state_dict_keys()`, `merge_state_dicts()`

### Architecture

```
CheckpointContext enters with checkpoint_data (from source stage) + model
        ↓
  Loading strategy extracts state_dict, applies to model
        ↓
CheckpointContext exits with model weights loaded + metadata populated
```

All strategies extend `LoadingStrategy(PipelineStage)` and implement `async process(context) -> context`.

### Design decisions

- **ColdStartLoader inherits WeightsOnlyLoader**: Same weight loading + explicit state reset. No code duplication.
- **WarmStartLoader is standalone**: Does NOT inherit from WeightsOnlyLoader — fundamentally different flow (strict loading, optimizer/scheduler restore). Forced inheritance would require overriding everything.
- **Non-mutating state dict filter**: TransferLearningLoader builds a new filtered dict instead of `del state_dict[key]`. Source `checkpoint_data` is preserved — other stages may need it.
- **`_apply_anemoi_metadata` on base class**: `_ckpt_model_name_to_index` and `weights_initialized` preserved by all strategies via shared helper. Missing metadata breaks inference.
[...]

Issue #1139, opened 2026-05-22
feature: AzureSource for checkpoint acquisition


JesperDramsch:
Follow-up to PR #464.

Adds an `AzureSource` checkpoint stage mirroring the existing
`S3Source` and `HTTPSource` patterns, using `anemoi.utils` for transport
where possible (or `azure-storage-blob` if no anemoi-utils Azure backend
exists yet).

### Scope

- New `training/src/anemoi/training/checkpoint/sources/azure.py` with
  `AzureSource(CheckpointSource)`
- Lazy import; declare an `[azure]` optional-dependency extra
- Unit tests using the same `sys.modules` injection pattern as the
  refreshed `S3Source` tests
- (Optional) one network-marked integration test against a public Azure
  blob, if one is available

### Out of scope

- `GCSSource` — explicitly deferred (no active use case per @mchantry
  on PR #464)

Picked up from the discussion on PR #464.

PR #1196, opened 2026-06-16
feat(training)!: integrate the checkpoint pipeline as the single checkpoint surface (#495)

### LINKED ISSUE #495: Checkpoint System Integration and Migration (Phase 3)
## Overview
Integrate the checkpoint components into a unified API and replace the legacy load path at Phase 3 of the 5-PR checkpoint refactor.

## Parent Issue
Part of #248 - Checkpoint System Refactor

## Dependencies — all merged
- Pipeline Infrastructure #493 (PR #501) ✅
- Checkpoint Acquisition #458 (PR #464) ✅
- Loading Orchestration #494 (PR #998) ✅
- Model Modifiers #410 (PR #442) ✅

## Delivery model
Phase 3 lands as **one core activation PR** plus off-critical-path **satellite PRs**. The core PR is a single coherent change; satellites carry orthogonal work and never block it.

### Core activation PR
Config wiring → run-lineage resolver (`run_id`/`fork_run_id`) → trainer flip onto the pipeline → `freeze_submodule_by_name` retirement → removal of the legacy-path guards → unified `CheckpointManager` API. **Replace-in-place**: the trainer uses the pipeline; the legacy path is retired. Per-gate deliverables, acceptance tests, and file layout are tracked in the Phase 3 gate plan (GATES.md §Phase 3) — not duplicated here.

### Satellites
- CLI tools — `inspect` / `convert` / `diff`
- Performance benchmarks
- Migration guide

## Migration — replace-in-place
No coexistence adapter. Existing configs keep working: each renamed config key emits a deprecation warning carrying its copy-pasteable replacement, and is removed after two minor releases. A migration guide ships as a satellite.

## Config (illustrative — contract is the Phase 3 gate plan)
```yaml
checkpoint:
  source:
    type: s3
    bucket: ecmwf-models
    key: anemoi/pretrained.ckpt
  loading:
    type: transfer_learning
    strict: false
    skip_mismatched: true

model_modifier:
  modifiers:
    - type: freeze
      layers: [encoder, processor.0]
# PEFT/LoRA modifiers are roadmap — see #248, not Phase 3 scope.
```

## Checklist
**Core (one PR):**
- [ ] Config wiring + pipeline builder
- [ ] Run-lineage resolver (`run_id`/`fork_run_id`)
- [ ] Trainer flip onto the pipeline (replace-in-place)
- [ ] `freeze_submodule_by_name` retirement + guard removal
- [ ] Unified `CheckpointManager` API
- [ ] Config-key deprecation shims
- [ ] Integration / end-to-end tests

**Satellites (separate PRs):**
- [ ] CLI tools (inspect / convert / diff)
- [ ] Migration guide
- [ ] Performance benchmarks

## Branch
- Core: `feature/checkpoint-integration-migration`

Live status: linked PRs.

JesperDramsch:
## Summary

Completes the checkpoint integration (issue #495) and makes
`training.checkpoint.{source,loading,modifiers}` the **single** configuration
surface and the **single owner** of checkpoint loading. The source → loading →
modifier pipeline is built from configuration and run by the trainer; every
legacy checkpoint keyword is removed and rejected at config validation. Builds
on the merged acquisition (#464), loading (#998) and modifier (#442) layers.

This is a breaking change — see **Breaking changes** below.

## What this delivers

**Builder & config group**

- `checkpoint/builder.py` — `build_checkpoint_pipeline(cfg, *, parent_run_server2server, fork_run_server2server)`
  assembles a `CheckpointPipeline` from `training.checkpoint`, ordering stages
  source → loading → modifiers. An absent `training.checkpoint` yields an empty,
  no-op pipeline. Runtime server-to-server lineage is merged onto a `RunSource`
  config here (only onto a RunSource, only non-`None` values) rather than by
  mutating built stages.
- `config/training/checkpoint/{source,loading,modifiers}/*` — the canonical
  Hydra group (sources, four loading strategies, the freezing modifier), wired
  as opt-in default groups in every shipped preset (default `null` → no pipeline).

**Single load path**

- The trainer `.model` property builds and runs the pipeline when
  `training.checkpoint` is configured; otherwise it is a plain fresh run. There
  is no legacy fallback.
- `last_checkpoint` reads the resolved path from the executed pipeline context
  (`context.checkpoint_path`, cached as `_resolved_ckpt_path`) instead of
  re-walking the source config, so source dispatch and the rank-0
  missing-checkpoint policy live in one place (`RunSource`). Rank-0 detection
  matches `lightning_fabric._get_rank()` for DDP parity.
- `RunSource` expresses resume/fork by run id and loads via `LocalSource`; an
  explicit file is a `LocalSource{path}`. `_derive_run_identity` lowers a
  `RunSource` to the internal `run_id`/`fork_run_id` contract so MLflow tags and
  output paths are byte-identical to the equivalent legacy keys.

**Single owner of warm start**

- The pipeline owns *applying the checkpoint to the model* — weights, the
  format/processor parity steps, and modifiers — run once at model-build.
- Lightning's `ckpt_path` owns *restoring training progress* — optimizer,
  scheduler and loop counters — which can only be applied at `fit()`. This is
  selected by a declared `LoadingStrategy.restores_training_state` (True only on
  `WarmStartLoader`), replacing the previous loader-name string match in
  `_skip_lightning_restore`.
- `on_load_checkpoint` skips the parity steps when the pipeline already applied
  them (`weights_initialized`), so warm start no longer runs them twice; the body
  remains as a fallback for a direct, non-pipeline Lightning load.

**Parity, validation, schema**

- The Lightning-parity logic (format/edge-perm migrations, processor refresh,
  metadata restoration) is deduplicated into shared functions in
  `checkpoint/loading/base.py`; the loading strategies and
  `AnemoiLightningModule.on_load_checkpoint` both delegate to them.
- `CheckpointPipeline` composition checks now **raise** `CheckpointConfigError`
  for ordering/conflict violations (source-before-loader, modifier-before-loader,
  multiple loaders); multiple sources still warns. The duplicated
  `CheckpointContext.__post_init__` consistency warnings are removed —
  `validation.py` is the single structural-check home.
- `training.checkpoint` is typed as `CheckpointPipelineSchema` (source/loading
  require `_target_`, modifiers is a list), replacing the untyped `dict` passthrough.

## Breaking changes

- **Removed config keys**, rejected at config validation (a `_DEPRECATED_KEYS`
  before-validator on `SchemaCommonMixin`, firing on key presence even when set
  to `null`, on both the validated `BaseSchema` and `UnvalidatedBaseSchema`):
  `training.run_id`, `training.fork_run_id`, `system.input.warm_start`,
  `training.load_weights_only`, `training.transfer_learning`,
  `training.submodules_to_freeze`. Each error names its `training.checkpoint`
  replacement; there is no deprecation-warning grace period.
- **Removed public API**: `CheckpointPipeline.from_config` and the `config=`
  constructor branch (no production caller); the loading-layer `TrainingState`
  (its optimizer/epoch restore is owned by Lightning's `ckpt_path`);
  `utils.checkpoint.transfer_learning_loading` (zero callers).
- **Exception type**: a configured-but-missing local checkpoint now raises
  `CheckpointNotFoundError` (was `FileNotFoundError`); a missing run checkpoint
  still raises `RuntimeError` on rank 0.

## Using it

```bash
# opt in via Hydra group selection (default is null = keyless)
anemoi-training train training/checkpoint/loading=weights_only
anemoi-training train training/checkpoint/source=local training/checkpoint/loading=transfer_learning

# resume a run / fork a new run from an existing run's weights
anemoi-training train training/checkpoint/source=run +training.checkpoint.source.run_id=<id>
anemoi-training train training/checkpoint/source=run +training.checkpoint.source.run_id=<id> +training.checkpoint.source.fork=true
```

or inline:

```yaml
training:
  checkpoint:
    source: { _target_: anemoi.training.checkpoint.sources.local.LocalSource, path: /path/to.ckpt }
    loading: { _target_: anemoi.training.checkpoint.loading.strategies.WeightsOnlyLoader, strict: false }
    modifiers:
      - _target_: anemoi.training.checkpoint.modifiers.freezing.FreezingModifierStage
        submodules_to_freeze: [encoder]
```

## Notes for reviewers

- **Single warm-start owner** is the main design decision: the pipeline applies
  weights+parity (once, at build); Lightning's `ckpt_path` restores
  optimizer/scheduler/loop state (the part the pipeline cannot, since those
  objects exist only at `fit()`). The one accepted residual is that `ckpt_path`
  re-reads the weights (idempotent); fully removing it would mean
  re-implementing Lightning's resume in the pipeline, which is deliberately not
  done. Covered by `test_warm_start.py`,
  `test_on_load_checkpoint_skips_parity_when_pipeline_already_applied`, and the
  `_skip_lightning_restore` parametrization.
- **`_derive_run_identity` parity**: a `RunSource` lowers to the same internal
  `run_id`/`fork_run_id` the legacy keys produced, so MLflow tags and output
  paths are unchanged (`utils/mlflow_sync.py` still indexes the internal keys
  untouched). Covered by `test_derive_run_identity_*` and `test_restarting_run.py`.
- **Removed-key errors** are covered on both schema paths by
  `tests/unit/schemas/test_deprecated_keys.py`; the typed surface by
  `CheckpointPipelineSchema` cases in `test_training_schemas.py`.
- **Path resolution** moved to the acquisition layer: `RunSource.resolve_path` /
  missing-checkpoint behaviour is tested in `sources/test_run.py` and
  `sources/test_local.py`; the trainer reads the resolved path back
  (`test_checkpoint_loading.py`).

## Testing

- Checkpoint, schema and trainer unit suites pass locally (≈466 passed, 2
  skipped, excluding the env-gated `test_s3.py` below); `ruff` clean on all
  changed files; `pre-commit` clean except as noted next.
- **`docsig`**: the refactor commit was made with `SKIP=docsig`. `docsig` trips
  on pre-existing docstring-parameter violations (`SIG203/202/402`) in unrelated
  `BaseTrainingModule` methods (`_compute_loss`, `_compute_metrics`, …) that this
  change only touches the file alongside — it surfaces them because `docsig` runs
  per changed file. Not a regression from this PR; flagged for a separate fix.
- Two env-gated local skips, both provided by CI: `tests/unit/checkpoint/sources/test_s3.py`
  needs `anemoi-utils>=0.5.7` (`settings_schema.object_storage`), and the
  `mocker`-fixture tests need `pytest-mock`.

----
📚 Documentation preview 📚: https://anemoi-training--1196.org.readthedocs.build/en/1196/

----
📚 Documentation preview 📚: https://anemoi-training--1196.org.readthedocs.build/en/1196/

----
📚 Documentation preview 📚: https://anemoi-graphs--1196.org.readthedocs.build/en/1196/

----
📚 Documentation preview 📚: https://anemoi-models--1196.org.readthedocs.build/en/1196/


JesperDramsch:
# Checkpoint pipeline parity

  Switches training checkpoint loading from the legacy in-line path to the `source → loading → modifier` pipeline, and retires the legacy checkpoint keys (`load_weights_only`, `transfer_learning`, `resume_from_checkpoint`, `submodules_to_freeze`, `run_id`/`fork_run_id`) to hard errors with migration messages. Ships as one core activation PR.

  Verified behaviour-equivalent to the legacy path:
  - Unit: full checkpoint/train/schema suite with 684 passed.
  - GPU: resume across GNN, GraphTransformer, LAM, ensemble/CRPS, multi-dataset; weights-only / cold-start / transfer-learning loading; inference-checkpoint round-trip; warm-start resume on 2× A100 (DDP).
  - Cross-version: legacy run_id resume vs pipeline resume are bitwise-identical (140/140 tensors, diff 0.0).
  - MLflow: browsable parity matrix in experiment 731: baseline + resume / fork / weights-only / cold-start / transfer-learning, with resume/fork lineage.

## 1. PR changes

- Retires the legacy checkpoint keys (`load_weights_only`, `transfer_learning`, `resume_from_checkpoint`, `submodules_to_freeze`, `run_id`/`fork_run_id`) to hard errors with a migration message; lineage is now expressed as a `RunSource` (`fork: true|false`) under `training.checkpoint.source`.
- Wires `train.py` to build and run the pipeline (`build_checkpoint_pipeline`), with warm-start ownership split cleanly: the pipeline loads weights at model-build; Lightning `ckpt_path` restores optimizer/scheduler/loop at fit, gated by each strategy `restores_training_state`.
- Deduplicates the Lightning parity helpers into shared functions in `loading/base.py`.
- Ships as one core activation PR (D17) on top of `origin/main`.

## 2. Verification ladder

| Layer | What ran | Result |
|---|---|---|
| Unit | full checkpoint + train + schema suite | **684 passed, 2 skipped** (skips = CI-only `pytest-mock`) |
| Resume | GNN + GraphTransformer via `RunSource`+`WarmStartLoader` | ✅ job 33156672 |
| Loaders | weights-only / cold-start / transfer-learning on a real GNN ckpt | ✅ job 33249665 (3 passed) |
| Architectures | LAM, ensemble/CRPS, multi-dataset resume | ✅ job 33204199 |
| Inference | save-side → `inference-*.ckpt` metadata round-trip | ✅ job 33204199 |
| DDP | warm-start resume on 2× A100 | ✅ job 33167803 |
| Cross-version | legacy `run_id` resume vs pipeline resume | ✅ **bitwise-identical** (job 33239197, 140/140 tensors, diff 0.0) |

Every one of the 436 checklist aspects was checked by execution, not just mapped to a test:
233 were already asserted by the existing suite and 180 new tests close the rest. Per-subsystem
traceability is in `PARITY-EXECUTED-COVERAGE.md`; sbatch harnesses are in
`checkpoint-sprint/parity-jobs/`.

Running the verification for real rather than assuming it caught three genuine defects that
mapping-only coverage had missed — a warm-start resume crash from a double-prefixed state dict
(found on GPU across architectures), a metadata-extraction crash on a null checkpoint, and a
permission-error path that raised on every DDP rank instead of only rank 0 — all fixed and
regression-guarded.

## 3. MLflow parity runs

Experiment logged on the server in [exp 731](https://mlflow.ecmwf.int/#/experiments/731)

Six small GNN runs (`madj/checkpoint-parity-20260630`), all derived from one baseline, make the loading-strategy behaviour comparable in the UI. Each derived run uses `RunSource` off the baseline checkpoint, varying only the loading strategy, so the experiment is itself the parity matrix.

| Run | Loading path | Lineage in UI | Demonstrates |
|---|---|---|---|
| R0-baseline-gnn | fresh train, 4 ep | — (parent) | baseline trajectory |
| R1-warmstart-resume | `WarmStartLoader`, resume | `resumedRun`, nested under R0 | continues R0's run to 8 ep; loss curve extends |
| R2-fork-warmstart | `WarmStartLoader`, fork | `forkedRun` → R0 | branches R0's full state into a new run |
| R3-weights-only | `WeightsOnlyLoader`, fork | `forkedRun` → R0 | R0 weights, fresh optimizer/epoch |
| R4-cold-start | `ColdStartLoader`, fork | `forkedRun` → R0 | weights + explicit epoch/step reset |
| R5-transfer-learning | `TransferLearningLoader`, fork | `forkedRun` → R0 | non-strict shape-tolerant load |

R3 and R4 converge to the same training loss (0.0886) under a fixed seed — expected, since
both are "fresh start from R0's weights". The resume (R1) and warm-start fork (R2) instead
continue R0's optimizer state, so they pick up the loss curve where R0 left off.


<details>
<summary>Reproducible on GPU (~2.5 min).</summary>

```python
#!/usr/bin/env python
"""Drive curated checkpoint scenarios into one offline MLflow experiment.

Produces a baseline parent run (R0) plus one run per loading path off it:
resume (warm start), fork, weights-only, cold-start, transfer-learning. All runs
land in a single offline MLflow file store so resume/fork lineage tags resolve and
the runs are comparable in the UI. The store is synced to the ECMWF server afterwards.

Mirrors the proven integration recipe (training/tests/integration: gnn_config +
test_restart_training), adding offline MLflow logging.
"""

from __future__ import annotations

import os

# Must be set before anemoi.utils.testing is imported: _offline() is lru_cached and
# skip_if_offline is evaluated at import time. MLFLOW_ALLOW_FILE_STORE is mandatory
# for the offline file store on mlflow>=3 (FileStore raises otherwise).
os.environ["IN_CI_HPC"] = "1"
os.environ["ANEMOI_BASE_SEED"] = "42"
os.environ.setdefault("MLFLOW_ALLOW_FILE_STORE", "true")

import traceback
from copy import deepcopy
from pathlib import Path

from hydra import compose, initialize_config_dir
from omegaconf import DictConfig, OmegaConf, open_dict

WORKTREE = Path(
    "/etc/ecmwf/nfs/dh1_perm_a/madj/software/anemoi/checkpoint-sprint/worktrees/checkpoint-integration-migration",
)
CONFIG_DIR = WORKTREE / "training/src/anemoi/training/config"

# test_global.yaml / testing_modifications.yaml are loaded relative to CWD by the recipe.
os.chdir(WORKTREE)

from anemoi.training.train.train import AnemoiTrainer  # noqa: E402
from anemoi.utils.testing import (
    GetTestArchive,
    GetTestData,
    TemporaryDirectoryForTestData,
)

# --- output locations (shared filesystem) --------------------------------------
OUT_ROOT = Path(os.environ["PARITY_OUT"])
RUN_ROOT = OUT_ROOT / "runs"  # training output root (checkpoints land under here)
# The offline MLflow logger resolves its file store to <system.output.root>/logs/mlflow
# regardless of a save_dir override, and never appends the run id to it, so every run
# lands in this one store and resume/fork lineage tags resolve against it.
STORE = RUN_ROOT / "logs" / "mlflow"
DATA_CACHE = Path(os.environ.get("PARITY_DATA_CACHE", str(OUT_ROOT.parent / "test-data-cache")))
EXPERIMENT = os.environ.get("PARITY_EXPERIMENT", "checkpoint-parity")
for _d in (RUN_ROOT, DATA_CACHE):
    _d.mkdir(parents=True, exist_ok=True)

DATASET_RELPATH = (
    "anemoi-integration-tests/training/datasets/aifs-ea-an-oper-0001-mars-o96-2017-2017-6h-v8-testing.zarr"
)

RUN_SRC = "anemoi.training.checkpoint.sources.run.RunSource"
LOADERS = {
    "warm_start": "anemoi.training.checkpoint.loading.strategies.WarmStartLoader",
    "weights_only": "anemoi.training.checkpoint.loading.strategies.WeightsOnlyLoader",
    "cold_start": "anemoi.training.checkpoint.loading.strategies.ColdStartLoader",
    "transfer_learning": "anemoi.training.checkpoint.loading.strategies.TransferLearningLoader",
}


def build_base_cfg() -> DictConfig:
    """Build the proven gnn_config recipe with offline MLflow logging enabled."""
    with initialize_config_dir(version_base=None, config_dir=str(CONFIG_DIR), job_name="parity"):
        template = compose(config_name="config")

    testing_mods = OmegaConf.load(WORKTREE / "training/tests/integration/config/testing_modifications.yaml")
    testing_mods.system.output.root = str(RUN_ROOT)

    use_case = OmegaConf.load(WORKTREE / "training/tests/integration/config/test_global.yaml")
    name_dataset = Path(DATASET_RELPATH).name
    extracted = Path(TemporaryDirectoryForTestData(DATA_CACHE)(DATASET_RELPATH + ".tgz", archive=True)) / name_dataset
    use_case.system.input.dataset = str(extracted)

    cfg = OmegaConf.merge(template, testing_mods, use_case)
    OmegaConf.resolve(cfg)
    assert isinstance(cfg, DictConfig)
    cfg.diagnostics.plot.callbacks = []  # datashader/matplotlib-heavy; tested elsewhere
    cfg.diagnostics.callbacks = []  # RolloutEval; tested elsewhere

    with open_dict(cfg):
        cfg.diagnostics.log.mlflow.enabled = True
        cfg.diagnostics.log.mlflow.offline = True
        cfg.diagnostics.log.mlflow.authentication = False
        cfg.diagnostics.log.mlflow.tracking_uri = None
        cfg.diagnostics.log.mlflow.experiment_name = EXPERIMENT
        cfg.diagnostics.log.mlflow.system = False
    return cfg


def download_dataset() -> None:
    tdd = TemporaryDirectoryForTestData(DATA_CACHE)
    get_test_archive = GetTestArchive(tdd, GetTestData(tdd))
    get_test_archive(DATASET_RELPATH + ".tgz")


def run_scenario(
    base: DictConfig,
    name: str,
    run_name: str,
    max_epochs: int,
    *,
    parent_id: str | None = None,
    fork: bool | None = None,
    loader: str | None = None,
) -> None:
    cfg = deepcopy(base)
    cfg.training.max_epochs = max_epochs
    with open_dict(cfg):
        cfg.diagnostics.log.mlflow.run_name = run_name
        if parent_id is not None:
            cfg.training.checkpoint = {
                "source": {"_target_": RUN_SRC, "run_id": parent_id, "fork": fork},
                "loading": {"_target_": LOADERS[loader]},
            }
    print(
        f"\n===== {name}: {run_name} (epochs={max_epochs}, parent={parent_id}, fork={fork}, loader={loader}) =====",
        flush=True,
    )
    AnemoiTrainer(cfg).train()


def summarise_store() -> None:
    print("\n===== OFFLINE STORE SUMMARY =====", flush=True)
    import mlflow

    store = STORE
    if not (store.exists() and any(store.glob("*/meta.yaml"))):
        # Fallback: the logger roots the store under system.output.root; find the
        # experiment-level meta.yaml carrying our name and take its store dir.
        for meta in RUN_ROOT.glob("**/meta.yaml"):
            if f"name: {EXPERIMENT}" in meta.read_text():
                store = meta.parent.parent
                break
    client = mlflow.tracking.MlflowClient(tracking_uri=f"file:{store}")
    exp = client.get_experiment_by_name(EXPERIMENT)
    if exp is None:
        print(f"!!! experiment {EXPERIMENT!r} not found in {store}", flush=True)
        return
    lineage_keys = ("resumedRun", "forkedRun", "forkedRunId", "mlflow.parentRunId")
    for r in client.search_runs([exp.experiment_id], order_by=["attribute.start_time ASC"]):
        tags = {k: v for k, v in r.data.tags.items() if k in lineage_keys}
        print(
            f"RUN run_id={r.info.run_id} name={r.data.tags.get('mlflow.runName')!r} "
            f"status={r.info.status} lineage={tags}",
            flush=True,
        )
    print(f"EXPERIMENT_ID={exp.experiment_id}", flush=True)
    print(f"STORE={STORE}", flush=True)


def main() -> None:
    base = build_base_cfg()
    download_dataset()

    # R0: baseline parent run.
    run_scenario(base, "R0", "R0-baseline-gnn", max_epochs=4)
    ckpt_root = Path(base.system.output.root + "/" + str(base.system.output.checkpoints.root))
    run_dirs = [d for d in ckpt_root.iterdir() if d.is_dir()]
    assert len(run_dirs) == 1, f"expected exactly one run dir after R0, found {[d.name for d in run_dirs]}"
    r0_id = run_dirs[0].name
    print(f"\nR0 run id (= mlflow run id = checkpoint dir): {r0_id}", flush=True)

    # Every loading path off the same parent. fork=True mints a fresh run from R0's
    # weights (forkedRun lineage); fork=False resumes the run itself (resumedRun lineage,
    # continues global_step). Forks run FIRST so they read R0's pristine checkpoint: the
    # resume reuses R0's run_id and writes later epochs into R0's checkpoint dir, so it
    # must run last. The warm-start fork restores R0's epoch (4) and so needs
    # max_epochs > 4; the weights-only/cold-start/transfer forks reset the epoch and train
    # a fresh 4. The resume restores epoch 4 and continues to 8.
    scenarios = [
        ("R2", "R2-fork-warmstart", 6, True, "warm_start"),
        ("R3", "R3-weights-only", 4, True, "weights_only"),
        ("R4", "R4-cold-start", 4, True, "cold_start"),
        ("R5", "R5-transfer-learning", 4, True, "transfer_learning"),
        ("R1", "R1-warmstart-resume", 8, False, "warm_start"),
    ]
    for name, run_name, epochs, fork, loader in scenarios:
        try:
            run_scenario(base, name, run_name, epochs, parent_id=r0_id, fork=fork, loader=loader)
        except Exception as exc:
            traceback.print_exc()
            print(f"!!! {name} ({run_name}) FAILED: {exc}", flush=True)

    summarise_store()


if __name__ == "__main__":
    main()
```

</details>

<details>
<summary>with the sbatch:</summary>

```bash
# ....
set -o pipefail
. /etc/bashrc
WT=/etc/ecmwf/nfs/dh1_perm_a/madj/software/anemoi/checkpoint-sprint/worktrees/checkpoint-integration-migration
VENV=/etc/ecmwf/nfs/dh1_perm_a/madj/software/anemoi/checkpoint-sprint/.venv
JOBS=/perm/madj/software/anemoi/checkpoint-sprint/parity-jobs
cd "$WT" || exit 2
source "$VENV/bin/activate" || exit 3

export IN_CI_HPC=1
export MLFLOW_ALLOW_FILE_STORE=true
export PARITY_OUT="$JOBS/mlflow-$SLURM_JOB_ID"
export PARITY_DATA_CACHE="$JOBS/test-data-cache"
export PARITY_EXPERIMENT="checkpoint-parity-20260630"

echo "=== node: $(hostname) | out: $PARITY_OUT ==="
python "$JOBS/mlflow_parity_runs.py" 2>&1
echo "=== driver exit code: $? ==="
echo "=== offline store at: $PARITY_OUT/mlruns ==="
```

</details>

JesperDramsch:
@fprill — heads-up on a change to the AICON integration config in this PR.

#1196 retires the legacy checkpoint keys (`training.run_id`, `training.fork_run_id`, `system.input.warm_start`, `load_weights_only`, `transfer_learning`, `submodules_to_freeze`) to hard errors with a migration message — resume/fork/warm-start now go through `training.checkpoint.source` (a `RunSource` for resume/fork, or a `LocalSource` for an explicit file).

`training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.yaml` still declared three of these (`run_id`, `fork_run_id`, `warm_start`, all `null`), so `test_config_validation_aicon` began failing schema validation. Since they were all `null` (no-op), I removed them — the config is behaviour-preserving and the test passes again (commit `75e544cd8`).

Nothing to do if your AICON runs never set those keys with real values. If they do, the new form is:
- resume: `training.checkpoint.source: {_target_: anemoi.training.checkpoint.sources.run.RunSource, run_id: <id>, fork: false}`
- fork: same with `fork: true`
- explicit file: `training.checkpoint.source: {_target_: anemoi.training.checkpoint.sources.local.LocalSource, path: <path>}`

Flagging so the AICON side is aware before this merges.

training/docs/user-guide/yaml/config_evaluate.yaml
```diff
@@ -1,29 +1,35 @@
...
 
 # Pattern B: point directly at a checkpoint file
-# warm_start takes precedence over run_id when both are set.
-# system:
-#   input:
-#     warm_start: /path/to/epoch=10-step=5000.ckpt
 # training:
-#   load_weights_only: True
+#   checkpoint:
+#     source:
```
cathalobrien, line 28: What about making the config syntax simpler?
```
checkpoint:
  source:  rundir
# corresponds to
# source:
#   _target_: anemoi.training.checkpoint.sources.run.RunSource
#   rundir: /rundir
```
```
checkpoint:
  source: /path/to/epoch=10-step=5000.ckpt
# corresponds to
# source:
#   _target_: anemoi.training.checkpoint.sources.local.LocalSource
#   path: /path/to/epoch=10-step=5000.ckpt
```

training/docs/user-guide/evaluation.rst
```diff
@@ -46,29 +45,39 @@ strategy), but replaces the final ``trainer.fit()`` call with
...
-   ``<checkpoints.root>/<run_id>/last.ckpt``.  Raises
-   :exc:`RuntimeError` if the file is not found.
+2. **``LocalSource``** (``training/checkpoint/source=local`` with a ``path``) —
+   load from an explicit file. Raises
+   :exc:`~anemoi.training.checkpoint.exceptions.CheckpointNotFoundError` if the
+   file does not exist.
 
-3. **Neither set** — raises :exc:`RuntimeError` immediately with a
-   descriptive message (unlike training, where a fresh start is valid).
+Evaluation expects a source to be configured; without one the model is left at
```
cathalobrien, line 61: shouldn't we keep the original behaviour? or is there a good reason to change. loudly crashing instead of doing something implicit seems good to me but you know more about this code and use cases

training/docs/user-guide/evaluation.rst
```diff
@@ -92,16 +101,18 @@ a config file and any Hydra overrides as positional arguments:
 
    anemoi-training evaluate \
        --config-name evaluate_ana_short \
-       training.run_id=<run_id>
+       training/checkpoint/source=run \
```
cathalobrien, line 104: my comment above about simplifying config syntax would also simplify this I think

training/docs/user-guide/training.rst
```diff
@@ -494,27 +494,35 @@ i.e. because the training has exceeded the time limit on an HPC system
...
-a specific checkpoint. This can either be a checkpoint from the same run
-or a checkpoint from a different run that you have run in the past or
-that you using for transfer learning. To do this, set
-``config.system.input.warm_start`` to be the path to the checkpoint they
-want to restart from.
-
-The above can be adapted depending on the use case and taking advantage
-of hydra, you can also reuse ``config.training.run_id`` or
-``config.training.fork_run_id`` to define the path to the checkpoint.
+This is done by setting ``training.checkpoint.source`` to a ``RunSource``
```
cathalobrien, line 499: stupid question but does `RunSource` mean: restart from a checkpoint stored remotely like in Mlflow (or Wandb or not?)?
if so, I think its not such a clear name. maybe the source could be called 'mlflow'?

training/docs/user-guide/training.rst
```diff
@@ -685,11 +716,17 @@ model are excluded from training. This is useful when certain parts of
 the model have been sufficiently trained or should remain unchanged for
 the current task.
 
-To specify which submodules to freeze, use the
-``config.training.submodules_to_freeze`` field in the configuration. List
-the names of submodules to be frozen. During model initialization, these
-submodules will have their parameters frozen, ensuring they are not
-updated during training.
+The recommended way to freeze submodules is through the
```
cathalobrien, line 722: `the recommended way`
is there another way?

training/docs/user-guide/training.rst
```diff
@@ -604,22 +613,44 @@ trained checkpoint. This is particularly useful when the new task is
 related to the old one, enabling faster convergence and often improving
 model performance.
 
-To enable transfer learning, set the config.training.transfer_learning
-flag to True in the configuration file.
+The recommended way to configure transfer learning is through the
```
cathalobrien, line 619: `The recommended way` again, is there another way?

cathalobrien (commented):
nice work, I just had a glance at some docs. will try look more later

training/docs/user-guide/yaml/config_evaluate.yaml
JesperDramsch: Hi @cathalobrien thanks for the review. This is specifically built to also enable retrieving checkpoints from:
- S3
- Run IDs (our old behaviour)
- local paths
- Azure (opened an issue in #1139 to keep track after @mchantry's suggestion in the source PR)

I would much prefer to not magically guess where the user wants to get their checkpoint from, but we can make reasonable default behaviour of course.

training/docs/user-guide/evaluation.rst
JesperDramsch: I don't think it will.

training/docs/user-guide/training.rst
JesperDramsch: Yes, it's from our RunID, we can rename it, but it's essentially the canonical identifier we had been using for a while. Can rename it of course, I'm not married to the name, just thought `RunIDSource` is ugly that's all.

training/docs/user-guide/training.rst
JesperDramsch: Many. You can always do it by hand before loading a checkpoint or via other Lightning utils (I believe).

training/docs/user-guide/training.rst
JesperDramsch: Use pipeline-stages outside of the loader or you can do good ol' checkpoint-surgery.

training/docs/user-guide/yaml/config_evaluate.yaml
cathalobrien: yeah that's fair. 
I was just thinking about the local and mlflow cases, where in which cases you wouldn't have to guess. you could check if the input was an existing path. and for s3 you could check for 's3' at the start of the url. not sure about azure though... so maybe this is good.

training/docs/user-guide/training.rst
cathalobrien: I think it reads a little confusingly. how about 'to configure transfer learning through the config ...'

training/docs/user-guide/training.rst
cathalobrien: again, I think its confusing to hint at the existence of other ways. Lets keep users on the straight-and-narrow. what about 'to freeze submodules through the config use the ``training.checkpoint`` pipeline, ...'

training/docs/user-guide/training.rst
JesperDramsch: You did not say it was confusing, you asked if there is another way. I'm happy to adjust now that you say it's confusing for you. I think it's nice to read that a package recommends a certain way of usage, but I'm happy to change it now that you've clarified.

training/docs/user-guide/evaluation.rst
anaprietonem: do users need's to pass both? or could they just do `+training.checkpoint.source.run_id=run_id`?

training/docs/user-guide/evaluation.rst
anaprietonem: I don't see changes in the evaluate command so I assume the `RuntimeError` should it still be raised? I agree with Cathal this behaviour shouldn't be changed

training/docs/user-guide/training.rst
```diff
@@ -494,27 +494,35 @@ i.e. because the training has exceeded the time limit on an HPC system
...
+checkpoints go in the same folder as the old ones:
+
+.. code:: yaml
+
+   training:
+     checkpoint:
+       source:
+         _target_: anemoi.training.checkpoint.sources.run.RunSource
+         run_id: <run_id>
+         fork: false   # resume; set true to fork into a new run
```
anaprietonem: does this always need to defined? as in could users/devs just do:
``
         _target_: anemoi.training.checkpoint.sources.run.RunSource
         run_id: <run_id>
``
and assume that's resuming because fork is default False?

training/src/anemoi/training/checkpoint/loading/strategies.py
```diff
@@ -68,16 +69,18 @@ async def process(self, context: CheckpointContext) -> CheckpointContext:
         """
         self._apply_format_migrations(context)
         self._refresh_checkpoint_processors(context)
+        self._apply_trainable_edge_perm_migration(context)
```
anaprietonem, line 72: what is this function about? is this one of the existing anemoi-models migration? is this is a new function compared to main?

training/src/anemoi/training/checkpoint/loading/strategies.py
```diff
@@ -68,16 +69,18 @@ async def process(self, context: CheckpointContext) -> CheckpointContext:
...
         state_dict = self._extract_state_dict(context)
 
         try:
             context.model.load_state_dict(state_dict, strict=self.strict)
         except RuntimeError as e:
-            msg = f"Failed to load state dict into model: {e}"
-            raise CheckpointLoadError(msg) from e
+            raise CheckpointLoadError(context.checkpoint_path or "<in-memory checkpoint>", e) from e
 
+        self._warn_on_hparams_divergence(context)
```
anaprietonem, line 81: similarly this is new functionality? Will this be checked even when forking runs? cause then I'd assume hparams can be configured differently

training/src/anemoi/training/checkpoint/sources/run.py
```diff
@@ -0,0 +1,188 @@
...
+        LOGGER.info("RunSource: resolved checkpoint path (%s): %s", resolution, path)
+        context.checkpoint_path = path
+        context.update_metadata(resolved_checkpoint_path=str(path), lineage_resolution=resolution)
+
+        # Delegate the actual torch.load + format detection to LocalSource so the
+        # load path is identical to an explicit local checkpoint.
+        return await LocalSource().process(context)
+
+
+def _is_rank_zero() -> bool:
```
anaprietonem, line 208: why having this and not relying in the is_rank_zero decorator we had before? what are the implications of settings this env vars when using DDP or other model parallelisation strategies?

training/src/anemoi/training/checkpoint/loading/strategies.py
```diff
@@ -170,88 +174,68 @@ async def process(self, context: CheckpointContext) -> CheckpointContext:
...
-        # 1. Model weights (strict — exact match expected for resume)
+        # Model weights (strict — exact match expected for resume)
         state_dict = self._extract_state_dict(context)
         try:
             context.model.load_state_dict(state_dict, strict=True)
         except RuntimeError as e:
             msg = f"WarmStart requires exact model match: {e}"
             raise CheckpointIncompatibleError(msg) from e
 
-        # 2. Optimizer state
```
anaprietonem, line 212: why is this being removed?

anaprietonem (commented):
Still going through the code, some comments above!

training/src/anemoi/training/checkpoint/loading/strategies.py
```diff
@@ -129,6 +132,7 @@ async def process(self, context: CheckpointContext) -> CheckpointContext:
 
         self._apply_format_migrations(context)
         self._refresh_checkpoint_processors(context)
```
anaprietonem, line 134: I probably missed this in the other review, but where is this `_refresh_checkpoint_processors` coming from?

training/src/anemoi/training/train/methods/base.py
```diff
@@ -33,6 +32,10 @@
 from anemoi.models.distributed.graph import gather_tensor
 from anemoi.models.interface import AnemoiModelInterface
 from anemoi.models.utils.config import get_multiple_datasets_config
+from anemoi.training.checkpoint.loading.base import apply_trainable_edge_perm_migration
+from anemoi.training.checkpoint.loading.base import extract_checkpoint_variables_metadata
+from anemoi.training.checkpoint.loading.base import preserve_anemoi_metadata
+from anemoi.training.checkpoint.loading.base import refresh_checkpoint_processors
```
anaprietonem, line 38: arent' those also called inside the loaders?

training/src/anemoi/training/train/train.py
```diff
@@ -677,6 +816,34 @@ def strategy(self) -> Any:
             static_graph=not self.config.training.accum_grad_batches > 1,
         )
 
+    def _skip_lightning_restore(self) -> bool:
```
anaprietonem, line 785: I am not sure I get the idea behind this or that's it's clear for the users/devs when this is being skipped

training/src/anemoi/training/train/train.py
```diff
@@ -417,41 +561,33 @@ def run_id(self) -> str:
...
-
-        if rank_zero_only.rank == 0:
-            msg = "Could not find last checkpoint: %s", checkpoint
-            raise RuntimeError(msg)
-
-        return None
+        # Building the model runs the checkpoint pipeline, whose source stage sets
+        # the resolved path (cached property: in the real flow the model is already
+        # built before ckpt_path is read, so this is a cache hit).
+        _ = self.model
```
anaprietonem: I understand the above but find it a bit counter intuitive we have to call the model generation to get this path

training/src/anemoi/training/train/train.py
```diff
@@ -57,6 +55,63 @@
...
+    prior_struct = OmegaConf.is_struct(config)
+    OmegaConf.set_struct(config, False)
+    try:
+        config.training.run_id = run_id
+        config.training.fork_run_id = fork_run_id
+    finally:
+        OmegaConf.set_struct(config, prior_struct)
+
+
+def _reject_unsupported_warm_start(pipeline_cfg: DictConfig) -> None:
```
anaprietonem: I'd move this as utility function to the checkpoint moduke

training/src/anemoi/training/train/train.py
```diff
@@ -57,6 +55,63 @@
 PL_VERSION = version.parse(pl.__version__)
 
 
+def _write_run_identity(config: DictConfig, run_id: str | None, fork_run_id: str | None) -> None:
+    """Write the internal run-identity keys onto the live config.
+
+    ``training.run_id`` / ``training.fork_run_id`` were removed from the schema,
+    so the validated config carries no slot for them; the MLflow logger, the
```
anaprietonem: can't we update the MlFlow logger part to accomodate to new keys? as in get the RunSource to return the run id and pass that to the Logger?

training/src/anemoi/training/train/train.py
```diff
@@ -351,61 +420,136 @@ def model(self) -> pl.LightningModule:
...
+        # Trainer-side parity until the dataset/units validators move into the
+        # pipeline: when weights were loaded, keep the current config's data
+        # indices and run the transfer-learning compatibility checks.
+        if has_loading:
+            loaded_model.data_indices = self.data_indices
+            self._validate_transfer_learning_datasets(loaded_model)
+            self._validate_transfer_learning_units(loaded_model)
+        return loaded_model
+
+    def _derive_run_identity(self) -> None:
```
anaprietonem: should this be part of the trainer or the Sources?

training/docs/user-guide/evaluation.rst
JesperDramsch: Evaluate still raises the moment `training.checkpoint.source` is missing 

https://github.com/ecmwf/anemoi-core/blob/ef792377b72c4e15dde6e9f6b5cdc554a9365b1f/training/src/anemoi/training/train/evaluate.py#L56-L63

Totally agree that this should still crash loudly.

training/src/anemoi/training/checkpoint/loading/strategies.py
JesperDramsch: It's the trainable-edge-permutation migration `on_load_checkpoint` already ran. The loader just replays it so a pipeline load matches the legacy path exactly.

training/src/anemoi/training/checkpoint/loading/strategies.py
JesperDramsch: Good question. This should not block a fork with different hparams as It only warns

https://github.com/ecmwf/anemoi-core/blob/ef792377b72c4e15dde6e9f6b5cdc554a9365b1f/training/src/anemoi/training/checkpoint/loading/base.py#L526-L531

So basically, It's a heads-up, not a gate. And the specific case you're picturing, fewer variables, is now an opt-in: `training.allow_variable_subset` (which is supposed to be the #838 fix in this PR).

training/src/anemoi/training/checkpoint/loading/strategies.py
JesperDramsch: It's the shared helper in `checkpoint/loading/base.py`. It re-injects the pre/post-processors from the current model when `update_ds_stats_on_ckpt_load` is set, the same parity step `on_load_checkpoint` runs. 

This is linked to the early-stop behaviour in https://github.com/ecmwf/anemoi-core/pull/1196#discussion_r3585662751

training/src/anemoi/training/train/methods/base.py
JesperDramsch: Yes, but they don't run twice.

`on_load_checkpoint` bails out early via `weights_initialized` once the pipeline has applied weights + parity at build. There's a test pinning that skip.

https://github.com/ecmwf/anemoi-core/blob/ef792377b72c4e15dde6e9f6b5cdc554a9365b1f/training/src/anemoi/training/train/methods/base.py#L468-L482

training/src/anemoi/training/checkpoint/loading/base.py
```diff
@@ -379,11 +333,199 @@ def _inject_model_weights(
...
+
+    Returns the (possibly rewritten) checkpoint. A no-op when the migration is
+    unavailable in the installed anemoi-models, and tolerant of incomplete
+    checkpoint shapes (raw ``state_dict`` saves / test fixtures lack the
+    ``hyper_parameters.config`` tree the migration reads), which are skipped
+    rather than raising.
+    """
+    if checkpoint is None:
+        return checkpoint
+    migrate = _load_chunking_fix_migration()
```
fprill, line 378: Does this implementation handle checkpoint migrations differently from the existing (`main`) behavior? The `WeightsOnlyLoader` seems to execute `models/src/anemoi/models/migrations/scripts/1762857428_chunking_fix.py` even when it is not needed (this specific migration fix appears to assume that a processor is present, which is problematic when loading autoencoder weights).

training/src/anemoi/training/config/training/checkpoint/source/http.yaml
```diff
@@ -0,0 +1,5 @@
+# Checkpoint source — download a checkpoint over HTTP(S). Set `url` to the
+# checkpoint location; optionally set `expected_checksum` (sha256) to verify
+# integrity after download.
+_target_: anemoi.training.checkpoint.sources.http.HTTPSource
```
JPXKQX, line 4: would it makes sense to remove the "http", "local", ... and other file names for the absolute path? I see they already already available through the `source/__init__.py`

JPXKQX (commented):
Hi, I took a look at the PR and it looks great to me. The 'Checkpoint Pipeline Configuration' page is amazing!

I just have one question: how do you think these changes will interact with having more and more components (encoders/decoders) inside the model? For example, I can imagine people using multiple loaders at the same time to load two decoders from different sources. I think these changes align well with the work done in this series of PRs, but I just wanted to check if there are any specific details that could conflict with this in a near future

training/src/anemoi/training/checkpoint/loading/strategies.py
JesperDramsch: Yes and No.

The warn-not-gate half holds: the `hparams` comparison only warns, on forks too. The `allow_variable_subset` half did not. With a weights-only load the strict `load_state_dict` raised on the size mismatch before that check was ever reached, so the #838 case only ever worked through `TransferLearningLoader`.

The latest push fixes that. `WeightsOnlyLoader` stays strict by default and gains `skip_mismatched: true`: every compatible parameter loads, the shape-mismatched ones are dropped with a warning and recorded in `metadata["skipped_params"]`, and an empty load is refused outright. `ColdStartLoader` inherits that behaviour. Then `allow_variable_subset` stays the dataset-side switch for the variable-order check. The presets, the `allow_subset` docstring and a troubleshooting entry now say which loaders make the reduction loadable.

training/src/anemoi/training/checkpoint/loading/base.py
JesperDramsch: You were right, on the file-backed path.

At the revision you reviewed (thanks for that!), a checkpoint whose ledger did not record `chunking_fix` was migrated in-process with `Migrator.sync` on the file. That applies every outstanding migration, whatever the processor. The in-memory fallback was already screened on processor geometry, so only the dict-only path was safe.

I removed the in-process sync. So now, the loader reads the ledger instead. A checkpoint that records every shipped migration is returned unchanged, so nothing re-runs on a current checkpoint. One that is behind gets a warning, naming the outstanding migrations and the `anemoi-models migration sync` command, and the load continues to the in-memory `chunking_fix` fallback. That fallback runs only when the checkpoint's processor declares `num_layers` and `num_chunks`. A `NoOpProcessor` declares neither, so autoencoder weights are never touched, on a weights-only load or on a resume through `on_load_checkpoint`.

Now tests pin both paths.

training/docs/user-guide/training.rst
JesperDramsch: Yes.

`fork` defaults to false, so `run_id: <id>` alone is a resume, and the docs now show it that way.

The other direction is covered too. Leaving `run_id` unset (the `source=run` preset ships it as `null`) fails at config time with a `CheckpointConfigError` naming `training.checkpoint.source.run_id`, instead of the old "corrupted checkpoint" message.

training/src/anemoi/training/checkpoint/sources/run.py
JesperDramsch: Because the acquisition layer does not import Lightning.

So it reads the same environment variables in the same priority order as `lightning_fabric`'s `_get_rank` (`RANK`, `LOCAL_RANK`, `SLURM_PROCID`, `JSM_NAMESPACE_RANK`) instead of using the decorator. Under DDP the behaviour is the legacy one: rank 0 raises on a missing checkpoint, the other ranks defer and the job is torn down with rank 0. That policy now has one home, `RunIdSource.resolve`. The trainer no longer re-derives the path.

Your question also turned up a real defect. The deferring ranks used to print "corrupted checkpoint". I now fixed it as follows: a rank that defers marks the context `source_deferred`, and a loading stage that runs afterwards aborts with a `CheckpointLoadError` saying rank 0 reported the missing file.

training/src/anemoi/training/checkpoint/loading/strategies.py
JesperDramsch: It wasn't possible.

The pipeline runs at model build, before the optimizer and scheduler exist. Lightning's `ckpt_path` restores them at `fit()`. What was left after that (`TrainingState`, `context.optimizer`/`.scheduler` and the checks on them) only suggested a restore happens in the pipeline when none does. 244f5eaa9 deletes it.

training/src/anemoi/training/train/train.py
JesperDramsch: Agreed, it was not clear.

I've reduced it to one rule: a checkpoint has one loader. On a resume (`WarmStartLoader`, or no `loading` block) `ckpt_path` is passed and Lightning loads weights, optimizer, scheduler and loop state in one go. With any other strategy the pipeline already applied the weights, so `ckpt_path` is withheld.

The decision lives in `builder.resumes_via_lightning`, shared by the builder and the trainer, so the two cannot disagree. The docstring now says exactly that.

training/src/anemoi/training/train/train.py
JesperDramsch: That's fair. `last_checkpoint` is the path the source stage resolved while the model was built, read back from the executed pipeline instead of derived a second time from the config. That is why it touches `self.model`. The two derivations used to disagree about `~`; now the file Lightning receives is the one the source checked.

The pipeline itself runs through `CheckpointPipeline.execute_sync`, so building the model from inside a running event loop (a notebook, an async launcher) works too.

training/src/anemoi/training/train/train.py
JesperDramsch: Done.

Path resolution now lives in the sources themselves (`CheckpointSource.resolve`, implemented per source), and the resume decision in `builder.resumes_via_lightning`. The trainer no longer inspects `_target_` strings for sources at all. `source_class_from_config` resolves the class, and two classmethods carry what the trainer needs: `run_identity` (run id and fork flag) and `with_run_lineage` (server-to-server ids, merged only by sources that accept them).

training/src/anemoi/training/train/train.py
JesperDramsch: The sources.

Each one implements `resolve(context)`, which makes the checkpoint reachable as a local file. `LocalSource` expands and canonicalises the path, `RunIdSource` finds the run's `last.ckpt`, `HTTPSource` and `S3Source` download to a node-local file and keep it until training finishes, then the trainer deletes it. The trainer only reads the resulting path.

training/src/anemoi/training/train/train.py
JesperDramsch: Done.

`RunIdSource.run_identity` returns the run id and fork flag (other sources return none), and the trainer passes that to the logger instead of matching `_target_` strings or writing keys back onto the config. `anemoi-training mlflow prepare` reads the same function.

One consequence: forking into a run id prepared by `mlflow prepare` is no longer expressible on this surface. The code now records that in `tracking.rst`, and removes the logger branch that served it, since nothing exercised it.

training/src/anemoi/training/checkpoint/loading/strategies.py
JesperDramsch: Addendum to my earlier answer.

The replay is the same `trainable_edge_perm_fix` migration main runs first in `on_load_checkpoint`. Since it is not replayed per strategy any more. `apply_checkpoint_corrections` runs format migrations, then this migration, then the processor refresh, in that order, on both the pipeline path and the resume hook.

training/docs/user-guide/evaluation.rst
JesperDramsch: Just the run id.

`training/checkpoint/source=run +training.checkpoint.source.run_id=<id>` is enough. A source with no `loading` block is a resume, so no loading override is needed, and evaluate reads the same resolved path the trainer would. The docs show that form now.

training/src/anemoi/training/config/training/checkpoint/source/http.yaml
JesperDramsch: Both work.

Hydra imports whatever the dotted path names, and `sources/__init__.py` re-exports `HTTPSource`, `LocalSource`, `RunIdSource` and `S3Source`, so `anemoi.training.checkpoint.sources.HTTPSource` resolves today (the builder goes through `hydra.utils.get_class`).

I kept the full module path on purpose. It is the form every other `_target_` in the shipped configs uses, it is what the error names when a target fails to import, and it survives a module split without the re-export having to follow.

Either is fine by me. If you prefer the short form I switch the four source presets and the four loading presets together, nine lines, no code impact. Which do you prefer?

training/docs/user-guide/training.rst
JesperDramsch: Renamed to `RunIdSource` in 0ea78c8ad.

training/docs/user-guide/training.rst
JesperDramsch: Reworded as you suggested. No "recommended way" wording is left in `training/docs/`.

JesperDramsch:
> Hi, I took a look at the PR and it looks great to me. The 'Checkpoint Pipeline Configuration' page is amazing!
> 
> I just have one question: how do you think these changes will interact with having more and more components (encoders/decoders) inside the model? For example, I can imagine people using multiple loaders at the same time to load two decoders from different sources. I think these changes align well with the work done in this series of PRs, but I just wanted to check if there are any specific details that could conflict with this in a near future

Thanks @JPXKQX , glad the configuration page is useful.

Nothing in this PR blocks it per sé, and nothing supports it yet either.

Today `training.checkpoint` is one `source`, one `loading` block and a list of `modifiers`. The pipeline enforces that shape: more than one loading strategy is a hard `CheckpointConfigError` at construction, more than one source only warns. The context carries one `checkpoint_path` and one `checkpoint_data`. So "decoder A from run X, decoder B from S3" has no config form right now.

What keeps it cheap to add is that none of that is baked into the stages. The pipeline is a plain ordered list, roles are decided by base class (`CheckpointSource`, `LoadingStrategy`, `ModelModifier`), the catalog discovers new stages the same way, and the state-dict filtering is key-based, so scoping a load to a submodule prefix is a small change. Modifiers already run as a list after loading with the model in hand; `FreezingModifierStage` takes a `submodule_root` for exactly this per-component reason.

A multi-component design would have to touch two things: the one-loader invariant (becomes one loader per submodule) and the single `checkpoint_data` on the context (becomes per-stage). Both are local. What would not change: a resume stays one Lightning load of one file. Stitching decoders from several checkpoints is a fresh-start case (weights-only or cold start), which is the path the pipeline already owns basically.

I would rather add that once the multi-decoder config shape exists than pre-build it wrong now. Which PR in your series defines how the components are named? I can check those prefixes against `filter_state_dict` and the `submodule_root` lookup so the two stay aligned.

PR #1200, opened 2026-06-18
feat: make cartopy projections flexible

anaprietonem:
## Description

Make flexible Cartopy projections for diagnostic plots

Summary

- Any cartopy.crs projection is now accepted via projection_kind in the plot config. Any cartopy.crs class name in snake_case (e.g. robinson, mollweide, orthographic) is resolved dynamically, in addition to the built-in equirectangular and lambert_conformal. The projection_kind schema field is widened from a Literal to str with updated docs.

- Cartopy axes render correctly: Projection.for_plot now passes geographic degree coordinates + a PlateCarree transform for Cartopy axes, letting Cartopy reproject internally. Coastlines and borders store both projected and degree-space line collections, using the degree-space one with a PlateCarree transform on Cartopy axes.

- datashader + non-equirectangular projection is now caught at the PlottingSettings.from_plot_config boundary with a warning and coercion to equirectangular, rather than silently inside _setup_figure_and_colormaps. _setup_figure_and_colormaps no longer takes a datashader parameter. A schema-level validator also raises an explicit error when config_validation=True.

- Global projections no longer crash: single_plot skips set_extent for near-global data (lat span > 150° or lon span > 340°), fixing a ValueError: Axis limits cannot be NaN or Inf when using projections like Robinson or Mollweide on global grids.

- Lambert Conformal guards: falls back to equirectangular with a warning when the domain spans more than 60° latitude or 180° longitude. Standard parallels are nudged when they would sum to zero (equator-straddling domains).

- plot_flat_sample axis hiding unified into a single ax[ii].axis("off") when data[ii] is None, fixing incorrect axis hiding for autoencoder and diagnostic-only fields.

Config
```
diagnostics:
  plot:
    projection_kind: robinson   # equirectangular, lambert_conformal, or any cartopy.crs snake_case name
    datashader: false           # must be false for any non-equirectangular projection
```

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1200.org.readthedocs.build/en/1200/

----
📚 Documentation preview 📚: https://anemoi-graphs--1200.org.readthedocs.build/en/1200/

----
📚 Documentation preview 📚: https://anemoi-models--1200.org.readthedocs.build/en/1200/


training/src/anemoi/training/diagnostics/projections.py
VeraChristina: I think we can remove this now?

training/src/anemoi/training/schemas/diagnostics.py
```diff
@@ -205,11 +205,19 @@ class PlotSchema(PydanticBaseModel):
     "Handle plotting tasks without blocking the model training."
     datashader: bool
     "Use Datashader to plot."
-    projection_kind: Literal["equirectangular", "lambert_conformal"] = Field(
+    projection_kind: str = Field(
```
VeraChristina, line 208: Can we do additional validation to make sure this is a projection? Perhaps too complicated

training/src/anemoi/training/diagnostics/projections.py
```diff
@@ -104,12 +108,45 @@ def lambert_conformal(cls, latlon: np.ndarray) -> "Projection":
...
+        crs_name = "".join(part.capitalize() for part in kind.split("_"))
+        crs_cls = getattr(ccrs, crs_name, None)
+        if crs_cls is None:
+            msg = (
+                f"Unknown projection_kind='{kind}'. "
+                f"Use 'equirectangular', 'lambert_conformal', or any cartopy.crs class name in snake_case "
+                f"(e.g. 'robinson', 'mollweide', 'orthographic')."
+            )
+            raise ValueError(msg)
+        return cls(crs_cls())
```
VeraChristina, line 194: Perhaps worth noting somewhere that the projection will be instantiated with defaults

VeraChristina (commented):
Thanks for adding this! It looks like the cartopy and non-cartopy paths are nicely separated.

A few comments below. What I haven't tested is whether the shared plotting paths work for cartopy projections. More generally, it might be good to add tests for the cartopy path as it doesn't seem to be covered by existing tests.

training/docs/user-guide/yaml/config_evaluate.yaml
```diff
@@ -0,0 +1,47 @@
+# Minimal Hydra override file for `anemoi-training evaluate`.
+#
+# A checkpoint MUST be specified (Pattern A or B below) or the command will
+# raise a RuntimeError before evaluation starts.
+#
+# Usage:
+#   anemoi-training evaluate --config-dir /path/to/overrides --config-name config_evaluate
```
frazane: `--config-dir` or `--config-path`?

training/src/anemoi/training/schemas/base_schema.py
```diff
@@ -170,6 +170,27 @@ def _check_deprecated_targets(cls, values: Any) -> Any:
             raise ValueError(msg)
         return values
 
+    @model_validator(mode="before")
+    @classmethod
+    def _check_datashader_projection_compat(cls, values: Any) -> Any:
+        """Raise if a non-equirectangular projection is combined with datashader."""
```
frazane: Here we raise an error but at runtime we warn and coerce to equirectangular. Is this intentional?

training/src/anemoi/training/diagnostics/plots.py
```diff
@@ -878,6 +873,8 @@ def plot_flat_sample(
                 datashader=datashader,
                 transform=transform,
             )
+        else:
+            ax[ii].axis("off")
 
 
 def single_plot(
```
frazane, line 880: In the PR comment you mention

> Global projections no longer crash: single_plot skips set_extent for near-global data (lat span > 150° or lon span > 340°), fixing a ValueError: Axis limits cannot be NaN or Inf when using projections like Robinson or Mollweide on global grids.

but I cannot find this, and I my tests with global projections are crashing.

frazane (commented):
Tested and it's working except when using global + non-equirectangular projection (see comment). 

On a more general note: I see here we are using three different terminologies: "crs", "projection" and "transform". I haven't given it much thought, but I wonder if it's a problem (I think we are also calling the exact same object with different terms). Also, "projection" could be mixed up with the other concept of sparse-matrix projections that is becoming prevalent in anemoi. Note that this is not a blocker but just wanted to open a discussion!

training/src/anemoi/training/schemas/base_schema.py
anaprietonem: Aha yes indeed fair point, I will update the schema same as runtime, so here we just coerce and warn!

training/src/anemoi/training/schemas/diagnostics.py
anaprietonem: I thought about this, the issue would be then we'd need to import cartopy and get the list of projections or store that list somewhere, and that's why I then keep it lean here and do the check in Projection.from_kind

training/docs/user-guide/yaml/config_evaluate.yaml
anaprietonem: path, fixed now!

training/src/anemoi/training/diagnostics/plots.py
anaprietonem: I had the changes locally, had forgotten to push them, should be good now, if you can please try again!

anaprietonem:
> Tested and it's working except when using global + non-equirectangular projection (see comment).
> 
> On a more general note: I see here we are using three different terminologies: "crs", "projection" and "transform". I haven't given it much thought, but I wonder if it's a problem (I think we are also calling the exact same object with different terms). Also, "projection" could be mixed up with the other concept of sparse-matrix projections that is becoming prevalent in anemoi. Note that this is not a blocker but just wanted to open a discussion!

Thanks Francesco, about the naming, yes that's true. The main driver to use those names is that because matplotlib subplot takes a [projection](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.subplot.html) and [scatter](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html) takes a transform, and the CRS is a term used by cartopy or pyproj to define the projection. I've tried to simplify things a bit without breaking the configs so that now:
- Projection is called  MapProjection to avoid clash with graph/sparse-matrix projections in anemoi)

And then crs I have tried to separate between data_crs and axes_crs to reduce the use of proj/transform terms. 
- crs_for_axes() changed method name to axes_crs()
- proj → axes_crs - what we pass to cartopy
- transform → data_crs (parameter/variable name throughout plots.py and maps.py, so for example ccrs.PlateCarree() would the data_crs describing the input data's coordinate system)
- crs → data_crs (parameter name in maps.py plot_continents/plot_borders/MapFeatures.plot)

training/src/anemoi/training/diagnostics/projections.py
anaprietonem: yes fair point, I have added now a comment to make that clear! the idea is that Lambert Conformal is adjusted, and if we need to do this with more regional we can extend but for all the general ones, as you say, we instante with the default cause that's sufficient

frazane (dismissed):
Thanks for the quick turnaround, Ana — re-checked against the latest push and both my blockers are resolved:

- The datashader/projection validator is gone from `base_schema.py`, so the unvalidated path is clean again and CI is green. Nice side effect: the config-time-raise vs runtime-warn divergence I was worried about is now a single behaviour (warn + coerce at runtime).
- The `is_global` guard is in `single_plot` now. I re-ran my projection matrix on this head: global robinson, mollweide and orthographic all render (they used to raise `Axis limits cannot be NaN or Inf`), and the maps look right — continents land where they should. The new regression test covers robinson + mollweide; orthographic only fell out of my local harness, so it might be worth adding to the parametrize list.

The terminology cleanup (`MapProjection`, `data_crs`/`axes_crs`) reads much better and dodges the sparse-matrix "projection" clash.

One non-blocking nit: `equirectangular_projection` in `projections.py` looks unused now (no callers in src or tests) and its docstring calls it a backward-compat wrapper — if that's not intentional it can go.

Approving — nice work.

anaprietonem:
thanks @frazane and @VeraChristina  for the review! integration tests are passing https://github.com/ecmwf/anemoi-core/actions/runs/28933485574 too!

PR #1201, opened 2026-06-18
test(training): Loss convergence test

mcocdawc:
## Description

This PR introduces a loss convergence integration test. These new set of tests will run nightly on a similar fashion as the existing integration and PB tests. 
It should detect subtle changes arising from changing defaults etc. that might affect the loss convergence.
It tracks deviation from a reference run, hence it has the assumption of deterministicity in the run.
Depending on GPU, backend libraries etc. this might not be always true. 
Perhaps it is necessary to relax the very strict element-wise comparison in the future.

In addition, there will surely be PRs that change the loss convergence for good. They will need to update the reference ID.

This is the first step to develop an accuracy testing suite, and follow up work will include extension to regional models and other use cases supported in anemoi apart from global forecasting. However we consider merging this is a first step is a good feature to start gathering data and better identify problems associated with accuracy in the future.

---------------------------

Other minor changes: 

The integration tests were split via a matrix.

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1201.org.readthedocs.build/en/1201/

----
📚 Documentation preview 📚: https://anemoi-graphs--1201.org.readthedocs.build/en/1201/

----
📚 Documentation preview 📚: https://anemoi-models--1201.org.readthedocs.build/en/1201/


training/tests/integration/test_training_cycle.py
```diff
@@ -413,33 +386,53 @@ def test_training_cycle_imerg_target(
...
+                "step": [m.step for m in history],
+                "loss": [m.value for m in history],
+            }
+        ).set_index("step")
+
+    def is_similar(run_id1, run_id2):
+        df1, df2 = get_loss_df(run_id1), get_loss_df(run_id2)
+        return np.allclose(df1.loc[:, "loss"], df2.loc[:, "loss"])
+
+    assert is_similar(runs[0].info.run_id, REFERENCE_ID), f"Loss curve for run {runs[0].info.run_id} does not match reference"
```
mcocdawc: here we need something more stable than just the last run. Upon submitting the calculation we imidiately want to obtain the run_id.

training/src/anemoi/training/utils/config.py
```diff
@@ -0,0 +1,69 @@
...
+from pathlib import Path
+
+from hydra import compose
+from hydra import initialize_config_dir
+from hydra.core.global_hydra import GlobalHydra
+from omegaconf import DictConfig
+from omegaconf import OmegaConf
+
+
+def load_config(
```
mcocdawc: Could potentially replace

Reuse / references found

 - Search-path plugin (makes defaults: resolve from the package): hydra_plugins/anemoi_searchpath/anemoi_searchpath_plugin.py:72.
 - Existing compose pattern: commands/config.py:194 (validate_config), :203 (dump_config).
 - Validation helper to recommend to callers: schemas/base_schema.py:195 (validate_schema).
 - Test compose patterns to mirror: training/tests/unit/conftest.py:24, training/tests/unit/commands/test_config.py:96.


(this list was AI generated)

.github/workflows/accuracy-test-hpc.yml
```diff
@@ -0,0 +1,96 @@
+name: accuracy-test-hpc-gpu
+
+ # Controls when the workflow will run
+on:
+  pull_request:
```
mcocdawc: Is executed on PRs

.github/workflows/integration-tests-hpc.yml
```diff
@@ -23,16 +23,63 @@ jobs:
...
+    name: ${{ matrix.test_name }}
     runs-on: hpc
+    timeout-minutes: ${{ matrix.timeout_minutes }}
     needs: check_date
     if: ${{ needs.check_date.outputs.should_run != 'false' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - test_name: test_accuracy
```
mcocdawc: do we want the test_accuracy twice? I don't thinks so

.github/workflows/accuracy-test-hpc.yml
cathalobrien: every commit or when?

cathalobrien:
great work @mcocdawc, we could enforce as much determinism as possible. from cublas and torch. this will increase the runtime by 2x and use about 50% more memory. but this is a small test so should be fine.
```
os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":16:8"
torch.backends.cudnn.benchmark = False
torch.backends.cudnn.deterministic = True
torch.use_deterministic_algorithms(True)
seed=42
torch.manual_seed(seed)
np.random.seed(seed)

```

training/tests/integration/test_training_cycle.py
cathalobrien: for the benchmark tests there is a function to find the last shared commit between the current pr and a database containing past runs
https://github.com/ecmwf/anemoi-core/blob/f0e8206038a827a48b771bc6813e5bf1e972cc5e/training/src/anemoi/training/diagnostics/benchmark_server.py#L167

training/tests/integration/test_accuracy.py
```diff
@@ -0,0 +1,99 @@
...
+            raise ValueError(msg)
+        # Align on step before comparing: np.allclose compares positionally and
+        # would silently broadcast-error (or mismatch) if the step sets differ.
+        aligned = df1.join(df2, how="inner", lsuffix="_1", rsuffix="_2")
+        if aligned.empty:
+            msg = f"Runs {run_id1} and {run_id2} share no common steps"
+            raise ValueError(msg)
+        return np.allclose(aligned.loc[:, "loss_1"], aligned.loc[:, "loss_2"])
+
+    assert is_similar(
```
cathalobrien: we could use torch.testing.close() instead. gives more information about how many elems didnt match and by how much
https://docs.pytorch.org/docs/main/testing.html#torch.testing.assert_close
```
Tensor-likes are not close!

Mismatched elements: 2 / 3 (66.7%)
Greatest absolute difference: 2.0 at index (1,) (up to 1e-05 allowed)
Greatest relative difference: 1.0 at index (1,) (up to 1.3e-06 allowed)
```

mcocdawc: The function looks nice but the problem is I want to print the run_id of the current run upon failure, to make it easier to change the reference.
Hence, I need to call assert with a custom message.

The torch function is `torch.test.assert_close` and calls assert itself. :-/

.github/workflows/accuracy-test-hpc.yml
mcocdawc: upon every push to a PR

.github/workflows/integration-tests-hpc.yml
mcocdawc: @cathalobrien @anaprietonem do you have toughts about it?

training/tests/integration/config/accuracy_testing/test_global.yaml
```diff
@@ -0,0 +1,59 @@
+defaults:
```
mcocdawc: could you @sahahner double check the input please?

Shall I remove more specifications from `training.optimization`?

training/tests/integration/config/atmo_integration_test.yaml
```diff
@@ -0,0 +1,59 @@
...
+    training: 400
+    validation: 100
+  training:
+    datasets:
+      data:
+        start: 2020
+training:
+  max_steps: 1000
+  optimization:
+    lr: 0.625e-4 #local_lr # LR for 8 GPUs and batchsize 2 # try 4 GPUS and batchsize 4
```
sahahner: This is the default, so we could remove it. 
I think we had in mind that we might need to adapt this, but if the default works well and leads to significant results, we should stick to it.

`lr: 0.625e-4`

training/tests/integration/config/atmo_integration_test.yaml
```diff
@@ -0,0 +1,59 @@
...
+    lr_scheduler:
+      warmup_t: 100 # number of optimizer steps after that LR reaches the maximum
+      t_initial: ${training.max_steps} # number of optimizer steps after that LR reaches the minimum
```
sahahner: This is also the default, so that line can be removed

`t_initial: ${training.max_steps}`

training/tests/integration/config/accuracy_testing/test_global.yaml
sahahner: Yes, that is a good point. Quite a few are the defaults.

Besides that, the config file of the deterministic training looks reasonable to me

training/tests/integration/config/atmo_integration_test.yaml
mcocdawc: roger wilco

training/tests/integration/config/atmo_integration_test.yaml
mcocdawc: yes, I removed it.

training/tests/integration/config/accuracy_testing/test_global.yaml
mcocdawc: thank you!

training/tests/integration/test_accuracy.py
cathalobrien: @mcocdawc 
the msg parameter could help here right? a callable which returns x+"your ruinid is ..."
>msg (Optional[Union[[str](https://docs.python.org/3/library/stdtypes.html#str), Callable[[[str](https://docs.python.org/3/library/stdtypes.html#str)], [str](https://docs.python.org/3/library/stdtypes.html#str)]]]) – Optional error message to use in case a failure occurs during the comparison. Can also passed as callable in which case it will be called with the generated message and should return the new message.

.github/workflows/integration-tests-hpc.yml
cathalobrien: why does it test twice?

.github/workflows/accuracy-test-hpc.yml
mc4117: Could I ask what the reasoning for this it? This seems a bit excessive to me especially if running on a gpu?

training/tests/integration/test_training_cycle.py
mcocdawc: I found another possibility, but thank you!

.github/workflows/accuracy-test-hpc.yml
mcocdawc: changed it to nightly runs.

training/tests/integration/test_accuracy.py
mcocdawc: replaced with torch's assert_close

.github/workflows/integration-tests-hpc.yml
```diff
@@ -107,22 +106,3 @@ jobs:
             **Time:** ${{ github.event.schedule }}
 
             cc: @ecmwf/anemoisecurity
-
-  notify-team-bm:
```
anaprietonem, line 111: why are we removing this?

.github/workflows/integration-tests-hpc.yml
```diff
@@ -23,16 +23,52 @@ jobs:
         if: ${{ github.event_name == 'schedule' }}
         run: test -z $(git rev-list  --after="24 hours"  ${{ github.sha }}) && echo "::set-output name=should_run::false"
```
anaprietonem, line 24: could you explain why are the changes below required for the accuracy testing? are those an improvement to the yaml definition?

training/pytest.ini
```diff
@@ -6,6 +6,8 @@ markers =
     auth: marks tests that require authentication (deselect with '-m "not auth"')
     gpu: marks tests that require a GPU (deselect with '-m "not gpu"')
     slow: mark test as slow (skipped unless --slow is used)
+    mlflow: marks tests requiring an MLflow test server (skipped unless --mlflow is used)
```
anaprietonem, line 9: thanks for catching this Oskar!

training/tests/integration/test_accuracy.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@pytest.mark.mlflow
+@pytest.mark.slow
+def test_config_build(tmp_path: Path, mlflow_server: str) -> None:
+
+    # Load without resolving: interpolations such as
+    # system.input.graph = ${system.output.root}/... depend on system.output.root,
+    # which is only set below. Resolving eagerly would collapse them to MISSING.
+    config = load_config(
```
anaprietonem: why can't we leverage the existing functionality for the integration tests and just maxing adjustments for the data part to load them from disk rather than download from s3 as we do there?

training/tests/integration/test_accuracy.py
```diff
@@ -0,0 +1,100 @@
...
+    trainer.train()
+
+    client = AnemoiMlflowClient(mlflow_server, authentication=True)
+
+    reference_id: Final = "684814cd86ed45f383bed3f0a87a782b"
+    metric: Final = "train_multi_dataset_loss_step"
+
+    # Printed so a failing run's ID can be promoted to the new reference_id.
+    LOGGER.info("Run ID from trainer: %s", trainer.run_id)
+    print(f"Run ID from trainer: {trainer.run_id}")  # noqa: T201
```
anaprietonem: better use the logger for this too.

training/tests/integration/config/accuracy_testing/test_global.yaml
```diff
@@ -0,0 +1,56 @@
...
+- data: zarr
```
anaprietonem: I'd suggest putting this file under config/accuracy_testing/test_global.yaml . To keep the naming consistent with the other set ups and since in the future we'd probably be extending this test suite with other models/uses cases

training/tests/integration/test_accuracy.py
anaprietonem: the other reason to motivate this question, is cause what we'd like in the near future as next steps is to have a fixture that then we parametrise (as we do for the Benchmarking and integration tests)

anaprietonem (commented):
Left a few comments to clarify! thanks a lot for looking into this @mcocdawc, great work!

.github/workflows/integration-tests-hpc.yml
mcocdawc: In the beginning I put the accuracy test into this file as well. 
This meant that there were three very similar tests which I refactored into a loop/matrix.

IMHO the loop structure is still a bit nicer than before, but I can also revert it.

.github/workflows/integration-tests-hpc.yml
mcocdawc: It's still there, just once instead of twice.

training/tests/integration/config/accuracy_testing/test_global.yaml
mcocdawc: done

training/tests/integration/test_accuracy.py
mcocdawc: done

training/pytest.ini
mcocdawc: :-)

training/tests/integration/test_accuracy.py
mcocdawc: Unfortunately I don't fully get the comment. Because I thought that I am **not** using the S3 storage in the test. Would be happy to have a chat.

mcocdawc (commented):
Adressed your comments @anaprietonem, probably it's best if we have a chat about the remaining two:
- the matrix strategy in the integration test
- the comment about the S3 storage data loading.

.github/workflows/integration-tests-hpc.yml
VeraChristina: If you want to remove the separate step `notify-team-bm`, can you adapt the logic in `notify-team` to also notify about failures in the benchmark tests (and not just the integration tests)? Nevermind, the names of the jobs have changed, my bad.

.github/workflows/integration-tests-hpc.yml
VeraChristina: On this note, why did you choose to put the convergence tests into a separate workflow? I think we might want them to have the same triggers as the integration and benchmark tests. (I.e., for now, they should also be triggered when the integration test workflow is dispatched manually on a branch. And in the future we might want to trigger them on (certain) PRs before merging. In any case following the same logic as the other long running tests.)

training/tests/integration/test_accuracy.py
```diff
@@ -0,0 +1,100 @@
...
+# Lightning's deterministic=True (set via config below) does not cover this.
+torch.backends.cudnn.benchmark = False
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+@pytest.mark.mlflow
+@pytest.mark.slow
+def test_config_build(tmp_path: Path, mlflow_server: str) -> None:
```
VeraChristina: Can we give this a different name, e.g. `test_loss_convergence_global_model` or something along those lines?

training/tests/integration/test_accuracy.py
VeraChristina: A variant of Ana's question: are we sure that `load_config` produces the same config as loading with hydra would? (Just to make sure we don't deviate too much from the code we run via the anemoi cli, i.e. we test the path we want to be testing.)
Or should we follow the config loading logic we use in the integration or benchmark tests?

(On re-using integration test functionality, I think it might be simpler to not reuse the integration tests fixtures directly, but perhaps worth re-using some of the logic there -- partly to make it easier for people to keep an overview of the different test set-ups.

On Ana's second point, we could extract the config build into a fixture. We can also do this in a future PR that adds more tests.)

VeraChristina (commented):
Thanks for adding this! It'll be a good additional layer to our testing. Just a few comments below.

.github/workflows/integration-tests-hpc.yml
mcocdawc: as discussed with @VeraChristina and @anaprietonem  in person, the accuracy test is now included into the integration test. The matrix is kept though.

training/tests/integration/test_accuracy.py
mcocdawc: as discussed with you in person, I reverted to using hydra and removed the load_config function.

training/tests/integration/test_accuracy.py
mcocdawc: already named it test_accuracy, is that fine?

training/tests/integration/test_accuracy.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.slow
+def test_accuracy(tmp_path: Path, mlflow_server: str) -> None:
+
+    # compose leaves interpolations such as
+    # system.input.graph = ${system.output.root}/... unresolved; they depend on
+    # system.output.root, which is only set below, and we resolve afterwards.
+    # config_path is relative to this file; the autouse set_working_directory fixture
+    # (integration/conftest.py) chdirs to the repo root so Path.cwd() resolves the yaml.
+    with initialize(version_base=None, config_path="../../src/anemoi/training/config", job_name="test_accuracy"):
+        template = compose(config_name="config", overrides=["model=graphtransformer"])
```
VeraChristina, line 50: This means we use the defaults from `anemoi/training/config/config.yaml`. I think that's a fine solution to make hydra work but it might be worth removing the defaults from the config file `test_global.yaml` you added for this test (just for clarity)

VeraChristina (dismissed):
Thanks for addressing the feedback! This looks good to me. Just a minor point below

training/tests/integration/test_accuracy.py
mcocdawc: > This means we use the defaults from anemoi/training/config/config.yaml

I think this is crucial for this test and we should really do this.

>  I think that's a fine solution to make hydra work but it might be worth removing the defaults from the config file test_global.yaml 

Honestly, here I just followed the existing tests and don't have any opinion on it. I am also very happy removing it. I understood you that this is mostly to switch off certain logging and callback options.

training/tests/integration/config/accuracy_testing/test_global.yaml
```diff
@@ -0,0 +1,56 @@
...
+- dataloader: native_grid
+- diagnostics: evaluation
+- system: example
+- graph: multi_scale
+- model: graphtransformer
+- task: forecaster
+- training: single
+- _self_
+
```
VeraChristina: I think my comment above wasn't very clear. I just meant removing this list of defaults since it is not used in the current set-up

mcocdawc: done

training/tests/integration/test_accuracy.py
mcocdawc: done

anaprietonem (approved):
Thanks @mcocdawc for the nice work! And to everyone for their feedback and time looking at the PR, LGTM to be merged!

PR #1202, opened 2026-06-18
fix(training): close checkpoint loading parity gaps (#495)

JesperDramsch:
## Summary

Closes a set of parity gaps in the new checkpoint **loading** layer so it matches the
legacy `transfer_learning_loading` / `on_load_checkpoint` behaviour. Part of the checkpoint
pipeline integration (#495); independent of the config-wiring PR (#1196) — this touches only
`checkpoint/loading/`.

## Behaviour changes (note for users)

- **Weights-only loading is now strict by default.** `WeightsOnlyLoader()` rejects a checkpoint
  with missing keys (raising `CheckpointLoadError`); pass `strict=False` for the previous
  fill-model leniency.
- **Pre-multi-dataset checkpoints are rejected.** A checkpoint whose
  `hyper_parameters.data_indices` is a single `IndexCollection` (from a pre-multi-dataset
  anemoi-core) now raises a `TypeError` pointing at the fix:
  `anemoi-models migration sync <checkpoint>` upgrades it to the multi-dataset format (or
  re-export with current anemoi-core).

## What else changed

- Fixed the `CheckpointLoadError` wrapper (`strategies.py`): it was called with one argument
  but the exception requires `(path, original_error)`, so a strict-load failure crashed with a
  `TypeError` instead of surfacing `CheckpointLoadError`.
- Warn when the checkpoint's stored model hparams diverge from the run config (the
  silent-divergence case fill-model loading would otherwise hide).
- Apply the runtime `trainable_edge_perm_fix` migration on load (model-dependent; idempotent).
- Re-inject `model_output_idx` buffers from the live model during the processor-stats refresh,
  matching the legacy refresh.
- Extract `_ckpt_variables_metadata` on transfer-learning load (feeds the unit-compatibility check).

## Tests

- New `test_parity_gap_closure.py` covers each behaviour.
- Existing loading tests that asserted the prior weights-only / metadata behaviour are updated
  to the new contract.
- `pytest training/tests/unit/checkpoint/loading/` → 54 passed; the rest of the checkpoint unit
  suite is green (the unrelated `sources/test_s3.py` collection error is a pre-existing
  `anemoi-utils` version mismatch in the local env, not this change).

### DIFF
--- a/training/src/anemoi/training/checkpoint/loading/base.py
+++ b/training/src/anemoi/training/checkpoint/loading/base.py
@@ -166,9 +166,12 @@ class LoadingStrategy(PipelineStage):
           attribute becomes ``{dataset_name: ic.name_to_index for ...}``,
           which matches what ``DiagnosticsSanityCallback`` indexes by
           dataset name.
-        - **Single-dataset** (legacy): ``hyper_parameters["data_indices"]``
-          is a single ``IndexCollection`` with a ``.name_to_index``
-          attribute. The flat mapping is assigned unchanged.
+        - **Single-dataset** (legacy, pre-multi-dataset):
+          ``hyper_parameters["data_indices"]`` is a single ``IndexCollection``.
+          This is rejected with a ``TypeError`` — the current loaders require
+          the dataset-keyed dict, matching legacy ``transfer_learning_loading``
+          (utils/checkpoint.py). The checkpoint must be upgraded first
+          (``anemoi-models migration sync``).
 
         If neither shape is recognised, the attribute is left alone and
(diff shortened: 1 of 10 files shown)


JesperDramsch:
Superseded by #1196 — folded in via merge `7dcfae999`; both commits (`6bdf79870`, `f34dc6276`) are ancestors of #1196's branch, and `LineageResolver` was refactored into `RunSource` (`sources/run.py`) per D18. Its work now ships through the single Phase 3 core PR, so closing here to avoid a duplicate review surface.

PR #1203, opened 2026-06-18
test(training): transfer learning checkpoint loading integration test

JesperDramsch:
## Description
Adds one integration test file covering the transfer learning checkpoint loading path end to end.

It drives a real `LocalSource -> TransferLearningLoader` pipeline over an on-disk checkpoint (`torch.save` round-trip, toy `nn.Module`s, no network/GPU).

`TransferLearningLoader` is on `main` via #998, but the existing unit tests in `test_transfer_learning.py` leave a gap: `test_transfer_learning_skips_mismatched_shapes` actually uses a key that is **absent from the target** (`old_head`), so it only exercises the "key not in target" branch of `filter_state_dict` - never a real same-key shape mismatch. As a result, neither the genuine shape-mismatch skip nor the strict-mode error path was covered, and nothing exercised the
source-to-loader pipeline composition.

This adds test coverage for:
- true shape mismatch (same key, different shape) skipped when `skip_mismatched=True`;
- `skip_mismatched=False` raising `CheckpointIncompatibleError`, with the offending key named;
- matched weights loaded verbatim while the mismatched target layer keeps its fresh init;
- source-only keys reported as "Key not in target";
- optimizer/scheduler discarded by the transfer-learning path.

#998 and the full merge of #1196 

##  Additional notes ##


anaprietonem:
@dietervdb-meteo @dnerini, Jesper has started to look at implementing a transfer learning integration test. Since we discussed this at Vaals I think it could be very useful to get your input here or what could be a good example for a realistic TL example!

mchantry:
@fprill maybe of interest for you to review?

PR #1208, opened 2026-06-24
fix(training/profilers): ensure dir on multi GPU is created

mcocdawc:
## Description

Fixes two bugs in the nightly benchmark tests.

One was about missing directories on rank-n nodes (n not root)

The other one was about failed logging to MLflow when `offline=True`.

Note that this is a regression in functionality for profiler offline logging. 

In `profilers.py:309` we have 
```python
    @rank_zero_only
    def create_output_path(self) -> None:
        self.dirpath = Path(self.config.system.output.profiler)
        self.dirpath.mkdir(parents=True, exist_ok=True)
```
note the `@rank_zero_only`! On a shared filesystem this is not a problem, but if rank-0 and rank-n (n != 0) write to different directories, it is non-existent for rank-n.

In `profiler.py:336` we see that the path is constructed as 
```
        if self.run_id:  # when using mlflow only rank0 will have a run_id except when resuming runs
            # Multi-gpu new runs or forked runs - only rank 0
            # Multi-gpu resumed runs - all ranks
            self.config.system.output.profiler = Path(self.config.system.output.profiler, self.run_id)
```
which means, that using mlflow leads to different profile output paths for the different ranks.

This left two situations in which the bug remained unnoticed:

- No mlflow / no run_id → the path stays at the static <base>/profiler without a run-specific subdirectory. That directory is created on all ranks as part of the normal output-path setup, so every rank can write its traces there without issue.
- mlflow enabled, but rank 0 and non-root ranks resolve different paths → rank 0 writes to <base>/profiler/<run_id>/, while the other ranks (whose run_id is empty) continue writing to <base>/profiler/, which already exists. Again, nothing fails. The only consequence is that the traces end up in different locations (and most people probably only inspect the rank-0 trace anyway.)
<- Might this be related to the issues you @anaprietonem saw in memory tracing?

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/train/profiler.py
+++ b/training/src/anemoi/training/train/profiler.py
@@ -167,7 +167,13 @@ class AnemoiProfiler(AnemoiTrainer):
             self.to_wandb()
 
         elif self.logger and self.logger.logger_name == "mlflow":
-            self.to_mlflow()
+            if self.config.diagnostics.log.mlflow.offline:
+                # Offline mlflow writes to a throwaway local file store with no server backing it;
+                # logging the profiler reports there has no value and the offline FileStore is only
+                # partially initialised (e.g. missing the .trash folder that log_table scans).
+                LOGGER.info("MLflow is offline; skipping profiler report logging to the local store.")
+            else:
+                self.to_mlflow()
 
     def report(self) -> str:
         """Print report to console."""
@@ -200,10 +206,6 @@ class AnemoiProfiler(AnemoiTrainer):
     def to_mlflow(self) -> None:
         """Log report into MLFlow."""
         LOGGER.info("logging to MLFlow Profiler report")
-        # Ensure the (offline) MLflow store exists and the run is materialised before logging
-        if self.mlflow_logger.save_dir is not None:
-            Path(self.mlflow_logger.save_dir).mkdir(parents=True, exist_ok=True)
-        _ = self.mlflow_logger.experiment  # force-create the run in the tracking store
         self.write_benchmark_profiler_report()
         # check this https://stackoverflow.com/questions/71151054/how-to-log- d da-table-of-metrics-into-mlflow
 
--- a/training/src/anemoi/training/train/profiler.py
+++ b/training/src/anemoi/training/train/profiler.py
@@ -200,6 +200,10 @@ class AnemoiProfiler(AnemoiTrainer):
     def to_mlflow(self) -> None:
         """Log report into MLFlow."""
         LOGGER.info("logging to MLFlow Profiler report")
+        # Ensure the (offline) MLflow store exists and the run is materialised before logging
+        if self.mlflow_logger.save_dir is not None:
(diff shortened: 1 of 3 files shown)


training/src/anemoi/training/train/profiler.py
```diff
@@ -200,6 +200,10 @@ def _log_reports_to_mlflow(self, run_id: str, data: pd.DataFrame, artifact_file:
     def to_mlflow(self) -> None:
         """Log report into MLFlow."""
         LOGGER.info("logging to MLFlow Profiler report")
+        # Ensure the (offline) MLflow store exists and the run is materialised before logging
+        if self.mlflow_logger.save_dir is not None:
+            Path(self.mlflow_logger.save_dir).mkdir(parents=True, exist_ok=True)
+        _ = self.mlflow_logger.experiment  # force-create the run in the tracking store
```
mcocdawc: does this make sense?

mcocdawc: no it doesn't :-D

training/src/anemoi/training/train/profiler.py
```diff
@@ -167,7 +167,13 @@ def export_to_logger(self) -> None:
             self.to_wandb()
 
         elif self.logger and self.logger.logger_name == "mlflow":
-            self.to_mlflow()
+            if self.config.diagnostics.log.mlflow.offline:
```
anaprietonem, line 170: I think this is not on the critical path, but this was here to allow to push the results of the profiler to mlflow, and this should be something we can do with both offline and online runs. @cathalobrien thoughts?

cathalobrien: I agree it should be possible, but this seems more like a problem with mlflow. maybe we should try pin an older version of mlflow and see if that fixes the partial initialisation error? could also avoid the annoying 'allow_file_store' error

mcocdawc: I am not sure how to proceed now. You know more mlflow infrastructure, to make this decision.

anaprietonem: @cathalobrien can you run the profiler from online mode here to confirm the logging to mlflow still works? If that's okey, then I would add clearly in the comment above and PR description that there is a regression functionality for profiler with mlflow offline logging, but online works fine, and I can approve the PR and we merge.

anaprietonem: @gmertes and I have started to discuss the issues with mlflow offline logging to give the sqlite backend a go. So we could then look at this regression and fix it

cathalobrien (approved):
LGTM great work

training/src/anemoi/training/train/profiler.py
anaprietonem: Okey Cathal did the check, so online still works happy to resolve my comment and merge

anaprietonem (approved):
Thanks a lot @mcocdawc! nice work! LGTM

PR #1209, opened 2026-06-24
fix(training): By default, explicitly run on gpu 

cathalobrien:
## Description
small PR to change the default accelerator from 'auto' to 'GPU'. This is to prevent users from accidentally running on CPU (the existing check if users are running on CPU does not fire if the 'auto' accelerator resolves to 'CPU').

Accidentally running on CPU can happen easily if torch does not support cuda. On Arm systems the default torch install does not support cuda, so this is becoming a more common issue. Users end up silently running on CPU with terrible throughput

Now, when a user starts training anemoi will default to running on GPU. If the the torch install does not support running on GPUs, a helpful error message is given telling them how to fix their install.

----
📚 Documentation preview 📚: https://anemoi-training--1209.org.readthedocs.build/en/1209/

----
📚 Documentation preview 📚: https://anemoi-graphs--1209.org.readthedocs.build/en/1209/

----
📚 Documentation preview 📚: https://anemoi-models--1209.org.readthedocs.build/en/1209/

### DIFF
--- a/training/docs/introduction/installing.rst
+++ b/training/docs/introduction/installing.rst
@@ -61,7 +61,7 @@ functionality:
    .. code:: bash
 
       pip install torch torchvision --index-url https://download.pytorch.org/whl/cu126
-   
+
    For full instructions on installing PyTorch with CUDA support, see the
    `PyTorch installation guide <https://pytorch.org/get-started/locally/>`__.
 
--- a/training/docs/introduction/installing.rst
+++ b/training/docs/introduction/installing.rst
@@ -45,6 +45,26 @@ functionality:
    python -m pip install "anemoi-training[profile]" # Install optional dependencies for profiling gpu usage
    python -m pip install "anemoi-training[docs]" # Install optional dependencies for generating docs
 
+.. warning::
+
+   After installation, verify that your PyTorch build has CUDA (or ROCm) support:
+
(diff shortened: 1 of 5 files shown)


mchantry:
@cathalobrien nice contribution. I think it would be good to put something in docs about this behaviour too. Please could you add it.

cathalobrien:
[integration tests](https://github.com/ecmwf/anemoi-core/actions/runs/28089463976/job/83163575658) pass

cathalobrien:
@mchantry good idea https://anemoi--1209.org.readthedocs.build/projects/training/en/1209/introduction/installing.html#instructions

training/docs/introduction/installing.rst
```diff
@@ -45,6 +45,26 @@ functionality:
    python -m pip install "anemoi-training[profile]" # Install optional dependencies for profiling gpu usage
    python -m pip install "anemoi-training[docs]" # Install optional dependencies for generating docs
 
+.. warning::
+
+   After installation, verify that your PyTorch build has CUDA (or ROCm) support:
```
mchantry:
```suggestion
   After installation, verify that your PyTorch build has CUDA or ROCm support:
```

HCookie (approved):

PR #1212, opened 2026-06-24
feat(training)!: speedup multiscale loss

cathalobrien:
## Description
This PR introduces a few changes to various losses with the aim of speeding up the multiscale loss in the temporal downscaler.

The main change is grouping the N sparse matmuls in sparse_projector.py to 1 large sparse matmul. the multi scale loss was having to load the projection matrix into L2 memory with each call to sparse matmul. 

Additionally the projection matrices are now converted to CSR format during multiscale loss init time. This further speeds up the loading of the projection matrices, as well as reducing peak memory usage.

After these changes, the individual CRPS loss computations begin to dominate. Compiling them gives another speedup (that's set in the config not in this PR)

some rough timings for the loss alone, using a standalone bm script. 

|  version | time (s)  | peak memory (GB)  | 
|---|---|---|
| Initial |  3.77 |  49 | 
| grouped sparse matmul  |  3.30 |  49 |  
| + CSR format  |  2.06 | 44  |  
| + compile CRPS |  1.08 | 45  |  

The results roughly match what I see in the traces from full training runs. The end-to-end throughput of a full training run has gone from 0.22 it/s to 0.62 it/s (0.55 without compiling)

Breaking change because I move training.recompile_limit to models.recompile_limit. since the components to be compiled live in the models config, it makes sense to keep the recompile limit there


japols:
Nice! Should we consider adding the CRPS loss as a default compile option in `config.model.compile`?

(probably need to revisit compile defaults when observations are merged, specifically dynamic=False)

ssmmnn11:
I had [this](https://github.com/ecmwf/anemoi-core/commit/d2cd23587770b0a9bba77bad170677e77a4f1501) in a branch to allow for directly creating the matrix in csr ; which is relevant when creating large matrices. Can you check how this would fit here as well?

mc4117:
just for context, we're still doing loss correctness tests on this branch and I will post the mlflows here once we have the comparisons

mc4117:
These are the mlflow runs pre and post pr (using the same seed) https://mlflow.ecmwf.int/#/compare-runs?runs=[%220bd1757800c34e00981f388f8afb1442%22,%221f0b2da799ff41b1b8f872239392d5ee%22]&experiments=[%22277%22]

cathalobrien:
I made some changes to make torch compile here more predictable and to get integration tests passing.

the integration tests pass locally. they are passing ⁠[now](https://github.com/ecmwf/anemoi-core/actions/runs/28574837831).

main thing change is changing from 'x_subset = x[subset_indices]' to index select to get it compiling in losses/scaler_tensor.py.
 and then once it was compiling I had to reduce the recompilations. this was done by selectively marking dimensions of individual tensors as dynamic
 ```
dynamic_indices = True  # TODO(cathal) set as true only for validation
        if dynamic_indices:
            torch._dynamo.mark_dynamic(pred_filtered, -1)
            torch._dynamo.mark_dynamic(target_filtered, -1)

            # prevent changing stride forcing specialisation
            pred_filtered = pred_filtered.contiguous()
            target_filtered = target_filtered.contiguous()
 ```

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -536,65 +550,59 @@ def _build_from_graph(
...
-        row_index = edge_index[0].long()
-        # edge_index[0] contains row indices (targets) for COO tensor format
-        norm = total.scatter_add_(0, row_index, weights)
-        norm = norm[row_index]
-        return weights / (norm + 1e-8)
+        matrix = matrix.tocsr(copy=True)
+        row_sums = np.asarray(matrix.sum(axis=1)).ravel()
+        inv_row_sums = np.zeros_like(row_sums, dtype=np.float32)
+        non_zero = row_sums != 0
+        inv_row_sums[non_zero] = 1.0 / row_sums[non_zero]
```
anaprietonem, line 610: why are just taking the non-zero entries here? is it because of the CRS format, if so maybe a line explaining this in the docstring could be good

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -536,65 +550,59 @@ def _build_from_graph(
...
-        src_size: int,
-        dst_size: int,
-        row_normalize: bool,
-    ) -> None:
-        """Create sparse projection matrix."""
-        row_index = edge_index[0].long()
-        edge_index = torch.stack([row_index, edge_index[1].long()])
+    def _create_matrix_from_scipy(self, matrix: spmatrix, row_normalize: bool) -> None:
+        """Create sparse projection matrix from a SciPy sparse matrix."""
+        matrix = matrix.astype(np.float32, copy=False).tocsr()
```
anaprietonem, line 575: can't we do this already when creating the matrix in line 553?

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -536,65 +550,59 @@ def _build_from_graph(
...
-            device=edge_index.device,
-        ).coalesce()
-
-        self._edge_dim = self.projection_matrix.shape[1]
+            matrix = self._row_normalize_matrix(matrix)
 
-        row_sums = torch.zeros(src_size, device=weights.device).scatter_add_(0, row_index, weights)
-        if not torch.allclose(row_sums, torch.ones_like(row_sums), atol=1e-5):
+        row_sums = np.asarray(matrix.sum(axis=1)).ravel()
+        if not np.allclose(row_sums, np.ones_like(row_sums), atol=1e-5):
```
anaprietonem, line 582: if row_normalisation is not True, this would fail right?

models/src/anemoi/models/layers/sparse_projector.py
```diff
@@ -27,34 +27,66 @@ def __init__(self, autocast: bool = False) -> None:
...
         """Apply sparse projection.
 
         Parameters
         ----------
         x : torch.Tensor
-            Input tensor
+            Input tensor with shape ``[..., input_nodes, channels]``.
         projection_matrix : torch.Tensor
             Sparse projection matrix (assumed to be on the correct device)
+        num_chunks : int
```
anaprietonem: is this something now configurable at the config level? or what's the guidance to choose the num_chunks?

models/pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
   "anemoi-utils>=0.1.9",
   "einops>=0.6.1",
   "hydra-core>=1.3",
+  "scipy>=1.6",
```
anaprietonem, line 42: since anemoi-models depends of anemoi-graphs, which already needs scipy, I don't think we need this one here

models/src/anemoi/models/schemas/models.py
```diff
@@ -295,6 +295,8 @@ class BaseModelSchema(PydanticBaseModel):
     "Residual connection schema."
     compile: Optional[list[dict[str, Any]]] = Field(None)
     "Modules to be compiled"
+    recompile_limit: PositiveInt = 32
```
anaprietonem: why removing those from the config and adding them here?

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -536,65 +550,59 @@ def _build_from_graph(
...
                 "Projection matrix rows do not sum to 1 (min=%.4f, max=%.4f, mean=%.4f). "
                 "This is unexpected; please check your matrix. "
-                "Consider using row_normalize=True or pre-normalized weights.",
+                "Consider using pre-normalized weights or row_normalize=True.",
                 row_sums.min().item(),
                 row_sums.max().item(),
                 row_sums.mean().item(),
             )
 
+        self.projection_matrix = torch.sparse_csr_tensor(
```
anaprietonem, line 592: before this was define as COO tensor and now that's a CSR (right?), does that have any implications on the downstream code accessing this projection_matrix?

training/src/anemoi/training/config/model/transformer_ens.yaml
```diff
@@ -156,6 +156,15 @@ compile:
   - module: anemoi.models.layers.normalization.ConditionalLayerNorm
     options:
       dynamic: false
+  - module: anemoi.training.losses.CRPS
```
anaprietonem, line 159: could we add this to the graphtransformer_ens_compile_config in test_compile?

training/tests/integration/conftest.py
```diff
@@ -31,6 +32,28 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def pytest_configure(config):
```
anaprietonem: fine to have it as a TODO, but found this in case it's helpful https://docs.pytorch.org/docs/main/generated/torch._logging.set_logs.html#torch._logging.set_logs https://github.com/pytorch/pytorch/issues/94788#issuecomment-1439471999

training/tests/integration/conftest.py
```diff
@@ -31,6 +32,28 @@
...
+    # suppress logging spam when using torch compile
+    # TODO(cathal) are there better ways to suppress then setting to warning?
+    logging.getLogger("torch.__trace").setLevel(logging.WARNING)
+    logging.getLogger("torch.__trace").propagate = False
+
+
+@pytest.fixture(autouse=True)
+def _reset_torch_compile(tmp_path, monkeypatch) -> None:
```
anaprietonem: I would add a comment about not deleting this fixture even if it's not called anywhere, cause the autouse True is handling that. Maybe it's obvious, but I am afraid someone might try to delete it

training/src/anemoi/training/train/train.py
```diff
@@ -665,10 +665,14 @@ def prepare_compilation(self) -> None:
 
         if hasattr(self.config.model, "compile"):
             self.model = mark_for_compilation(self.model, self.config.model.compile)
-        if hasattr(self.config.training, "recompile_limit"):
-            torch._dynamo.config.cache_size_limit = int(self.config.training.recompile_limit)
-            torch._dynamo.config.accumulated_cache_size_limit = max(8 * int(self.config.training.recompile_limit), 256)
-            LOGGER.info("Recompile limit set to %d", torch._dynamo.config.cache_size_limit)
+        if hasattr(self.config.model, "recompile_limit"):
```
anaprietonem: So we are moving the config entry from training to model? is that a config breaking change?

models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -536,65 +550,59 @@ def _build_from_graph(
...
-        # -> swap edge_index to [targets, sources] for COO tensor
-        edge_index_for_coo = torch.stack([sub_graph.edge_index[1], sub_graph.edge_index[0]])
-
-        self._create_matrix(
-            edge_index_for_coo,
-            weights,
-            graph[edges_name[2]].num_nodes,  # dst_size (targets) = rows
-            graph[edges_name[0]].num_nodes,  # src_size (sources) = cols
-            row_normalize,
+        matrix = coo_matrix(
```
anaprietonem, line 557: what's the difference now between this and the SparseProjector?

training/src/anemoi/training/losses/scaler_tensor.py
```diff
@@ -533,12 +533,41 @@ def resolve(self, ndim: int) -> Self:
...
+            # torch.compile/Dynamo does not support the < / >= operators on enums.
+            int_dims = [int(d) for d in dims]
+            if any(d < 0 for d in int_dims):
+                int_dims = [d if d >= 0 else ndim + d for d in int_dims]
+            resolved_scalers[name] = (int_dims, scaler)
 
         return ScaleTensor(**resolved_scalers)
 
+    @staticmethod
+    def _subset_tensor(
```
anaprietonem: it does seems like it could be simplied :D

anaprietonem (commented):
Left a few comments, nice work @cathalobrien !

training/src/anemoi/training/config/model/transformer_ens.yaml
cathalobrien: good catch! yes I will

training/tests/integration/conftest.py
cathalobrien: haha good catch, can't be too careful :p

models/src/anemoi/models/layers/graph_provider.py
cathalobrien: line 553 creates the matrix in COO format, and then this function converts it to a CSR format.
I will update the function name to ` _create_csr_matrix_from_scipy`

models/src/anemoi/models/layers/graph_provider.py
cathalobrien: Not if the weights are already normalised.

Currently we default to row_normalize=False, and assume that the weights are pre-normalised. The user can optionally enable row_normalize=True, but this is expensive

models/src/anemoi/models/layers/graph_provider.py
cathalobrien: there are some changes, but they don't seem to impact anemoi
```
# accessing indices has changed.
# old: coo_matrix.indices()
# new: csr_matrix.crow_indices()
# new: csr_matrix.col_indices()

# the matrix layout has changed
# old: coo_matrix.layout == torch.sparse_coo
# new: csr_matrix.layout == torch.sparse_csr
```
but I greped for  `indices()` and `layout` in models and training and they aren't used

models/src/anemoi/models/layers/graph_provider.py
cathalobrien: yeah exactly. I will add a line

models/src/anemoi/models/layers/sparse_projector.py
cathalobrien: yes, I suppose it could be.
less chunks will be faster but use more memory.
I would default to 1 and try to add a custom error message if you OOM in sparse_projector.py telling you to double the number of chunks

models/src/anemoi/models/schemas/models.py
cathalobrien: the list of which modules to compile are set in the models config. so makes sense to put the recompile limit in the same place

models/pyproject.toml
cathalobrien: maybe is it good to be explicit in this case? idk?

training/src/anemoi/training/train/train.py
cathalobrien: yeah, because the list of modules to compile is in the model config. so makes sense to keep recompile limit there too. my mistake putting it in training in the first place.
Yeah I suppose it is a config-breaking change (but I wonder how many people use this key). what is the procedure here?

training/tests/integration/conftest.py
cathalobrien: thanks for sharing, didn't know about that. I think I'll keep it as is for now, because torch._logging is an internal function and might break on old versions. useful to know for developing tho!

models/src/anemoi/models/layers/sparse_projector.py
cathalobrien: done

models/src/anemoi/models/layers/graph_provider.py
ssmmnn11: the current flow is:

projection graph provider -> provides matrix / edges -> sparse_projector applies

anaprietonem: thanks Simon, @cathalobrien could you add some docs about this? At least explaining the flow mentioned by Simon above could be useful

models/pyproject.toml
anaprietonem: Well the issue is a version mistmatch between scipy coming from graphs and models, but since we are directly importing scipy in models, I guess it's fine to list it here (since if in the future graphs stop using it we don't want to accidentally break models)

models/src/anemoi/models/layers/graph_provider.py
anaprietonem: thank you!

models/src/anemoi/models/layers/graph_provider.py
anaprietonem: right yes, renaming makes that cleared, thanks Cathal!

training/src/anemoi/training/train/train.py
anaprietonem: I agree probably not many and that models seems a more intuitive place. Still if this a breaking change we should mark that clearly in the PR title doing - feat(training)!: (so adding the ! symbol). Also for anyone using a config with the old `recompile_limit` in training it would be good if the code catches that and return a meaning error saying this is now deprecated and defined under models. 

also the num_chunks is now also a new config entry? We could do that one so that there is a default in the code so not breaking, but the user can still overwrite it at the config level. 
All these should be test with pydantic on and off

training/src/anemoi/training/losses/scaler_tensor.py
anaprietonem: also what's the difference between subset_index and subset_indices? could be good to add a docstring

anaprietonem: Also this changes and the casting below are related to the compiler right? Wouldn't it be cleaner to have a compiler object that handles these operations?

anaprietonem (commented):
For the changes in the scalers and variable mapper, I think it could be useful to extend and check the unit-tests to ensure the changes here (related to the use of the compiler) are okey

training/src/anemoi/training/losses/scaler_tensor.py
cathalobrien: > Wouldn't it be cleaner to have a compiler object that handles these operations?

I'm not sure about an object, but I like the idea of moving the subset function to models/utils/compile.py because you're right it is related to compilation. And then I can reuse it in other functions

cathalobrien: and can write tests for it too

cathalobrien: done 
```
training/utils/compile.py
```

models/src/anemoi/models/layers/graph_provider.py
cathalobrien: I added some comments in the docstrings, to make the flow more explicit

training/src/anemoi/training/train/train.py
cathalobrien: about the num_chunks for the SparseProjector, currently we instantiate SparseProjector in code not the config. so not sure where `num_chunks` would go in the config. I am tempted to leave it like this, but to give users a notification to change the code if they OOM in this part. what do you think?

.github/workflows/integration-tests-hpc.yml
```diff
@@ -34,14 +34,14 @@ jobs:
         include:
           - test_name: test_benchmark
             pip_extras: all,tests,profile
-            pytest_cmd: srun python -m pytest -v training/tests/integration/test_benchmark.py --slow --multigpu
+            pytest_cmd: srun --ntasks=2 --gpus-per-task=1 --gpu-bind=single:1 python -m pytest -v training/tests/integration/test_benchmark.py --slow --multigpu
```
anaprietonem: @cathalobrien  why this change?

anaprietonem (dismissed):
left a comment to clarify in the CI file otherwise LGTM!

.github/workflows/integration-tests-hpc.yml
cathalobrien: that was trying to fix the bm error before I realised it was a node issue. reverted now.

training/src/anemoi/training/train/train.py
cathalobrien: I have wired num_chunks up to the config

mc4117 (approved):
LGTM! Great work!

PR #1213, opened 2026-06-26
feat: reuse processing callbacks

anaprietonem:
## Description
This PR is a simplification of https://github.com/ecmwf/anemoi-core/issues/1135, where we noticed some issues with the memory management raised during the execution of the integration tests. Main changes:
Introduce a cross-callback process() cachet . `on_validation_batch_end` creates a fresh processed_cache={} per batch and passes it through to each _plot() call. BasePlotAdditionalMetrics.process() reads from and writes to this cache keyed on (dataset_name, members). Subsequent callbacks calling process() for the same dataset/members pair get the already-computed (data, output_tensor) tuple without re-running post-processors. The cache dict is not stored on self, it is local to each batch invocation, so there is no memory accumulation between batches and no shared mutable state between concurrent async invocations.

this PR is co-developed and based on the work of @VeraChristina 

Continuation of the callbacks refactor and improvement. This PR is focused on solving https://github.com/ecmwf/anemoi-core/issues/1135

https://github.com/ecmwf/anemoi-core/issues/1135

##  Additional notes ##

### DIFF
--- a/training/tests/integration/conftest.py
+++ b/training/tests/integration/conftest.py
@@ -8,11 +8,13 @@
 # nor does it submit to any jurisdiction.
 
 
+import gc
 import logging
 import os
 from pathlib import Path
 from typing import Union
 
+import psutil
 import pytest
 import torch
 from hydra import compose
@@ -27,6 +29,20 @@ from anemoi.utils.testing import GetTestData
 from anemoi.utils.testing import TemporaryDirectoryForTestData
 
 LOGGER = logging.getLogger(__name__)
+
+
+@pytest.fixture(autouse=True)
+def log_memory_usage(request: pytest.FixtureRequest) -> None:
+    """Log CPU RSS before and after each test to help debug memory leaks."""
+    process = psutil.Process()
+    rss_before = process.memory_info().rss / 1024**3
+    LOGGER.info("MEMORY [%s] before: %.2f GB RSS", request.node.name, rss_before)
+    yield
+    gc.collect()
+    rss_after = process.memory_info().rss / 1024**3
+    LOGGER.info(
+        "MEMORY [%s] after: %.2f GB RSS (delta: %+.2f GB)", request.node.name, rss_after, rss_after - rss_before,
+    )
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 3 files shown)


training/tests/integration/conftest.py
```diff
@@ -8,11 +8,13 @@
 # nor does it submit to any jurisdiction.
 
 
+import gc
 import logging
 import os
 from pathlib import Path
 from typing import Union
 
+import psutil
```
mc4117, line 17: just wondering, do these packages need to be added to the pyproject toml?

training/tests/unit/diagnostics/test_plotting_callbacks.py
```diff
@@ -405,6 +405,134 @@ def test_process_temporal_downscaler_multi_out_squeeze():
     assert output_tensor.shape == (pl_module.task.num_output_timesteps, 1, 1, nlatlon, nvar), output_tensor.shape
 
 
+# ---- process() cache ----
+
+
+def test_process_cache_shared_across_callbacks():
+    """A shared processed_cache avoids redundant post-processing across PlotSample, PlotSpectrum, PlotHistogram.
```
mc4117, line 412: PlotLoss also inherits from BasePerBatchPlotCallback and doesn't seem to have the process_cache in it, so I'm not sure if it is actually working? Would it be possible to add PlotLoss to this test and if it is not working to fix it?

training/tests/integration/conftest.py
anaprietonem: Fair question, but the reason is that psutil is already listed as dependency of a few packages we use in training, like [datasets](https://github.com/ecmwf/anemoi-datasets/blob/main/pyproject.toml) or also [torch-geometric ](https://github.com/pyg-team/pytorch_geometric/blob/master/pyproject.toml) 
Screenshot 2026-06-26 at 11 14 24

training/tests/unit/diagnostics/test_plotting_callbacks.py
anaprietonem: Yes this was indeed wrong, I have pushed a fix now and checked that the Integration tests are passing okey https://github.com/ecmwf/anemoi-core/actions/runs/28225387981

mc4117: Could you add it to the test?

anaprietonem: good idea, I have added it now!

mc4117 (approved):
Great work! Looks good to me!

PR #1216, opened 2026-06-26
feat(training): add --config-path to config validate and unify config composition

### LINKED ISSUE #1184: anemoi-training config validate has no --config-path, unlike config dump and train
`anemoi-training config validate` cannot be pointed at a config directory, unlike `config dump` and `train`.

Among the commands that consume an existing config, `validate` is the only one with no way to say where the config lives. `train` takes Hydra's `--config-path` (via `@hydra.main`), and `config dump` has its own `--config-path/-i`. But `config validate` exposes only `--config-name`, `--overwrite` and `--mask_env_vars`, and its implementation calls the Hydra Compose API with an empty primary path:

```python
def validate_config(self, config_name, mask_env_vars):
    with initialize(version_base=None, config_path=""):
        cfg = compose(config_name=config_name)
        ...
        BaseSchema(**cfg)
```

In practice the config is then only discoverable through the `AnemoiSearchPathPlugin` — i.e. the packaged defaults and the current working directory. To validate a config that lives anywhere else you have to `cd` into its directory first; there is no flag for it. The `config_path=""` argument also does no useful work (discovery is entirely the plugin's job) and is misleading to read.

This is worth fixing on top of #1181 (#1093). That change establishes `--config-path` as the canonical "point Hydra at my config directory" lever — `--config-path` > CWD > packaged defaults — for `train`, but `validate` has no `--config-path` to honour it, so it silently sits outside that model. Aligning `validate` is the natural next step, and doing it after #1181 avoids overlapping edits in `commands/config.py` and the search-path plugin.

Proposed direction:

- Add `--config-path/-i` to the `validate` subparser, mirroring `dump`.
- When it is given, compose from that directory (`initialize_config_dir`); when it is omitted, keep today's behaviour and let the search-path plugin supply CWD and the packaged defaults. This also lets the misleading `config_path=""` go away.

The packaged-preset and CWD cases are unchanged; this only adds the ability to validate a config in an arbitrary location without changing directory, and makes `validate` consistent with its sibling subcommands.

frazane:
## Description

Add `--config-path/-i` to `anemoi-training config validate` so a config living in any directory can be validated without first `cd`-ing into it, and route both `config validate` and `config dump` through a single Hydra-composition helper. This brings `validate` in line with `train` (which honours Hydra's `--config-path`) and `dump` (which already exposes `--config-path/-i`). Closes #1184.

A new, backward-compatible feature plus an internal cleanup.

`validate` was the only config-consuming subcommand with no way to say where the config lives: it exposed only `--config-name`, `--overwrite` and `--mask_env_vars`, and composed with an empty primary path (`initialize(version_base=None, config_path="")`), so a config was reachable only through `AnemoiSearchPathPlugin` — the current working directory and the packaged defaults. Validating a config anywhere else meant changing directory first.

When `--config-path` is given, the config is now composed from that directory via Hydra's `initialize_config_dir`, which installs it as the primary (`main`) search-path entry — the same precedence `@hydra.main` gives `train`'s `--config-path`: `--config-path` > current working directory > packaged defaults (the model #1181 established). When it is omitted, behaviour is unchanged, and the misleading `config_path=""` (which did no useful work) is dropped.

```bash
anemoi-training config validate --config-path /path/to/configs --config-name debug.yaml
```

Closes #1184. Follows #1181, which made the user-specified `--config-path` take precedence and simplified the search-path hierarchy for `train`.

##  Additional notes ##

`validate` and `dump` now share one composition helper — `initialize_config_dir(dir)` when a `--config-path` is given, plain `initialize()` otherwise. This replaces `dump`'s previous approach of copying the directory's files into a temporary directory and `chdir`-ing into it, a workaround that existed only because Hydra's `initialize` resolves `config_path` relative to the calling module and so cannot take an arbitrary user directory. `dump`'s observable behaviour is unchanged — it still composes the named config from the given directory, or from the working directory and packaged defaults when none is given — but it no longer copies files or mutates the process working directory.

Tests: `tests/unit/commands/test_config.py` covers `validate` composing a config discovered only via `--config-path`, and `dump` composing from a `--config-path` directory. Docs: a note and example were added to `docs/user-guide/configuring.rst`.

---

----
📚 Documentation preview 📚: https://anemoi-training--1216.org.readthedocs.build/en/1216/

----
📚 Documentation preview 📚: https://anemoi-graphs--1216.org.readthedocs.build/en/1216/

----
📚 Documentation preview 📚: https://anemoi-models--1216.org.readthedocs.build/en/1216/

### DIFF
--- a/training/docs/user-guide/configuring.rst
+++ b/training/docs/user-guide/configuring.rst
@@ -199,6 +199,15 @@ run using the following command:
 
    anemoi-training config validate --config-name debug.yaml
 
+By default the config is looked up on the search path described above (the
+current working directory and the packaged defaults). To validate a config
+that lives somewhere else without changing directory, point ``--config-path``
+at its directory, exactly as for ``anemoi-training train``:
+
+.. code:: bash
+
+   anemoi-training config validate --config-path /path/to/configs --config-name debug.yaml
+
 This will check that the configuration is valid and that all the
 required fields are present. If your config is correctly defined then
 the command will show an output similar to:
--- a/training/src/anemoi/training/commands/config.py
+++ b/training/src/anemoi/training/commands/config.py
@@ -15,13 +15,13 @@ import logging
 import os
 import re
 import shutil
-import tempfile
 from collections.abc import Generator
 from pathlib import Path
 from typing import Any
 
 from hydra import compose
 from hydra import initialize
+from hydra import initialize_config_dir
 from omegaconf import DictConfig
 from omegaconf import OmegaConf
 from pydantic import BaseModel
@@ -52,6 +52,7 @@ class ConfigGenerator(Command):
         validate = subparsers.add_parser("validate", help=help_msg, description=help_msg)
 
         validate.add_argument("--config-name", help="Name of the primary config file")
+        validate.add_argument("--config-path", "-i", type=Path, default=None, help="Configuration directory")
         validate.add_argument("--overwrite", "-f", action="store_true")
         validate.add_argument(
             "--mask_env_vars",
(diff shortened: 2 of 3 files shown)


anaprietonem (approved):
Had a look at the changes, and looks fine to me! thanks @frazane, good to be merged from my side

PR #1218, opened 2026-06-30
fix: config precedence

mc4117:
The documented Hydra config search-path hierarchy for anemoi-training is:

1. --config-path (highest priority)
2. Current working directory (CWD)
3. Packaged default configs (lowest priority / fallback)

In practice, this hierarchy was not respected. All three @hydra.main entry points (train.py, evaluate.py, profiler.py) used config_path="../config", which caused Hydra to register pkg://anemoi.training/config as the primary search path entry (position 2) — before AnemoiSearchPathPlugin ran. Since the plugin only appends CWD and the package fallback, the final order was:

## Description
Change config_path="../config" → config_path=None on all three @hydra.main decorators. With no primary path pre-populated, the plugin becomes the single authority on search path ordering, producing the correct layout:

The packaged config.yaml and all config group defaults are still found via the fallback at position 4.

Three new tests added to test_search_path_plugins.py to check this functionality does not break again

This meant user config overrides in CWD (e.g. diagnostics/log/mlflow.yaml) were silently ignored in favour of the packaged defaults, even though the documentation promised the opposite.

### DIFF
--- a/training/src/anemoi/training/train/evaluate.py
+++ b/training/src/anemoi/training/train/evaluate.py
@@ -98,7 +98,7 @@ class AnemoiEvaluator(AnemoiTrainer):
         LOGGER.debug("---- DONE. ----")
 
 
-@hydra.main(version_base=None, config_path="../config", config_name="config")
+@hydra.main(version_base=None, config_path=None, config_name="config")
 def evaluate(config: DictConfig) -> None:
     AnemoiEvaluator(config).evaluate()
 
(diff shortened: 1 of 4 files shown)


gmertes (dismissed):
@frazane ?

frazane (dismissed):
Tested and it works, thank you @mc4117!

mc4117:
integration tests pass https://github.com/ecmwf/anemoi-core/actions/runs/28443630612

gmertes (approved):

PR #1223, opened 2026-07-02
feat: Introduction of `anemoi-metadata`

### LINKED ISSUE #1207: Introduction of `anemoi-metadata?`
# Anemoi-Metadata
**Unification of metadata standards and interfaces to a lightweight self contained package**

## Problem statement

Various packages interface with checkpoints and the metadata contained within, yet there is no shared representation. This leads to upgrade issues, and provides friction in implementing features across the stack. A mitigation to this would be to share the code that interfaces with said metadata.

## Proposal

This issue seeks to propose and seek a discussion on the creation of a new package for anemoi, one to contain the metadata interfaces and standards. The first goal would be to encapsulate the expectation of "what is a checkpoint", providing the necessary interfaces and abstractions to allow `training` to write, and `inference` to read. A secondary goal would be to migrate the `Variable` class from `transforms` to make this package a more holisitic representation of those objects throughout `anemoi` which need a unified representation.

This package could live in core alongside the primary drivers of changes to the metadata to ease changes and reduce the need for synced PR's.

### Structure

A rough directory is shown below, showcasing how the raw schema will be hidden behind a top level interface with clear versioning.

```
src/anemoi/metadata/
├── __init__.py          # Public API
│
│  # Layer 1: Raw Schema
├── base.py              # BaseMetadata ABC
├── registry.py          # Version registry (semver)
├── versions/
│   └── v1.py            # V1 schema (from existing spec)
├── migrations/          # Migration functions
│   └── v1_to_v2.py      # (when v2 exists)
|
├── variable/          # Variable metadata (moved from anemoi-transforms)
├── dataset/          # Dataset metadata (moved from anemoi-datasets) TBD
│
│  # Layer 2: Abstraction
├── interface.py         # Metadata class (user API)
│
│  # Infrastructure
├── checkpoint.py        # I/O (moved from anemoi-utils)
├── migration.py         # MetadataMigrator (sequential chaining)
└── commands/
    └── inspect.py
```

## Reasoning

While the current interface in `inference` works, there is no shared contract between `training` and `inference`, which causes issues with major updates to the model, and associated configuration. Additionally, other tools looking to use / inspect a checkpoint rely upon `inference` which is not a light package and pulls in `torch` among many others. This causes a burden on what could be thin web interfaces or other tools, where `torch` is not needed.

## Steps

- [x] Determine if to include dataset metadata
		- Nope
- [ ] Create raw schema
- [ ] Implement migration patterns
- [ ] Abstract interface and checkpoint
- [ ] Adoption across anemoi

HCookie:
# Anemoi-Metadata

Unification of metadata standards and interfaces to a lightweight self contained package.

Migrates the interface in inference, and the utilities in utils, to a shared place to ease use of checkpoints and the metadata contained within.

```python
   from anemoi.metadata import Metadata

   # Load from checkpoint (handles legacy checkpoints without schema_version)
   metadata = Metadata.from_checkpoint("model.ckpt")

   # Typed inference properties
   print(metadata.dataset_names)     # ['data']
   print(metadata.timestep)          # '6h'
   print(metadata.multi_step_input)  # 3
   print(metadata.variables[:5])     # ['10u', '10v', '2d', '2t', ...]

   # Category-based variable selection
   prognostic = metadata.select_variables(include=["prognostic"])

   # Per-dataset access
   cfg = metadata.dataset_config("data")
   print(cfg.shapes.variables)     # 309
   print(cfg.timesteps.timestep)   # '6h'

   # Permissive section access
   lr = metadata.get("config", "training")

```

Various packages interface with checkpoints and the metadata contained within, yet there is no shared representation. This leads to upgrade issues, and provides friction in implementing features across the stack. A mitigation to this would be to share the code that interfaces with said metadata.

Closes #1207 

This PR was developed with agentic assistance. The tests so far have been entirely agentically written 

----
📚 Documentation preview 📚: https://anemoi-training--1223.org.readthedocs.build/en/1223/

----
📚 Documentation preview 📚: https://anemoi-graphs--1223.org.readthedocs.build/en/1223/

----
📚 Documentation preview 📚: https://anemoi-models--1223.org.readthedocs.build/en/1223/


metadata/src/anemoi/metadata/registry.py
```diff
@@ -0,0 +1,296 @@
...
+        if not re.match(r"^\d+\.\d+$", version):
+            raise ValueError(
+                f"Version '{version}' does not match the required format 'major.minor' "
+                f"(e.g. '1.1', '2.3'). Version must be two numeric components separated by a dot."
+            )
+
+        if version in cls._versions:
+            raise ValueError(f"Version {version} already registered")
+
+        base_version = ".".join(version.split(".")[:-1]) + ".0"
```
bdvllrs, line 135: If I do `@MetadataRegistry.register("1.0")` then  `@MetadataRegistry.register("1.1")` (minor but with schema change), then `MetadataRegistry.register_minor("1.2")` the metadata_cls will not be the one from 1.1 but from 1.0 which is not what is expected I think?

metadata/src/anemoi/metadata/migration.py
```diff
@@ -0,0 +1,376 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Sequential migration system for metadata versions.
+
+Migrations can be defined between any versions (direct multi-hop jumps
+or adjacent steps). Multi-version migrations are resolved by preferring
+the longest registered direct jump that doesn't overshoot, with automatic
+no-op version bumps for schema-sharing minors (registered via
+``MetadataRegistry.register_minor``).
```
bdvllrs, line 16: Why this choice to allow multiple migrations that span the same versions? I don't see use cases where chaining unit migration (one version to the next) would not be preferred.

Compared to the checkpoint migrator, this is the biggest change that could be hard to unify.

metadata/src/anemoi/metadata/migration.py
```diff
@@ -0,0 +1,376 @@
...
+        Parameters
+        ----------
+        metadata : MetadataContract
+            Source metadata instance.
+        target_version : str
+            Target version string.
+        allow_stop : bool, optional
+            If True, migrate as far as possible along the resolvable prefix of
+            the path and stop (with a ``UserWarning``) instead of raising when
+            no complete path to the target exists. Default is False.
```
bdvllrs, line 219: I don't find this description very clear. Especially the term "resolvable prefix of the path". Maybe just remove this part of the sentence as the rest is quite clear?

metadata/src/anemoi/metadata/base.py
```diff
@@ -0,0 +1,566 @@
...
+                    f"but this metadata is version {current}"
+                )
+            return func(self, *args, **kwargs)
+
+        return wrapper  # type: ignore[return-value]
+
+    return decorator
+
+
+class MetadataContract(BaseModel, ABC):
```
bdvllrs, line 84: I'm wondering whether the advantages of having versioned metadata schemas is worth the effort. I agree that the contributor experience would be better when writing migrations with the auto-completion, but at the same time for each metadata change you will need at the worth case:
- a new schema
- a new migration.

This feels redundant. Why not having only one schema that evolves over versions to the actual latest schema, and let the migrations handle upgrades from older to newer?

metadata/src/anemoi/metadata/registry.py
```diff
@@ -0,0 +1,296 @@
...
+        The base version is derived automatically by replacing the last version
+        component with "0" (e.g. "1.1" -> "1.0").
+        The schema class must use ``extra="allow"`` so that new optional fields
+        are accepted without breaking validation.
+
+        This is the recommended pattern for backwards-compatible additions that
+        don't warrant a new schema class.  Migration from the base version is
+        handled automatically by the migrator (a no-op version bump), so no
+        migration registration is needed for schema-sharing minors unless you
+        need to populate or transform fields.
```
bdvllrs, line 105: I don't see what the purpose of registering minor versions that don't come with a schema change and doesn't need a migration.

metadata/src/anemoi/metadata/registry.py
```diff
@@ -0,0 +1,296 @@
...
+
+    >>> MetadataRegistry.list_versions()
+    ['1.0']
+
+    >>> cls = MetadataRegistry.get_version("1.0")
+    >>> cls.__name__
+    'MetadataV1'
+    """
+
+    _versions: dict[str, type["MetadataContract"]] = {}
```
bdvllrs, line 47: Maybe a more general question about this "version explicit" system. Why you need them for the metadata schemas? As a user, you never give explicit version: you always migrate up to latest, or as close as possible.

metadata/src/anemoi/metadata/migration.py
```diff
@@ -0,0 +1,376 @@
...
+                current = nxt
+                continue
+
+            # Unresolvable from here: no migration and schemas differ.
+            return steps, (current, nxt)
+
+        return steps, None
+
+    @classmethod
+    def migrate(cls, metadata: "MetadataContract", target_version: str, allow_stop: bool = False) -> "MetadataContract":
```
bdvllrs, line 202: With checkpoint migrations, I have tried to keep a trace of what happened to the checkpoint (the list of migration that were executed). This might be useful to debug future bugs: is the metadata fresh or was it modified by a faulty migration? It could be interesting to track executed migrations as part of the metadata.

bdvllrs (changes requested):
Thanks for this! I'm reviewing this with the idea to potentially converge to a common way to handle migrations in anemoi and I'm sorry in advance if I am very critical about the implementation. I tried to tackle everything that I felt was significantly different than the checkpoint migrations!

As a remainder, for now there is the checkpoint migration tool, this one, and I'm currently working on config migration.

I've started extracting some common parts between the checkpoint migrator and the config migration that I'm building in ecmwf/anemoi-utils#339 with the actual implementation in #1350 for checkpoints and #1353 for config.

This implementation of migrations has some significant differences with the checkpoint one and it will not be trivial to combine. 

My hope is that we can converge to a better abstraction for all these use cases that would fit all our constraints!

PR #1228, opened 2026-07-03
fix(training): protect against seed overflow

### LINKED ISSUE #1217: Per-rank seed overflows uint32 at scale → init crash
### What happened?

`DDPEnsGroupStrategy.seed_rnd` does `initial_seed = base_seed * (model_comm_group_id + 1)` and passes it unbounded to `pl.seed_everything`. At scale it exceeds `2**32 - 1`; one rank raises `ValueError`, the rest hang on the next collective, NCCL watchdog kills the job. Hit easily by default since `get_base_seed()` falls back to the large `SLURM_JOB_ID` when `ANEMOI_BASE_SEED` is unset (overflow at only ~220 groups).

Any multi-node ensemble run with `ANEMOI_BASE_SEED` unset past ~220 model-comm groups crashes at init — visible symptom is an NCCL timeout hang, not the seed error.

**Error** (512 ranks, `num_gpus_per_model=2`, `base_seed = SLURM_JOB_ID = 19525198`):
```
strategy.py:62 in seed_rnd → pl.seed_everything(initial_seed)
ValueError: 4295543560 is not in bounds, numpy accepts from 0 to 4294967295
```
`4295543560 = 19525198 * 220` — first group id crossing `2**32`.

**Fix Used:**
```python
initial_seed = int(base_seed * (model_comm_group_id + 1)) % (2**32)
```

**Possible Workaround:** set `ANEMOI_BASE_SEED=<small int>` (`AIFS_BASE_SEED` is not read by ≥0.7.0).

### Version

**`anemoi-training` 0.7.0** (pytorch-lightning 2.6.1, numpy 2.2.6)

Platform: HPE Cray EX (Shasta) x86_64 | OS: SUSE Linux Enterprise Server 15 SP6 (Kernel 6.4.0-150600.23.73_15.0.14-cray_shasta_c)

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

ssmmnn11:
fixes #1217 

----
📚 Documentation preview 📚: https://anemoi-training--1228.org.readthedocs.build/en/1228/

----
📚 Documentation preview 📚: https://anemoi-graphs--1228.org.readthedocs.build/en/1228/

----
📚 Documentation preview 📚: https://anemoi-models--1228.org.readthedocs.build/en/1228/


training/src/anemoi/training/utils/seeding.py
```diff
@@ -10,6 +10,14 @@
 
 import os
 
+MAX_SEED = 2**32 - 1
+_SEED_MODULUS = MAX_SEED + 1
+
+
+def normalize_seed(seed: int) -> int:
+    """Bring a seed into the range accepted by Lightning, NumPy, and PyTorch."""
+    return int(seed) % _SEED_MODULUS
```
mcocdawc: why do you cast `int(seed)` if you only accept `int` seeds?

training/src/anemoi/training/utils/seeding.py
```diff
@@ -10,6 +10,14 @@
 
 import os
 
+MAX_SEED = 2**32 - 1
```
mcocdawc: might be worth to declare them as final.

mcocdawc (dismissed):
Just a few small comments, feel free to incorporate or ignore them, they are non-blocking.

Yesterday, the HPC was down, so I don't know if you could fully test this new code yet.
I think it would be worth doing so.

training/docs/user-guide/training.rst
```diff
@@ -105,10 +105,12 @@ seed can replay the same sequence of random choices after restart. This
 is usually not a concern when the seed comes from ``SLURM_JOB_ID``,
 because a new SLURM job normally gets a new job ID.
 
-Seeds below ``1000`` are multiplied by ``1000`` before use, so a
-fallback seed of ``42`` appears in logs as an effective seed of
-``42000``. This normalized base seed is logged during training and
-stored in checkpoint metadata.
+Anemoi Training derives 32-bit runtime seeds from the selected base seed
```
mcocdawc: I would add unsigned 32-bit to the docstring

If possible I would perhaps even return a `np.uint32`. At the moment it is a python `int` which is only by convention forced into the range [0, 2**32 - 1].

mcocdawc (dismissed):
I have a super nitpicky comment. If you want to proceed quickly, feel free to ignore it.

training/docs/user-guide/training.rst
ssmmnn11: Done, we keep int for comparability with python.random()

mcocdawc (approved):
cool cool :-)

Issue #1114, opened 2026-05-12
Encoder/decoder structure: multiple datasets per encoder/decoder


Sooph-knmi:
### Is your feature request related to a problem? Please describe.

Currently, anemoi-core accepts multiple input datasets but the amount of encoders/decoders is not configurable. By definition, each dataset gets a separate encoder/decoder. Allowing the same encoder for multiple datasets would enable multi-domain training functionality (see issue #919 and PR #1107). For some cases, the similarity between datasets would allow for having a joint encoder/decoder (such as when using multiple similar high resolution datasets). These datasets can be small and not overlapping in time, making the current set-up unsuitable. 

### Describe the solution you'd like

A config similar to the one below could provide a basis for allowing multiple datasets per encoder:
```
training:
  encoder_0:
    era5_0:
      dataset: ${system.input.dataset_a}
      frequency: ${data.frequency}
    meps:
      dataset: ${system.input.dataset_b}
      frequency: ${data.frequency}
   input_dim: 3
   edge_dim: 4
  encoder_1:
    era5_1:
      dataset: ${system.input.dataset_c}
      frequency: ${data.frequency}
    arome-arctic:
      dataset: ${system.input.dataset_d}
      frequency: ${data.frequency}
   input_dim: 3
   edge_dim: 4
```

During the initialization of the encoders and decoders, one could loop over the encoders/decoders and initialize the same encoder for multiple datasets, while keeping a unique graph_provider for each dataset. 

Note that for the above config to work, changes need to be made to the dataloader too, such that the batch could look like:
```
{encoder_0: {era5_0: dataset_a}, encoder_1: {arome_arctic: dataset_d}}
```

The datasets within each encoder would be shuffled, as proposed in issue #919 and PR #1107.

### Organisation

KNMI
MetNorway

JPXKQX:
Thanks all, some thoughts for context: 
- Per-dataset configurability is already missing from the main branch, and it is a valid use case (e.g. users may want to specify a larger or smaller decoder depending on the dataset)

Once we update the model configuration, it would be good to consider incoming functionality, such as multi-domain or observations
- We should no longer assume a 1-to-1 mapping between datasets and encoders/decoders. Both multi-domain and observations would benefit from sharing one encoder for multiple datasets.
- It may also be beneficial to include additional config fields, such as 'optional', to perform extra checks during the forward pass. 
[...]

PR #1092, opened 2026-05-05
feat: Add support for shared nested configs and optional enc/dec prefixes

HCookie:
## Summary

Extends the dictionary-style encoder/decoder configuration (from the parent branch) with optional/shared semantics and runtime validation.

### Optional prefix (`~`)

Mark a dataset's encoder/decoder as optional so it need not be present at forward time. Prefix the dataset key with `~` in the config:

```yaml
defaults:
  - encoder@encoder.data: gt16        # required
  - encoder@encoder.~other: gnn       # optional
```

This produces:

```yaml
encoder:
  data:
    _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
    ...
  ~other:
    _target_: anemoi.models.layers.mapper.GNNForwardMapper
    ...
```

At forward time, `_validate_required` raises if any non-optional dataset is missing from the input batch.

### Shared config (`,`)

Multiple datasets can share the same configuration (each getting an independently instantiated module) by joining their names with `,`:

```yaml
encoder:
  dataset1,dataset2:
    _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
    ...
```

This is equivalent to writing the same config under each dataset key individually. Each dataset receives a separate module instance built from the shared config.

### Shared instance (`&`) (not yet enabled)

Define a single encoder/decoder instance shared across multiple datasets:

```yaml
encoder:
  dataset1&dataset2:
    _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
    ...
```

Unlike `,`, this instantiates the encoder **once** and reuses the same module for both datasets. The build loop skips instantiation if the key already exists, and `_get_from_dict_with_and` resolves lookups at runtime. Currently gated behind `NotImplementedError` pending forward pass support.

### Runtime validation

Each model's `forward` method now calls `_validate_required` to check that all non-optional encoders and decoders have matching datasets in the input. Flags are stored in `encoder_flags` / `decoder_flags` dicts per dataset.

### Migration

`1777984565_add_enc_dec_flags.py` adds `encoder_flags` and `decoder_flags` to old checkpoints by inferring dataset names from encoder state dict keys, defaulting all to `{"optional": False}`.

### Files changed

All changes are in `models/src/anemoi/models/`:
- `models/base.py` -- `PREFIX_TO_MEANING`, updated `_get_nested_configuration` (returns 3-tuple with key + flags), `_validate_required`, `_get_from_dict_with_and`, `,` separator for shared configs
[...]

PR #1229, opened 2026-07-03
feat(configs)!: expand configurability for models with multiple encoders/decoders

JPXKQX:
## Description

The purpose of this PR is to introduce a new configuration for multiple encoders/decoders, providing greater flexibility and paving the way for new use cases in the coming weeks/months (downsampling, multi-domain and dynamic obs).
The proposal agreed within the WG discussions was: 
```
encoder:
  0: # user-defined key. It will appear in the state-dict
    dataset: [ "data1", "data2" ]
    dataset_fusing_strategy: "not_supported"
    mapper:
        ....

decoder:
  0: # user-defined key. It will appear in the state-dict
    dataset: [ "data1" ]
    input_target_nodes: [ ??? ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs 
    mapper:
       ...
```

To avoid new breaking changes on the config side for the coming PRs we have decided to add some keys with partial functionality (raising meaningful errors).

### Latent Aggregator
We have created a new interface, `BaseLatentAggregator`, that acts as a bridge between multiple encoded latents and the input to the processor. This provides 3 first three classes: `SumAggregator`, `MeanAggregator` and `ConcatAggregator`, but can be extended to more sophisticated blending methods. This is now specified in the configs under `config.model.latent_aggregator`. One consequence of these changes is that we can now use different `num_channels` for each encoder. The `config.model.num_channels` is no longer used by the code. It is still in the configs, but used as a placeholder to broadcast the values across all the encoders/decoders.

### Input Target features (Decoder)

Currently, the input target features for the decoder are defined by the encoder. New use cases like having a decoder but not an encoder for that same dataset require more flexibility with this conditioning. Also, this was the only difference between the `AnemoiModelEncProcDec` and `AnemoiModelAutoEncoder`. In this PR, we propose a new key `input_target_features` that defines these input features. This will be a list of strings where each string represents a different set of variables.
- `encoded_data` (default). This includes the output of the encoder for the same dataset. If the dataset is not encoded, an error will be raised.
- `coordinates`. This includes the coordinates of the nodes (specifically their sine & cosine)
- `trainable_parameteres`. This includes the node-trainable parameters for that dataset.
- `forcings`. This would include the forcing variable. These could be input/target forcings (?)
- `prognostics`.  This would include the prognostic variables.

As a consequence of this feature, we can remove the autoencoder.py and hierarchical_autoencoder.py. This also means that the `AnemoiEnsModelEncProcDec` could be used as an autoencoder without requiring an extra class.

### Handling multiple input datasets for the same encoder

Added a new key `dataset_fusing_strategy` ONLY for encoders to specify how we want to proceed when 2 datasets are passed during the forward pass. For now, it will only support "not_supported", this will raise an error when 2 datasets are passed to the same encoder in the forward pass. In the future, we can add more options like:
- `concat_channels`: to concatenate the channels of both decoders. (i.e. downscaling where you want to encode an upsampled low-res field and a forcing high-res field)
- `concat_grids`: similar to GraphDOP
- `sequential`: use the same encoder for both datasets and aggregate the encoded latents.

### Trainable Parameters

Now, the edge trainable parameters must be specified under each `encoder/decoder.mapper` configuration. The old `trainable_parameters` field is renamed to `node_trainable_parameters` and contains only the node trainable parameters (both from data and hidden nodes). The `edge_trainable_parameters` is kept as a placeholder for these options to be broadcasted easily across multiple mappers. Under `node_trainable_parameters`, we need to specify the nodes name, `data` will no longer be broadcasted to all data nodes (e.g. "era" and "cerra" in a multiple-datasets approach). We should specify the following
```
node_trainable_parameters:
    era: 8
    cerra: 8
    hidden: 8 # or hidden_1 and hidden_2 if we use a hierarchical graph.
```

The `node_trainable_parameters` will be set to 0 by default if not present in the configuration file. The code will not fail. This will be logged in the console.

### Other components

There are 3 other components that have been updated to support this per-dataset configurability: `residual`, `bounding` and `output_mask`. Now, all 3 require definition per-dataset, as their configuration is no longer broadcast along all dataset names.

The bounding will fail if a variable is specified and it's not in the expected prediction.

Any feedback on the new names proposed is welcome.

### Next steps

We still identify 3 points to investigate further:
- Investigating options for using defaults & per-dataset overwrites (yaml references?, hydra subdirectories?, new default key? ...)
- Simplify instantiation of the same mapper config for multiple datasets (something like "," proposed in #1092) 

- Flexibility to define an arbitrary number of encoders and decoders.
- Flexibility to define encoders/decoder with different hyperparameters.
- Flexibility to merge encoded latents.

#1114
#1092

Thanks to all involved! @havardhhaugen @einrone @Sooph-knmi @sahahner @OpheliaMiralles @icedoom888 @HCookie 

----
📚 Documentation preview 📚: https://anemoi-training--1229.org.readthedocs.build/en/1229/

----
📚 Documentation preview 📚: https://anemoi-graphs--1229.org.readthedocs.build/en/1229/

----
📚 Documentation preview 📚: https://anemoi-models--1229.org.readthedocs.build/en/1229/


fprill:
Hello @JPXKQX, many thanks for this implementation! In my opinion, the architecture is quite flexible - almost flexible enough for a temporal processor to advance time purely within that latent space (this is something we are working on experimentally). For this application, I would like to use the `BaseLatentAggregator` class and its descendants to merge the processor output with other embeddings.

To anticipate this, wouldn't it be reasonable to formulate the currently "datasets"-oriented comments in `models/src/anemoi/models/layers/aggregator.py` in a more general way, in the sense that this just a aggregation of latent space vectors


I also have a few minor things:

- I see the benefit of keeping `config.model.num_channels` in the configs as a placeholder, although it is no longer used by the code. However, such overlapping settings can also be confusing, and a slightly more verbose script could help (me) facilitate understanding here.
- I understand that the `ConcatAggregator` is order‑dependent, reflecting the insertion order of the definition in the config (guaranteed by Python 3.7+, I think). Alternatively, the `ConcatAggregator` could explicitly specify the order via a special key. What do you think?
- I did not quite understand your point about the "same mapper config for multiple datasets": Wouldn't it be sufficient to define a YAML anchor with `&`, and define an alias which can be reused later with `*` syntax? 
- I'd prefer that the definition of a non‑trivial `emb_nodes_src` be made known by a log message (could be at debug level).

... finally: The title of this PR could be phrased a bit more accurately, I think.

anaprietonem (commented):
@JPXKQX, very nice work! I was wondering if it could be worth separating this into two PRs. So in the first part we focus on expanding the configurability for the model components, and then we look at a second one where we introduce the latent_aggregator and expands the functionality there?

JPXKQX:
Hi Florian, thanks for taking the time to look into the PR! 
> Hello @JPXKQX, many thanks for this implementation! In my opinion, the architecture is quite flexible - almost flexible enough for a temporal processor to advance time purely within that latent space (this is something we are working on experimentally). For this application, I would like to use the `BaseLatentAggregator` class and its descendants to merge the processor output with other embeddings.
> 
> To anticipate this, wouldn't it be reasonable to formulate the currently "datasets"-oriented comments in `models/src/anemoi/models/layers/aggregator.py` in a more general way, in the sense that this just a aggregation of latent space vectors

The main motivation for adding the interface for the aggregator in this PR is to support different `num_channels` for each encoder/decoder and open the space for more flexibility here. If you have a clear use case you are interested in, it would be nice to have an issue with this.  

> I also have a few minor things:
> 
> * I see the benefit of keeping `config.model.num_channels` in the configs as a placeholder, although it is no longer used by the code. However, such overlapping settings can also be confusing, and a slightly more verbose script could help (me) facilitate understanding here.

I understand your point. Our motivation here was to keep something high-level, so current users working on main don't suffer much from the change of options, but still have space for more advanced users to specify this configuration directly without referencing to `model.num_channels`. A comment may be helpful in this line of the config. Anyway, we are happy to remove this if people think it is clearer
> * I understand that the `ConcatAggregator` is order‑dependent, reflecting the insertion order of the definition in the config (guaranteed by Python 3.7+, I think). Alternatively, the `ConcatAggregator` could explicitly specify the order via a special key. What do you think?

The order is important, yes. I don't think we need to support configuring the order, but we should ensure the order is consistent across all model passes.
> * I did not quite understand your point about the "same mapper config for multiple datasets": Wouldn't it be sufficient to define a YAML anchor with `&`, and define an alias which can be reused later with `*` syntax?

yes, that's one option. Another option would be to use another placeholder field for the mapper. This point was about exploring all these options and their pros & cons.
> * I'd prefer that the definition of a non‑trivial `emb_nodes_src` be made known by a log message (could be at debug level).

That makes sense!
> ... finally: The title of this PR could be phrased a bit more accurately, I think.

Open to any suggestions

JPXKQX:
> @JPXKQX, very nice work! I was wondering if it could be worth separating this into two PRs. So in the first part we focus on expanding the configurability for the model components, and then we look at a second one where we introduce the latent_aggregator and expands the functionality there?

yes, that's up to us. The reason to put this together was to avoid multiple breaking changes in configs. There are 2 changes (related to the code) that could be moved to follow-up PRs:
- Aggregator: this is useful for adding support for different encoder/decoder thicknesses (i.e. `num_channels`). We have added a simple interface to handle the input latents as we are doing now. This interface will also check the number of channels of these latents;
- Define which information is used as the target nodes for the decoder. This is missing, but we have added a proposal for the config now in the PR description

models/src/anemoi/models/layers/mapper.py
```diff
@@ -992,8 +1001,8 @@ def __init__(
             Input channels of the source node
         in_channels_dst : int
             Input channels of the destination node
-        hidden_dim : int
-            Hidden dimension
+        num_channels : int
```
sahahner, line 1004: How do the `num_channels` relate to the `self.hidden_dim` from the aggregator of the latents?

JPXKQX: the motivation here was to standardise the naming of all components (encoders & processor) and the current `model.num_channels`, but they are the same.

models/src/anemoi/models/layers/aggregator.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def forward(self, latents: dict[str, Tensor]) -> Tensor:
+        values = list(latents.values())
+        if len(values) == 1:
+            return values[0]
+        return torch.stack(values).mean(dim=0)
+
+
+class ConcatAggregator(BaseLatentAggregator):
+    """Concatenate latent representations and project back to the original channel dimension."""
```
sahahner: I think "project back" can be removed from the docstring?

I propose
`Concatenate latent representations along the channel dimension.`

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
```diff
@@ -0,0 +1,213 @@
...
+      graph_attention_backend: "triton"  # Options: "triton", "pyg"
+      edge_pre_mlp: False
+
+latent_aggregator:
+  _target_: anemoi.models.layers.aggregator.SumAggregator
+
+decoders:
+  global: # user-defined key. It will appear in the state-dict
+    datasets: [ "era5" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner: I think there is a small error in the options in almost all config files:

```suggestion
    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoded_data
```

`encoder_outputs` is not an option, see `VALID_TARGET_FEATURES = ["coordinates", "forcings", "prognostics", "trainable_parameters", "encoded_data"]`

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
```diff
@@ -0,0 +1,213 @@
...
+      qk_norm: False
+      cpu_offload: ${model.cpu_offload}
+      gradient_checkpointing: True
+      layer_kernels: ${model.layer_kernels}
+      shard_strategy: "edges"
+      graph_attention_backend: "triton"  # Options: "triton", "pyg"
+      edge_pre_mlp: False
+  regional: # user-defined key. It will appear in the state-dict
+    datasets: [ "cerra" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner:
```suggestion
    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoded_data
```

training/src/anemoi/training/config/model/graphtransformer_transport.yaml
```diff
@@ -55,44 +56,57 @@ processor:
...
-  qk_norm: True
-  cpu_offload: ${model.cpu_offload}
-  gradient_checkpointing: True
-  layer_kernels: ${model.layer_kernels}
-  shard_strategy: "edges"
-  graph_attention_backend: "triton"  # Options: "triton", "pyg"
+decoders:
+  0:
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner:
```suggestion
    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoded_data
```

training/src/anemoi/training/config/model/graphtransformer_transport_tendency.yaml
```diff
@@ -56,44 +57,57 @@ processor:
...
-  qk_norm: True
-  cpu_offload: ${model.cpu_offload}
-  gradient_checkpointing: True
-  layer_kernels: ${model.layer_kernels}
-  shard_strategy: "edges"
-  graph_attention_backend: "triton"  # Options: "triton", "pyg"
+decoders:
+  0:
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner:
```suggestion
    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoded_data
```

training/src/anemoi/training/config/model/transformer_ens.yaml
```diff
@@ -66,40 +67,51 @@ processor:
...
+      cpu_offload: ${model.cpu_offload}
+      gradient_checkpointing: True
+      layer_kernels: ${model.layer_kernels}
+      shard_strategy: "edges"
+      graph_attention_backend: "triton"  # Options: "triton", "pyg"
+
+decoders:
+  0:
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner:
```suggestion
    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoded_data
```

training/src/anemoi/training/config/model/transformer_transport.yaml
```diff
@@ -57,44 +58,56 @@ processor:
...
-  qk_norm: True
-  cpu_offload: ${model.cpu_offload}
-  gradient_checkpointing: True
-  layer_kernels: ${model.layer_kernels}
-  shard_strategy: "edges"
-  graph_attention_backend: "triton"  # Options: "triton", "pyg"
+decoders:
+  0:
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner:
```suggestion
    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoded_data
```

training/src/anemoi/training/config/model/transformer_transport_tendency.yaml
```diff
@@ -58,43 +59,54 @@ processor:
...
-  qk_norm: True
-  cpu_offload: ${model.cpu_offload}
-  gradient_checkpointing: True
-  layer_kernels: ${model.layer_kernels}
-  shard_strategy: "edges"
-  graph_attention_backend: "triton"  # Options: "triton", "pyg"
+decoders:
+  0:
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner:
```suggestion
    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoded_data
```

training/src/anemoi/training/config/model/graphtransformer_ens.yaml
```diff
@@ -74,44 +73,63 @@ processor:
...
+      graph_attention_backend: "triton"  # Options: "triton", "pyg"
+      edge_pre_mlp: False
+
+latent_aggregator:
+  _target_: anemoi.models.layers.aggregator.SumAggregator
+
+decoders:
+  0:
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner: encoder_outputs not an option

training/src/anemoi/training/config/model/graphtransformer.yaml
```diff
@@ -40,46 +41,63 @@ processor:
...
+      edge_pre_mlp: False
+
+
+latent_aggregator:
+  _target_: anemoi.models.layers.aggregator.SumAggregator
+
+decoders:
+  0: # user-defined key. It will appear in the state-dict
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner: encoder_outputs not an option

training/src/anemoi/training/config/model/gnn.yaml
```diff
@@ -31,36 +31,52 @@ processor:
...
-  cpu_offload: ${model.cpu_offload}
-  gradient_checkpointing: True
-  layer_kernels: ${model.layer_kernels}
+latent_aggregator:
+  _target_: anemoi.models.layers.aggregator.SumAggregator
+
+decoders:
+  0:
+    datasets: [ "data" ]
+    input_target_features: [ "encoded_data" ] # options: coordinates, forcings, prognostics, trainable_parameters, encoder_outputs
```
sahahner: encoder_outputs not an option

models/src/anemoi/models/models/base.py
```diff
@@ -89,18 +92,57 @@ def __init__(
...
         # Instantiation of model output bounding functions (e.g., to ensure outputs like TP are positive definite)
         # Multi-dataset: create ModuleDict with ModuleList per dataset
-        self.boundings = build_boundings(model_config, self.data_indices, self.statistics)
+        self.boundings = build_boundings(
+            get_multiple_datasets_config(model_config.model.get("bounding", [])),
+            data_indices=self.data_indices,
+            statistics=self.statistics,
+        )
+
+    def _build_dataset_routing(self, encoders_config: DotDict, decoders_config: DotDict) -> None:
```
sahahner: What about splitting this into two routines, one for the encoder routing and one for the decoder routing? A clearer separation inside the routine might already help to increase readibility.

models/src/anemoi/models/layers/mapper.py
sahahner: Okay. I think we need to make sure, for example, with a test, that the logic of the concatenation of encoder latents and the asserts works well, as this changes the hidden_dim/num_channels between encoders and the latent representation

models/src/anemoi/models/models/base.py
sahahner: Maybe we could even factor out the validation into another separate function, and also add an assert for the encoder routing. Is it possible to check whether datasets_to_encode match/line up with self.dataset_names that are extracted from the data_indices?

models/src/anemoi/models/layers/mapper.py
sahahner: I think it makes sense to be clear about terminology here and unify where possible. I think there is still a bit of a back and forth between num_channels and hidden_dim. This can of course be justified, but we should just clarify what is what

models/src/anemoi/models/layers/aggregator.py
JPXKQX: Yes, totally! Good point. Initially, there was a MLP there. Thanks!

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
JPXKQX: True, thanks!

training/src/anemoi/training/config/model/point_wise.yaml
```diff
@@ -68,34 +81,36 @@ attributes:
...
-  # is not 0. For example, in case of the sea surface temperature we don't use
-  # [0, infinity), buth rather [-2C, infinity). We do not want the water
-  # temperature to be below the freezing temperature.
-
-  # - _target_: anemoi.models.layers.bounding.NormalizedReluBounding
-  #   variables: [sst]
-  #   min_val: [-2]
-  #   normalizer: ['mean-std']
+  datasets:
+    data:
```
sahahner, line 89: It is nice, that the boundings are now applied per dataset. The bounding configuration still allows listing variables in the config file, that are not actually present in the output. Do we want to disable this feature now, as it might lead to configuration errors because of typos, for example?

sahahner:
Thank you @JPXKQX for enabling the configuration of multiple encoders and decoders. This will be very useful!

I made a few comments. On a more general note, I would recommend including a few docstrings in the model files `base.py` and `encoder_processor_decoder.py`. Some of the configuration functions carry more complexity now.

Have you considered moving the asserts into a `_validate_model_configuration` function?

training/src/anemoi/training/config/model/point_wise.yaml
JPXKQX: I see that is expected in the code, we are testing that the missing variables are skipped from bounding. Or do you mean this was introduced because of the per-dataset configurability of the bounding, so we should remove it now?

JPXKQX: Updated in [this commit](https://github.com/ecmwf/anemoi-core/pull/1229/changes/11452f782c1607046c53dc3a817528865f8b6e09). Added to the PR description.

models/src/anemoi/models/layers/mapper.py
JPXKQX: The motivation here was to keep `hidden_dim` as the object property for the mappers to avoid issues with migration (not sure if renaming would cause problems). On the other hand, we tried to unify all exposed parameters to `num_channels`, which is more consistent with in_channels_src, in_channels_dst and out_channels_dst. `hidden_channels` could be another option for this argument.

models/src/anemoi/models/models/base.py
JPXKQX: I have split the routines and added a check to ensure the datasets in the encoders/decoders are in `data_indices`.

training/src/anemoi/training/config/model/graphtransformer.yaml
```diff
@@ -40,46 +41,64 @@ processor:
...
+      shard_strategy: "edges"
+      graph_attention_backend: "triton"  # Options: "triton", "pyg"
+      edge_pre_mlp: False
+
+
+latent_aggregator:
+  _target_: anemoi.models.layers.aggregator.SumAggregator
+
+decoders:
+  0: # user-defined key. It will appear in the state-dict
```
MeraX: Isn't this key violating `DatasetDict: TypeAlias = dict[Literal["datasets"], dict[str, T]] | T`?

Is this key referenced elsewhere in the config?

models/src/anemoi/models/models/target_features.py
```diff
@@ -0,0 +1,328 @@
...
+    """Register a :class:`DecodingTargetFeature` subclass under ``name``."""
+
+    def decorator(cls: type[DecodingTargetFeature]) -> type[DecodingTargetFeature]:
+        if name in TARGET_FEATURE_REGISTRY:
+            raise ValueError(f"Target feature '{name}' is already registered.")
+        TARGET_FEATURE_REGISTRY[name] = cls
+        cls.name = name
+        return cls
+
+    return decorator
```
MeraX, line 45: This is the first time, I come across registered target options in the core. Is there a reason not to use the full class name instead?

training/src/anemoi/training/config/model/graphtransformer.yaml
```diff
@@ -40,46 +41,64 @@ processor:
...
+  datasets:
+    data:
+      _target_: anemoi.models.layers.residual.SkipConnection # options: SkipConnection, TruncatedConnection
+      step: -1 # use the latest step as skip connection
 
 output_mask:
-  _target_: anemoi.training.utils.masks.NoOutputMask
+  datasets:
+    data:
+      _target_: anemoi.training.utils.masks.NoOutputMask
```
MeraX, line 101: By explicitly configuring each dataset, could one deprecate https://github.com/ecmwf/anemoi-core/blob/d7f623353cd901298d6727d5756c02f07c9f69ef/models/src/anemoi/models/utils/config.py#L19 ?

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
```diff
@@ -0,0 +1,215 @@
...
+  qk_norm: False
+  cpu_offload: ${model.cpu_offload}
+  gradient_checkpointing: True
+  layer_kernels: ${model.layer_kernels}
+  graph_attention_backend: "triton"  # Options: "triton", "pyg"
+  edge_pre_mlp: False
+
+encoders:
+  global: # user-defined key. It will appear in the state-dict
+    datasets: [ "data1", "data2", "data3" ]
```
MeraX: where are these datasets defined?

models/docs/usage/create_model.rst
```diff
@@ -26,38 +26,59 @@ First, let's take the model configuration ``transformer.yaml``:
...
-   decoder:
-     _target_: anemoi.models.layers.mapper.GraphTransformerBackwardMapper
-     trainable_size: 8
-     sub_graph_edge_attributes: ${model.attributes.edges}
-     num_chunks: 1
-     mlp_hidden_ratio: 4
-     num_heads: 16
+   encoders:
+     0:                        # user-defined group name (appears in the state-dict)
+       datasets: [ "data" ]           # datasets encoded by this group
```
MeraX, line 37: Just a thought: What about using a key that is different from the DatasetDict meta-key here? It might help to keep thinks apart. e.g `input_datasets`for `encoders` and `output_datasets` for `decoders`. Or `target_datasets` if you want the same key for both.

MeraX (commented):
Hi @JPXKQX, this PR sound reasonable. I had a look at the (ever growing) configs and left a few comments.

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
JPXKQX: These keys have to match to existing keys in the dataloader.training/validation (but also `data`)

models/docs/usage/create_model.rst
JPXKQX: yes, good idea!

training/src/anemoi/training/config/model/graphtransformer.yaml
JPXKQX: Yes, you're right, but I would postpone that change. 
- For now, the old configs without the `datasets:` are still supported. This change would break that support. Having said that, I don't think there's a problem with this, and it's something that will happen in this PR or very soon.
- The main point is that we are still not sure how to handle the defaults after this PR (to be discussed in a follow-up chat), but one option is to add a `default` key at the level of the `datasets` key. In that case, it would be very easy to use the existing function to implement this across all config fields.

models/src/anemoi/models/models/target_features.py
JPXKQX: yes, good point!, Thinking about how it will be used, it's a list of single values where you don't need to set extra params for each value. 

For this, it seems like a cleaner solution, but I would also see the point of keeping a list of instantiable objects for homogeneity with the rest of the configs

training/src/anemoi/training/config/model/graphtransformer.yaml
JPXKQX: The encoders/decoders don't follow this schema but a regular `dict[str, ]` ([ref](https://github.com/ecmwf/anemoi-core/blob/feat/multi-dataset-configs/models/src/anemoi/models/schemas/models.py#L317)) Is what you were asking?

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
MeraX: but as far as I can see, there is just `era5`and `cerra` defined, no `data{1,2,3}`.

models/docs/usage/create_model.rst
JPXKQX: what about `source_datasets` and `target_datasets` ?

training/src/anemoi/training/config/model/graphtransformer.yaml
MeraX: Okay, but still the key should be str `"0"` not int `0`.

models/docs/usage/create_model.rst
MeraX: fine with me. just something that I can grep for.

training/src/anemoi/training/config/model/graphtransformer.yaml
JPXKQX: We are casting those with a schema hook to avoid later conversions in the code when referencing the encoders/decoders. Do you think we should update the schema type hint?

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
JPXKQX: ahh, I see what you mean. Yes, this has to be an error

training/src/anemoi/training/config/model/graphtransformer.yaml
MeraX: Alright. I didn't know that the schemas do type casting. Does this work with and without config validation?

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
MeraX: this should be `era5`, shouldn't it?

training/src/anemoi/training/config/model/graphtransformer.yaml
JPXKQX: I would say so but we should check it

JPXKQX: If config_validation is disabled, these keys are not cast. We have updated the configs to use str

MeraX: Thank you. But so far, these keys are not used as references anywhere, aren't they? My point is, the encoders could just be a list instrad of a dict. Nonetheless I see the advantage of a dict for future extensions. I just want to understand the concept here.

JPXKQX: Yes, they are used by the code. These keys define a mapping between datasets and encoder/decoders. At some point, hydra/omegaconf/anemoi-utils will transform these keys into strings. So the motivation was to do that conversion as early as possible to avoid casting it in the code with `str(decoder_name)`.

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
```diff
@@ -0,0 +1,215 @@
...
+      step: -1
+
+output_mask:
+  datasets:
+    era5:
+      _target_: anemoi.training.utils.masks.NoOutputMask
+    cerra:
+      _target_: anemoi.training.utils.masks.NoOutputMask
+
+trainable_parameters:
```
havardhhaugen: I think this should be made more explicit to better match the structure of the rest of the config. Something like 
```
node_trainable_parameters:
  hidden: 8
  datasets:
    data: 8 # single dataset
    era5: 8 # multi dataset 
    cerra: 8 # multi dataset
```
And then setting the edge trainable features explicitly in the mappers only

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
```diff
@@ -0,0 +1,215 @@
...
+      qk_norm: False
+      cpu_offload: ${model.cpu_offload}
+      gradient_checkpointing: True
+      layer_kernels: ${model.layer_kernels}
+      shard_strategy: "edges"
+      graph_attention_backend: "triton"  # Options: "triton", "pyg"
+      edge_pre_mlp: False
+  regional: # user-defined key. It will appear in the state-dict
+    source_datasets: [ "cerra" ]
+    dataset_fusing_strategy:  "not_supported" # Options: "not_supported" (only option now)
```
havardhhaugen, line 65: Is this a placeholder to allow asymmetric perceiver style attention? Would it be cleaner to use `input_target_features`  like in the decoder? This would then be e.g.
```
input_target_features: ['coordinates', 'trainable_parameters`] # standard encoding
input_target_features: ['encoded_data_global'] # asymmetric attention
```
and then `model.latent_aggregator` can be used to set the explicit data routing for the asymmetric attention case

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
MeraX: As far as I understand, these values here are just used for omegaconf to "interpolate" settings like
`model.decoders.0.trainable_size: ${model.trainable_parameters.hidden2data}`. I.e. This `model.trainable_parameters`section should become fully optional in the schema, as a user might whish to set `model.decoders.0.trainable_size`directly.

havardhhaugen: I thought only the edge trainable parameters were interpolated? I can't see any reference to `trainable_parameters.data` or `trainable_parameters.hidden` elsewhere in the config.

JPXKQX: The edge trainable parameters are the ones that are optional now. Do we want to keep them here? If not, we can rename `trainable_parameters` to `node_trainable_parameters`

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
JPXKQX: We think the perceiver-style attention is closer to the `latent_aggregator` than to the `input_target_features`. Here, we would like to control how multiple input datasets are handled for the same encoder. The following cases are envisaged:

> __concat_channels__: concatenate the channels of both decoders. This would be useful for downscaling, for example, when you want to encode an upsampled low-resolution field and a high-resolution field.
__concat_grids__: similar to GraphDOP.
__sequential__: use the same encoder for both datasets and aggregate the encoded latents.

Perceiver-style attention is more about grouping the outputs of multiple encoders, so latent_aggregator may be a better fit for that. Does this make sense to you? 

P.S. If users have only 1 encoder that processes multiple datasets, they could use the "sequential" strategy

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
havardhhaugen: I would remove them. The only worry would be that they are less visible and people forget to turn them off for multi-stage training. Do you still use these in the global models, or could it make sense to set them to 0 by default?

JPXKQX: Perhaps we could change the default setting to 0. We currently use them for AIFS v2, but their usage in future versions is under discussion.

training/src/anemoi/training/config/model/graphtransformer_multi.yaml
havardhhaugen: I see, that makes sense!

models/src/anemoi/models/models/hierarchical.py
```diff
@@ -71,124 +93,142 @@ def _build_networks(self, model_config):
...
-        self.upscale = nn.ModuleDict()
-        self.upscale_graph_providers = nn.ModuleDict()
-        for i in range(1, self.num_hidden):
+        # Downscale
+        self.downscale = nn.ModuleDict()
+        self.downscale_graph_providers = nn.ModuleDict()
+        for i in range(0, self.num_hidden - 1):
             src_nodes_name = self._graph_name_hidden[i]
-            dst_nodes_name = self._graph_name_hidden[i - 1]
+            dst_nodes_name = self._graph_name_hidden[i + 1]
```
VeraChristina, line 172: Do you intend to swap the order of src and dst nodes here?

JPXKQX: The hierarchical integration test should be working now with the latest change

models/docs/usage/multi_dataset_models.rst
```diff
@@ -0,0 +1,215 @@
...
+       dataset_fusing_strategy: "not_supported"
+       mapper:
+         _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
+         num_channels: 1024
+         # ... mapper configuration
+
+   decoders:
+     global:
+       datasets: [ "era5" ]
+       input_target_features: [encoded_data]
```
sahahner: This should be a string

models/docs/usage/multi_dataset_models.rst
```diff
@@ -0,0 +1,215 @@
...
+       mapper:
+         _target_: anemoi.models.layers.mapper.GraphTransformerBackwardMapper
+         num_channels: 1024
+         # ... mapper configuration
+     regional:
+       datasets: [ "cerra" ]
+       input_target_features: [encoded_data]
```
sahahner: this should be a string

models/docs/usage/multi_dataset_models.rst
```diff
@@ -0,0 +1,215 @@
...
+   concatenated in the configured dataset order, independently of
+   the order of the mapping passed to the model. Every configured dataset
+   must be active because omitting one would change the output width.
+
+*************************
+ Decoder target features
+*************************
+
+Each decoder group builds its input from an ordered list of
+``input_target_features``. This controls what the decoder receives in
```
sahahner: If I understand correctly these "live" on the output grid, which is why they are called "target" features. If this is the case, could we specify this in the explanation, please?

models/src/anemoi/models/layers/target_features.py
```diff
@@ -0,0 +1,326 @@
...
+
+def create_decoding_target_features(
+    features: list[str],
+    decoder_datasets_names: list[str],
+    model: BaseGraphModel,
+) -> DecodingTargetFeature:
+    """Create a dict of :class:`DecodingTargetFeature` instances from a list of registry keys."""
+    invalid = set(features) - VALID_TARGET_FEATURES
+    assert not invalid, (
+        f"Decoder has invalid input_target_features: {invalid}. " f"Valid options: {sorted(VALID_TARGET_FEATURES)}"
```
sahahner: I still find it slightly confusing that in the config file the list is called input_target_features, but the object that is created is a DecodingTargetFeature. I think we should harmonise this. 
Intuitively, input and target are contradictory in one feature name. Also, I think the information that these belong to the decoder should be reflected in their name.

sahahner (commented):
Hi @JPXKQX , @VeraChristina and I had a look at this together and we were both slightly confused by the naming of the DecodingTargetFeature/input_target_features, see specific comments as well. 

Another question would be whether the new DecodingTargetFeatures are covered by tests? This might be me, but I cannot find it.

Otherwise, the functionality is great, thank you for adding and updating the tests, and I am happy to approve once the questions regarding the naming and the tests are cleared up.

models/docs/usage/multi_dataset_models.rst
```diff
@@ -0,0 +1,215 @@
...
+   dimension is the **sum** of the per-dataset channel dimensions, so
+   the processor ``num_channels`` must be sized accordingly. Channels are
+   concatenated in the configured dataset order, independently of
+   the order of the mapping passed to the model. Every configured dataset
+   must be active because omitting one would change the output width.
+
+*************************
+ Decoder target features
+*************************
+
```
sahahner, line 130: What about using this:

Each decoder group builds its input from an ordered list of ``input_target_features``. These are features defined on the decoder's target (output) grid and supplied as inputs to the decoder. They control what the decoder receives in addition to, or instead of, the processed latent representation.

Here, target grid refers to the grid on which the decoder produces its output; it does not refer to the prediction targets of the forecasting model.

models/docs/usage/create_model.rst
```diff
@@ -26,38 +26,59 @@ First, let's take the model configuration ``transformer.yaml``:
...
+         num_chunks: 1
+         mlp_hidden_ratio: 4
+         num_heads: 16
+
+   latent_aggregator:
+     _target_: anemoi.models.layers.aggregator.SumAggregator
+
+   decoders:
+     0:
+       datasets: [ "data" ]
```
VeraChristina: should it be `target_datasets`?

models/docs/modules/models.rst
```diff
@@ -22,6 +22,39 @@ processor, and decoder components (`Lang et al. (2024a)
...
+   and is effectively a no-op, so no special residual class is needed.
+#. In the **model** config, set the decoder ``input_target_features`` to
+   ``[forcings, coordinates]`` (the base model default is
+   ``[encoded_data]``).
+
+.. code:: yaml
+
+   decoders:
+     global:
+       datasets: [era5]
```
VeraChristina, line 49: should it be `target_datasets`?

models/docs/usage/multi_dataset_models.rst
```diff
@@ -0,0 +1,215 @@
...
+       datasets: [ "cerra" ]
+       dataset_fusing_strategy: "not_supported"
+       mapper:
+         _target_: anemoi.models.layers.mapper.GraphTransformerForwardMapper
+         num_channels: 1024
+         # ... mapper configuration
+
+   decoders:
+     global:
+       datasets: [ "era5" ]
```
VeraChristina: `target_datasets`?

VeraChristina (commented):
Just a minor question about docs and configs, otherwise looking good!

models/src/anemoi/models/layers/target_features.py
JPXKQX: after discussing this, `target_node_features` is the new proposed name.

models/docs/usage/create_model.rst
JPXKQX: yes

JPXKQX:
Thanks @VeraChristina & @sahahner ! I think the config is now cleaner with the suggested changes. Documentation has been updated, and tests have been added for the target features.

VeraChristina (approved):
Thanks for your work and for addressing all the feedback!

JPXKQX:
[Benchmark tests passed](https://github.com/ecmwf/anemoi-core/actions/runs/33599504631/attempts/1) yesterday. Similarly, the `CheckpointError` raised by the edm diffusion configs is not raised locally. This issue is related to the compiled code for a fwd operation that is used for 2 different blocks due to the LRU cache. A solution to this issue in the CI will be explored in the coming weeks

PR #1230, opened 2026-07-03
feat(models, training): introduce sharding support for spectral losses

ssmmnn11:
support sharding for spectral losses

### DIFF
--- a/training/tests/unit/losses/test_loss_function.py
+++ b/training/tests/unit/losses/test_loss_function.py
@@ -498,6 +498,104 @@ def test_fft2d_spectral_losses_shape_and_validation(target: str) -> None:
         _ = loss(*wrong, squash=True)
 
 
+@pytest.mark.parametrize(
+    ("loss_cls", "ensemble_size"),
+    [
+        (SpectralL2Loss, 1),
+        (LogSpectralDistance, 1),
+        (FourierCorrelationLoss, 1),
+        (SpectralCRPSLoss, 4),
+    ],
+)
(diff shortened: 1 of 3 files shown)


havardhhaugen (dismissed):
Really nice work! I did not have time to test the branch explicitly and I'll be away on vacation, but the code changes look good to me. 

Maybe an idea for future work is to allow loss / scaler sharding along arbitrary dimensions to avoid the reshard operation at the end.

training/tests/unit/losses/test_loss_function.py
```diff
@@ -498,6 +498,104 @@ def test_fft2d_spectral_losses_shape_and_validation(target: str) -> None:
...
+    mocker: MockerFixture,
+) -> None:
+    x_dim, y_dim = 4, 4
+    group = object()
+    grid_shard_sizes = [8, 8]
+    channel_shard_sizes = [1, 1]
+    spectral_shard_sizes = [8, 8]
+    local_grid = grid_shard_sizes[0]
+
+    loss = loss_cls(transform="fft2d", x_dim=x_dim, y_dim=y_dim)
```
sahahner: If we use the SHT transform here, we should also be able to include the AMSE to the tested loss functions without adding another test.

sahahner (approved):
Looks good to me. Thank you @ssmmnn11 

As far as I see, PR https://github.com/ecmwf/anemoi-core/pull/1189 and https://github.com/ecmwf/anemoi-core/pull/992 will need to be consolidated with a few changes from here now.

PR #1234, opened 2026-07-07
feat(models): comm primitives tests

joherics:
## Unit tests for Anemoi distributed communication primitive 

This PR adds unit tests for Anemoi's distributed communication primitives
 defined in `models/distributed/primitives.py`.
 
The logic for spawning multiple processes is separated from the test implementations and supports both pytorch's GPU (nccl) and CPU (Gloo) backend. On top of this, new pytest options and fixtures have been introduced to mark tests as distributed and to make it easier to configure the test setup of distributed tests. Especially, this makes it easy to include and exclude the tests of the primitives in CI runners or when running pytest from the command line. In the future it could be useful to homogenise the configuration and setups logic, across anemoi, for all the test suites that use GPU and distributed feature .
                                                                         
The tests can be run with:                                               
```                                                                      
srun \                                                                           
  --nodes=1 \                                                                    
  --ntasks=1 \                                                                   
  --cpus-per-task=8 \                                                            
  --gres=gpu:2 \                                                                 
  --gpus-per-node=2 \                                                            
  --mem=0 \                                                                      
  --qos=dg \                                                                     
  --time=0:10:00 \                                                               
  python -m pytest -q models/tests/distributed/test_communication_primitives.py \
  --distributed \                                                                
  --distributed-backend=nccl \                                                   
  --distributed-world-size=2                                                                                                                      
```                                                                      
                                                                         
                                                                         
There is another PR coming that will sit on top of this for testing the sharding and communication functionality that builds on top of this. 


joherics:
@ssmmnn11 @japols @cathalobrien - Feel free to have a look at this, I have implemented tests for the communication primitives on this branch. The tests for the primitives can be extended, but I am mainly interested in whether you agree with me that this is a good design for the distributed testing in Python, especially the way I have structured it with the test launcher/runners in `distributed_runner.py` and `conftest.py` handling config logic.

models/tests/distributed/distributed_runner.py
```diff
@@ -0,0 +1,144 @@
...
+        if torch.cuda.device_count() < world_size:
+            msg = (
+                f"NCCL backend requested with world_size={world_size}, "
+                f"but only {torch.cuda.device_count()} CUDA device(s) are visible."
+            )
+            raise RuntimeError(msg)
+
+    fd, path = tempfile.mkstemp(prefix="anemoi_dist_init_", suffix=".tmp")
+    os.close(fd)
+    init_file = Path(path)
```
cathalobrien, line 65: making the initfile under $TMPDIR and not a shared filesystem means that these tests only work within a single node, not across nodes, right?

maybe that's sufficient or perhaps we also want multimode tests. That can be setup from slurm env vars e.g. https://github.com/ecmwf/anemoi-inference/blob/90d79111a71963c560e026b67e9625ef195d2fbc/src/anemoi/inference/runners/parallel.py#L196

cathalobrien (commented):
nice work!
How long do the tests take to run? currently you have this higher order function run_distributed_test which takes in a test and recreates the proc groups for every test. I wonder if that has a performance impact.
One could also imagine having a single file which creates a single proc group and reuses that proc group for all tests. I'm not sure which is better.

models/tests/distributed/test_communication_primitives.py
```diff
@@ -0,0 +1,281 @@
...
+    use_fp32: bool,
+    dtype: torch.dtype = torch.float32,
+    atol: float = GLOBAL_DEFAULT_ATOL,
+    rtol: float = GLOBAL_DEFAULT_RTOL,
+) -> None:
+    base = torch.arange(math.prod(shape), dtype=dtype, device=device).reshape(shape)
+    local = base + rank
+
+    actual = _reduce(local, use_fp32=use_fp32, group=group)
+    expected = sum(base + other for other in range(world_size))
```
japols: we should use `torch.sum` here and also specify `dtype=torch.float32` when `use_fp32 is True`

models/tests/distributed/test_communication_primitives.py
```diff
@@ -0,0 +1,281 @@
...
+    )
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize(("shape", "dtype"), LOW_PRECISION_REDUCE_CASES)
+def test_reduce_fp32_accumulation_supports_low_precision_inputs(
+    shape: tuple[int, ...], dtype: torch.dtype, distributed_backend: str, distributed_world_size: int
+) -> None:
+    if distributed_backend != "nccl":
+        pytest.skip("Low-precision fp32 accumulation is only validated on the NCCL/CUDA path.")
```
japols: what's the reason to skip this here?

japols (dismissed):
Thanks for adding this! I would agree that we should only call the distributed innit once instead of per test.

Some minor comments :)

joherics:
> nice work! How long do the tests take to run? currently you have this higher order function run_distributed_test which takes in a test and recreates the proc groups for every test. I wonder if that has a performance impact. One could also imagine having a single file which creates a single proc group and reuses that proc group for all tests. I'm not sure which is better.

Thank you, they take a bit more than 1 minute to run. I was also thinking that it might be preferable to have a single proc group, however that comes with disadvantage of making the tests less modular and it becomes more complex to handle failures and error reporting. Even though it is possible to do so, I think it is cleaner to actually have the tests launch new processes and tear them down if we can afford it.

models/tests/distributed/distributed_runner.py
joherics: Yes, you are right about that and these tests only run locally on one node, like the tests in the feat/benchmark_sharded branch, which follows a similar launching pattern. I think it makes a lot of sense to keep these tests single node only, as it allows use to cleanly parametrise the tests with pytest and pytorch is the only requirement.

That said, I think it would also make sense to add a separate multinode runner, specifically for slurm multinode jobs. I suggest that we add that after the next level of communication wrapper tests has been added.

models/tests/distributed/test_communication_primitives.py
joherics: Good catch, I was a bit confused about this fp32 accumulation and was thinking that lower precision tests are only relevant for GPUs. However, I don't think it makes sense to skip it based on that. So I removed these lines now.

models/tests/distributed/test_communication_primitives.py
joherics: Yes, good catch, I have updated all the tests so that they cover all the functionality of their respective primitive and also uses pytorch functions wherever possible.

joherics:
I have rebased this over main and replied to your comments @cathalobrien and @japols. It would be good if you could have a look at this again.

The tests are now more extensive and have been refactored to use Pytorch methods as much as possible. When doing so I found that the  `reduce` primitive could not handle discontiguous tensors (but the other primitives do) so I added support for this to the `reduce` primitive as well.

The total runtime is about ~2 mins 30 s for a multigpu test with 2 gpus and a total of 39 tests.

I also added CI workflow now that has to be triggered manually, but we should still decide when to run this. One idea would be to run this on PRs that change the code in `models/dsitributed/*`.

cathalobrien (approved):
looks good to me, great work

PR #1235, opened 2026-07-07
fix: graph open dataset

mc4117:
## Description
Pass the full `reader_cfg` to `data_node_cfg.node_builder.dataset` when the config is a mapping, so that all dataset options are preserved and forwarded to open_dataset. The plain path string is only used as a fallback when `reader_cfg` is not a dict.

When building the graph only the "dataset" key (a plain path string) was passed to `data_node_cfg.node_builder.dataset`. This meant that any additional open_dataset options present in the dataset config — such as check_variables_compatibility — were silently dropped and never forwarded to the graph node builder.


training/src/anemoi/training/train/train.py
```diff
@@ -222,7 +222,26 @@ def graph_data(self) -> HeteroData:
...
+                            "statistics",
+                            "step_start",
+                            "step_end",
+                            "step_frequency",
+                        }
+                        extra = {k: v for k, v in dict(reader_cfg).items() if k not in _schema_keys and v is not None}
+                        if extra:
+                            graph_dataset_value = {"dataset": dataset_path, **extra}
+                    data_node_cfg.node_builder.dataset = graph_dataset_value
             else:
```
HCookie: Do we need this distinction, could we not define `graph_dataset_value` before the check, and then just call .update?

training/src/anemoi/training/train/train.py
```diff
@@ -222,7 +222,26 @@ def graph_data(self) -> HeteroData:
...
+                        _schema_keys = {
+                            "dataset",
+                            "frequency",
+                            "drop",
+                            "select",
+                            "statistics",
+                            "step_start",
+                            "step_end",
+                            "step_frequency",
+                        }
```
HCookie: This will be quite fickle if datasets changes a key name, can we just pass all?

mc4117: we can't pass all because these are the ones that need to be dropped, but they're now being read in from the schema

training/src/anemoi/training/train/train.py
mc4117: changed

mc4117:
integration tests pass https://github.com/ecmwf/anemoi-core/actions/runs/28861879138

training/tests/unit/train/test_graph_loading.py
```diff
@@ -43,3 +46,59 @@ def test_existing_graph_validation_detects_fused_graph_from_loaded_file(tmp_path
     loaded_graph = trainer.graph_data
 
     assert set(loaded_graph.node_types) == {"era5", "cerra"}
+
+
+def test_graph_build_forwards_full_dataset_config_to_node_builder() -> None:
```
HCookie: One last thing, could you add a test that ensures that the expected dropped keys are indeed dropped.

HCookie (approved):

JPXKQX:
Thanks Mariana, this looks good to me! I'll work on a follow-up PR to refactor how we pass the graph config to the `GraphCreator`.

JPXKQX (approved):

PR #772, opened 2025-12-24
refactor(graphs): Expose Python API

radiradev:
## Description
This PR introduces an alternative way to construct graphs by instantiating objects directly, in addition to the existing YAML-based configuration approach.

Creating graphs via direct object instantiation improves transparency: docstrings are visible, IDE enables us to “go to definition”, and objects can be inspected directly during development.  Furthermore this way we can utilise static type checking, reducing the need for Pydantic schema validation.

Sidenote: The GraphCreator could already be instantiated directly on `main` in Python, but this requires to pass in a `DotDict` which is basically a parsed config. This PR instead explores using the edge, node and attribute objects directly. 

This is my first attempt as part of the broader plan in #768 and would serve as a test-bed for ideas and general feedback to a more pythonic `anemoi`.

Example graph with the new Python API:

```Python
from anemoi.graphs.create import GraphBuilder
from anemoi.graphs.nodes.builders.from_reduced_gaussian import ReducedGaussianGridNodes
from anemoi.graphs.nodes.attributes.area_weights import SphericalAreaWeights
from anemoi.graphs.edges.attributes import EdgeLength, EdgeDirection
from anemoi.graphs.edges import CutOffEdges, KNNEdges

def create_graph():
    data_nodes = ReducedGaussianGridNodes(
        grid="o48",
        name="data",
        attributes=[
            SphericalAreaWeights(norm="unit-max")
        ]
    )

    hidden_nodes = ReducedGaussianGridNodes(
        grid="o48",
        name="hidden",
        attributes=[
            SphericalAreaWeights(norm="unit-max")
        ]
    )

    edge_data_hidden = CutOffEdges(
        source_name="data",
        target_name="hidden",
        cutoff_factor=0.6,
        attributes=[
            EdgeLength(norm="unit-std"),
            EdgeDirection(norm="unit-std"),
        ]
    )

    edge_hidden_hidden = KNNEdges(
        source_name="hidden",
        target_name="hidden",
        num_nearest_neighbours=8,
        attributes=[
            EdgeLength(norm="unit-std"),
            EdgeDirection(norm="unit-std"),
        ]
    )

    edge_hidden_data = KNNEdges(
        source_name="hidden",
        target_name="data",
        num_nearest_neighbours=3,
        attributes=[
            EdgeLength(norm="unit-std"),
        ]
    )

    builder = GraphBuilder(
        nodes=[data_nodes, hidden_nodes],
        edges=[edge_data_hidden, edge_hidden_hidden, edge_hidden_data]
[...]

Issue #768, opened 2025-12-19
Decouple library code from configuration


radiradev:
Currently, the configuration in `anemoi-core` serves three distinct purposes:
 - As the primary user interface
 - Providing metadata for downstream packages (e.g `anemoi-inference`)
 - Defining default values and documenting the API 

While this unified interface allows users to read configurations without deserialization, the tight coupling between configuration and code creates significant friction. It forces both users and developers to constantly context-switch between YAML and Python.

---
**Pain Points**

- **For users**: When setting up runs, users often encounter unknown parameters. To resolve this, they must check documentation or go through the source code. `anemoi-datasets` is currently attempting to address this by adding a Python layer to generate configs, to me this feels like a workaround. https://github.com/ecmwf/anemoi-datasets/pull/420
- **For Developers** : Passing raw configuration objects through the codebase decreases readability. It obscures types and makes it difficult to inspect the actual objects instantiated by `hydra`. This also makes it hard to inspect objects instantiated by `hydra` when developing new features (e.g models). 

---
 **Proposal**
Instead of building a user layer solely to produce configurations, we should expose the underlying Python API as a first-class citizen. This decouples the core logic from the configuration system.

In my view, this would come with some nice benefits:
 - **IDE Support**: Users can leverage autocomplete, docstrings, and static type checking to guide their experiments.
 - **Programmability**: Users can create and inject their own compliant objects directly into the codebase. https://github.com/ecmwf/anemoi-datasets/issues/398
 - **Readability**: Developers can instantiate and inspect objects directly, removing the ambiguity of passing config dictionaries.
 - **Unit Testing**: Simplifies writing tests by removing the need to mock the configuration.

To maintain the ability to track metadata (currently handled by the config), we can serialize the metadata into checkpoints and deserialize it back to the Python API structure when needed. 

I'd leave configuration alternatives for another issue but this is for instance how [`fiddle` does it](https://colab.research.google.com/github/google/fiddle/blob/main/fiddle/examples/colabs/visualization_printing_codegen.ipynb) - can also be [serialized to json](https://github.com/google/fiddle/blob/4956f24776a8abf0641ba2aa1eefe842b09c9806/fiddle/_src/experimental/serialization.py).

---
I have a (backwards-compatible) proof of concept on `anemoi-graphs` on such an approach here #772 

Thanks a lot to my MeteoSwiss coworkers for the feedback, as well as, @gmertes and @anaprietonem for letting me voice this idea in person a few weeks ago.

[...]

PR #1236, opened 2026-07-07
refactor(graphs): Expose Python API

MeraX:
This is the continuation of https://github.com/ecmwf/anemoi-core/pull/772 with the branch transferred to ecmwf/anemoi-core

## Description
This PR introduces an alternative way to construct graphs by instantiating objects directly, in addition to the existing YAML-based configuration approach.

Creating graphs via direct object instantiation improves transparency: docstrings are visible, IDE enables us to “go to definition”, and objects can be inspected directly during development.  Furthermore this way we can utilise static type checking, reducing the need for Pydantic schema validation.

Sidenote: The GraphCreator could already be instantiated directly on `main` in Python, but this requires to pass in a `DotDict` which is basically a parsed config. This PR instead explores using the edge, node and attribute objects directly. 

This is my first attempt as part of the broader plan in #768 and would serve as a test-bed for ideas and general feedback to a more pythonic `anemoi`.

Example graph with the new Python API:

```Python
from anemoi.graphs.create import GraphBuilder
from anemoi.graphs.nodes.builders.from_reduced_gaussian import ReducedGaussianGridNodes
from anemoi.graphs.nodes.attributes.area_weights import SphericalAreaWeights
from anemoi.graphs.edges.attributes import EdgeLength, EdgeDirection
from anemoi.graphs.edges import CutOffEdges, KNNEdges

def create_graph():
    data_nodes = ReducedGaussianGridNodes(
        grid="o48",
        name="data",
        attributes=[
            SphericalAreaWeights(norm="unit-max")
        ]
    )

    hidden_nodes = ReducedGaussianGridNodes(
        grid="o48",
        name="hidden",
        attributes=[
            SphericalAreaWeights(norm="unit-max")
        ]
    )

    edge_data_hidden = CutOffEdges(
        source_name="data",
        target_name="hidden",
        cutoff_factor=0.6,
        attributes=[
            EdgeLength(norm="unit-std"),
            EdgeDirection(norm="unit-std"),
        ]
    )

    edge_hidden_hidden = KNNEdges(
        source_name="hidden",
        target_name="hidden",
        num_nearest_neighbours=8,
        attributes=[
            EdgeLength(norm="unit-std"),
            EdgeDirection(norm="unit-std"),
        ]
    )

    edge_hidden_data = KNNEdges(
        source_name="hidden",
        target_name="data",
        num_nearest_neighbours=3,
        attributes=[
            EdgeLength(norm="unit-std"),
        ]
    )

    builder = GraphBuilder(
        nodes=[data_nodes, hidden_nodes],
        edges=[edge_data_hidden, edge_hidden_hidden, edge_hidden_data]
    )
    return builder.create()
```
`YAML` equivalent
```YAML
---
nodes:
  data:
    node_builder:
      _target_: anemoi.graphs.nodes.ReducedGaussianGridNodes
      grid: o48
    attributes:
      area_weight:
        _target_: anemoi.graphs.nodes.attributes.SphericalAreaWeights
        norm: unit-max
  hidden:
    node_builder:
      _target_: anemoi.graphs.nodes.ReducedGaussianGridNodes
      grid: o48
    attributes:
      area_weight:
        _target_: anemoi.graphs.nodes.attributes.SphericalAreaWeights
        norm: unit-max

edges:
- source_name: data
  target_name: hidden
  edge_builders:
  - _target_: anemoi.graphs.edges.CutOffEdges
    cutoff_factor: 0.6
  attributes:
    edge_length:
      _target_: anemoi.graphs.edges.attributes.EdgeLength
      norm: unit-std
    edge_dirs:
      _target_: anemoi.graphs.edges.attributes.EdgeDirection
      norm: unit-std

- source_name: hidden
  target_name: hidden
  edge_builders:
  - _target_: anemoi.graphs.edges.KNNEdges
    num_nearest_neighbours: 8
  attributes:
    edge_length:
      _target_: anemoi.graphs.edges.attributes.EdgeLength
      norm: unit-std
    edge_dirs:
      _target_: anemoi.graphs.edges.attributes.EdgeDirection
      norm: unit-std

- source_name: hidden
  target_name: data
  edge_builders:
  - _target_: anemoi.graphs.edges.KNNEdges
    num_nearest_neighbours: 3
  attributes:
    edge_length:
      _target_: anemoi.graphs.edges.attributes.EdgeLength
      norm: unit-std
```

This an experimental new feature that is fully backwards compatible with the `YAML` configuration

#768 
##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1236.org.readthedocs.build/en/1236/

----
📚 Documentation preview 📚: https://anemoi-graphs--1236.org.readthedocs.build/en/1236/

----
📚 Documentation preview 📚: https://anemoi-models--1236.org.readthedocs.build/en/1236/


MeraX:
Dear @radiradev, dear @leifdenby,

I took the liberty to take up your excellent work. I updated your work from #772 to the latest main and added a test based on Radi's example. You find the new test in `test_create.py`. Unfortunately, I could not push to the old PR, thus I opened this new one. But I'm happy for Radi to take over the updates so that we could proceed from #772. I had a a look at Leif previous comment and Radi's replies and was tending to agree more with Radi's current implementation. 

Thus I would be glad if we could give this work another review and try to get this merged rather soon-ish. I think renaming and restructuring could come in following PRs. I personally could imagine to use the new API to easily add dedicated tests of the ICON-based graphs.

This work was approved ATS a while ago and the general notion was to take this work as a first step towards a more open APIs. More refactoring of the core is yet to come.

anaprietonem:
@leifdenby the github access issue should now be solved, but let me know if you still have issues looking at the PR!

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -46,11 +45,13 @@ def __init__(
         target_name: str,
         source_mask_attr_name: str | None = None,
         target_mask_attr_name: str | None = None,
+        attributes: list[BaseEdgeAttributeBuilder] | None = None,
```
leifdenby, line 49: I appreciate you are not changing `BaseEdgeBuilder` much here, but maybe we should create a docstring for `__init__` while we are modifying this function? It would be good to write down what these call arguments will be used for in the base class construction

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -99,38 +100,43 @@ def register_edges(self, graph: HeteroData) -> HeteroData:
         graph[self.name].edge_type = edge_type
         return graph
 
-    def register_attributes(self, graph: HeteroData, config: DotDict) -> HeteroData:
+    def register_attributes(self, graph: HeteroData, attributes: list | None = None) -> HeteroData:
         """Register attributes in the edges of the graph specified.
```
leifdenby: what does "registering" an attribute mean? Can we give brief explanation?

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -99,38 +100,43 @@ def register_edges(self, graph: HeteroData) -> HeteroData:
...
+        attributes : list
+            List of instantiated attribute objects.
 
         Returns
         -------
         HeteroData
             The graph with the registered attributes.
         """
-        for attr_name, attr_config in config.items():
+        attributes = attributes or []
```
leifdenby: add a default empty list argument instead of replacing `None` (which is the current default arg) to expose what the default actually is

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -99,38 +100,43 @@ def register_edges(self, graph: HeteroData) -> HeteroData:
...
         """Register attributes in the edges of the graph specified.
 
         Parameters
         ----------
         graph : HeteroData
             The graph to register the attributes.
-        config : DotDict
-            The configuration of the attributes.
+        attributes : list
+            List of instantiated attribute objects.
```
leifdenby, line 106: can we give the type of the "attribute" objects? It looks from the code below that they behave like dicts. Maybe the type annotation would then be `List[AttributeCollection]` or something like that?

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -99,38 +100,43 @@ def register_edges(self, graph: HeteroData) -> HeteroData:
...
+            graph[self.name][attr_obj.name] = attr_obj(
                 x=(graph[self.name[0]], graph[self.name[2]]), edge_index=edge_index
             )
         return graph
 
-    def update_graph(self, graph: HeteroData, attrs_config: DotDict | None = None) -> HeteroData:
+    def update_graph(
+        self,
+        graph: HeteroData,
+        attributes: list[BaseEdgeAttributeBuilder] | None = None,
```
leifdenby, line 117: wouldn't it be more accurate to name the argument `attributes_builder` instead of `attributes` if it does some sort of "building"?

graphs/src/anemoi/graphs/edges/builders/base.py
leifdenby: there are some verbs that need defining in the docstring too I think:

- what does *building an attribute* mean?
- what does *registering an attribute* mean?

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -142,9 +148,11 @@ def update_graph(self, graph: HeteroData, attrs_config: DotDict | None = None) -
         t1 = time.time()
         LOGGER.debug("Time to register edge indices (%s): %.2f s", self.__class__.__name__, t1 - t0)
 
-        if attrs_config is not None:
+        attributes = self.attributes if attributes is None else attributes
```
leifdenby, line 138: again set actual default argument value instead of replacing in place here.

Also, it appears that the effect of providing `attributes` to this method is to override what is stored on `self.` - that should be documented in the docstring, as why one might want to do that.

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -99,38 +100,43 @@ def register_edges(self, graph: HeteroData) -> HeteroData:
...
             The graph with the registered attributes.
         """
-        for attr_name, attr_config in config.items():
+        attributes = attributes or []
+
+        for attr_obj in attributes:
             edge_index = graph[self.name].edge_index
-            edge_attribute_builder = instantiate(attr_config)
-            graph[self.name][attr_name] = edge_attribute_builder(
+            graph[self.name][attr_obj.name] = attr_obj(
```
leifdenby, line 110: based on this operation i feel like maybe "register" isn't the correct verb here. Maybe "adding" attributes (not sure where they are coming from through...) is more correct?

graphs/src/anemoi/graphs/edges/attributes.py
```diff
@@ -119,9 +121,11 @@ class DirectionalHarmonics(EdgeDirection):
         Compute directional harmonics from edge directions.
     """
 
-    def __init__(self, order: int = 3, norm: str | None = None, dtype: str = "float32") -> None:
+    def __init__(
+        self, name: str | None = None, order: int = 3, norm: str | None = None, dtype: str = "float32"
+    ) -> None:
         self.order = order
-        super().__init__(norm=norm, dtype=dtype)
+        super().__init__(name, norm=norm, dtype=dtype)
```
leifdenby: why not named argument?

graphs/src/anemoi/graphs/generate/tri_icosahedron.py
```diff
@@ -162,12 +162,22 @@ def add_1_hop_edges(
     """Adds edges for x_hops = 1 relying on trimesh only."""
 
     hop_1_edges = []
+    nodes_coords_rad_np = nodes_coords_rad.cpu().numpy()
+    tree = BallTree(nodes_coords_rad_np, metric="haversine")
```
leifdenby: this looks like a different implementation - is that intentional?

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -284,10 +284,11 @@ class MaskedPlanarAreaWeights(PlanarAreaWeights):
     def __init__(
         self,
         mask_node_attr_name: str,
+        name: str | None = None,
         norm: str | None = None,
         dtype: str = "float32",
     ) -> None:
-        super().__init__(norm, dtype)
+        super().__init__(name, norm, dtype)
```
leifdenby: named arguments would be safer

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -343,7 +345,7 @@ def __init__(
         assert (
             isinstance(radius, float) or isinstance(radius, int)
         ) and radius > 0, f"radius must be a positive value, but radius={radius}"
-        super().__init__(norm, dtype)
+        super().__init__(name, norm, dtype)
```
leifdenby: named arguments would be safer

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -418,12 +420,13 @@ class CosineLatWeightedAttribute(BaseLatWeightedAttribute):
...
     def __init__(
         self,
+        name: str | None = None,
         min_value: float = 1e-3,
         max_value: float = 1,
         norm: str | None = None,
         dtype: str = "float32",
     ) -> None:
-        super().__init__(norm, dtype)
+        super().__init__(name, norm, dtype)
```
leifdenby: named arguments would be safer

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -418,12 +420,13 @@ class CosineLatWeightedAttribute(BaseLatWeightedAttribute):
 
     def __init__(
         self,
+        name: str | None = None,
```
leifdenby, line 423: missing docstring for `name` also for `dtype`

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -331,6 +332,7 @@ class SphericalAreaWeights(BaseAreaWeights):
...
```
leifdenby, line 335: missing in docstring

graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
```diff
@@ -284,10 +284,11 @@ class MaskedPlanarAreaWeights(PlanarAreaWeights):
     def __init__(
         self,
         mask_node_attr_name: str,
+        name: str | None = None,
```
leifdenby, line 287: missing in docstring

graphs/src/anemoi/graphs/nodes/attributes/base_attributes.py
```diff
@@ -70,5 +72,5 @@ def compute(self, graph: HeteroData, nodes_name: str, **kwargs) -> torch.Tensor:
 class BooleanBaseNodeAttribute(BaseNodeAttribute, ABC):
     """Base class for boolean node attributes."""
 
-    def __init__(self) -> None:
-        super().__init__(norm=None, dtype="bool")
+    def __init__(self, name: str | None = None) -> None:
```
leifdenby, line 75: missing docstring, what is `name` for?

graphs/src/anemoi/graphs/nodes/attributes/boolean_op.py
```diff
@@ -24,8 +24,8 @@
 class BooleanOperation(BooleanBaseNodeAttribute, ABC):
     """Base class for boolean operations."""
 
-    def __init__(self, masks: MaskAttributeType | list[MaskAttributeType]) -> None:
-        super().__init__()
+    def __init__(self, masks: MaskAttributeType | list[MaskAttributeType], name: str | None = None) -> None:
```
leifdenby, line 27: missing docstring, `masks` and `name`

graphs/src/anemoi/graphs/nodes/attributes/base_attributes.py
```diff
@@ -27,7 +28,8 @@ class BaseNodeAttribute(ABC, NormaliserMixin):
 
     norm_by_group: bool = False
 
-    def __init__(self, norm: str | None = None, dtype: str = "float32") -> None:
+    def __init__(self, name: str | None = None, norm: str | None = None, dtype: str = "float32") -> None:
+        self.name = name or camel_to_snake(self.__class__.__name__)
```
leifdenby, line 32: if this is the default we should document that in docstring

graphs/src/anemoi/graphs/nodes/attributes/masks.py
leifdenby, line 1: missing docstrings on most of the classes

graphs/src/anemoi/graphs/nodes/attributes/masks.py
```diff
@@ -113,7 +113,7 @@ def get_grid_sizes(self, nodes):
         return open_dataset(nodes["_dataset"]).grids
 
     @staticmethod
-    def get_mask_from_grid_sizes(grid_sizes: tuple[int], masked_grids_posisitons: list[int]):
+    def get_mask_from_grid_sizes(grid_sizes: tuple[int, ...], masked_grids_posisitons: list[int]):
```
leifdenby: why is the ellipsis needed here?

graphs/src/anemoi/graphs/nodes/builders/base.py
```diff
@@ -34,14 +33,19 @@ class BaseNodeBuilder(ABC):
...
         The area of interest mask builder, if any. Defaults to None.
     """
 
     hidden_attributes: set[str] = set()
+    _init_attributes: list = None
 
-    def __init__(self, name: str) -> None:
+    def __init__(self, name: str, attributes: list[BaseNodeAttribute] | None = None) -> None:
         self.name = name
+        self.attributes = attributes or []
```
leifdenby, line 46: what is the difference between a `BaseNodeAttribute` and `BaseNodeAttributeBuilder`? This class has `self.attributes` being a list of the former, but in edge.builders classes `self.attributes` refers to "attribute builders". This make me think even mor that the attribute should be called `self.attribute_builders` in the latter case

graphs/src/anemoi/graphs/nodes/builders/base.py
```diff
@@ -69,15 +73,15 @@ def register_nodes(self, graph: HeteroData) -> HeteroData:
 
         return graph
 
-    def register_attributes(self, graph: HeteroData, config: DotDict | None = None) -> HeteroData:
+    def register_attributes(self, graph: HeteroData, attributes: list | None = None) -> HeteroData:
         """Register attributes in the nodes of the graph specified.
```
leifdenby: describe what "registering" an attribute means

graphs/src/anemoi/graphs/nodes/builders/base.py
```diff
@@ -121,15 +127,19 @@ def reshape_coords(
         coords = torch.stack([latitudes, longitudes], axis=-1).reshape((-1, 2))
         return torch.deg2rad(coords)
 
-    def update_graph(self, graph: HeteroData, attrs_config: DotDict | None = None) -> HeteroData:
+    def update_graph(
```
leifdenby, line 114: how does "updating a graph" lead to attributes being "registered"? Wouldn't it be more correct to call this `register_attributes_on_graph` if we want to go with the "register" verb?

graphs/src/anemoi/graphs/nodes/builders/from_file.py
```diff
@@ -43,10 +43,10 @@ class AnemoiDatasetNodes(BaseNodeBuilder):
         Update the graph with new nodes and attributes.
     """
 
-    def __init__(self, dataset: DictConfig, name: str) -> None:
+    def __init__(self, dataset: DictConfig | str, name: str, attributes: list | None = None) -> None:
```
leifdenby: missing docstring

graphs/src/anemoi/graphs/nodes/builders/from_file.py
```diff
@@ -76,12 +76,19 @@ class TextNodes(BaseNodeBuilder):
...
     """
 
-    def __init__(self, dataset: str | Path, name: str, idx_lon: int = 0, idx_lat: int = 1) -> None:
+    def __init__(
+        self,
+        dataset: str | Path,
+        name: str,
+        idx_lon: int = 0,
+        idx_lat: int = 1,
+        attributes: list | None = None,
```
leifdenby: `name` and `attributes` missing in docstring

graphs/src/anemoi/graphs/nodes/builders/from_file.py
```diff
@@ -172,12 +182,13 @@ def __init__(
...
         margin_radius_km: float = 100.0,
+        attributes: list | None = None,
     ) -> None:
-        super().__init__(npz_file, name, lat_key, lon_key)
+        super().__init__(npz_file, name, lat_key, lon_key, attributes)
 
         self.area_mask_builder = AreaMaskBuilder(reference_node_name, margin_radius_km, mask_attr_name)
 
-    def register_nodes(self, graph: HeteroData) -> None:
+    def register_nodes(self, graph: HeteroData) -> HeteroData:
```
leifdenby: needs a docstring, is `graph` modified in place? That should be written in docstring. If so I don't think we should return the object as well, might give the impression that aren't modifying inplace

graphs/src/anemoi/graphs/nodes/builders/from_healpix.py
```diff
@@ -42,10 +42,10 @@ class HEALPixNodes(BaseNodeBuilder):
         Update the graph with new nodes and attributes.
     """
 
-    def __init__(self, resolution: int, name: str) -> None:
+    def __init__(self, resolution: int, name: str, attributes: list | None = None) -> None:
```
leifdenby, line 44: `attributes` missing in docstring

graphs/src/anemoi/graphs/processors/post_process.py
```diff
@@ -331,7 +333,10 @@ def compute_mask(self, graph: HeteroData) -> torch.Tensor: ...
 
     def recompute_attributes(self, graph: HeteroData, graph_config: dict) -> HeteroData:
         """Recompute attributes"""
```
leifdenby, line 313: needs explaination of args in docstring

graphs/src/anemoi/graphs/create.py
```diff
@@ -17,32 +17,31 @@
...
 
 
-class GraphCreator:
-    """Graph creator."""
-
-    config: DotDict
+class GraphBuilder:
+    """Create a graph using without a config."""
 
     def __init__(
```
leifdenby, line 42: missing docstring, I think we need to explain really clearly here what is done with the arguments

graphs/src/anemoi/graphs/create.py
```diff
@@ -17,32 +17,31 @@
...
+        self.nodes = nodes or []
+        self.edges = edges or []
+        self.post_processors = post_processors or []
 
     def update_graph(self, graph: HeteroData) -> HeteroData:
         """Update the graph.
 
-        It instantiates the node builders and edge builders defined in the configuration
-        file and applies them to the graph.
+        It iterates over the node builders and edge builders and applies them to the graph.
```
leifdenby: to do what?

graphs/src/anemoi/graphs/create.py
```diff
@@ -17,32 +17,31 @@
...
-    """Graph creator."""
-
-    config: DotDict
+class GraphBuilder:
+    """Create a graph using without a config."""
 
     def __init__(
         self,
-        config: str | Path | DotDict | DictConfig,
+        nodes: list[BaseNodeBuilder] | None = None,
```
leifdenby: I think these should be named `node_builders` at least, they are not "nodes" yet. But I am not sure "building" is the right name, don't they just extract attributes? Or what do they return?

graphs/src/anemoi/graphs/create.py
```diff
@@ -52,26 +51,16 @@ def update_graph(self, graph: HeteroData) -> HeteroData:
...
         HeteroData
-            The updated graph with new nodes and edges added based on the configuration.
+            The updated graph with new nodes and edges added.
         """
-        for nodes_name, nodes_cfg in self.config.get("nodes", {}).items():
-            graph = instantiate(nodes_cfg.node_builder, name=nodes_name).update_graph(
-                graph, attrs_config=nodes_cfg.get("attributes", {})
-            )
+        for node in self.nodes:
+            graph = node.update_graph(graph)
```
leifdenby: aren't all the "updates" in place? If so then this assignment isn't needed and is miss-leading. If not then every iteration here is leading to the creation of a new object which would be slower than reusing hte same object

leifdenby: would it be better to have the function that takes the graph as an argument to "add nodes" be called a "node adder"?

Then the call would be:
```python
for add_nodes in self.node_adders:
   add_nodes(graph)
```

graphs/src/anemoi/graphs/create.py
```diff
@@ -100,7 +89,7 @@ def clean(self, graph: HeteroData) -> HeteroData:
     def post_process(self, graph: HeteroData) -> HeteroData:
         """Allow post-processing of the resulting graph.
 
-        This method applies any configured post-processors to the graph,
+        This method applies any post-processors to the graph,
```
leifdenby, line 124: what does "post-processing" the graph mean? We should explain that in the docstring

graphs/src/anemoi/graphs/create.py
```diff
@@ -176,6 +160,88 @@ def create(self, save_path: Path | None = None, overwrite: bool = False) -> Hete
         return graph
 
 
+class GraphCreator(GraphBuilder):
```
leifdenby: what is the difference between a "graph builder" and a "graph creator"? To me this looks like something that could be a class method on `GraphCreator` in that method that works like a specialised `__init__` for using the `GraphBuilder`, so it could be a method on the parent class `GraphBuilder.create_graph_from_config(cls, config)`, decorated with `@classmethod`.

This is also what I suggested in https://github.com/radiradev/anemoi-core/pull/3. We can then add support for creating graphs from other things too that way easily.

graphs/src/anemoi/graphs/create.py
```diff
@@ -176,6 +160,88 @@ def create(self, save_path: Path | None = None, overwrite: bool = False) -> Hete
...
+            if attributes_cfg:
+                for attr_name, attr_cfg in attributes_cfg.items():
+                    attributes.append(instantiate(attr_cfg, name=attr_name))
+
+            node = instantiate(node_builder_cfg, name=node_name, attributes=attributes)
+            _nodes.append(node)
+    return _nodes
+
+
+def _parse_edges(cfg: DotDict) -> list[BaseEdgeBuilder]:
```
leifdenby, line 206: missing docstring

graphs/src/anemoi/graphs/utils.py
```diff
@@ -30,7 +48,7 @@ def get_distributed_device() -> torch.device:
         local_rank = int(os.environ.get("SLURM_LOCALID", 0))
         device = torch.device(f"cuda:{local_rank}")
     else:
-        device = "cpu"
+        device = torch.device("cpu")
```
leifdenby, line 52: was there a bug here?

graphs/src/anemoi/graphs/commands/create.py
```diff
@@ -54,7 +54,7 @@ def run(self, args):
                 GraphDescriptor(args.save_path).describe()
             return
 
-        graph_creator = GraphCreator(config=args.config)
+        graph_creator = GraphCreator(config_path=args.config)
         graph_creator.create(save_path=args.save_path, overwrite=args.overwrite)
```
leifdenby: I don't see the need for naming a `.create()` method on a "graph creator". The main role of the "graph creator" is to "create graphs", so we can just use `__call__(...)`, so that this line becomes:

`graph_creator(save_path=args.save_path, overwrite=args.overwrite)`

graphs/src/anemoi/graphs/edges/builders/base.py
```diff
@@ -99,38 +100,43 @@ def register_edges(self, graph: HeteroData) -> HeteroData:
         graph[self.name].edge_type = edge_type
         return graph
 
-    def register_attributes(self, graph: HeteroData, config: DotDict) -> HeteroData:
+    def register_attributes(self, graph: HeteroData, attributes: list | None = None) -> HeteroData:
```
leifdenby: is `register_attributes` meant to be called externally or only internally in `update_graph(...)`? If so I would prefix this with an underscore

leifdenby (changes requested):
This is really great @MeraX! Thank you for taking up this work. I've left quite a lot of comments, many a just relating to docstrings others are about naming and the overall design here.

I think we should take the time to make the docstrings really useful now since the aim is to be able to make use of these in for example our editors.

A generally point also is to use `__call__` on the classes where the main role of an object is to do something which is already described by its name. There is not need to have `edge_builder.build_edges(...)` we can just have `edge_builder(...)`

Hope this review is useful :)

graphs/src/anemoi/graphs/utils.py
MeraX: no bug here. I'll revert.

graphs/src/anemoi/graphs/nodes/builders/from_file.py
MeraX: name and attributes will be added.

graphs/src/anemoi/graphs/generate/tri_icosahedron.py
MeraX: no this change was accidental, but related to another bug that altered a test. not it is resolved.

graphs/src/anemoi/graphs/edges/attributes.py
MeraX: resolved.

graphs/src/anemoi/graphs/nodes/builders/base.py
MeraX: I think there is no `BaseNodeAttributeBuilder`. The Builder is for Edge attributes. I think `BaseEdgeAttributeBuilder` is kind of a factory, but `BaseNodeAttribute` is the thing that can actually `compute` attributes. But now when looking at the code, the `BaseEdgeAttributeBuilder` also has a `compute` method. @JPXKQX could you help us to understand the software design beter?

graphs/src/anemoi/graphs/utils.py
MeraX: well, yes. the return type mismatched.

MeraX (commented):
Hi @leifdenby , I followed your suggestion of making the `graph` manipulation explicit. And at least I can understand the flow of data and the code much better now. However, this extends the change set of this PR quite a bit.

graphs/src/anemoi/graphs/edges/builders/base.py
MeraX: I believe they should all inherit from `BaseEdgeAttributeBuilder`. @JPXKQX can you confirm this? (I updated the type annotation accordingly)

graphs/src/anemoi/graphs/create.py
MeraX: I fully get your point and refactored this new code. There is no more new `GraphBuilder`, but an `initialize_from_config` class method. I think this would be a nice software pattern elsewhere, too.

graphs/src/anemoi/graphs/create.py
MeraX: What would you like to know here? is it worth letting DAWID add the following?
```
    """Create node builders from the node configuration.

    Parameters
    ----------
    cfg : DotDict
        Graph configuration containing node definitions.

    Returns
    -------
    list[BaseNodeBuilder]
        Instantiated node builders.
    """
```
Perhaps only the last line is worth it as it contains slightly more information than conveyed in the function signature.

MeraX (commented):
@JPXKQX I think your expertise is needed to get this PR over the finish line. Could you help us to understand the structure and motivation behind all the various Builders better so that we can add the missing docsctrings?

graphs/src/anemoi/graphs/edges/builders/base.py
JPXKQX: exactly, they should all inherit from `BaseEdgeAttributeBuilder`.

graphs/src/anemoi/graphs/edges/builders/base.py
JPXKQX: We can change the name of this method, but for consistency, I would also rename `register_edges`, `register_nodes`.

The attribute values are normally derived from node features.

graphs/src/anemoi/graphs/nodes/builders/base.py
JPXKQX: Hi, both edge and node attributes are created/added/registered in a similar way. Looking at the names, renaming the `BaseNodeAttribute` to `BaseNodeAttributeBuilder` would make more sense or even removing the "Base" for all base classes and leaving them as "EdgeBuilder", "NodeBuilder", "NodeAttributeBuilder", "EdgeAttributeBuilder".

graphs/src/anemoi/graphs/nodes/attributes/masks.py
MeraX: Do you mean regarding the Arguments?

PR #1239, opened 2026-07-10
feat(transport): add experimental piecewise sigma schedule (for diffusion downscaling)

### LINKED ISSUE #1318: Add piecewise diffusion sampler for downscaling

For diffusion downscaling we need a sampling strategy, where sampling should:

- traverse the high-noise regime quickly;
- spend most computation in the low-noise regime where fine-scale fidelity is determined.

Introduce a piecewise sigma schedule:

- rapid collapse through the high-noise regime;
- Karras-style refinement in the low-noise regime.

This is intended as a purely additive change consisting of one new sampling schedule in models/src/anemoi/models/transport/schedules.py

_No response_

### Additional context

_No response_

### Organisation

_No response_

JoffreyDumontLeBrazidec:
Adds a piecewise EDM noise schedule for diffusion-based downscaling.
The schedule uses a fast high-noise collapse followed by finer Karras-style refinement at low noise.

When to use this scheduler

1 The model was trained with a noise distribution that puts real mass at high σ (e.g. a heavy-tailed log-normal), so sampling from a very large σ_max exploits a regime the denoiser has actually learned — this is where large-scale structure and extreme values are committed.

2 Inference cost matters (large ensembles, high resolution), so you want to spend the step budget at low noise, where step density controls fine-scale fidelity, rather than on the high-noise descent.


JPXKQX:
Hi Joffrey, thanks for the pull request, it looks good to me. The only comment I have is about the 'experimental' tag. Is this because no scientific results have yet been obtained, or because the implementation has not yet been fully tested? We don't yet have a proper system for experimental features, but I would remove the 'experimental' prefix from the scheduler name.

JoffreyDumontLeBrazidec:
Thanks for you review Mario, I agree and I removed the experimental tag

anaprietonem:
@JoffreyDumontLeBrazidec thanks for the PR! is this something you already trained some models with? it would if you could if you could share more insights on this or extend the PR description to understand when is it recommended to users to use this scheduler

JoffreyDumontLeBrazidec:
added a description @anaprietonem !

PR #1245, opened 2026-07-17
feat(model): fileGraphProvider - loading graphs from disk

### LINKED ISSUE #1150: feat(models): Support loading graphs from disk

Currently, the graph provider supports either static graphs or dynamically created graphs. 
When scaling up to a large amount of (large-sized) graphs with either multi-dataset or multi-domain, using static graphs can cause out of memory exceptions, and dynamically creating graphs causes and extreme slow-down in training. 

Pre-generating graphs and loading them from disk would solve both problems by reducing memory footprint with minimal time increase. Initial tests show 250 300MB graphs can be loaded and offloaded in 182.0 ms (4 workers, pinned memory). 

The proposed solution would be to add a graph provider `FileGraphProvider` allowing users to specify a folder with pre-generated graphs that would be loaded dynamically. 

```
path = self.paths[idx]
graph = torch.load(path, weights_only=False, map_location="cpu")
```

An alternative solution would be to load all graphs in CPU memory. However, there remains a limit to how many graphs can be stored (experiments show a maximum of about 50 300MB graphs). Sharing memory across GPUs would increase the limit to 200 graphs, but no more. 

### Additional context

Use cases:
- Training with large (high-resolution) graphs
- Scaling up the amount of graphs (with multi-domain or multi-dataset training)

### Organisation

KNMI

Sooph-knmi:
## Description
Supporting loading graphs from disk during training. 

If the provided graph in `system.input.graph` is a `Path`, the graphs are loaded from that folder with the new graph provider `FileGraphProvider` during the model forward. 

Use cases:

- Training with large (high resolution) graphs
- Scaling up the amount of graphs (multi-dataset/multi-domain training)

No breaking changes. 

Currently, the graph provider supports either static graphs or dynamically created graphs. When scaling up to a large amount of (possibly also large-sized) graphs, using static graphs can cause out of memory exceptions, and dynamically creating graphs causes an extreme slow-down in training.

#1150 

##  Additional notes ##

Right now, the model/traning initialization can still be slow because many graph attributes are loaded independently (see also the #1150 discussion). This might be something for a separate discussion. 

Current implementation is only for `AnemoiModelEncProcDec`. Can be extended to the other models if there is interest from the community. 

----
📚 Documentation preview 📚: https://anemoi-training--1245.org.readthedocs.build/en/1245/

----
📚 Documentation preview 📚: https://anemoi-graphs--1245.org.readthedocs.build/en/1245/

----
📚 Documentation preview 📚: https://anemoi-models--1245.org.readthedocs.build/en/1245/


models/src/anemoi/models/layers/graph_provider.py
```diff
@@ -760,3 +787,217 @@ def from_config(
...
+        self.graph_name = dataset_name if dataset_name is not None else "graph"
+        self.edge_attributes = edge_attributes
+        self.trainable_size = trainable_size
+        assert self.edge_attributes is not None, "Edge attributes must be provided"
+
+        self.graph_dir = Path(graph_dir)
+
+        self._dataset = _GraphFileDataset(self.graph_dir, extension=extension)
+        self.names = self._dataset.names
+        graph = self._dataset[self.graph_name]
```
einrone: The attr self.graph_name is either a dataset name or defaulted to "graph". However isnt self._dataset an iterator? hence later it get called in the __getitem__ with self.names[idx], or what is the idea here

PR #1248, opened 2026-07-20
fix(models): normalized relu bounding multidataset

sahahner:
## Description

Enable NormalizedReluBounding (and NormalizedLeakyReluBounding) in multi-dataset setups.
Other bounding classes already tolerate variables missing from a dataset via BaseBounding._create_index, which silently filters variables against name_to_index. 
NormalizedReluBounding didn't use this method and therefore direct name_to_index[var] lookups raised KeyError. 

## Solution
This PR applies the same silent-skip to variables, min_val, and normalizer. Config validation now runs before filtering so errors are dataset-independent.

## Tests
Added a parametrised test covering the skip-missing behavior across all bounding classes. This test would have caught the error.


models/src/anemoi/models/layers/bounding.py
```diff
@@ -124,49 +124,53 @@ def __init__(
         name_to_index_stats : dict
             A dictionary mapping the variable names to their corresponding indices in the statistics dictionary.
         """
+        # Validate raw config before filtering so errors are dataset-independent.
```
mc4117: just wondering why has this now moved before the super init?

models/src/anemoi/models/layers/bounding.py
```diff
@@ -124,49 +124,53 @@ def __init__(
...
-            raise ValueError(
-                "The length of the normalizer list must match the number of variables in NormalizedReluBounding."
-            )
-        if len(self.min_val) != len(variables):
-            raise ValueError(
-                "The length of the min_val list must match the number of variables in NormalizedReluBounding."
-            )
+        # Silently skip variables absent from this dataset (matches BaseBounding._create_index).
+        kept = [(ii, var) for ii, var in enumerate(variables) if var in name_to_index]
+        kept_variables = [var for _, var in kept]
```
mc4117: should self.variables be updated to kept_variables?

models/src/anemoi/models/layers/bounding.py
sahahner: I moved it up to test the normalized relu bounding specific inputs directly.
The comment can be removed, though, as it does not belong here.

models/src/anemoi/models/layers/bounding.py
sahahner: There are two options (both would work for me)
1. Align self.variables with how it is used for the other bounding functions and do not filter per dataset. Per-dataset arguments are then stored in kept_variables, self.min_val, self.normalizer.
2. Align self.variables with self.min_val and self.normalizer and overwrite with kept_variables.

mc4117: I have to say I prefer option 2 but if the other bounding functions use option 1 then maybe it's better to leave how it is

sahahner: I do agree with your comment, that aligning self.variables to self.min_val and self.normalizer is possibly more feasible than aligning it with the other binding functions.

mc4117 (approved):
LGTM!

PR #1250, opened 2026-07-21
Feat/the great refactoring

b8raoult:
## Description

See https://github.com/ecmwf/anemoi-core/pull/1270, which explain why this draft PR exists.

Experimental refactoring:

- isolates hydra code in one class, and provide other ways of parametrising models and graphs creation, as well as training;
-  paves the way to the creation of models/graphs/training programmatically;
-  will allow for pickle less checkpoints

##  Additional notes ##


radiradev:
Hi @b8raoult thank you for working in this direction, I like this! Is the idea in the future to also to directly pass some of the "configuration" as constructor arguments - similar to what I had done for `graphs`?  As a concrete example I think it would be great if `layer_kernels` instead of being a dict (or a Parametrisation) arguments were directly passed into the class. As an example: 
```
kernels = {
    "Linear": {"_target_": "torch.nn.Linear"},
    "LayerNorm": {"_target_": "torch.nn.LayerNorm"},
    "Activation": {"_target_": "torch.nn.GELU"},
    "QueryNorm": {
        "_target_": "anemoi.models.layers.normalization.AutocastLayerNorm",
        "_partial_": True,
        "bias": False,
    },
    "KeyNorm": {
        "_target_": "anemoi.models.layers.normalization.AutocastLayerNorm",
        "_partial_": True,
        "bias": False,
    },
}
```
```
SomeModel(..., layer_kernels=kernels) 
```

Would become instead (with defaults whenever we don't specify that constructor argument)

```
SomeModel(..., activation=torch.nn.ReLU, query_norm=anemoi.models.layers.normalization.AutocastLayerNorm) 
```

Do you think this is feasible? In my mind working in that direction would make things more transparent and would allow for easier testing and better user experience.

PR #1251, opened 2026-07-21
feat(training): use local synchronization for communication groups

ssmmnn11:
use local synchronization for communication groups to speed up init of training


cathalobrien (approved):
looks good, nice work

JPXKQX:
Should this be tagged somehow to show that it's changing the default behaviour? I don't think breaking change is an appropriate label for this, but it would be nice to have something to tag this

ssmmnn11:
> Should this be tagged somehow to show that it's changing the default behaviour? I don't think breaking change is an appropriate label for this, but it would be nice to have something to tag this

It changes default behaviour, but there is no impact on users apart from sometimes faster startup.

joherics (approved):
Looks good to me! This seems like a useful change to the comm group creation defaults.

PR #1254, opened 2026-07-22
feat(training): ignore warmup iterations when calculating throughput

cathalobrien:
## Description
currently throughput is a rolling avg of all batch times. however the initial batches (first 100) are typically slow (torch compile, cuda memory allocations) and variable. when running a ~200 iterations these initial batches drag down the reported throughput, this can cause throughput results to vary across runs.

this pr ignores the first 100 iterations when calculation throughput. if you're running less then 100 iterations it ignores the first half.


anaprietonem (approved):
LGTM!

PR #1259, opened 2026-07-23
fix(training): expandable segments warning

cathalobrien:
## Description
compiling the ensemble crps while using expandable segments gives a null pointer error for torch 2.10 and older. this PR does two things:
1) disable expandable segments by default in benchmark tests (we dont need it since we made the tests smaller)
2) when using compilation, check if expandable segments is enabled and warning users that this error might occur and how to fix it

with this PR the ensemble crps runs and [the benchmark tests are passing](https://github.com/ecmwf/anemoi-core/actions/runs/29996078754/job/89193053150#step:3:5393)
```
11:56:10 INFO PASS. Local value for avThroughputIterPerS has improved compared to the reference value (1.23iter/s local vs 0.39iter/s reference)
11:56:10 INFO PASS. Local value for peakMemoryMB has improved compared to the reference value (14975.36MB local vs 16335.80MB reference)
```

*Note* this error appears to be fixed in torch 2.12, so as soon as anemoi can go past 2.10 then it will be possible to use expandable segments + torch compile

### DIFF
--- a/training/src/anemoi/training/diagnostics/benchmark_server.py
+++ b/training/src/anemoi/training/diagnostics/benchmark_server.py
@@ -33,7 +33,6 @@ from pytorch_lightning.utilities.rank_zero import rank_zero_only
 from torch.cuda import memory_stats
 
 os.environ["ANEMOI_BASE_SEED"] = "42"  # need to set base seed if running on github runners
-os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"  # reduce memory fragmentation
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


cathalobrien:
don't merge yet, need to run integration tests still

anaprietonem (commented):
I assume before `expandable_segments:True` was used for all the tests, and now it's a problem on the crps one because there we are compiling the loss. But wouldn't still be good to keep these true for the other tests (ie have some way to specify the Pytorch env variables and settings per test?).

cathalobrien:
> I assume before `expandable_segments:True` was used for all the tests, and now it's a problem on the crps one because there we are compiling the loss. But wouldn't still be good to keep these true for the other tests (ie have some way to specify the Pytorch env variables and settings per test?).

yep, that's right! it should be straightforward to just disable expandable_segments for the ensemble tests. but I was thinking about the future, where we might(/hopefully) have more compiled code. globally disabling them seems more straightforward. 
currently the bm tests just use 6-11GB per GPU depending on the test out of 40GB. so I don't think there's a need for expandable_segments to squeeze out every last bit of memory

anaprietonem (approved):

PR #1264, opened 2026-07-24
refactor(training)!: repair combined loss summation

ssmmnn11:
Refactor to repair combined loss and prevent silent broadcast bug. 

Note that the labels under which the loss values are logged will change.


ssmmnn11:
we will remove logging of multiple scales instead.

PR #1265, opened 2026-07-24
fix(training)!: aggregate multiscale loss outputs

ssmmnn11:
MultiscaleLossWrapper now returns a scalar when squashed and a per-variable vector when unsquashed. This prevents a bug when combined with other losses via the CombinedLoss. Per-scale validation metrics are no longer logged. Please note that this will change the labeling of some ml flow metrics - it will impact the labels only.


mc4117 (dismissed):
LGTM! Thanks for making this simpler

mc4117 (approved):
LGTM!

PR #1267, opened 2026-07-24
fix(training): fix race condition in benchmark tests

cathalobrien:
## Description
sometimes the final benchmark test will fail, when process 0 can't find the files it needs to analyse the performance. that's bc rank 1 has already finished and deleted the files. this PR adds a barrier at the end of the test (outside the timer) to prevent this.

also I renamed the dataloader benchmark so it runs after the other benchmarks, bc I got annoyed at waiting


anaprietonem (approved):
@cathalobrien can you add a link to the run where this fix solves the problem associated to the file not found error on the last test please. Otherwise LGTM, feel free to merge!

cathalobrien:
sure, in [this test](https://github.com/ecmwf/anemoi-core/actions/runs/30014562282/job/89249104524#step:3:7173) run with this branch there is no 'file not found' error during the last test, just an ensemble error due to runtime variability being too high.

In [this test](https://github.com/ecmwf/anemoi-core/actions/runs/30133658199/job/89613142399#step:3:8811) run on main, there is a file not found error during the last test

PR #1268, opened 2026-07-24
feat(training): graph scores

ssmmnn11:
Graph scores and more ...

Screenshot 2026-07-24 at 18 06 55

- Adds GraphEnergyScoreLoss, GraphEdgeEnergyScoreLoss, GraphEdgeCRPSLoss, GraphVariogramScoreLoss, and GlobalEnergyScoreLoss, plus coefficient-magnitude scoring for SpectralCRPSLoss.
- These ensemble scores assess spatial structure across graph neighbourhoods, edge differences, global fields, variables (global energy) and spectral amplitudes
- The graph scores can be composed with CombinedLoss and used as objectives in MultiscaleLossWrapper.
- Performant via `torch.compile` and full sharding support

Pls see https://arxiv.org/abs/2607.19161v1 for a description of the score options.

----
📚 Documentation preview 📚: https://anemoi-training--1268.org.readthedocs.build/en/1268/

----
📚 Documentation preview 📚: https://anemoi-graphs--1268.org.readthedocs.build/en/1268/

----
📚 Documentation preview 📚: https://anemoi-models--1268.org.readthedocs.build/en/1268/


training/src/anemoi/training/losses/kcrps.py
```diff
@@ -168,7 +168,7 @@ def forward(
         without_scalers: list[str] | list[int] | None = None,
         grid_shard_slice: slice | None = None,
         group: ProcessGroup | None = None,
-        squash_mode: Squash_mode = "sum",
+        squash_mode: Squash_mode = "avg",
```
mchantry, line 172: Why is this changing?

ssmmnn11: We had a bug where the loss wrapper was overwriting this anyway. So this is the current "default" by accident.

training/src/anemoi/training/losses/graph_edge_energy_score.py
```diff
@@ -0,0 +1,104 @@
...
+from torch_geometric.data import HeteroData
+
+from anemoi.training.losses.graph_energy_score_base import BaseGraphEnergyScoreLoss
+from anemoi.training.losses.graph_score_base import csr_matmul
+from anemoi.training.losses.graph_score_base import safe_sqrt
+from anemoi.training.losses.graph_score_base import scale_node_differences
+from anemoi.training.losses.graph_score_graph import GraphScoreGraph
+
+
+class GraphEdgeEnergyScoreLoss(BaseGraphEnergyScoreLoss):
```
VeraChristina, line 20: Perhaps worth noting somewhere that this doesn't inherit from BaseGraphEdgeScoreLoss, just because the name suggests it does. (I think it makes sense that it doesn't and the name makes sense, it's more for clarity)

training/src/anemoi/training/losses/graph_edge_energy_score.py
```diff
@@ -0,0 +1,104 @@
...
+        assert matrix is not None
+        assert row_weight_sum is not None
+
+        safe_differences = differences
+        if node_valid is not None:
+            safe_differences = torch.where(node_valid, differences, torch.zeros_like(differences))
+
+        # Edge differences are invariant to a constant spatial offset.
+        # Improve stability of computation
+        safe_differences = safe_differences - safe_differences[..., :1, :]
```
VeraChristina, line 75: Does this work in both ignore_nan=True/False cases?

training/src/anemoi/training/losses/graph_score_graph.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @classmethod
+    def from_definition(
+        cls,
+        graph_definition: Mapping[str, object] | None,
+        graph_data: HeteroData | None,
+        *,
+        graph_name: str,
+        allow_none: bool = False,
+        require_square: bool = True,
```
VeraChristina, line 81: Do we use/allow require_square = False anywhere? There seem to be places where we assume square shapes, see below.

training/src/anemoi/training/losses/graph_score_base.py
```diff
@@ -0,0 +1,360 @@
...
+                f"Prediction and target shapes are incompatible: "
+                f"{tuple(y_pred_ens.shape)} and {tuple(y_target.shape)}."
+            )
+            raise ValueError(msg)
+
+    def _validate_graph_grid_size(self, y_pred_ens: torch.Tensor) -> None:
+        if self.graph is None:
+            return
+        grid_size = y_pred_ens.shape[TensorDim.GRID]
+        expected_shape = (grid_size, grid_size)
```
VeraChristina, line 176: This assumes a square shape, see other comment on require_square

training/src/anemoi/training/losses/graph_score_graph.py
```diff
@@ -0,0 +1,249 @@
...
+            raise ValueError(msg)
+        if num_src_nodes != num_dst_nodes:
+            msg = (
+                "Graph score losses require a grid-preserving loss graph with the same number "
+                "of source and target nodes."
+            )
+            raise ValueError(msg)
+
+    @staticmethod
+    def _row_normalize_weights(
```
VeraChristina, line 191: Is this consistent with `_row_normalize_matrix` on the ProjectionGraphProvider? (if it needs to be)

VeraChristina (commented):
Thanks for adding this! I have looked at the schemas, changes made to `training.py`, model layers, and existing losses; all of these changes are small and look good. The class hierarchy of the added losses, and the basic structure of the graph based losses make sense to me. A few minor comments/questions below.

I have not reviewed or tested all the different graph based losses. Perhaps it would be good for potential users to review/test the different added losses before merging, to see if the scores behave as expected and to check config options.

training/src/anemoi/training/losses/graph_edge_energy_score.py
ssmmnn11: Ok. Added to doc string. thx!

training/src/anemoi/training/losses/graph_edge_energy_score.py
ssmmnn11: yes, added a test as well.

training/src/anemoi/training/losses/graph_score_graph.py
ssmmnn11: yes all need it so I removed the switch. good spot! thx.

training/src/anemoi/training/losses/graph_score_graph.py
ssmmnn11: yes, I also added a test to check that they are consistent.

ssmmnn11:
[tests](https://github.com/ecmwf/anemoi-core/actions/runs/32253198619)

models/src/anemoi/models/utils/compile.py
```diff
@@ -84,7 +87,11 @@ def mark_for_compilation(model: Module, compile_config: DictConfig | None) -> Mo
             # It is just marked for JIT-compilation later
             # It will be compiled before its first forward pass
 
-            module.compile(**options)
+            compile_for_training = getattr(module, "compile_for_training", None)
```
cathalobrien, line 90: Nice idea, this is for compiling methods which aren't forward? does it work? the 4 lines after this insert the compiled module into the model, but I dont see the same for this. And is it used? I don't see a 'compile_for_training' method elsewhere in this PR

ssmmnn11: Yes, it is defined here:

https://github.com/ecmwf/anemoi-core/blob/09095d0f48eeefd44c5f9108fd857fd35982d149/training/src/anemoi/training/losses/graph_score_base.py#L95

ssmmnn11: compile_for_training Is maybe not the best name though

cathalobrien: ah nice, missed that.

VeraChristina (approved):
Thanks for addressing the feedback and for your contribution! As discussed, it's fairly isolated and it will be good to have the functionality so more people can use and test it.

PR #1271, opened 2026-07-27
feat(models): Allow torch compile to go through triton GT

cathalobrien:
## Description
Before if you tried to compile a region of code containing the triton GT you would get an error. This happened because the compiler was not able to trace through the triton GT. 
To enable this, you have to register [a custom pytorch operator](https://docs.pytorch.org/tutorials/advanced/python_custom_ops_functional.html). This does two things

1. defines '_fake' versions of forward and backward to be used during tracing. these just do the required memory allocations
```
@graph_transformer_attention.register_fake
def _graph_transformer_attention_fake(
    q: Tensor,
    k: Tensor,
    v: Tensor,
    e: Tensor,
    row: Tensor,
    colptr: Tensor,
    rowptr: Tensor,
    edge_ids: Tensor,
    edge_dst: Tensor,
) -> tuple[Tensor, Tensor, Tensor]:
    N_dst, H, C = q.shape
    out = torch.empty((N_dst, H, C), device=q.device, dtype=q.dtype)
    out_saved = torch.empty((N_dst, H, C), device=q.device, dtype=torch.float32)
    m = torch.empty((N_dst, H), device=q.device, dtype=torch.float32)
    return out, out_saved, m
```
2. Defines a '_setup_context' function which defines which tensors must be saved for the backward pass
```
def _graph_transformer_attention_setup_context(ctx, inputs, output):
    q, k, v, e, row, colptr, rowptr, edge_ids, edge_dst = inputs
    _out, out_saved, m = output
    ctx.save_for_backward(q, k, v, e, out_saved, m, row, colptr, rowptr, edge_ids, edge_dst)
```
Once after these changes, the triton GT can now be traced by torch compile.

I removed the triton GT module, and replaced it with an autograd function. The syntax changed from
```
 - self.conv = GraphTransformerFunction.apply
 + self.conv = graph_transformer_attention_conv
```
I updated the integration tests to use this new syntax, and added a test [recommended by pytorch](https://docs.pytorch.org/tutorials/advanced/python_custom_ops_functional.html#testing-python-custom-operators) to ensure the operator was registered correctly.

### DIFF
--- a/models/src/anemoi/models/layers/block.py
+++ b/models/src/anemoi/models/layers/block.py
@@ -51,7 +51,7 @@ from anemoi.models.triton.utils import is_triton_available
 from anemoi.utils.config import DotDict
 
 if is_triton_available():
-    from anemoi.models.triton.gt import GraphTransformerFunction
+    from anemoi.models.triton.gt import graph_transformer_attention_conv
 
 LOGGER = logging.getLogger(__name__)
 
@@ -613,7 +613,7 @@ class GraphTransformerBaseBlock(BaseBlock, ABC):
 
         if self.graph_attention_backend == "triton":
             LOGGER.info(f"{self.__class__.__name__} using triton graph attention backend.")
-            self.conv = GraphTransformerFunction.apply
+            self.conv = graph_transformer_attention_conv
         else:
             self.conv = GraphTransformerConv(out_channels=self.out_channels_conv)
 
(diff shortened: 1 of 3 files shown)


cathalobrien:
I did some n320-L6 GT tests comparing this branch to main. there might be a tiny slowdown, but very minor. likely not able to measure that accurately.
```
feat/triton_gt_custom_op: 1.66265 it/s
 
main: 1.66889 it/s
 ```

japols (approved):
looks good! thanks

Issue #1276, opened 2026-07-28
`anemoi-training profile` crashes with `TypeError` when `progress_bar` is unset


dietervdb-meteo:
### What happened?

Running anemoi-training profile crashes during callback setup when the config has no diagnostics.progress_bar block:

TypeError: 'NoneType' object does not support item assignment  File ".../train/profiler.py", line 281, in callbacks    self.config.diagnostics.progress_bar["_target_"] = ...
The schema default for diagnostics.progress_bar is None (ProgressBarSchema | None, default None), so any config that omits the block — which is valid and works fine for train — makes the profiler fail immediately.

Root cause: AnemoiProfiler.callbacks mutates config.diagnostics.progress_bar["_target_"] to force ProfilerProgressBar before calling super().callbacks. It assumes the block is always a populated mapping. train tolerates None via the shared, null-safe _get_progress_bar_callback; the profiler pre-empts that guard by mutating the config first.

Expected: profile should behave like train and not require a progress_bar block to be present.

### What are the steps to reproduce the bug?

Run `anemoi-training profile` on a config with progress bar unset.

### Version

current main

### Platform (OS and architecture)

irrelevant

### Relevant log output

```shell

```

### Organisation

RMI

dietervdb-meteo:
Fixed in #1277

PR #1277, opened 2026-07-28
fix(training): set default progress bar for profiler

dietervdb-meteo:
## Description
`AnemoiProfiler.callbacks` forced its own progress bar by assigning into `config.diagnostics.progress_bar["_target_"]` before delegating to `super().callbacks`. When the config omits the `progress_bar` block (pydantic schema default `None`), this raises an error. The block is optional and valid to omit when running `anemoi-training train` so `profile` should too.

Fix:
Stop mutating the config. Instead, let `super().callbacks` resolve the progress bar through the existing shared null-safe guard (`_get_progress_bar_callback`), then filter out whatever `ProgressBar` it produced and substitute the profiler's own `ProfilerProgressBar`. This keeps exactly one `ProgressBar` for Lightning (satisfies the Lightning ≥ 2.6 single-bar requirement) without duplicating the `None`-handling logic.

#1276

### DIFF
--- a/training/src/anemoi/training/train/profiler.py
+++ b/training/src/anemoi/training/train/profiler.py
@@ -278,10 +278,12 @@ class AnemoiProfiler(AnemoiTrainer):
 
     @cached_property
     def callbacks(self) -> list[pl.callbacks.Callback]:
-        self.config.diagnostics.progress_bar["_target_"] = (
-            ProfilerProgressBar.__module__ + "." + ProfilerProgressBar.__name__
-        )
         callbacks = super().callbacks
+
+        # Force the profiler's own progress bar.
+        callbacks = [c for c in callbacks if not isinstance(c, pl.callbacks.ProgressBar)]
+        callbacks.append(ProfilerProgressBar())
+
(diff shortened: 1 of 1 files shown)


dietervdb-meteo:
Running integration tests just to be sure (since the profiler progress bar is used to measure speed? and hence might affect benchmarks?): https://github.com/ecmwf/anemoi-core/actions/runs/30371644750
 :white_check_mark:

dietervdb-meteo:
I realized this implementation behaves slightly differently to what was there before in two edge cases:
1.custom `refresh_rate`: If a config set e.g. `progress_bar.refresh_rate: 10`, the old code would carry that into `ProfilerProgressBar(refresh_rate=10)`. The new code always uses the default 1. I guess that would make sense for the profiler, is this indeed desirable behavior?
2. `enable_progress_bar: False` in the config is now ignored by the profiler. But I guess that is actually desired behavior?

@cathalobrien, @japols?

japols (approved):
Hi @dietervdb-meteo I think this behaviour is fine, thanks for the fix!

PR #1278, opened 2026-07-30
perf(training): disable broadcast buffers

cathalobrien:
## Description
by default when using DDP all the registered buffers are broadcast from rank 0 to all other ranks at the start of every iteration. this synchronisation is required for batchNorm, but we dont use that. so it should be fine to disable.

when sharding over multiple nodes, the broadcasts start to hurt

**16 gpus per model, 1024c n320-l6 forward pass. the broadcast is 10% of the forward pass**
Screenshot 2026-07-30 at 08 58 47

after this PR, the broadcasts are gone. Throughput has increased from 6.15it/s to 6.27it/s
Screenshot 2026-07-30 at 09 20 29

### DIFF
--- a/training/src/anemoi/training/distributed/strategy.py
+++ b/training/src/anemoi/training/distributed/strategy.py
@@ -93,6 +93,7 @@ class BaseDDPStrategy(DDPStrategy):
         num_gpus_per_model: int,
         read_group_size: int,
         use_local_synchronization: bool = True,
+        broadcast_buffers: bool = False,
         **kwargs: dict,
     ) -> None:
         """Initialise the distributed strategy.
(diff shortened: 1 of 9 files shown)


training/src/anemoi/training/distributed/strategy.py
```diff
@@ -108,7 +108,7 @@ def __init__(
         **kwargs : dict
             Additional keyword arguments.
         """
-        super().__init__(**kwargs)
+        super().__init__(**kwargs, broadcast_buffers=False)
```
ssmmnn11: I think we should make this configurable.

ssmmnn11 (dismissed):
Great that you  spotted the performance implication of this!

training/src/anemoi/training/distributed/strategy.py
cathalobrien: good idea, I added the `training.strategy.broadcast_buffers` key, defaulting to `False`

training/src/anemoi/training/distributed/strategy.py
```diff
@@ -105,10 +106,12 @@ def __init__(
             Number of GPUs per reader group.
         use_local_synchronization : bool, optional
             Use synchronization local to the group when creating process groups.
+        broadcast_buffers : bool, optional
+            Whether to broadcast module buffers at the start of each iteration. Defaults to False.
         **kwargs : dict
             Additional keyword arguments.
         """
-        super().__init__(**kwargs)
```
HCookie, line 111: Given this passes up kwargs, the explicit is not strictly necessary, but I guess is helpful to showcase it is a setting

HCookie (dismissed):

HCookie (approved):
LGTM

HCookie (approved):

PR #1279, opened 2026-07-30
perf(training): minimise broadcasts from lightning checkpoints

cathalobrien:
## Description
When we use time based checkpointing, pytorch lightning adds a global broadcast across all ranks at the end of each batch. When running over many gpus, this can become noticeable 
```
# this will enable the broadcasts each step
diagnostics:
  checkpoint:
    every_n_minutes:
      save_frequency: 30
```
The purpose of the constant broadcasting is to keep the ranks in sync, so that the non-zero ranks know when rank 0 is checkpointing.

for an n320-o96 model across 16 GPUs you can see the broadcasts in light green, before the forward in purple, it takes up about 10% of the forward pass
Screenshot 2026-07-30 at 09 17 47
If I disable the broadcast, throughput increases from 6.27 to 6.4it/s

[Here](https://github.com/Lightning-AI/pytorch-lightning/blob/f9751d2726a1e838fabe5faf90c138112beeb57c/src/lightning/pytorch/callbacks/model_checkpoint.py#L435) is the underlying pytorch lightning code

one possible way to fix this would be to only do the broadcast every N steps. This means that your checkpoint wouldn't be produced exactly at frequency T, instead it would be written at the first multiple of step N after time T. 
I have suggested N=10 in this PR

the integration tests [pass](https://github.com/ecmwf/anemoi-core/actions/runs/30527146377/job/90822543998), except for an unrelated error

### DIFF
--- a/training/src/anemoi/training/config/diagnostics/evaluation.yaml
+++ b/training/src/anemoi/training/config/diagnostics/evaluation.yaml
@@ -16,7 +16,7 @@ debug:
 enable_checkpointing: True
 checkpoint:
   every_n_minutes:
-    save_frequency: 30 # Approximate, as this is checked at the end of training steps
+    save_frequency: 30 # Approximate, as this is checked at the end of every 10 training steps
     num_models_saved: 3 # If set to k, saves the 'last' k model weights in the training.
 
   every_n_epochs:
--- a/training/src/anemoi/training/config/diagnostics/evaluation.yaml
+++ b/training/src/anemoi/training/config/diagnostics/evaluation.yaml
@@ -16,7 +16,7 @@ debug:
 enable_checkpointing: True
 checkpoint:
   every_n_minutes:
-    save_frequency: 30 # Approximate, as this is checked at the end of every 10 training steps
+    save_frequency: 30 # Approximate, as this is checked at the end of every 50 training steps
     num_models_saved: 3 # If set to k, saves the 'last' k model weights in the training.
 
   every_n_epochs:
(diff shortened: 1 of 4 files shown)


cathalobrien:
I tested some checkpointing and it still produces checkpoints:
```yaml
diagnostics:
  enable_checkpointing: True
  checkpoint:
    every_n_minutes:
      save_frequency: 1 # Approximate, as this is checked at the end of 10 training steps
      num_models_saved: -1
    every_n_train_steps:
      save_frequency: 100
      num_models_saved: -1
```
```
(v3-bm-ag) [naco@ag6-411 v3-bm]$ ls /lus/h2resw01/scratch/naco/aifs/o96/checkpoint/f9798ecc-15c8-442e-8f0e-70c01d151d22/
anemoi-by_epoch-epoch_000-step_000250.ckpt  anemoi-by_time-epoch_000-step_000250.ckpt		  inference-anemoi-by_time-epoch_000-step_000210.ckpt
anemoi-by_step-epoch_000-step_000100.ckpt   inference-anemoi-by_epoch-epoch_000-step_000250.ckpt  inference-anemoi-by_time-epoch_000-step_000250.ckpt
anemoi-by_step-epoch_000-step_000200.ckpt   inference-anemoi-by_step-epoch_000-step_000100.ckpt   inference-last.ckpt
anemoi-by_step-epoch_000-step_000250.ckpt   inference-anemoi-by_step-epoch_000-step_000200.ckpt   last.ckpt
anemoi-by_time-epoch_000-step_000210.ckpt   inference-anemoi-by_step-epoch_000-step_000250.ckpt
```

cathalobrien:
I checked the traces and confirmed that the broadcast call is now only happening every 10 iterations

HCookie:
Even 10 feels too frequent, if the speed is 6.24 it/s is that not about every 2 seconds? 20 or 50 seem okay if we are only looking at saving at most each minute, and we already werent at the exact minute due to running only at every batch end

HCookie (dismissed):
LGTM

mc4117 (approved):
LGTM!

PR #1282, opened 2026-07-30
perf(training): all-to-all interface for torch.compile() 

cathalobrien:
## Description
Makes torch compile able to trace through all-to-all-transpose(). this is required to run the compiled processor across multiple nodes

### DIFF
--- a/models/src/anemoi/models/distributed/primitives.py
+++ b/models/src/anemoi/models/distributed/primitives.py
@@ -294,89 +294,91 @@ def _alltoallwrapper(output_list: list, input_list: list, group: ProcessGroup):
                 reqs.append(dist.isend(input_list[j], group_dst=j, group=group))
                 reqs.append(dist.irecv(output_list[j], group_src=j, group=group))
             else:
-                output_list[rank].copy_(input_list[rank])
+                output_list[rank] = input_list[rank]
         for req in reqs:
             req.wait()
     else:
         dist.all_to_all(output_list, input_list, group=group)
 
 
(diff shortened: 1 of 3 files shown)


models/src/anemoi/models/distributed/primitives.py
```diff
@@ -370,6 +357,76 @@ def _alltoall_transpose(
...
+    dim_split = dim_split % ndim
+    dim_concat = dim_concat % ndim
+    out_shape = list(input_.shape)
+    out_shape[dim_split] = split_sizes[myrank]
+    out_shape[dim_concat] = sum(concat_sizes)
+    return input_.new_empty(out_shape)
+
+
+def _resolve_group_name(group: Optional[ProcessGroup]) -> str:
+    if group is None:
```
japols, line 352: I think we want to keep the behaviour of `group is None` means single gpu i.e. no communication

japols (changes requested):
nice! one question: wouldn't it make more sense to write the interface for the `_alltoallwrapper` instead of just the transpose?

models/src/anemoi/models/distributed/primitives.py
cathalobrien: yeah I agree. the rest of the file uses
```
comm_size = dist.get_world_size(group=group)
    if comm_size == 1:
        return grad_output
```
In this case, group is None which resolves to the global comm group containing all GPUs (1 here). `_resolve_group_name` also returns the global comm group

japols (approved):
Looks good now, thanks!

PR #1283, opened 2026-07-30
fix(training): handle NaN masks correctly with sharding

ssmmnn11:
When training with sharded batches, each GPU created NaN masks for its own part of the grid. These masks were stored as PyTorch buffers, so distributed training copied the masks from rank 0 to the other ranks. Those ranks then applied masks belonging to different spatial locations, which gave the loss incorrect weights.

The fix stores batch dependent masks as runtime tensors instead of buffers. They are no longer synchronised between ranks or saved in checkpoints. Each rank now keeps the masks created from its own data shard, while inference recomputes them from its current input.


sahahner:
Thank you for finding and fixing that bug. 

I ran a few tests, and indeed the model-sharded imputer is using wrong NaN locations on main. I checked with the release we used to train v2, and there the sharding of the imputer worked without any problems, so the operational model is fine. 

The moving of the masks between devices is working well for the callbacks, which have been enabled for all tests.

For reference, here are a few model-sharded and unsharded runs using different versions of the code: https://mlflow.ecmwf.int/#/experiments/795/runs?searchFilter=&orderByKey=attributes.start_time&orderByAsc=false&startTime=ALL&lifecycleFilter=Active&modelVersionFilter=All+Runs&datasetsFilter=W10%3D

It is difficult for me to trace back what happened in between the releases that broke this feature, but I was thinking that it might be useful to add the imputer to the multi-GPU tests. Any thoughts on that?

ssmmnn11:
> Thank you for finding and fixing that bug.
> 
> I ran a few tests, and indeed the model-sharded imputer is using wrong NaN locations on main. I checked with the release we used to train v2, and there the sharding of the imputer worked without any problems, so the operational model is fine.
> 
> The moving of the masks between devices is working well for the callbacks, which have been enabled for all tests.
> 
> For reference, here are a few model-sharded and unsharded runs using different versions of the code: https://mlflow.ecmwf.int/#/experiments/795/runs?searchFilter=&orderByKey=attributes.start_time&orderByAsc=false&startTime=ALL&lifecycleFilter=Active&modelVersionFilter=All+Runs&datasetsFilter=W10%3D
> 
> It is difficult for me to trace back what happened in between the releases that broke this feature, but I was thinking that it might be useful to add the imputer to the multi-GPU tests. Any thoughts on that?

I hope once we have a set of sharding tests of the full model, these bugs will become less likely. We currently don't have tests that would have highlighted the issue. In terms of releases: the bug was triggered by a series of unfortunate events / our design. What triggered it in the end was #448 I think.

sahahner:
Okay, makes sense. Thank you for the answer. 

Fine for me to cover the sharded imputer in tests that will be implemented later.

HCookie (commented):
The `scaler` implementation looks fine to me, it's been a bit since I looked at the code, and am concerned it's getting a bit out of hand, perhaps a refactor is due soon.

sahahner:
A test on the imputer in inference was also successful, both for using model sharding and not using model sharding.

sahahner (approved):
LGTM

PR #1288, opened 2026-08-03
fix: forecaster and tendency statistics

ssmmnn11:
remove workaround that was introduced ->  breaks transport models

##  Additional notes ##

### DIFF
--- a/training/src/anemoi/training/data/datamodule.py
+++ b/training/src/anemoi/training/data/datamodule.py
@@ -79,14 +79,10 @@ class AnemoiDatasetsDataModule(pl.LightningDataModule):
         tendency_delta = getattr(self.task, "tendency_delta", None)
         tendency_delta_str = frequency_to_string(tendency_delta) if tendency_delta is not None else None
 
-        # When no tendency_delta is set, use the output step frequency (first lead
-        # time) for all offsets so tendencies represent step-to-step differences.
-        step_freq = lead_times[0]
-
         stats_by_dataset: dict[str, dict | None] = {}
         for dataset_name, dataset in self.ds_train.data_readers.items():
             stats_by_lead = {
-                lead_time: dataset.statistics_tendencies(tendency_delta_str or step_freq) for lead_time in lead_times
+                lead_time: dataset.statistics_tendencies(tendency_delta_str or lead_time) for lead_time in lead_times
             }
             if all(stats is None for stats in stats_by_lead.values()):
                 stats_by_dataset[dataset_name] = None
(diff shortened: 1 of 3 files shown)


mishooax (dismissed):

training/src/anemoi/training/data/datamodule.py
```diff
@@ -79,14 +79,10 @@ def statistics_tendencies(self) -> dict[str, dict | None] | None:
...
 
-        # When no tendency_delta is set, use the output step frequency (first lead
-        # time) for all offsets so tendencies represent step-to-step differences.
-        step_freq = lead_times[0]
-
         stats_by_dataset: dict[str, dict | None] = {}
         for dataset_name, dataset in self.ds_train.data_readers.items():
             stats_by_lead = {
-                lead_time: dataset.statistics_tendencies(tendency_delta_str or step_freq) for lead_time in lead_times
+                lead_time: dataset.statistics_tendencies(tendency_delta_str or lead_time) for lead_time in lead_times
```
mc4117, line 85: This was introduced because in the temporal downscaler when we have [1h, 2h, 3h, 4h, 5h] offsets, I want it to use the 1h tendency statistics for all outputs rather than the 4h one for example. I think what you are proposing would break this again?

mc4117 (changes requested):
This would now break the temporal downscaler use case. Would be good to try and find a workaround for both cases?

training/src/anemoi/training/data/datamodule.py
VeraChristina: I think you introduced the tendency_delta_str for the case where we have a constant delta offset, so that behaviour should be preserved. (Do we need tendency statistics for the temporal downscaler? Are they used outside of tendency transport models (which are forecasting specific as of now)?)

mc4117: We use tendency statistics in the temporal downscaler (it's tested in the benchmark test too). I have now updated to use the tendency_delta and will check if the benchmark tests now pass

ssmmnn11: [benchmark tests](https://github.com/ecmwf/anemoi-core/actions/runs/30816441425) complete without the config change already so will revert and you can do a follow up

ssmmnn11:
> This would now break the temporal downscaler use case. Would be good to try and find a workaround for both cases?

I think we are fine, see [tests](https://github.com/ecmwf/anemoi-core/actions/runs/30816441425).

mc4117:
> > This would now break the temporal downscaler use case. Would be good to try and find a workaround for both cases?
> 
> I think we are fine, see [tests](https://github.com/ecmwf/anemoi-core/actions/runs/30816441425).

I'm surprised the benchmark tests are passing as they were failing in a different version when I implemented this, but maybe some defaults changed. I think they're picking up the wrong value?

mc4117 (approved):

PR #1291, opened 2026-08-03
perf(training): reuse shared env for integration test 

cathalobrien:
## Description

This PR reuses a shared venv across the workflow, benchmark and correctness integration tests. It moves the venv from $SCRATCH to $PERM and builds it on a cpu job (fractional queue). so this saves about 30 minutes of GPU time.

before:
Screenshot 2026-07-28 at 10 35 23
after:
Screenshot 2026-07-28 at 10 35 06

the env is deleted only once all 3 tests have passed. this allows you to rerun failing tests. When you rerun a test, git pull is used inside anemoi-core to fetch the latest commit.

there is 1 venv per integration test run.

the integration tests [pass](https://github.com/ecmwf/anemoi-core/actions/runs/30817207601)

### DIFF
--- a/.github/workflows/integration-tests-hpc.yml
+++ b/.github/workflows/integration-tests-hpc.yml
@@ -22,18 +22,66 @@ jobs:
         name: check latest commit is less than a day
         if: ${{ github.event_name == 'schedule' }}
         run: test -z $(git rev-list  --after="24 hours"  ${{ github.sha }}) && echo "::set-output name=should_run::false"
+  build-env:
+    name: build-env
+    runs-on: hpc
+    timeout-minutes: 120
+    needs: check_date
+    if: ${{ needs.check_date.outputs.should_run != 'false' }}
+    steps:
+      - uses: ecmwf-actions/reusable-workflows/ci-hpc-generic@v2
+        with:
+          template: |
+            env | grep SLURM # helps with debugging
+            set -eux
+            module load python3/3.12.9-01
+            # Shared per-commit working dir on $PERM. The clone + virtualenv are
+            # built once here and reused by every benchmark job, then removed by
+            # the cleanup-env job. Keyed on the commit sha so parallel runs of
+            # different commits do not collide.
+            WORKDIR="$PERM/TESTING/integration-venvs/${{ github.sha }}"
+            REPO_NAME=anemoi-core
+            REPO_BRANCH=${{ github.head_ref || github.ref_name }}
+            # Build off /dev/shm: the default $TMPDIR is RAM-backed tmpfs and counts
+            # against the SLURM --mem cgroup, so installing torch + the nvidia CUDA
+            # stack into it stalls. Redirect to a disk-backed per-job scratch dir.
+            export TMPDIR="$SCRATCH/ci-tmp/$SLURM_JOB_ID"
+            mkdir -p "$TMPDIR"
+            trap 'rm -rf "$TMPDIR"' EXIT
+            # Start from a clean shared dir so a rebuild is always reproducible.
+            rm -rf "$WORKDIR"
+            mkdir -p "$WORKDIR"
+            cd "$WORKDIR"
+            git clone -b $REPO_BRANCH https://${{ secrets.GH_REPO_READ_TOKEN }}@github.com/ecmwf/${REPO_NAME}.git
+            python3 -m venv .venv
+            source .venv/bin/activate
+            pip install --upgrade pip
+            # Install the superset of extras required by all benchmark jobs
+            # (profile is only needed by test_benchmark). Editable installs point
+            # at the shared clone, so the clone must persist until cleanup.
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests-hpc.yml
```diff
@@ -13,27 +13,72 @@ jobs:
...
+      date: ${{ steps.date.outputs.date }}
     steps:
       - uses: actions/checkout@v4
       - name: print latest_commit
         run: echo ${{ github.sha }}
+      - id: date
+        name: compute shared venv date key
+        # Computed once here and shared with every job so all jobs agree on the
+        # same WORKDIR even if a long-running test crosses midnight.
+        run: echo "date=$(date +%F)" >> "$GITHUB_OUTPUT"
```
VeraChristina: Are we using this? Looks like we use the `github.run_id` below

.github/workflows/integration-tests-hpc.yml
```diff
@@ -74,33 +117,43 @@ jobs:
...
-            git clone -b $REPO_BRANCH https://${{ secrets.GH_REPO_READ_TOKEN }}@github.com/ecmwf/${REPO_NAME}.git
-            cd $REPO_NAME
-            python3 -m venv .venv
-            source .venv/bin/activate
-            pip install --upgrade pip
-            pip install -e ./training[${{ matrix.pip_extras }}] -e ./models[all,tests] -e ./graphs[all,tests]
+            # Reuse the shared clone + virtualenv built by the build-env job.
+            WORKDIR="$PERM/TESTING/integration-venvs/${{ github.run_id }}"
+            cd "$WORKDIR/anemoi-core"
+            git pull # update to the latest commit on the branch
```
VeraChristina, line 118: Do we want to do an additional git pull here? I guess it makes sense if we want to re-run jobs on later commits and not re-create the venv. But it also means venv and re-runs can diverge, and it means we don't know on which commits we run the different matrix jobs (if the user commits between the build-venv job and the git pull in the different matrix jobs). (Also, I'm not sure how concurrent git pulls on the same repo behave)

.github/workflows/integration-tests-hpc.yml
```diff
@@ -74,33 +117,43 @@ jobs:
...
           troika_user: ${{ secrets.HPC_CI_INTEGRATION_USER }}
-
+  cleanup-env:
+      name: cleanup-env
+      runs-on: hpc
+      timeout-minutes: 30
+      needs: [check_date, integration-tests]
+      # Only clean up when every integration test succeeded.
+      # if any test fails or is cancelled then failed/cancelled
+      # then leave the venv on filesystem to allow tests to be rerun.
```
VeraChristina, line 137: This is good for debugging. Have we thought about cleaning up venvs from failed runs once in a while?

VeraChristina (commented):
Thanks for adding this! A few small comments below.
My main comment is that this restricts the options we have to enable running only one of the matrix jobs. But with the GPU time this saves that might be a good trade off, especially since smoke tests can still be run locally.

.github/workflows/integration-tests-hpc.yml
cathalobrien: I suppose some manual cleaning will have to be done every once in a while e.g. if a test is left in a failing state, the venv will never be cleaned up. maybe I can set up something so that when the test is deleted the venv is closed.

probably automating the cleaning by date is overcomplicating. We have a lot of space on the filesystem where I am building the venvs (for context, on my personal allocation here I have 83 venvs lol and have only used 13% of the allocation)

.github/workflows/integration-tests-hpc.yml
mcocdawc: Fully agree with Vera.
I would also not do a `git pull` here. (concurrent `git pull` should be a non issue)

If we use the shared environment, then individual jobs should not change it.

.github/workflows/integration-tests-hpc.yml
cathalobrien: ah sorry, that was left over from a prev commit where I made 1 venv per day and shared it across all tests. that was too far :p

I'll delete that, good catch

.github/workflows/integration-tests-hpc.yml
VeraChristina: As discussed offline, this doesn't modify current behaviour. However, it would be good to remove it once we can run select matrix jobs on this workflow. Can you document this in an issue to be a follow up?

cathalobrien: on discussion we decided to keep the git pull temporarily until we have the ability to run individual tests. and there will be an open issue to track the fact that individual tests can run on different commits if you rerun, which we all agree shouldnt be the case. when we can run individual tests we will remove the git pull

mcocdawc (approved):
cool cool, nice nice

PR #1293, opened 2026-08-03
fix: tendency temporal downscaler

mc4117:
## Description

Stores `output_timestep` as a string attribute on `TemporalDownscaler` and sets `stdev_tendency.timestep: ${task.output_timestep}` in both temporal downscaler configs so the tendency scaler always uses the per-step (output resolution) statistics rather than cumulative lead-time statistics.

Removes the `tendency_delta` workaround from `datamodule.py` — all tasks now use cumulative per-lead-time tendency statistics consistently. Base scaler configs declare `timestep: null` to allow config-level overrides without OmegaConf struct errors.

Using `stdev_tendency` with `TemporalDownscaler` previously relied on a `tendency_delta` workaround in the datamodule to force all lead times to use the same 1h statistics. The scaler `timestep` is now set explicitly in the config via `${task.output_timestep}`, making the intent clear and removing the hidden workaround.

Follows on from #1288.

----
📚 Documentation preview 📚: https://anemoi-training--1293.org.readthedocs.build/en/1293/

----
📚 Documentation preview 📚: https://anemoi-graphs--1293.org.readthedocs.build/en/1293/

----
📚 Documentation preview 📚: https://anemoi-models--1293.org.readthedocs.build/en/1293/


training/src/anemoi/training/losses/scalers/variable_tendency.py
```diff
@@ -64,10 +70,13 @@ def __init__(
...
-            lead_times = list(lead_times)
-            assert lead_times, "lead_times must be a non-empty list"
+            assert list(lead_times), "lead_times must be a non-empty list"
             if timestep is None:
-                timestep = lead_times[0]
+                tendency_delta = getattr(task, "tendency_delta", None)
+                if tendency_delta is not None:
+                    timestep = frequency_to_string(tendency_delta)
+            if timestep is None:
+                timestep = next(iter(lead_times))
```
mc4117: This change from lead_times[0] to this was suggested by the pre-commit hook but I'm not convinced it's right?

training/src/anemoi/training/config/temporal_downscaler.yaml
```diff
@@ -43,6 +43,15 @@ data:
               - "sin_local_time"
               - "insolation"
 
+training:
```
mc4117: I wouldn't add this, as this is something specific to our set-up for AIFS. ALso we only use the stdev_tendency

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -67,3 +67,10 @@ model:
 
 training:
   ensemble_size_per_device: 2
+  scalers:
```
mc4117, line 73: see comment on temporal_downscaler.yaml

training/tests/unit/losses/test_loss_scaling.py
```diff
@@ -415,29 +415,6 @@ def test_tendency_scaler_uses_configured_timestep(
     assert scalers["additional_scaler"][1][variable_idx] == expected_scaling
 
 
-def test_tendency_scaler_timestep_overrides_all_lead_times() -> None:
```
mc4117: I added this to make sure it doesn't break again. It seemed like a minor test to add?

training/tests/unit/tasks/test_temporal_downscaler.py
```diff
@@ -80,9 +80,3 @@ def test_temporal_downscaler_input_offsets_are_boundary_pair() -> None:
     """Input offsets are always [0h, input_timestep] regardless of output settings."""
     task = TemporalDownscaler(input_timestep="6h", output_timestep="2h")
     assert task._input_offsets == [datetime.timedelta(0), datetime.timedelta(hours=6)]
-
-
-def test_temporal_downscaler_stores_output_timestep_string() -> None:
```
mc4117: again I added this to check that it doesn't break again although I agree this test is probably over the top

training/src/anemoi/training/config/training/scalers/global.yaml
```diff
@@ -35,11 +35,9 @@ datasets:
     # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
     stdev_tendency:
       _target_: anemoi.training.losses.scalers.StdevTendencyScaler
-      timestep: null
```
mc4117: this is required otherwise you will get schema errors

training/src/anemoi/training/config/training/scalers/lam.yaml
```diff
@@ -38,11 +38,9 @@ datasets:
     # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
     stdev_tendency:
       _target_: anemoi.training.losses.scalers.StdevTendencyScaler
-      timestep: null
```
mc4117: required otherwise you will get errors

training/src/anemoi/training/config/training/scalers/multi.yaml
```diff
@@ -33,11 +33,9 @@ datasets:
     # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
     stdev_tendency:
       _target_: anemoi.training.losses.scalers.StdevTendencyScaler
-      timestep: null
```
mc4117: required otherwise you will get errors

training/src/anemoi/training/config/training/scalers/stretched.yaml
```diff
@@ -39,11 +39,9 @@ datasets:
     # if using this option 'variable_loss_scalings' should all be set close to 1.0 for prognostic variables
     stdev_tendency:
       _target_: anemoi.training.losses.scalers.StdevTendencyScaler
-      timestep: null
```
mc4117: required otherwise you will get errors trying to override this config

ssmmnn11: Where do you get the error? If you want to override it in the cli you can just use:

+training.scalers...timestep=2H

training/tests/unit/tasks/test_temporal_downscaler.py
ssmmnn11: I removed these because they were not used.

training/src/anemoi/training/config/temporal_downscaler.yaml
ssmmnn11: I think we ultimately have to make a decision, secure it like this, or let the user activate it how they like. We can also switch it off per default.

cathalobrien (approved):
nice work LGTM

mc4117:
integration tests pass https://github.com/ecmwf/anemoi-core/actions/runs/31095746095

PR #1295, opened 2026-08-04
feat: add label-based gating + selective running of individual tests

### LINKED ISSUE #485: Revisit ci for integration tests in training

There have been a couple of discussion about this. This is to capture some of the questions raised and start gathering the solutions proposed.

Should integration tests in training run on PRs before merging? Can we use labels to mark PRs that require integration tests before being merged? How much of this can we / do we need to automate?

mcocdawc:
## Description

Run the integration tests, if the label `"run-expensive-tests"` is attached to a PR.
In order to merge one of the labels`"run-expensive-tests"` or `"expensive-tests-not-needed"` have to be present.

After "run-expensive-tests" has been added, the tests are run upon every commit to a PR.

In addition, the code makes it easier to run individual tests manually and moves the `pip freeze` into the construction of the environment (where it belongs now, since the `git pull` in every job is gone.)

Note: The check for the labels reports failure a bit differently, than the check for the ATS labels. Namely it reports green, if everything is ok and pending otherwise.
It should block PRs from being merged, but not report a red CI.
This means that people can have the labels unset while iterating in the early stages of a PR and don't get spurios CI failures.

- [ ] *After merging* block merge of PRs upon the commit status of the check. This has to be done by someone with admin permissions.

##  Additional notes ##


.github/workflows/run-integration-test-hpc.yml
```diff
@@ -0,0 +1,41 @@
...
+      pytest_cmd:
+        required: true
+        type: string
+      timeout_minutes:
+        required: true
+        type: number
+      sbatch_options:
+        required: true
+        type: string
+
```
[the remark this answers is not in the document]
mcocdawc: secrets: inherit makes declaration unnecessary; the secrets resolve correctly.

.github/workflows/pr-label-expensive-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - labeled
+      - unlabeled
+
```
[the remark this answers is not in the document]
mcocdawc: done and verified that pull_request_target is actually safe here.

.github/workflows/pr-label-expensive-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+            const ok = isBot || decided;
+            await github.rest.repos.createCommitStatus({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              sha: pr.head.sha,
+              context: 'expensive-tests-label',
+              state: ok ? 'success' : 'pending',
+              description: ok
+                ? "Expensive-tests decision recorded"
+                : "Add 'run-expensive-tests' or 'expensive-tests-not-needed' to merge",
```
HCookie, line 59: From a semantics point of view, I feel as if `expensive-tests-needed` might be better, thoughts?

```suggestion
                : "Add 'expensive-tests-needed' or 'expensive-tests-not-needed' to merge",
```

.github/workflows/pr-label-expensive-tests.yml
```diff
@@ -0,0 +1,61 @@
+# Gate that a PR declares whether the expensive integration tests are needed,
+# via exactly one of these labels:
+# - "run-expensive-tests": the expensive HPC tests run on every push (see
```
cathalobrien, line 3: running every push could be excessive. I thought the idea was to require the user to run them once before merging? If they push, then requiring them to run again is reasonable.

but I'm fine to wait and see how this turns out.

.github/workflows/pr-label-expensive-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+#   integration-tests-hpc.yml).
+# - "expensive-tests-not-needed": the author has opted out.
+#
+# Rather than *failing* when a label is missing (a red X that desensitises
```
cathalobrien, line 7: this comment isn't clear to me, and feels out of place. should it be in documentation instead maybe? 

I suppose some docs explaining the new process would also be good

cathalobrien (commented):
really nice work. excited to try the individual tests myself. I think we also need some documentation explaining the new process

.github/workflows/pr-label-expensive-tests.yml
VeraChristina: I think it's not clear that the expensive tests run on each commit once you put an `expensive-tests-needed` label on a PR. (We'll need to educate people about this anyways but I think it would be good for this to be explicit in the label, otherwise we might get many runs.)

.github/workflows/pr-label-expensive-tests.yml
mcocdawc: > running every push could be excessive

It **could** be excessive, but premature optimization is the root of all evil. :-D 

If we see in the future that our computing resources are strained too much, then we can add some remedies. For example any run could remove the label, so it has to be manually added again to trigger a test.

.github/workflows/pr-label-expensive-tests.yml
mcocdawc: We @cathalobrien discussed in person, that for this case ideally the whole documentation lies in the web UI and error message.

However, unfortunately some of this job, can only be seen after merging, as `pull_request_target` workflows only trigger from the **base** branch.

If there is anything unclear after merging, I am ready to revisit and add to the documentation.

cathalobrien (approved):
very nice, well done

Issue #698, opened 2025-11-25
Qhull Voronoi error when enabling lam_area_weight on stretched grid


cerodell:
## Qhull Voronoi error when enabling `lam_area_weight` on stretched grid

Hi anemoi team :)

I am running into a persistent error when enabling `lam_area_weight` on a stretched grid configuration. The issue occurs when generating the graph for a stretched grid model that uses a 2km WRF domain in a Lambert conformal projection as the LAM dataset, combined with a global ERA5 N320 dataset as the forcing dataset.

### Description of the problem

When I enable the following block in both `stretched_grid.yaml` and `training.scale.stretched.yaml`:

**stretched_grid.yaml**
```yaml
## TODO get this working, could not define
lam_area_weight:
  _target_: anemoi.graphs.nodes.attributes.MaskedPlanarAreaWeights
  mask_node_attr_name: cutout_mask
  norm: unit-max
```

**training.scale.stretched.yaml**
```yaml
## TODO get this working, could not define
mse_inside_lam:
  _target_: anemoi.training.losses.MSELoss
  scalers: ['lam_node_weights']
  ignore_nans: True
```

The graph creation step fails with a Qhull Voronoi error during the area weight computation. The traceback indicates numerical instability when constructing the Voronoi diagram from the stretched grid lat lon coordinates.

Error
```bash
scipy.spatial._qhull.QhullError: QH6229 qhull input error:
51 attempts to construct a convex hull with joggled input.
Increase joggle above 'QJ0.087' or modify qh_JOGGLE parameters...
```

This happens consistently whenever `lam_area_weight` is enabled.
If I comment out this block in both YAML files, Anemoi runs normally and training proceeds.

### Current workaround

I must comment out the lam_area_weight section in:
- stretched_grid.yaml
- training.scale.stretched.yaml

Otherwise the graph fails during Voronoi area weight computation.

### My setup
- 2 km WRF Lambert conformal as dataset
- ERA5 N320 global as forcing dataset

### Environment
```txt
=== Python and System ===
Python version: 3.12.12
Platform: Linux-4.18.0

=== Anemoi Packages ===
Anemoi Training: 0.7.0
Anemoi Datasets: 0.5.28
Anemoi Models: 0.10.0
Anemoi Graphs: 0.7.2

=== PyTorch and CUDA ===
PyTorch version: 2.9.1+cu128
CUDA available: False
```

### Request
Could you advise whether:
- The WRF Lambert projection might be causing the instability
[...]

PR #1298, opened 2026-08-04
fix(graphs): replace qhull joggle with merged facets in PlanarAreaWeights

### LINKED ISSUE #690: Increase in node joggle required when computing Voronoi region in Planar Area Weights when dealing with 1km datasets.

frazane:
## Description

`PlanarAreaWeights` builds its Voronoi tessellation with `qhull_options="QJ Pp"`. This PR switches it to `Qbb Qc Qz Pp`, qhull's merged-facet precision handling and SciPy's own default for 2-D Voronoi.

* Allows creation of stretched-grid graphs that were previously failing unless one was manually patching the QHull joggle. 
* Brings a significant performance improvement: for our n320 + ICON 1km graph (~1.7M points) we go from ~300s to ~20s. When using a large joggle like 0.5 in main, we could start from lower values (~60s).

No longer rely on joggle to be able to build very high-resolution stretched-grid graphs. It was taking longer and leading to wrong results (due to the joggle itself).

Closes #690. Same root cause as #698.

## Additional notes

**Could someone with a different stretched or LAM configuration confirm the graphs still look right on their side?** I have only checked a 1 km LAM cut into ERA5 N320.

---

### DIFF
--- a/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
+++ b/graphs/src/anemoi/graphs/nodes/attributes/area_weights.py
@@ -166,10 +166,7 @@ class PlanarAreaWeights(BaseAreaWeights):
         SciPy returns each 2-D region's vertices in boundary order, so for a convex cell
         this matches ``ConvexHull(...).volume`` to float-rounding level. A region whose
         stored vertex order is not convex would make shoelace under-count, so those are
-        detected and recomputed exactly with ``ConvexHull``. Under the merged-facet qhull
-        options used by :meth:`compute_area_weights` this guard is not expected to fire
-        (it does not on N320 or on 1.7M-node stretched grids); it is kept as a cheap
-        safety net against near-collinear cell vertices.
+        detected and recomputed exactly with ``ConvexHull``.
 
         Parameters
         ----------
@@ -242,14 +239,8 @@ class PlanarAreaWeights(BaseAreaWeights):
         boundary_points = self._get_boundary_ring(latlons, resolution)
 
         # Build the Voronoi tessellation over all points (boundary ring included).
-        #
-        # "Qbb Qc Qz" is qhull's merged-facet precision handling (SciPy's own default for
-        # 2-D Voronoi) and "Pp" silences its precision warnings. Do NOT swap this for the
-        # joggle ("QJ"): joggle displaces every input coordinate by up to 1% of the
-        # *bounding box*, which is unrelated to the node spacing. On a stretched grid
-        # (global extent, 1 km nodes) that is ~40x the spacing, so qhull fails to build a
-        # hull at all and asks for a larger joggle -- and raising it, as QH6229 suggests,
-        # returns silently meaningless areas rather than fixing anything. See #690.
+        # Merged facets, not the joggle: joggle scales with the bounding box rather than
+        # the node spacing, so it fails on stretched grids (#690).
         extended_points = np.vstack([latlons, boundary_points])
         v = Voronoi(extended_points, qhull_options="Qbb Qc Qz Pp")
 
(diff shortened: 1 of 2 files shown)


JPXKQX:
Hi @frazane , thanks for this PR, I think the documentation and test you added look good to me. It would be nice to see a difference map of main with this branch for n320 to check whether we can expect any difference during training

frazane:
> Hi @frazane , thanks for this PR, I think the documentation and test you added look good to me. It would be nice to see a difference map of main with this branch for n320 to check whether we can expect any difference during training

Hi @JPXKQX, this PR touches the PlanarAreaWeights so in general I would guess it won't be used for global grids like n320 (we would rather use SphericalAreaWeights), right ? Still, see the quick comparison below. Note that the differences are somewhat expected because we no longer have the joggle. 

image

I would also wait to see if others working with stretched-grids can report with no issues. @bastien-francois is looking at this!

caglarkucuk:
Thanks for the fix @frazane, this came just in time :) Had a look at the branch, and compared graphs created here and main (using n320+ara-2.5km). 

Graphs are byte-identical except `lam_area_weight`, which actually seems to return mostly noise in the main branch. Your argument we've had offline about large joggles applied in main introducing that noisy output makes perfect sense.

Graph creation for N320+ara-1km failed in main, but succeeded in this branch :white_check_mark:

Didn't see any significant differences in runtime, though.

caglarkucuk (approved):
Already tested and reported in my previous post.
LGTM!

JPXKQX:
Yes, true! Thanks, and amazing work (again)!
> Hi @JPXKQX, this PR touches the PlanarAreaWeights so in general I would guess it won't be used for global grids like n320 (we would rather use SphericalAreaWeights), right ? Still, see the quick comparison below. Note that the differences are somewhat expected because we no longer have the joggle.

bastien-francois:
Also tested with HARMONIE-AROME data. Except changes of lam_area_weight values, everything else (node coordinates, masks, the spherical area_weight) is identical.

However, I am not sure to get the impact on training performances, should it be necessarily improved with the new graph?

Thanks

PR #1299, opened 2026-08-05
feat(models): Tests for the autograd-aware communication functions in ``anemoi.models.distributed.graph``

joherics:
*This PR is the next step for testing of the distributed functionality in `models/distributed`, following PR https://github.com/ecmwf/anemoi-core/pull/1234.*

New tests of the autograd-aware communication functions. The tests use the same pattern as the tests for the communication primitives to launch with a distributed runner. The distributed primitives hpc workflow action has also been updated and now runs a on a full node with 4 gpus to cover more complex communication patterns but still has to be triggered manually, see: https://github.com/ecmwf/anemoi-core/actions/runs/30915143895/job/92011275993. 

The new tests cover:

- `ensure_sharded`
- `shard_tensor`
- `gather_tensor`
- `reduce_tensor`
- `sync_tensor`
- `reduce_shard_tensor`
- `all_to_all_transpose`

** Note, there is another PR coming after this with tests for graph-topology communication, including `halo_exchange`, alongside the graph-sharding and `HaloInfo` tests.** 

### DIFF
--- a/.github/workflows/distributed-primitive-tests-hpc-gpu.yml
+++ b/.github/workflows/distributed-primitive-tests-hpc-gpu.yml
@@ -24,10 +24,10 @@ jobs:
             pip install --upgrade pip
             pip install -e ./training[all,tests] -e ./models[all,tests] -e ./graphs[all,tests]
 
-            python3 -m pytest -q models/tests/distributed/test_communication_primitives.py \
+            python3 -m pytest -q models/tests/distributed \
               --distributed \
               --distributed-backend=all \
-              --distributed-world-size=2
+              --distributed-world-size=4
 
             deactivate
             rm -rf $REPO_NAME
@@ -36,7 +36,7 @@ jobs:
             #SBATCH --nodes=1
             #SBATCH --ntasks=1
             #SBATCH --cpus-per-task=4
-            #SBATCH --gpus-per-node=2
+            #SBATCH --gpus-per-node=4
             #SBATCH --time=00:30:00
             #SBATCH --qos=ng
             #SBATCH --mem=64G
--- a/models/tests/distributed/test_graph_comm.py
+++ b/models/tests/distributed/test_graph_comm.py
@@ -321,3 +321,93 @@ def test_shard_tensor_no_backward_gather(
         dim=dim,
     )
 
+
+@pytest.mark.parametrize(
+    "sizes",
+    [
+        pytest.param(None, id="sizes_none"),
+        pytest.param([4], id="explicit_single_rank_sizes"),
+    ],
+)
+def test_gather_tensor_identity_without_group(sizes: list[int] | None) -> None:
+    """With ``mgroup=None``, values and gradients pass through unchanged.
+
+    Covers the local fallback of ``_GatherParallelSection``; the shard-size
+    metadata is ignored on this path.
+    """
+    x = _make_range_tensor((4, 3))
+    expected = x.clone()
+    grad_output = _make_grad_output((4, 3))
+    x.requires_grad_(True)
+
+    gathered = gather_tensor(x, dim=0, sizes=sizes, mgroup=None)
+
+    assert gathered.size() == expected.size()
+    assert gathered.dtype == expected.dtype
+    assert gathered.device == expected.device
+    torch.testing.assert_close(gathered, expected)
+
+    loss = (gathered * grad_output).sum()  # d(loss)/d(gathered) == grad_output
+    loss.backward()
+    torch.testing.assert_close(x.grad, grad_output)
+
+
+def _test_gather_tensor_rank(
+    *,
+    rank: int,
+    world_size: int,
+    device: torch.device,
+    group: dist.ProcessGroup,
+    shape: tuple[int, ...],
+    dim: int,
+    atol: float = GLOBAL_DEFAULT_ATOL,
+    rtol: float = GLOBAL_DEFAULT_RTOL,
+) -> None:
+    full = _make_range_tensor(shape, device)
+    sizes = get_balanced_partition_sizes(full.size(dim), world_size)
+    local = torch.split(full, sizes, dim=dim)[rank].contiguous().clone().requires_grad_(True)
+
+    gathered = gather_tensor(local, dim=dim, sizes=sizes, mgroup=group)
+
+    assert gathered.size() == full.size()
+    assert gathered.dtype == full.dtype
+    assert gathered.device == full.device
+    torch.testing.assert_close(gathered, full, atol=atol, rtol=rtol)
+
+    grad_output = _make_grad_output(shape, device, scale=rank + 1)
+    loss = (gathered * grad_output).sum()  # d(loss)/d(gathered) == grad_output
+    loss.backward()
+
+    expected_grad = torch.split(grad_output, sizes, dim=dim)[rank].clone()
+    torch.testing.assert_close(local.grad, expected_grad, atol=atol, rtol=rtol)
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize(
+    ("shape", "dim"),
+    [
+        pytest.param((1025, 1024), 0, id="dim0_uneven_1m"),
+        pytest.param((1024, 1025), 1, id="dim1_uneven_1m"),
+        pytest.param((64, 128, 129), -1, id="negative_last_dim_3d_1m"),
+    ],
+)
+def test_gather_tensor_splits_gradients(
+    shape: tuple[int, ...],
+    dim: int,
+    distributed_backend: str,
+    distributed_world_size: int,
+) -> None:
+    """Gather shards forward; split each rank's own grad_output backward.
+
+    Every rank reconstructs the full tensor forward. Backward involves no
+    communication: with rank-scaled grad_outputs the expected gradient is the
+    rank's slice of its own grad_output, so a spurious cross-rank all-reduce
+    would produce the slice of the rank-summed grad_outputs and fail.
+    """
+    run_distributed_test(
+        _test_gather_tensor_rank,
+        backend=distributed_backend,
+        world_size=distributed_world_size,
+        shape=shape,
+        dim=dim,
+    )
--- a/models/tests/distributed/test_graph_comm.py
+++ b/models/tests/distributed/test_graph_comm.py
@@ -411,3 +411,72 @@ def test_gather_tensor_splits_gradients(
         shape=shape,
         dim=dim,
     )
+
+
+def test_reduce_tensor_identity_without_group() -> None:
+    """With ``mgroup=None``, values and gradients pass through unchanged.
+
+    Covers the local fallback of ``_ReduceParallelSection``.
+    """
+    x = _make_range_tensor((4, 3))
+    expected = x.clone()
+    grad_output = _make_grad_output((4, 3))
+    x.requires_grad_(True)
+
+    reduced = reduce_tensor(x, mgroup=None)
+
+    assert reduced.size() == expected.size()
+    assert reduced.dtype == expected.dtype
+    assert reduced.device == expected.device
+    torch.testing.assert_close(reduced, expected)
+
+    loss = (reduced * grad_output).sum()  # d(loss)/d(reduced) == grad_output
+    loss.backward()
+    torch.testing.assert_close(x.grad, grad_output)
+
+
+def _test_reduce_tensor_rank(
+    *,
+    rank: int,
+    world_size: int,
+    device: torch.device,
+    group: dist.ProcessGroup,
+    shape: tuple[int, ...],
+    atol: float = GLOBAL_DEFAULT_ATOL,
+    rtol: float = GLOBAL_DEFAULT_RTOL,
+) -> None:
+    base = _make_range_tensor(shape, device)
+    local = (base + rank).clone().requires_grad_(True)
+
+    reduced = reduce_tensor(local, mgroup=group)
+
+    expected = torch.stack([base + q for q in range(world_size)]).sum(dim=0)
+    assert reduced.size() == expected.size()
+    assert reduced.dtype == expected.dtype
+    assert reduced.device == expected.device
+    torch.testing.assert_close(reduced, expected, atol=atol, rtol=rtol)
+
+    grad_output = _make_grad_output(shape, device, scale=rank + 1)
+    loss = (reduced * grad_output).sum()  # d(loss)/d(reduced) == grad_output
+    loss.backward()
+
+    torch.testing.assert_close(local.grad, grad_output, atol=atol, rtol=rtol)
+
+
+@pytest.mark.distributed
+def test_reduce_tensor_backward_is_identity(
+    distributed_backend: str,
+    distributed_world_size: int,
+) -> None:
+    """Sum same-shaped rank contributions forward; keep grad_output unchanged backward.
+
+    The identity backward is deliberately not the adjoint of the forward sum.
+    With rank-scaled grad_outputs, a spurious backward all-reduce would
+    return the rank-summed grad_outputs instead of the rank's own and fail.
+    """
+    run_distributed_test(
+        _test_reduce_tensor_rank,
+        backend=distributed_backend,
+        world_size=distributed_world_size,
+        shape=(1025, 1024),
+    )
--- a/models/tests/distributed/test_graph_comm.py
+++ b/models/tests/distributed/test_graph_comm.py
@@ -639,3 +639,91 @@ def test_sync_tensor_no_forward_gather(
         gather_in_fwd=gather_in_fwd,
         with_sizes=with_sizes,
     )
+
+
+def test_reduce_shard_tensor_identity_without_group() -> None:
+    """With ``mgroup=None``, values and gradients pass through unchanged.
+
+    Covers the local fallback of ``_ReduceShardParallelSection``.
+    """
+    x = _make_range_tensor((4, 3))
+    expected = x.clone()
+    grad_output = _make_grad_output((4, 3))
+    x.requires_grad_(True)
+
+    shard_sum = reduce_shard_tensor(x, dim=0, sizes=None, mgroup=None)
+
+    assert shard_sum.size() == expected.size()
+    assert shard_sum.dtype == expected.dtype
+    assert shard_sum.device == expected.device
+    torch.testing.assert_close(shard_sum, expected)
+
+    loss = (shard_sum * grad_output).sum()  # d(loss)/d(shard_sum) == grad_output
+    loss.backward()
+    torch.testing.assert_close(x.grad, grad_output)
+
+
+def _test_reduce_shard_tensor_rank(
+    *,
+    rank: int,
+    world_size: int,
+    device: torch.device,
+    group: dist.ProcessGroup,
+    shape: tuple[int, ...],
+    dim: int,
+    atol: float = GLOBAL_DEFAULT_ATOL,
+    rtol: float = GLOBAL_DEFAULT_RTOL,
+) -> None:
+    base = _make_range_tensor(shape, device)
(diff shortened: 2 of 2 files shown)


models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+    scale: float = 1.0,
+) -> torch.Tensor:
+    """Create an integer-valued grad_output: a repeating ``1..251`` pattern times ``scale``."""
+    # Small bounded values keep expectations exact in float32; variation exposes mis-routed slices.
+    numel = torch.Size(shape).numel()
+    values = torch.arange(numel, dtype=torch.float32, device=device) % 251 + 1
+    return (values * scale).reshape(shape)
+
+
+def test_ensure_sharded_computes_sizes_without_group() -> None:
```
cathalobrien, line 82: I think all the non-distributed tests could be moved to a different file. like `models/tests/distributed/test_graph.py`

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+
+
+@pytest.mark.distributed
+def test_ensure_sharded_shards_replicated_tensor(
+    distributed_backend: str,
+    distributed_world_size: int,
+) -> None:
+    """Shard a replicated tensor with balanced sizes; pass an existing shard through.
+
+    With ``shard_sizes=None`` the tensor is sharded and the computed sizes are
```
cathalobrien, line 144: would it be worth also adding a test where `shard_sizes != None`?

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+
+
+@pytest.mark.parametrize(
+    "gather_in_backward",
+    [
+        pytest.param(True, id="gather_in_backward"),
+        pytest.param(False, id="no_gather_in_backward"),
+    ],
+)
+def test_shard_tensor_identity_without_group(gather_in_backward: bool) -> None:
```
cathalobrien, line 164: seems to be a lot replicated between this local function and the distributed test `_test_shard_tensor_rank`. if we want to keep local tests, you could maybe run the distributed test with a model comm group =None? or remove the local versions. suppose there might be an issue where when dist is initialised, mgroup=None => global comms

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+
+    assert sharded.size() == expected.size()
+    torch.testing.assert_close(sharded, expected, atol=atol, rtol=rtol)
+
+    # each rank's grad_output is its slice of one full grad_output
+    grad_output_full = _make_grad_output(shape, device)
+    grad_output_local = torch.split(grad_output_full, sizes, dim=dim)[rank].contiguous()
+    loss = (sharded * grad_output_local).sum()  # d(loss)/d(sharded) == grad_output_local
+
+    with patch(
```
cathalobrien, line 281: what is this doing? maybe add a comment

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+    start = sum(sizes[:rank])
+    actual_local_slice = full.grad.narrow(normalized_dim, start, sizes[rank])
+    torch.testing.assert_close(actual_local_slice, grad_output_local, atol=atol, rtol=rtol)
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize(
+    ("shape", "dim"),
+    [
+        pytest.param((1025, 1024), 0, id="dim0_uneven_1m"),
```
cathalobrien, line 301: should you add `negative_last_dim_3d_1m` case used in the previous test?

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+    assert gathered.dtype == expected.dtype
+    assert gathered.device == expected.device
+    torch.testing.assert_close(gathered, expected)
+
+    loss = (gathered * grad_output).sum()  # d(loss)/d(gathered) == grad_output
+    loss.backward()
+    torch.testing.assert_close(x.grad, grad_output)
+
+
+def _test_gather_tensor_rank(
```
cathalobrien, line 355: maybe change `_rank` to `_dist` ? or remove it

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+        shape=shape,
+        dim=dim,
+    )
+
+
+@pytest.mark.parametrize(
+    "sizes",
+    [
+        pytest.param(None, id="sizes_none"),
+        pytest.param([4], id="explicit_single_rank_sizes"),
```
cathalobrien, line 329: this local test is small enough to be a unit test. maybe another reason to move to another file?

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+
+    grad_output = _make_grad_output(shape, device, scale=rank + 1)
+    loss = (reduced * grad_output).sum()  # d(loss)/d(reduced) == grad_output
+    loss.backward()
+
+    torch.testing.assert_close(local.grad, grad_output, atol=atol, rtol=rtol)
+
+
+@pytest.mark.distributed
+def test_reduce_tensor_backward_is_identity(
```
cathalobrien, line 467: could also just call this `test_reduce_tensor` maybe? :D

models/tests/distributed/test_graph_comm.py
```diff
@@ -0,0 +1,839 @@
...
+
+    The identity backward is deliberately not the adjoint of the forward sum.
+    With rank-scaled grad_outputs, a spurious backward all-reduce would
+    return the rank-summed grad_outputs instead of the rank's own and fail.
+    """
+    run_distributed_test(
+        _test_reduce_tensor_rank,
+        backend=distributed_backend,
+        world_size=distributed_world_size,
+        shape=(1025, 1024),
```
cathalobrien, line 481: might be useful to talk to @japols about some relevant sizes for these tests

cathalobrien (commented):
nice work, great tests. I have some small comments, mostly about if we need the small local tests, and if so should we move them to a different file

models/tests/distributed/test_graph_comm.py
joherics: As, discussed offline, I think it is better that we keep them in here in order to group all tests of the same functionality and let the pytest markers handle the filtering of distributed vs normal tests.

models/tests/distributed/test_graph_comm.py
joherics: Yes, it is handled in another test.

models/tests/distributed/test_graph_comm.py
joherics: This is testing that `anemoi.models.distributed.graph._gather` isn't called inside the `with patch(...):` scope. I used `patch` which is part of the std libraries mocking feature. I agree that this could warrant a comment!

models/tests/distributed/test_graph_comm.py
joherics: I have updated the tests now to better cover the different interesting code paths. Especially, the tests of the primitives, which are now testing more configs than the autograd aware wrappers.

models/tests/distributed/test_graph_comm.py
joherics: I am using `_rank` for the local functions that are per process and launched with distributed launcher, so I think it is a good name.

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             #SBATCH --nodes=1
             #SBATCH --ntasks=1
             #SBATCH --cpus-per-task=4
-            #SBATCH --gpus-per-node=2
```
cathalobrien, line 39: I have a slight preference to keep it at 2 gpus per node, unless we really need 4 gpus for some reason

joherics:
I have updated the tests now and following an offline discussion with @japols, and your comments @cathalobrien, I have introduced more test cases (especially for the primitives) but somewhat reduced the test matrix for the autograd aware wrappers so that we don't test the same thing twice. Nonetheless, it would nice if @japols could have a look at this and see if you think the important test paths are covered.

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
cathalobrien: ah I read to the end, I suppose 4 gpus is to test the asymmetric shards with an empty shard?

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
```diff
@@ -24,10 +24,10 @@ jobs:
             pip install --upgrade pip
             pip install -e ./training[all,tests] -e ./models[all,tests] -e ./graphs[all,tests]
 
-            python3 -m pytest -q models/tests/distributed/test_communication_primitives.py \
+            python3 -m pytest -q models/tests/distributed \
```
cathalobrien, line 27: should we change the name of the tests to 'distributed-tests-hpc' instead of 'distributed-primitive-tests-hpc-gpu' since we now test more then just the primitives?

models/tests/distributed/test_graph_comm.py
joherics: Yes, if we go down that route. Otherwise I think we should add this file to the normal "unit tests" and skip the tests that are marked as distributed.

models/tests/distributed/test_communication_primitives.py
```diff
@@ -89,6 +90,30 @@ def test_split_distributes_full_tensor_to_rank_local_slice(
     )
 
 
+@pytest.mark.distributed
+@pytest.mark.parametrize(
+    "shard_size_pattern",
+    [
+        pytest.param((128, 384, 256, 512), id="dim=0,irregular-shard-sizes=[128,384,256,512]"),
```
cathalobrien, line 97: what happens if you try to run a test which requires 4 gpus but it only has 1 or 2 available? maybe you skip it somewhere in the distributed runner?

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
joherics: Yes, I forgot to add a comment about that. I suggest we do that as a final commit before merging. If I change it now, then I don't think I will be able to trigger the workflow manually anymore as it is not on main yet.

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
joherics: Exactly, to capture some of the non-trivial comm patterns we need 4. I agree that we don't want to use too much gpu resources and I have reduced the total number of test cases and was able to finish the tests within a 10 min allocation on ac. The workflow action took ~13 mins https://github.com/ecmwf/anemoi-core/actions/runs/32947676685/job/98111859430 which may be acceptable?

models/src/anemoi/models/distributed/graph.py
```diff
@@ -78,8 +78,8 @@ def shard_tensor(
         dimension along which to shard
     sizes : ShardSizes
         Per-rank shard sizes
-    mgroup : ProcessGroup
-        model communication group
+    mgroup : ProcessGroup or None
+        Model communication group. Pass ``None`` explicitly for the local identity fallback.
```
ssmmnn11, line 82: can we re-word "local identity fallback"? I don't know if everyone understands what that means.

models/tests/distributed/distributed_test_utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+import torch
+
+
+def shard_sizes_from_pattern(shard_size_pattern: tuple[int, ...], world_size: int) -> list[int]:
+    """Repeat a shard-size pattern until there is one size per rank."""
+    return [shard_size_pattern[rank % len(shard_size_pattern)] for rank in range(world_size)]
+
+
+def torch_version_less_than(major: int, minor: int) -> bool:
```
ssmmnn11, line 20: are we using something like this elsewhere? It could be useful in other places, but can also stay here.

joherics: I didn't find a "shared/common anemoi util for this" but I would assume similar function(s) are used elsewhere in the code and I agree that it would make sense to move this somewhere else, but I am not sure what the right place is?

models/src/anemoi/models/distributed/graph.py
joherics: Yes, I could also drop this commit that updates the type hints and docs if we prefer it the way it was (it is a clean commit that only introduced these changes).

models/tests/distributed/test_communication_primitives.py
ssmmnn11, line 97: would it make sense to do something like [127, 257, 63, 577] - more awkward, also for future comm routines?

ssmmnn11 (changes requested):
Great work! Pls see minor suggestions.

models/src/anemoi/models/distributed/graph.py
ssmmnn11: I don't have a strong opinion

ssmmnn11 (dismissed):
Already answered my questions.

japols (changes requested):
Very nice! we could also test that wrong shard_sizes not matching the tensor shapes raises an error. This would fail right now so we would need to change the primitives accordingly.

joherics:
> Very nice! we could also test that wrong shard_sizes not matching the tensor shapes raises an error. This would fail right now so we would need to change the primitives accordingly.

Yes, but this is an error raised by Pytorch and not by anemoi. I am not sure whether we want to enforce this at the anemoi layer as well? Currently we do a "mix of both" in the primitives. We assert that the ``dim_``  parameter is within range (even though this is also verified in ``torch.split``) but we don't assert that ``sizes_`` match the size of the dimensions but let ``torch.split`` raise the error in that case.
Furthermore, when looking at this more closely I realise that the assertions in the primitives aren't "complete". Consider this exerpt from ``primitives.py``, which does not capture the error in indexing when ``dim_ < -input.dim()``.
```python
48     # sanity checks                                                                                            
49     assert dim_ < input_.dim(), f"Error, cannot split along {dim_} for tensor with {input_.dim()} dimensions." 
```


Personally, I think that we should let torch handle the errors for sizes and dimensions as these checked are already built-in to torch (see examples below).  Alternatively, if we want to raise errors, I think we should raise an ``IndexError`` and a ``ValueError`` for these args. I could also add a check that the process group is initialized if the default ``group=None`` is passed, as this otherwise leads to a failure in ``dist.get_world_size(group)``.  

What do you think? @japols @ssmmnn11 @cathalobrien


### Pure Torch errors (already informative)

**Error raised when ``torch.split`` is called with wrong ``sizes``**
```python
import torch                                
                                            
x = torch.arange(10)                        
sizes = [3,3,3]                             
torch.split(x, split_size_or_sections=sizes)
```

output:

```
> python minimal_split_error_example.py                                                   
Traceback (most recent call last):
  File "xxx/anemoi-dev/anemoi-core/minimal_split_error_example.py", line 9, in <module>                          
    torch.split(x, split_size_or_sections=sizes)
  File "xxx/venvs/ac-anemoi-venv/lib/python3.12/site-packages/torch/functional.py", line 173, in split                           
    return tensor.split(split_size_or_sections, dim)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx/venvs/ac-anemoi-venv/lib/python3.12/site-packages/torch/_tensor.py", line 1030, in split                             
    return torch._VF.split_with_sizes(self, split_size, dim)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: split_with_sizes expects split_sizes to sum exactly to 10 (input tensor's size at dimension 0), but got split_sizes=[3, 3, 3]
```


**Error raised when ``torch.split`` is called with wron ``dim``**

```python
import torch                                           
                                                       
x = torch.arange(10)                                   
sizes = [3,3,4]                                        
y = torch.split(x, split_size_or_sections=sizes, dim=1)

```

output:
```
> python minimal_split_error_example.py                          
Traceback (most recent call last):
  File "xxx/anemoi-dev/anemoi-core/minimal_split_error_example.py", line 9, in <module> 
    y = torch.split(x, split_size_or_sections=sizes, dim=1)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx/venvs/ac-anemoi-venv/lib/python3.12/site-packages/torch/functional.py", line 173, in split  
    return tensor.split(split_size_or_sections, dim)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxx/venvs/ac-anemoi-venv/lib/python3.12/site-packages/torch/_tensor.py", line 1030, in split    
    return torch._VF.split_with_sizes(self, split_size, dim)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)

File: models/tests/distributed/test_communication_primitives.py (hunk 12 above)
Comment (joherics) [reply models/tests/distributed/test_communication_primitives.py:97]: Good question, the function that generates the shard sizes from a pattern will use that pattern as a "base" for the shard sizes and repeat the pattern to determine the shard sizes if `world_size>len(shard_size_pattern)`, otherwise it will use the first `world_size` elements of the pattern for the shard sizes, so it works with `world_size < shard_size_pattern` as well.

Comment (joherics) [reply models/tests/distributed/test_communication_primitives.py]: Yes, I have done this now and confirmed that it works, also added a an unbalanced case with a shard of size 1 now for each primitive as well.

File: .github/workflows/distributed-tests-hpc.yml [hunk 15]
```diff
@@ -24,10 +24,10 @@ jobs:
             pip install --upgrade pip
             pip install -e ./training[all,tests] -e ./models[all,tests] -e ./graphs[all,tests]
 
-            python3 -m pytest -q models/tests/distributed/test_communication_primitives.py \
+            python3 -m pytest -q models/tests/distributed \
```
joherics: I have renamed it now

joherics:
@ssmmnn11 @cathalobrien @japols Thank you for the reviews, I have updated the branch according to your comments and suggestions. On top of that I have updated the error and assertion handling and also renamed the workflow and test files to match anemoi's test structure and other workflows better.

In detail, the following changes have been made:
* Made the docstrings clearer about the option to pass ``None`` as process group in the autograd comm routines
* Added or changed shard sizes to be even more irregular in tests
* Added test cases with 0 sized shards
* Updated the exception/assertion logic, following https://github.com/ecmwf/anemoi-core/issues/1367
* Renamed, the test files for ``graph.py`` and ``primitives.py`` to simply be ``test_graph.py`` and ``test_primitives.py``, doing so forced me to add a new empty ``__init__.py`` in  the test folder, in order avoid name clashes. This follows the pattern already in use in ``training/tests/unit/*``
* Renamed the GH workflow for the distributed tests from ``distributed-primitive-tests-hpc-gpu`` to ``distributed-tests-hpc`` and also made the slurm job id and name more coherent with the rest of anemoi's workflows.
* Added all tests under ``models/tests/distributed`` to the workflow, this includes
    - the recently merged "torch.compile alltoall test" added in https://github.com/ecmwf/anemoi-core/pull/1282 @cathalobrien please, let me know if you don't think this should be included in the workflow.
    - the newly added spatial preprocess distributed tests from https://github.com/ecmwf/anemoi-core/pull/1305

PR #1305, opened 2026-08-05
feat: spatial preprocessor

### LINKED ISSUE #1304: Add spatial preprocessors for downscaling

For diffusion spatial downscaling we need a way to interpolate low resolution input data to the target grid.

Follow the suggestion in point 1 of https://github.com/ecmwf/anemoi-core/issues/1274

That is,

### 1. Spatial preprocessing

Map (batch, time, ensemble, source points, features)
to (batch, time, ensemble, target points, features)
while preserving every non-spatial axis and supporting both source-grid and target-grid sharding.
Approach

Introduce a spatial preprocessor (using a cross-grid processor) registered on the model interface.

Spatial preprocessing is applied before all other preprocessors so that every dataset reaches the model on the same grid and existing normalisation strategies (particularly for tendency models) continue to work unchanged.

An interpolation connection as implemented here: https://github.com/ecmwf/anemoi-core/pull/1242

### Additional context

_No response_

### Organisation

_No response_

VeraChristina:
## Description
Add a spatial preprocessor for spatial downscaling

This is a building block for the implementation of diffusion spatial downscaling. The preprocessor interpolates a low resolution input to the target grid resolution. Compare https://github.com/ecmwf/anemoi-core/issues/1304

https://github.com/ecmwf/anemoi-core/issues/1304

##  Additional notes ##
Note that an example configuration will be added later as part of an example configuration for the diffusion spatial downscaler.


training/src/anemoi/training/train/methods/base.py
ssmmnn11: just to check: are we here using the sizes after projection now?

training/src/anemoi/training/train/methods/base.py
```diff
@@ -328,9 +328,12 @@ def __init__(
 
         self.shard_sizes, self.grid_sizes = {}, {}
         for dataset_name in self.dataset_names:
-            self.grid_sizes[dataset_name] = graph_data[
-                dataset_name
-            ].num_nodes  # TODO(Mario): Replace by dataset.grid_size
+            if dataset_name in self.model.spatial_pre_processors:
```
ssmmnn11, line 333: we could check if the grid the encoder expects machtes the output_grid_size of the projector ; But maybe not needed.

training/src/anemoi/training/schemas/data.py
```diff
@@ -44,3 +45,5 @@ class DataSchema(PydanticBaseModel):
     "Time frequency requested from the dataset. Must be null when using trajectory (forecast) datasets."
     datasets: dict[str, DatasetDataSchema] | None = None
     "Dictionary mapping dataset names to their configurations."
+    spatial_processors: dict[str, SpatialProcessorSchema] | None = Field(default=None)
```
ssmmnn11: would it make sense to do it like this instead?

```
  data:
    datasets:
      low_res:
        processors: ...
        spatial_processor:
          _target_: ...CrossGridProjector
          file_path: ...
```

ssmmnn11 (changes requested):
Great work and exciting feature!

training/src/anemoi/training/train/methods/base.py
ssmmnn11: we could do here:

```
  def allgather_batch(self, batch, grid_shard_sizes):
      if grid_shard_sizes is None or self.reader_group_size == 1:
          return batch

      expected_size = grid_shard_sizes[self.reader_group_rank]
      if batch.shape[self.grid_dim] != expected_size:
          raise ValueError(
              f"Expected grid shard of size {expected_size}, "
              f"got {batch.shape[self.grid_dim]}"
          )

      return gather_tensor(
          batch,
          self.grid_dim,
          grid_shard_sizes,
          self.reader_groups[self.reader_group_id],
      )
```

then in _setup_batch_sharding (before projection) we would pass self.shard_sizes[dataset_name]

and in plotting (after projection) we would pass pl_module.grid_shard_sizes[dataset_name]

VeraChristina: updated in [19a75cb](https://github.com/ecmwf/anemoi-core/pull/1305/commits/19a75cbe5099a177f5ce6db2c1396191a3171a6f) to pass grid shard sizes explicitly

training/src/anemoi/training/train/methods/base.py
VeraChristina: added a check in [19a75cb](https://github.com/ecmwf/anemoi-core/pull/1305/commits/19a75cbe5099a177f5ce6db2c1396191a3171a6f)

training/src/anemoi/training/schemas/data.py
VeraChristina: yes, I agree this is better, updated in [b361653](https://github.com/ecmwf/anemoi-core/pull/1305/commits/b361653320b495d017ca63b4234dec9f28e7495d)

training/src/anemoi/training/train/methods/base.py
```diff
@@ -1281,15 +1281,14 @@ def log_optimizer(optimizer: torch.optim.Optimizer) -> None:
         LOGGER.info("Optimizer settings: %s", defaults_to_log)
 
     def _validate_spatial_processor_target_grid(self, graph_data: HeteroData) -> None:
-        """Check each spatial projector's target grid against the encoder's node set."""
+        """Check each spatial projector's target grid against its dataset's graph nodes."""
         for dataset_name, projector in self.model.spatial_pre_processors.items():
-            node_set = self.model.model.encoder_node_set(dataset_name)
-            encoder_grid_size = graph_data[node_set].num_nodes
-            if projector.output_grid_size != encoder_grid_size:
+            dataset_grid_size = graph_data[dataset_name].num_nodes
```
VeraChristina, line 1286: I think that graph_data[dataset_name].num_nodes will be the low res number of nodes (or not exist), since we build the graph based on the input datasets -- so that's not what we want to check against. The encoder input will have grid size of the high res target dataset. Maybe there's a better place to validate that these match

VeraChristina: This is the correct check for the current set up. i've just added to the error message that it's worth checking that the graph is built on the correct nodes (those after projection) as well as the projection matrix

ssmmnn11 (approved):
LGTM, thank you for  the contribution!

PR #1308, opened 2026-08-05
fix: Add AttrType to safe globals for torch serialisation

HCookie:
## Description
Addresses failing integration tests caused by #1303 as seen in https://github.com/ecmwf/anemoi-core/actions/runs/31021131375/job/92360048174?pr=1295

> [!CAUTION]
> Integration tests failed
> https://github.com/ecmwf/anemoi-core/actions/runs/31024878202

## New graph to replace old test graph
https://sites.ecmwf.int/ecm1947/files/Anemoi/lam-graph-rebuilt.pt

### DIFF
--- a/graphs/src/anemoi/graphs/create.py
+++ b/graphs/src/anemoi/graphs/create.py
@@ -9,6 +9,7 @@
 
 
 import logging
+from collections import defaultdict
 from itertools import chain
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


cathalobrien:
thanks for the quick fix! once [these](https://github.com/ecmwf/anemoi-core/actions/runs/31024878202) pass I'm happy to merge.

HCookie:
This is failing with 
```log
_pickle.UnpicklingError: Weights only load failed. In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file from a trusted source.
Please file an issue with the following so that we can make `weights_only=True` compatible with your use case: WeightsUnpickler error:

Can only SETITEMS for dict, collections.OrderedDict, collections.Counter, but got <class 'collections.defaultdict'>

Check the documentation of torch.load to learn more about types accepted by default with weights_only https://pytorch.org/docs/stable/generated/torch.load.html.
```
Without a change to pytorch, this is not fixable. 
It is possible that this is a problem particular to this checkpoint as checkpoints I have locally made with a recent version of anemoi-graphs do not raise this issue.

PR #1311, opened 2026-08-06
revert: "fix(graphs): use weights_only=True when loading graphs in anemoi-graphs"

HCookie:
Reverts ecmwf/anemoi-core#1303 due to issues identified in the attempt of a hotfix in #1308 

TLDR:
Some LAM graphs come with a `_cache_attr` property which is a `defaultdict`, it is not needed, but is included by `torch_geometric` causing issues with `torch.load` 

> [!TIP]
> Integration tests passed
> https://github.com/ecmwf/anemoi-core/actions/runs/31090338148


mc4117 (approved):
LGTM

PR #1281, opened 2026-07-30
fix: callbacks cache

anaprietonem:
## Description
This PR fixes redundant post-processing during validation plotting: previously each per-batch plot callback (PlotSample, PlotHistogram, etc.) created its own fresh processed_cache={}}, so post-processing ran once per callback per batch; now a shared _plot_cache dict is stashed on pl_module (keyed by _plot_cache_batch_idx, reset when the batch index changes), so all callbacks on the same validation batch reuse one cache and post-processing runs only once per (dataset, members) pair . A new unit test (test_on_validation_batch_end_shares_cache_across_callbacks) verifies the shared-cache behavior and reset-on-new-batch behavior, plus a clarifying docstring was added to the existing cache test

## What problem does this change solve?
Fix https://github.com/ecmwf/anemoi-core/issues/1280

##  Additional notes ##

***As a contributor to the Anemoi framework, please ensure that your changes include unit tests, updates to any affected dependencies and documentation, and have been tested in a parallel setting  (i.e., with multiple GPUs). As a reviewer, you are also responsible for verifying these aspects and requesting changes if they are not adequately addressed. For guidelines about those please refer to https://anemoi.readthedocs.io/en/latest/***

By opening this pull request, I affirm that all authors agree to the [Contributor License Agreement.](https://github.com/ecmwf/codex/blob/main/Legal/Contributor-License-Agreement.md)

PR #1319, opened 2026-08-11
revert(training): "feat: reuse processing callbacks (#1213)"

### LINKED ISSUE #1280: Caching system in plotting callbacks not properly working
### What happened?

At the moment the caching used in the process function in plotting works so that each per-batch plot callback (PlotSample, PlotHistogram, etc.) creates its own fresh processed_cache={}}, so post-processing ran once per callback per batch. This is a bug as the correct behaviour should be shared _plot_cache dict that just resets when the batch_idx changes

NA

### Version

main

ATOS

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

VeraChristina:
## Description
Reverts https://github.com/ecmwf/anemoi-core/pull/1213 which introduced a cache in the plotting callbacks that isn't working as intended, see https://github.com/ecmwf/anemoi-core/issues/1280.

Removes dead code (the not working cache in plotting callbacks)

https://github.com/ecmwf/anemoi-core/issues/1280

##  Additional notes ##
https://github.com/ecmwf/anemoi-core/pull/1281 proposes a fix for the cache. Since it is unclear whether the cache leads to a performance improvement, the suggestion is to remove the not working cache.

I have run the global smoke test to check I haven't missed any parts of the cache.


training/tests/integration/conftest.py
```diff
@@ -59,23 +57,6 @@ def _reset_torch_compile(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Non
     return
 
 
-@pytest.fixture(autouse=True)
-def log_memory_usage(request: pytest.FixtureRequest) -> None:
```
cathalobrien, line 63: why are we removing this? not opposed to it necessarily

cathalobrien (approved):
lgtm, thanks for this. just 1 question

training/tests/integration/conftest.py
VeraChristina: It was introduced in #1213, so I thought this would be a clean revert

VeraChristina: (also happy to keep it)

PR #1320, opened 2026-08-12
fix(graphs): correct planar area weights at the domain edge

frazane:
## Description

Two related fixes to the planar area weights. `MaskedPlanarAreaWeights` now tessellates the masked nodes on their own instead of tessellating every node and applying the mask afterwards. And the cells at the edge of the tessellation, which a planar Voronoi diagram cannot define, are no longer closed with a constructed ring of boundary points: each of them takes the area of its nearest well-defined cell.

Follows #1298, which fixed the qhull options in the same code path, but was an independent problem.

Bugfix, with a behaviour change. Both defects inflate the Voronoi cells at the edge of a domain, and under `norm: unit-max` the largest cell divides every other weight, so an inflated edge silently rescales the whole field. On a 1 km regional cutout in an N320 background, the edge cells reach into the coarse grid and come out up to ~500x too large, leaving essentially the whole domain below 10% of the maximum weight.

The fix asks the tessellation which cells it actually got right: a cell is trusted iff it is bounded and all its Voronoi vertices stay within reach of the node's own neighbour spacing. Every other node takes the area of its nearest trusted cell — on a quasi-uniform grid that is the exact answer. A node set with no trusted cell at all (fewer than three nodes, a collinear mask) gets uniform weights instead of aborting the build.

On the real stretched graph, `lam_area_weight` goes from median 0.0019 with a 526x spread to **median 0.786 with a 1.27x spread**:

image

*Shared logarithmic colour scale. Before, the inflated boundary ring (zoom) sets the normaliser and the whole domain sits near zero; after, the ring is gone and the field is uniform.*

Found while investigating #690. No separate issue.

## Additional notes

Anyone using `lam_area_weight` will see it change substantially — that is the point of the fix, but it will move trained-model numbers. **Could someone with a different regional or stretched configuration check their graphs?**

---

----
📚 Documentation preview 📚: https://anemoi-training--1320.org.readthedocs.build/en/1320/

----
📚 Documentation preview 📚: https://anemoi-graphs--1320.org.readthedocs.build/en/1320/

----
📚 Documentation preview 📚: https://anemoi-models--1320.org.readthedocs.build/en/1320/


bastien-francois:
Here are some plots to accompany @frazane's proposed fix, tested on UWC-West.

The fix successfully resolves the boundary artifact: lam_area_weight values that were spuriously inflated at the domain edges are now corrected, and substantially tightens the overall spread of values (previously spanning several orders of magnitude, now much more uniform).

Still, a band of high lam_area_weights is still present around longitude=0.

lam_area_weight_hist
lam_area_weight_spatial

bastien-francois:
Having a switch of longitudes from 359.99 to 0.01 in regional seems to be problematic with high lam_area_weights values at lon=0.

I looked for an additional fix to solve this issue: the idea is to actually recenter longitude values to avoid having the switch from 359.99 to 0.01 in our regional data.

After checking on UWC-West, it seems that it removes the artifacts at lon=0.
All the values are however affected due to the fact that the max lam_area_weight values have been changed (very likely to be one value sitting on this lon=0 line).

lam_area_weight_spatial

lam_area_weight_lon0_seam

lam_area_weight_hist

mchantry:
@frazane thanks for the contribution, this is looking great.
Is the argument for not running the expensive tests that this set of graph weights aren't currently tested in the integration tests?

frazane:
> @frazane thanks for the contribution, this is looking great. Is the argument for not running the expensive tests that this set of graph weights aren't currently tested in the integration tests?

Hi Matt! I haven't thought about it to be honest, I just wanted to disable them as long as this is a draft and would have enabled them later.

mchantry:
Makes sense.
To our shame, this feature isn't documented yet, with the proposed workflow. But this will be fixed asap.
The proposal is that you only need to put a label on just before merge.

havardhhaugen:
I compared a graph from this branch with the one I am currently training with, and the values of `lam_area_weights` have changed significantly - although this might also come from #1298 - is that an intended change?

Since this is only being used as a scaler in the validation - which by the looks of it is being normalized outside of graph creation anyways - I don't think it's a problem for stretched grid.
Screenshot from 2026-08-18 12-08-26

frazane:
> I compared a graph from this branch with the one I am currently training with, and the values of lam_area_weights have changed significantly - although this might also come from https://github.com/ecmwf/anemoi-core/pull/1298 - is that an intended change?

@havardhhaugen yes, this PR is a follow up to the other one. Basically before #1298 the values were noise dominated. After removing the noise, some other artifacts became more evident (they were a bit hidden away before), so this PR tries to fix them.

bastien-francois (approved):
Tested on stretched-grid after the last changes. I confirm the noise in lam_area_weight is gone, and so is the lon=0 artifact. The fix for lon=0 artifacts, based on recentering, assumes data doesn't span the full globe. When lam_area_weights isn't used as a training scaler, training is unaffected (also confirmed by @havardhhaugen).

havardhhaugen (approved):

PR #1325, opened 2026-08-13
feat: Optional usage of min_level. Often helpful for ICON-LAM hidden meshes…

svenulbrich:
add optional `min_level` argument for hidden LAM meshes

## Description
The optional argument min_level allows to define a coasening level up to which the connections are defined.

For small LAM areas the coarsening to the base icosaeder does not work and triggers a crash in icon_mesh.py .
This code does not resolve this issue in all cases, but for some like DWD's ICON-D2 grid.

##  Additional notes ##

### DIFF
--- a/graphs/src/anemoi/graphs/generate/icon_mesh.py
+++ b/graphs/src/anemoi/graphs/generate/icon_mesh.py
@@ -128,7 +128,7 @@ class ICONMultiMesh:
     def multi_mesh_edges(self) -> np.ndarray:
         """Returns the multi-mesh edges as an arrays of vertex indices."""
         # concatenate edge-vertex lists (= edges of the multi-level mesh):
-        edges = np.concatenate([edges for edges in self.edge_vertices[: self.max_level] if (edges.size > 0)], axis=0)
+        edges = np.concatenate([edges for edges in self.edge_vertices if (edges.size > 0)], axis=0)
         return np.concatenate([edges, np.fliplr(edges)])
 
     def _read_vertices_data(self):
--- a/graphs/src/anemoi/graphs/generate/icon_mesh.py
+++ b/graphs/src/anemoi/graphs/generate/icon_mesh.py
@@ -98,7 +98,7 @@ class ICONMultiMesh:
 
         self.reflvl_vertex = reflvl_vertex
         self.max_level = max_level if max_level is not None else reflvl_vertex.max()
-        self.min_level = min_level
+        self.min_level = min_level if min_level is not None else 0
 
         # restrict edge-vertex list to multi_mesh level "max_level":
         if self.max_level < self.reflvl_vertex.max():
--- a/graphs/src/anemoi/graphs/generate/icon_mesh.py
+++ b/graphs/src/anemoi/graphs/generate/icon_mesh.py
@@ -274,9 +274,10 @@ class ICONCellDataGrid:
     uuidOfHGrid: str
     nodeset: NodeSet  # set of ICON cell circumcenters
     max_level: int
+    min_level: int
     select_c: np.ndarray
 
-    def __init__(self, icon_grid_filename: str, max_level: int | None = None):
+    def __init__(self, icon_grid_filename: str, max_level: int | None = None, min_level: int | None = None):
         self.grid_filename = icon_grid_filename
 
         # open file, representing the finest level
@@ -292,6 +293,7 @@ class ICONCellDataGrid:
             self.uuidOfHGrid = ncfile.uuidOfHGrid
 
         self.max_level = max_level if max_level is not None else reflvl_cell.max()
+        self.min_level = min_level if min_level is not None else 0
 
         # restrict to level `max_level`:
         self.select_c = np.argwhere(reflvl_cell <= self.max_level)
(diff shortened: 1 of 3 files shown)


graphs/src/anemoi/graphs/generate/icon_mesh.py
```diff
@@ -97,6 +98,7 @@ def __init__(self, icon_grid_filename: str, max_level: int | None = None):
 
         self.reflvl_vertex = reflvl_vertex
         self.max_level = max_level if max_level is not None else reflvl_vertex.max()
+        self.min_level = min_level
```
MeraX:
```suggestion
        self.min_level = min_level if min_level is not None else 0
```

graphs/src/anemoi/graphs/schemas/node_schemas.py
```diff
@@ -85,6 +85,8 @@ class ICONMeshNodeSchema(BaseModel):
     "Name of NetCDF ICON grid file."
     max_level: int
     "Maximum refinement level of the multi mesh / cell grid."
+    min_level: int | None = 0
```
MeraX:
```suggestion
    min_level: int = 0
```

MeraX:
Please add a test. You might add this option to the aicon test and adjust https://github.com/ecmwf/anemoi-core/blob/main/training/tests/integration/aicon/test_cicd_aicon_04_icon-dream_medium.py#L154

graphs/src/anemoi/graphs/generate/icon_mesh.py
```diff
@@ -77,9 +77,10 @@ class ICONMultiMesh:
     uuidOfHGrid: str
     reflvl_vertex: np.ndarray
     max_level: int
+    min_level: int
     nodeset: NodeSet  # set of ICON grid vertices
 
-    def __init__(self, icon_grid_filename: str, max_level: int | None = None):
+    def __init__(self, icon_grid_filename: str, max_level: int | None = None, min_level: int | None = 0):
```
fprill, line 83: `min_level: int | None = 0` -  I would find it more consistent if the default value were `None` (see also `node_schemas.py`)

graphs/src/anemoi/graphs/generate/icon_mesh.py
```diff
@@ -126,7 +128,7 @@ def cell_vertices(self) -> np.ndarray:
     def multi_mesh_edges(self) -> np.ndarray:
         """Returns the multi-mesh edges as an arrays of vertex indices."""
         # concatenate edge-vertex lists (= edges of the multi-level mesh):
-        edges = np.concatenate([edges for edges in self.edge_vertices], axis=0)
+        edges = np.concatenate([edges for edges in self.edge_vertices if (edges.size > 0)], axis=0)
```
fprill, line 131: `if (edges.size > 0)`: Wouldn't this be better formulated as an assertion?

graphs/src/anemoi/graphs/nodes/builders/from_icon.py
```diff
@@ -21,8 +21,10 @@ class BaseICONNodeBuilder(BaseNodeBuilder):
 
     icon_node_class: type[ICONCellDataGrid] | type[ICONMultiMesh]
 
-    def __init__(self, name: str, grid_filename: str, max_level: int) -> None:
-        self.icon_nodes = self.icon_node_class(icon_grid_filename=grid_filename, max_level=max_level)
+    def __init__(self, name: str, grid_filename: str, max_level: int, min_level: int | None = 0) -> None:
```
fprill, line 24: I'd prefer not to define a default value for `min_level` here.

graphs/src/anemoi/graphs/generate/icon_mesh.py
```diff
@@ -151,8 +153,12 @@ def _restrict_multi_mesh_edges_and_cells(
         num_vertices = vertex_mask.shape[0]
         vertex_glb2loc = np.full(num_vertices, -1, dtype=int)
         vertex_glb2loc[vertex_mask] = np.arange(vertex_mask.sum())
+        vertex_glb2loc_tmp = [vertex_glb2loc[vertices] for vertices in edge_vertices[: self.max_level + 1]]
```
fprill, line 156: This seems to be slow for large arrays: It contains explicit loops over every edge and then over every vertex inside the edge and it creates temporary Python lists.

graphs/src/anemoi/graphs/generate/icon_mesh.py
svenulbrich: As edge_size=0 appears with min_level>0 and results in some error, which is the reason for this line. If I understand your suggestion correctly, an assertion would result in an error in the corresponding line. I assume hence, that I misunderstand your comment. If you have a particular different code change in mind to prevent the error, I would appreciate such suggestion!

graphs/src/anemoi/graphs/generate/icon_mesh.py
svenulbrich: You are right. The code has been updated accordingly.

graphs/src/anemoi/graphs/nodes/builders/from_icon.py
svenulbrich: You are right. The code has been updated accordingly.

graphs/src/anemoi/graphs/generate/icon_mesh.py
svenulbrich: The corresponding lines have been repolaced. It is now more efficient. Doesn't generate a temporary variable and should be faster than the previous solution. Does it satisfy the requirements, or do you have a better programming idea for this change?

Issue #1327, opened 2026-08-16
Reduce anemoi-training startup time


frazane:
Tracking issue for a set of small fixes reducing `anemoi-training` startup time, which hurts every debug iteration. We profiled where the time goes (importtime, py-spy, cold/warm A-B runs on an HPC cluster with a Lustre filesystem) and are sending the fixes as separate PRs; this issue holds the shared context and tracks their status.

Measured on a cold compute node, a debug iteration was ~2.5 min to the first optimizer step:

- `anemoi-training --help`: 38–40s cold, 2.5s warm — the CLI schemas ended up importing torch.
- importing the train stack: 44–78s cold, ~8s warm — dominated by pytorch_lightning (torchmetrics eagerly imports every metric domain, dragging scipy.signal and torchvision) and torch_geometric (eagerly imports its `datasets`/`llm` subtree).
- pre-fit setup: ~22s, of which 6.7s was a builtin `sum()` iterating a 1.7M-element bool tensor element-wise in Python.

Most of the cold cost is filesystem I/O (the same import chain runs in ~8s from node-local storage), which users can mitigate site-side; the fixes here go after what anemoi itself controls.

### Anemoi-side PRs

Each is independent; PR links are added here as they open.

- [x] #1328 — scaler: replace builtin `sum(scaling_mask)` with `scaling_mask.sum()` (6.7s per startup)
- [x] #1329 — models: stop importing torch on the CLI path (`--help` 2.4s → 0.5s warm)
- [ ] training: make the plotting dependencies truly optional, so the `_check_plotting_dependencies` install hints can actually fire

### Watching upstream

Reports filed against dependencies; status updated here as they move.

- [ ] torchmetrics eager imports: Lightning-AI/torchmetrics#3457
- [ ] torch_geometric eager `datasets` import: pyg-team/pytorch_geometric#10772

### Parked

The `timm` dependency exists only for the default `CosineLRScheduler` but keeps torchvision installed, which triggers the most expensive torchmetrics import branch. Replacing it touches default configs, so it needs its own discussion; it will be linked here if opened.

cathalobrien:
@frazane can we close now that the linked pr is merged?

frazane:
> [@frazane](https://github.com/frazane) can we close now that the issue is merged?

There is still the third item in the anemoi-side fixes that has not been addressed, and the issues in the dependencies have not been closed. I would keep this issue open as a way of tracking all efforts related to reducing startup time. Does it make sense?

PR #1328, opened 2026-08-16
perf(training): use tensor.sum() on the scaling mask in ReweightedGraphNodeAttributeScaler

frazane:
## Description

`ReweightedGraphNodeAttributeScaler.reweight_attribute_values` divided by `sum(scaling_mask)` — Python's builtin `sum`, which iterates the boolean node mask one element at a time. Replacing it with `scaling_mask.sum()` gives the identical result at a fraction of the cost.

Performance. On a ~1.7M-node grid this single line cost 6.7s of every training startup (found with py-spy); a microbenchmark on that size drops from 6.29s to 40ms. The scaler runs once per training start, so the saving is per-job, but it is paid on every debug iteration and grows with the number of nodes.

Related to #1327. Found while profiling `anemoi-training train` startup.

## Additional notes

Behaviour is unchanged: `torch.Tensor.sum` on a bool tensor returns the same count. The 18 tests in `training/tests/unit/losses/test_loss_scaling.py` pass.

---

### DIFF
--- a/training/src/anemoi/training/losses/scalers/node_attributes.py
+++ b/training/src/anemoi/training/losses/scalers/node_attributes.py
@@ -105,7 +105,7 @@ class ReweightedGraphNodeAttributeScaler(GraphNodeAttributeScaler):
         scaling_mask = self.nodes[self.scaling_mask_attribute_name].squeeze()
         unmasked_sum = torch.sum(values[~scaling_mask])
         weight_per_masked_node = (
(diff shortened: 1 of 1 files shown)


cathalobrien (approved):
nice LGTM.

a quick grep shows 37 other uses of this pattern across anemoi :D
``` bash
cd anemoi-core/
grep -Ir ' sum(' */src/ | wc -l
   37
```

cathalobrien:
running accuracy tests [here](https://github.com/ecmwf/anemoi-core/actions/runs/32134736798)...passed

PR #1329, opened 2026-08-16
perf(models): keep torch off the anemoi-training CLI import path

frazane:
## Description

Inline the 13 transport default values as literals in the pydantic fields of `TransportSourceConfig` and `TransportConfig`, instead of importing them from `anemoi.models.transport.settings`. Nothing else changes.

The schemas sit on the CLI path, and importing `anemoi.models.transport.settings` runs the `transport` package `__init__`, which imports torch. So every `anemoi-training` invocation — including `--help` — imported torch. With this change the CLI no longer imports torch: `import anemoi.training.__main__` goes from 1.93s to 0.51s warm, and much more is saved on a cold filesystem cache. `anemoi-training train` is unaffected (it imports torch anyway).

Related to #1327.

## Additional notes

`PlotSettingsSchema` already inlines its defaults the same way. A new test (`models/tests/schemas/test_transport_schema_defaults.py`) checks every inlined default against the corresponding dataclass attribute, so the two cannot drift apart. `model_dump()` output is unchanged and both existing import paths keep working.

---

### DIFF
--- a/models/src/anemoi/models/schemas/models.py
+++ b/models/src/anemoi/models/schemas/models.py
@@ -25,10 +25,6 @@ from pydantic import PositiveFloat
 from pydantic import PositiveInt
 from pydantic import model_validator
 
-from anemoi.models.transport.settings import EdmSettings
-from anemoi.models.transport.settings import NoiseConditioningSettings
-from anemoi.models.transport.settings import StochasticInterpolantSettings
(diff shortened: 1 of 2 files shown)


VeraChristina (approved):
LGTM, thanks for adding the tests to keep defaults consistent!

PR #1332, opened 2026-08-17
fix(training): honour diagnostics.enable_checkpointing in the trainer

frazane:
## Description

`diagnostics.enable_checkpointing: false` did not stop checkpoints from being written: `pl.Trainer` was never given `enable_checkpointing`, so it kept Lightning's default of `True`. This PR forwards the config value.

With the flag false there is no `ModelCheckpoint` in the callbacks, so Lightning appends its own default one — no `dirpath`, a full checkpoint every epoch in the working directory, `system.output.checkpoints.root` ignored. We hit this as a disk-quota failure. `AnemoiEvaluator` had the same gap and now passes `enable_checkpointing=False`.

##  Additional notes ##

---


MicheleCattaneo (approved):
Thanks for spotting this issue. Proposed fix looks good to me.

PR #1333, opened 2026-08-17
fix(graphs): use the correct CUDA device when building graphs on multiple GPUs

### LINKED ISSUE #1331: Graph building crashes in multi-rank GPU launches: torch-cluster ops run on the wrong CUDA device
Building a graph from several ranks crashes when the GPUs are in `Exclusive_Process` mode. `get_distributed_device()` returns `cuda:{SLURM_LOCALID}`, but the *current* CUDA device stays 0, and the torch-cluster ops (`radius`/`knn`) run on the current device's stream (no device guard). So every rank touches GPU 0, which rank 0 owns, and the other ranks die with `CUDA error: CUDA-capable device(s) is/are busy or unavailable`. On non-exclusive clusters ranks other than 0 still create a CUDA context on GPU 0; we have not tested the outcome there.

Two more problems in the same path:

- When SLURM binds one GPU per task, each task sees one device but `SLURM_LOCALID` runs 0..N-1, so `cuda:{SLURM_LOCALID}` is an invalid ordinal on every rank but 0.
- `CutOffEdges` with `cutoff_factor` and a mask attribute passes GPU tensors to scikit-learn: `TypeError: can't convert cuda:N device type tensor to numpy`. This one also hits single-task runs.

Minimal reproducer, `srun` with 4 tasks on exclusive-mode GPUs — ranks above 0 crash, rank 0 passes; with the `set_device` line uncommented all pass:

```python
import os
import torch
from torch_cluster import radius

dev = torch.device(f"cuda:{os.environ['SLURM_LOCALID']}")
# torch.cuda.set_device(dev)
x = torch.rand(100_000, 3, device=dev)
y = torch.rand(1_000, 3, device=dev)
radius(x, y, 0.01, max_num_neighbors=1)
torch.cuda.synchronize(dev)
```

Observed on 4x A100 (`Exclusive_Process`), torch 2.10.0, torch-cluster 1.6.3. PR #431 picked a per-rank device, but it is never made current. Related: issue #430. We have a fix and will open a PR.

frazane:
## Description

In a multi-rank launch every process's *current* CUDA device stays 0, and torch-cluster ops use it internally, so all ranks touch GPU 0. On GPUs in `Exclusive_Process` mode this crashes graph building on every rank but 0 with `CUDA-capable device(s) is/are busy or unavailable`.

Scopes the current CUDA device around the torch-cluster call sites, makes `get_distributed_device()` valid when SLURM binds one GPU per task, and moves coordinates to CPU before the scikit-learn calls (which raised on GPU tensors).

Tested on 4 ranks / 4 A100s in exclusive mode: graph rebuild crashed on ranks 1-3 before, all ranks build and training runs after. It would be great if someone could try this on their own cluster, especially one whose GPUs are not in exclusive mode.

Closes #1331. Related to #430/#431.

##  Additional notes ##

---


cathalobrien:
nice, stupid high level question but do we want each non-zero rank to build their own copy of the graph? otherwise we could have them block and then read the graph built by rank 0 once it's ready. I've never seen a use case where each model needs a different graph, but maybe you know of one

frazane:
> nice, stupid high level question but do we want each non-zero rank to build their own copy of the graph? otherwise we could have them block and then read the graph built by rank 0 once it's ready. I've never seen a use case where each model needs a different graph, but maybe you know of one

Not a stupid question, I have a separate issue drafted for exactly this. But yeah we could also do it here directly. What do you recommend?

japols:
nice! for sharded dynamic graph building (future feature for training from observations) we want exactly this behaviour where each GPU builds it's own independent subgraph.

frazane:
> nice! for sharded dynamic graph building (future feature for training from observations) we want exactly this behaviour where each GPU builds it's own independent subgraph.

Nice, I didn't know. So in a way this PR already documents how that could be accomplished. Then should we merge as is for now, or do we still want to avoid building the same graph multiple times?

graphs/src/anemoi/graphs/utils.py
```diff
@@ -28,27 +29,41 @@ def get_distributed_device() -> torch.device:
...
-        device = "cpu"
+        device = torch.device("cpu")
 
     return device
 
 
+def current_device_context(device: torch.device | str) -> contextlib.AbstractContextManager:
+    """Scoped switch of the current CUDA device; no-op for CPU.
+
+    Some extension ops (e.g. torch-cluster) use the current CUDA device internally, so
```
cathalobrien: I don't think the functions existence needs to be justified in the docstring

graphs/src/anemoi/graphs/utils.py
```diff
@@ -57,7 +72,7 @@ def get_nearest_neighbour(coords_rad: torch.Tensor, mask: torch.Tensor | None =
 
     nearest_neighbour = NearestNeighbors(metric="euclidean", n_jobs=4)
 
-    nearest_neighbour.fit(coords_rad)
+    nearest_neighbour.fit(coords_rad.cpu())
```
cathalobrien, line 69: why has the explicit move to cpu been added? was it happening implicitly before?

graphs/src/anemoi/graphs/utils.py
```diff
@@ -70,16 +85,16 @@ def get_grid_reference_distance(coords_rad: torch.Tensor, mask: torch.Tensor | N
...
-        mask to remove nodes, by default None
+        Mask to remove nodes, by default None.
 
     Returns
     -------
     float
         The reference distance of the grid.
     """
-    xyz = latlon_rad_to_cartesian(coords_rad)
+    xyz = latlon_rad_to_cartesian(coords_rad).cpu()
```
cathalobrien, line 91: same as above

graphs/src/anemoi/graphs/utils.py
```diff
@@ -28,27 +29,41 @@ def get_distributed_device() -> torch.device:
         import os
 
         local_rank = int(os.environ.get("SLURM_LOCALID", 0))
-        device = torch.device(f"cuda:{local_rank}")
+        device = torch.device(f"cuda:{local_rank % torch.cuda.device_count()}")
```
cathalobrien: so this is added to support running multiple processes per a single GPU? Seems out of scope for this PR. why should we change it from the current behaviour

cathalobrien (changes requested):
Nice PR thanks, I'm happy to proceed with this and build the graphs locally on each device. I have some comments

graphs/src/anemoi/graphs/utils.py
frazane: Apparently it wasn't implicit, I had `TypeError: can't convert cuda:0 device type tensor to numpy` without this.

graphs/src/anemoi/graphs/utils.py
frazane: same as above, too

cathalobrien (approved):
LGTM. thanks for this PR

PR #1339, opened 2026-08-21
fix(training): restore rollout and dataloader epoch on restart

ssmmnn11:
- Save and restore the dataloader epoch in training checkpoints
- Make it possible to configure if dataloader worker persist or not
- Automatically disable persistent workers when rollout changes between epochs
- Improve documentation of restart reproducibility
- Add tests for epoch restart and deterministic data shuffling
- Fix epoch rollout increment application

----
📚 Documentation preview 📚: https://anemoi-training--1339.org.readthedocs.build/en/1339/

----
📚 Documentation preview 📚: https://anemoi-graphs--1339.org.readthedocs.build/en/1339/

----
📚 Documentation preview 📚: https://anemoi-models--1339.org.readthedocs.build/en/1339/


mc4117 (dismissed):
LGTM!

training/src/anemoi/training/data/datamodule.py
```diff
@@ -162,15 +163,25 @@ def set_epoch(self, epoch: int) -> None:
...
+        """Return whether workers persist, disabling persistence for a changing rollout."""
+        persistent_workers = self.config.dataloader.persistent_workers
         rollout = getattr(self.task, "rollout", None)
-        if rollout is None:
-            return True
-        # Workers could also be persisted once rollout.step >= rollout.maximum,
-        # but that would make resumed runs behave differently from uninterrupted
-        # runs.
-        return rollout.epoch_increment == 0
+        if persistent_workers and rollout is not None and rollout.epoch_increment > 0:
```
pinnstorm, line 182: Nice! This also fixes a bug I was having trying to resume with `load_weights_only==false` and `rollout.epoch_increment > 0`, leading to the dataset not being resized and a slicing error on the resume. For checkpoints written before this PR comes in (which wont contain the `AnemoiDatasetsDataModule` key) I'm not sure if it is also worth us adding the 3-line fix I had for this as a fallback, I am happy either way. This is here: https://github.com/ecmwf/anemoi-core/blob/b88f88a13dc005a618eee4832fd4ec58c30a394e/training/src/anemoi/training/train/methods/base.py#L501

pinnstorm (dismissed):
This looks good. Just added a comment with the fix I had previously. Very happy to just go with this as is 👍

training/src/anemoi/training/data/datamodule.py
ssmmnn11: Indeed!

training/docs/user-guide/training.rst
```diff
@@ -98,19 +98,43 @@ parallel reductions can still differ slightly.
...
+The workers' random number generator state is not stored in checkpoints.
+Consequently, when ``persistent_workers`` is ``true``, a restarted job does not
+continue the same shuffle sequence as an uninterrupted run, even when it uses the
+same base seed.
+
+Anemoi does not store a dataloader's position within an epoch. A checkpoint
+written during an epoch therefore cannot resume the exact data sequence at the
+next batch and may repeat or skip samples. Use checkpoints saved at the end of an
+epoch and set ``persistent_workers`` to ``false`` when the data sequence must match
+uninterrupted training.
```
gabrieloks, line 137: With persistent_workers=True, we can force an identical data sequence across jobs provided every run restarts at the same controlled epoch boundaries (e.g. split a 60-epoch run at max_epochs 20, 40, 60) and uses the same base seed, though this sequence won't match an uninterrupted run.

Maybe we could add this info here

training/docs/user-guide/training.rst
```diff
@@ -544,35 +592,53 @@ specific timesteps.
...
+used.
 
 .. note::
 
-   When resuming from a checkpoint that contains saved rollout state,
-   ``rollout.start`` is ignored entirely — the rollout step is always
-   restored unconditionally from the checkpoint.
+   Fully resuming a training run restores the optimiser and scheduler state,
+   current rollout step, and dataloader epoch from the checkpoint. The saved
+   ``rollout.step`` overrides ``rollout.start`` from the current configuration.
```
gabrieloks, line 642: Should we add a test for this, to make sure that rollout.start is overwritten by rollout.step from the checkpoint ? Just to really make sure we enforce this behaviour

training/tests/unit/train/test_checkpoint_loading.py
```diff
@@ -439,10 +450,76 @@ def test_on_load_checkpoint_restores_rollout_step() -> None:
...
+
+    checkpoint = {
+        "task_state": {"rollout": {"step": 1, "last_increased_epoch": -1}},
+        "hyper_parameters": {"data_indices": {"data": DummyIndex()}},
+        "state_dict": {},
+    }
+    BaseTrainingModule.on_load_checkpoint(module, checkpoint)
+
+    assert task.rollout.step == 2
+    assert task.rollout._last_increased_epoch == -1
```
gabrieloks, line 486: The assert _last_increased_epoch == -1 doesn't really prove anything since the checkpoint already stores -1, which is the same as the fresh default, so if we accidentally restored it from the checkpoint the test would still be green. The step check is fine since the checkpoint's 1 differs from the configured 2.

training/tests/unit/train/test_checkpoint_loading.py
```diff
@@ -439,10 +450,76 @@ def test_on_load_checkpoint_restores_rollout_step() -> None:
...
+
+    BaseTrainingModule.on_train_start(module)
+
+    assert caplog.messages == ["Effective task rollout step: 2."]
+
+
+def test_on_load_checkpoint_load_weights_only_keeps_configured_rollout_step() -> None:
+    """Loading only model weights must not restore the checkpoint's rollout state."""
+    module, task = _make_module_with_forecaster_task(
+        {"start": 2, "epoch_increment": 0, "maximum": 2},
```
gabrieloks: We also never try weights-only together with epoch_increment > 0. Would be good to have a test that loads weights-only and confirms the schedule starts from a clean clock.

Could we bump the checkpoint to something clearly non-default (e.g. step: 4, last_increased_epoch: 3) with epoch_increment: 2, then drive on_train_epoch_end and assert the first increment lands where we expect?

gabrieloks:
Left some comments but looks good overall. We might want to complexify a little bit the rollout weights only tests to make sure we are covering edge cases.

training/docs/user-guide/training.rst
ssmmnn11: good point!

gabrieloks (approved):
LGTM! Great work

PR #1341, opened 2026-08-25
ci: run the HPC integration tests on fork pull requests

mcocdawc:
## Description

Put the integration tests behind `pull_request_target`.
This has serious security implications, hence use the `approved-for-ci` label to gate against it.

We retain the `run-expensive-tests` and `expensive-tests-not-needed` labels. These are relevant for both maintainers and outside developers and stick, i.e. are not removed upon a new push.

The `approved_for_ci` label is only relevant for outside contributors and is removed upon a push.

##  Additional notes ##


.github/workflows/integration-tests-hpc.yml
```diff
@@ -15,18 +15,53 @@ on:
...
+  # The action answers a different question -- may a fork's code run on our HPC
+  # account -- and FAILS when it may not. Never merge the two: an `if:` on the
+  # approval label would report Success for "nothing ran".
+  ci-approval:
+    name: ci-approval
+    if: ${{ github.event_name != 'pull_request_target' || contains(github.event.pull_request.labels.*.name, 'run-expensive-tests') }}
+    # No checkout, no Python; the action needs bash, jq, and gh only to revoke.
+    runs-on: ubuntu-slim
+    permissions:
+      pull-requests: write   # only so the label can be revoked again
```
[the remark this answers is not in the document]
mcocdawc: it is a github runner

.github/workflows/integration-tests-hpc.yml
```diff
@@ -16,48 +16,88 @@ on:
...
-            # here and reused by each test, then removed by the cleanup-env job.
+            # The virtualenv is built once here, next to the source the step
+            # above pushed, and reused by each test.
             WORKDIR="$PERM/TESTING/integration-venvs/${{ github.run_id }}"
-            REPO_NAME=anemoi-core
-            REPO_BRANCH=${{ github.head_ref || github.ref_name }}
-            rm -rf "$WORKDIR"
-            mkdir -p "$WORKDIR"
             cd "$WORKDIR"
-            git clone -b $REPO_BRANCH https://${{ secrets.GH_REPO_READ_TOKEN }}@github.com/ecmwf/${REPO_NAME}.git
```
mcocdawc, line 56: note I removed the git clone, that's why I added you @cathalobrien 

Instead we have the `checkout-under-test` action now, which correctly checks out the code of the fork PR.
Which requires `allow-unsafe-pr-checkout: 'true'` which is gated behind a `needs: ci-approval`

.github/workflows/integration-tests-hpc.yml
```diff
@@ -16,48 +16,88 @@ on:
           - test_training_cycle
           - test_accuracy
   pull_request:
-    types: [synchronize, labeled]
+    types: [synchronize]
```
VeraChristina, line 19: Why remove the 'labeled' here? If I understand correctly, this means that: When a developer adds the `run-expensive-tests label` before merging the PR, the tests won't run by default (unless an additional commit is pushed).
This would make it easier to forget to run them, counter to the intention of the labels.

.github/workflows/integration-tests-hpc.yml
```diff
@@ -16,48 +16,88 @@ on:
...
+  ci-approval:
+    name: ci-approval
+    # Comments must stay out here: inside the block below they would be part of
+    # the expression, which has no comment syntax.
+    #  - no pull request (schedule, dispatch, merge_group): always run
+    #  - a label event: only when the label just added is 'run-expensive-tests'
+    #  - a push: when the pull request carries 'run-expensive-tests'
+    if: >-
+      ${{
+        !github.event.pull_request ||
```
VeraChristina, line 39: does this include all pull_request_target triggers (i.e. without the expensive-tests label)?

VeraChristina (commented):
Thanks for making this work! Two questions on the workflow logic below

.github/workflows/integration-tests-hpc.yml
mcocdawc: because we have the labeled in `pull_request_target`.

I want to have as much as possible not behind `pull_request_target` and only trigger it if someone adds "approved-for-ci". Everything else can remain behind `pull_request`

Issue #1330, opened 2026-08-16
Non-deterministic CheckpointError when compiled modules sit inside gradient-checkpointed regions


frazane:
Training with `model.compile` and gradient checkpointing crashes non-deterministically with `torch.utils.checkpoint.CheckpointError: A different number of tensors was saved during the original forward and recomputation` (e.g. 204 vs 196). Some ranks fail with the sibling message instead — recomputed tensors have different metadata than during the forward pass, with a genuine shape mismatch at the same position (e.g. saved `[95344, 8]` vs recomputed `[95344, 4]`). Observed with a graphtransformer_ens config compiling `ConditionalLayerNorm` (gpm=2/gpe=4, 2 nodes x 4 A100, torch 2.10.0+cu126); roughly half of runs crash in a backward pass for a susceptible seed, others complete. Failed ranks then leave the survivors hanging in a collective, so the job wedges rather than exiting.

Root cause, confirmed with `TORCH_LOGS=recompiles`: dynamo guards compiled code on process-global state including the intra-op thread count. Checkpoint recomputation runs on autograd worker threads whose thread-local `num_threads` differs from the main thread's, so the compiled module's GLOBAL_STATE guard fails there and a second dynamo cache entry is created on the first backward (`Recompiling function forward in .../layers/normalization.py:75 ... GLOBAL_STATE changed: num_threads`, logged in every run). From then on forward and recomputation can select different compiled graphs (LRU tiebreak, pytorch/pytorch#166926; same symptom class as pytorch/pytorch#144035), and when they do the saved-tensor counts diverge. Whether a given backward crashes is a lottery, which makes the failure look flaky and seed-dependent. Setting `OMP_NUM_THREADS` does not help — the mismatch is thread-local ATen state, not the environment default.

Reproduction is probabilistic; with a susceptible seed we measured 3 crashes in 6 identical runs at current main, and the recompile fingerprint appears in passing runs too. Disabling compilation (`model.compile=[]`) avoids it at a performance cost.

cc: @cathalobrien @japols

cathalobrien:
Hey @frazane, what about setting 
```
export TORCHINDUCTOR_CACHE_DIR=$TMPDIR
```
the error only occurs in multi-node runs right? (or to be more specific, has only been observed in multi-node runs :D) 

[...]

PR #1344, opened 2026-08-26
fix(training): fix compile error when using activation checkpointing

cathalobrien:
## Description
Fixes an error when using torch compiled code inside a checkpointed region.

the error was reported [here](https://github.com/ecmwf/anemoi-core/issues/1330) with more background [here](https://github.com/pytorch/pytorch/issues/166926). The fix comes from [here](https://github.com/pytorch/torchtitan/pull/1991) and is only available for torch 2.10 and above. I have added a warning if you use compilation and torch < 2.10 which lists some workarounds. but the proper solution is upgrading torch.

It also fixes another cause of recompilation errors: `torch._inductor.config.shape_padding` can pad tensors based on micro-benchmarks. if you recompile inside a checkpointed reason this can lead to shape mismatches. I disable this feature when using gradient checkpointing. I also warn users about other settings e.g. "mode='max-autotune'" which could lead to errors when used with gradient checkpointing. 

Lastly, I set a default number of CPU threads. Otherwise this was being reset multiple times by PytorchLightning during runtime. Each time it was set, it would trigger a needless recompilation, which used up one of the 'model.recompile_limit' recompilations

I got Claude to generate a reproducer, using the compiled conditional layer norm in anemoi (see [here](https://github.com/ecmwf/anemoi-core/blob/3cb72df1f1767b8588b518b0c6e9af98f3e4893c/reproduce_compile_checkpoint_lru.py)) the error fires when the cache is enabled, and doesn't fire when the cache is disabled.


training/src/anemoi/training/utils/compile.py
```diff
@@ -102,6 +103,20 @@ def prepare_compilation(
...
+
+    # disable LRU cache, this is a fix for https://github.com/pytorch/pytorch/issues/166926
+    # The runtime impact of this should be marginal
+    if version.parse(torch.__version__) >= version.parse("2.10.0"):
+        torch._C._dynamo.eval_frame._set_lru_cache(False)
+    else:
+        LOGGER.warning(
+            "Could not disable torch compile LRU cache because torch version is < 2.10.0. "
+            "This may result in runtime errors when using torch.compile() alongside activation checkpointing. "
+            "If you encounter errors, consider upgrading to torch >= 2.10.0, disabling torch.compile() (model.compile=[]) "
```
ssmmnn11:
```suggestion
            "If you encounter errors, consider either upgrading to torch >= 2.10.0, or disabling torch.compile() (model.compile=[]) "
```

ssmmnn11 (dismissed):
Excellent. Minor suggestion only.

VeraChristina (dismissed):
Thanks for fixing this! LGTM

VeraChristina (dismissed):

VeraChristina (approved):
Disabling torch compile in the edm tests, which were the ones still failing with this error, seems like a good temporary solution (as that it also the solution suggested to the user in the logs)

PR #1347, opened 2026-08-27
some suggestions

ssmmnn11:
some fixes / suggestions


VeraChristina:
Thanks for the sharding fixes and suggestions!

PR #1348, opened 2026-08-27
feat: Add flexible grid defintion in schema

### LINKED ISSUE #1266: Restrictive grid specifications
At the moment the `node_schema` prescribes the grid supported in `ReducedGGNodes`, as utils / transforms is the provider of the grids, this feels overly restrictive, and should perhaps just do validation checking if it seems like a valid graph.

https://github.com/ecmwf/anemoi-core/blob/d0cfe4c88198d0aa9f9acf47009e4822734731b0/graphs/src/anemoi/graphs/schemas/node_schemas.py#L71

Furthermore, this grids code from utils seems deprecated so likely needs to be updated to use from transforms.

HCookie:
## Description

Add flexible grid definition in schema
- Maintain existing set as examples

Resolves #1266

### DIFF
--- a/graphs/src/anemoi/graphs/schemas/node_schemas.py
+++ b/graphs/src/anemoi/graphs/schemas/node_schemas.py
@@ -17,6 +17,7 @@ from typing import Optional
 from pydantic import Field
 from pydantic import PositiveFloat
 from pydantic import PositiveInt
+from pydantic import field_validator
 
 from anemoi.utils.schemas import BaseModel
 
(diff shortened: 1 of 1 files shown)


graphs/src/anemoi/graphs/schemas/node_schemas.py
```diff
@@ -68,9 +69,19 @@ class XArrayNodeSchema(BaseModel):
 class ReducedGaussianGridNodeSchema(BaseModel):
     target_: Literal["anemoi.graphs.nodes.ReducedGaussianGridNodes"] = Field(..., alias="_target_")
     "Nodes from NPZ grids class implementation from anemoi.graphs.nodes."
-    grid: Literal["o16", "o32", "o48", "o96", "o160", "o256", "o320", "n320", "o1280"]
+    grid: str = Field(..., examples=["o16", "o32", "o48", "o96", "o160", "o256", "o320", "n320", "o1280"])
```
mc4117, line 72: Does this change cause any problems later in the code? (ie. is the code only expecting one of these formats?)

JPXKQX: I think that we want the code to fail as early as possible when a grid (e.g. `o33`) is specified. But there are cases where the grid may make sense, but it's not available. In this case, the user should open an issue so a maintainer can pre-generate the grid so that it's available to users.

HCookie: So, do we merge this or keep it restrictive?

JPXKQX: I would lean towards keeping it as `Literal`

HCookie:
Keeping it as literal to break early rather than late

PR #1354, opened 2026-08-30
fix(training): replace memory efficient but unstable graph edge energ score formulation

ssmmnn11:
- Fix the unstable graph edge energy score formulation (can produce NaNs due to cancelation).
- Add a custom autograd operation to reduce memory of revised formulation.
- Replace the edge-score base class with shared helpers to simplify class hirachy.

revised class structure:
```
  BaseGraphScoreLoss
  ├── BaseGraphEnergyScoreLoss
  │   ├── GraphEnergyScoreLoss
  │   └── GraphEdgeEnergyScoreLoss
  └── BaseGraphEdgeScoreLoss
      ├── GraphEdgeCRPSLoss
      └── GraphVariogramScoreLoss

  Now:

  BaseGraphScoreLoss
  ├── BaseGraphEnergyScoreLoss
  │   ├── GraphEnergyScoreLoss
  │   └── GraphEdgeEnergyScoreLoss
  ├── GraphEdgeCRPSLoss
  └── GraphVariogramScoreLoss
```


cathalobrien:
nice contribution. so as I understand, you rewrote edge_differences because the old version was unstable. because with floating point arithmetic, when you had two large and similar but not identical values, the difference is very small and can be cancelled out by floating point rounding errors.
```
        return node_values[..., source_index, :] - node_values[..., destination_index, :]
```
the new version is more stable, but has to create lots of intermediate tensors (edge_difference, gathered_ab, row_max) to ensure stability. so its less memory efficient.

to partially mitigate this memory cost, you wrote the custom autograd function. Rather then saving the intermediate tensors until the backward pass, you recompute them in the backward.

On top of this, you moved a bunch of functions into a shared file and renamed some variables in all the other classes, which account for the other files changed.

you also added a test for the case where you have two large and similar but not identical numbers. 
```
prediction = torch.tensor(
        [[[[[0.0], [10_000.0], [10_001.0]], [[0.0], [10_000.0], [10_001.0]]]]],
        requires_grad=True,
    )
```

Seems like a reasonable fix, with appropriate mitigations for the increased memory required. LGTM

cathalobrien (approved):

PR #1355, opened 2026-08-31
pr(training): some fixes and suggestions

ssmmnn11:
some suggestions

----
📚 Documentation preview 📚: https://anemoi-training--1355.org.readthedocs.build/en/1355/

----
📚 Documentation preview 📚: https://anemoi-graphs--1355.org.readthedocs.build/en/1355/

----
📚 Documentation preview 📚: https://anemoi-models--1355.org.readthedocs.build/en/1355/

### DIFF
--- a/training/docs/modules/tasks.rst
+++ b/training/docs/modules/tasks.rst
@@ -139,17 +139,17 @@ Rollout behavior:
  OffsetForecaster
 ******************
 
-:class:`~anemoi.training.tasks.forecasting.OffsetForecaster` is an
+:class:`~anemoi.training.tasks.forecaster.OffsetForecaster` is an
 experimental generalisation of
-:class:`~anemoi.training.tasks.forecasting.Forecaster` that declares the
+:class:`~anemoi.training.tasks.forecaster.Forecaster` that declares the
 sample directly through time offsets, i.e. times relative to the forecast initialisation time:
 
 - ``input_offsets`` — input time offsets as frequency strings, e.g.
-  ``["-1H", "-30m", "0m"]``.
+  ``["-1H", "-30m", "0m"]``. The latest input must be ``0H``.
 - ``output_offsets`` — target time offsets as frequency strings, e.g. ``["30M",  "1h"]``.
 
 - ``rollout`` / ``validation_rollout`` — as for
-  :class:`~anemoi.training.tasks.forecasting.Forecaster`.
+  :class:`~anemoi.training.tasks.forecaster.Forecaster`.
 
 Examples
 ========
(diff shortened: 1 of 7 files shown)


training/docs/modules/tasks.rst
```diff
@@ -139,17 +139,17 @@ Rollout behavior:
...
-:class:`~anemoi.training.tasks.forecasting.OffsetForecaster` is an
+:class:`~anemoi.training.tasks.forecaster.OffsetForecaster` is an
 experimental generalisation of
-:class:`~anemoi.training.tasks.forecasting.Forecaster` that declares the
+:class:`~anemoi.training.tasks.forecaster.Forecaster` that declares the
 sample directly through time offsets, i.e. times relative to the forecast initialisation time:
 
 - ``input_offsets`` — input time offsets as frequency strings, e.g.
-  ``["-1H", "-30m", "0m"]``.
+  ``["-1H", "-30m", "0m"]``. The latest input must be ``0H``.
```
dietervdb-meteo: Ok with imposing the convention last offset is 0. This does mean we could drop it from the list in the config. Better to keep it to be more explicit to users, or go for the least info required (and saving on some users misconfiguring and having go through an errror message)?

training/src/anemoi/training/tasks/forecaster.py
```diff
@@ -465,19 +471,15 @@ def _convert_and_validate(
 
         # Check if the rollout shift is valid or replace "default" by the maximum valid shift.
 
-        # A shift S is valid if it is strictly positive, the shifted input offsets
-        # are contained in the union of input and output offsets, and no pairwise
-        # difference of output offsets is a multiple of S (which would cause the
-        # same output time step to be forecasted more than once across rollout steps).
+        # A shift S is valid if the shifted input offsets are contained in the
+        # union of input and output offsets and every output of one rollout step
```
dietervdb-meteo, line 475: Making the set of allowed `OffsetForecaster`s a little smaller by imposing the stronger chronological condition is ok, since the examples we exclude this way appear indeed rather fringe and possibly of no (current) interest. Still, is there any reason [other than save on a few lines of validation code] where you would see things go wrong for non chronological order of the output? Although I must admit it might complicate things in inference a bit. Happy to go for this stricter definition, just curious if this choice was driven by an actual technical constraint or rather to be prudent in generalizing.

training/docs/modules/tasks.rst
MeraX: Is this line self contradicting? If the latest input must be `0H`, how can it be `0m` in the example?

dietervdb-meteo: Since both are a string representation of a zero `timedelta` they are equivalent. But I agree that for documentation it might be a bit confusion stated this way.

dietervdb-meteo: Let me postpone this, we can do this with an absolutely minor PR in the future if we feel dropping 0 from the list would be an improvement.

dietervdb-meteo: I changed the wording to 'must be equivalent to'

training/src/anemoi/training/tasks/forecaster.py
dietervdb-meteo: Better to be safe than sorry, can be always relaxed later if a use case arises.

training/docs/modules/tasks.rst
MeraX: Sounds good to me. Thank you for your consideration.

PR #1357, opened 2026-09-01
perf: Results from prototype autoresearch improvements on GT triton kernel

sophie-xhonneux:
> [!WARNING]
> **TODO — needs testing on GPU before merge.** The Triton kernels in this PR have not been
> run from within anemoi-core yet, only from Jan's harness:. Everything below about performance is carried
> over from the autoresearch branch, and everything about correctness rests on a CPU
> simulation of the index arithmetic. Do not merge on that basis.
>
> - [ ] Run `models/tests/integration/triton/test_triton_gt.py` on a GPU (incl. the new
>       `..._src_tiling` case, which is the only coverage of the `BLOCK_E` heuristic)
> - [ ] Confirm the kernels actually compile for non-power-of-2 `H`/`C` — the tiled src pass
>       builds its masks through new `constexpr` branches that no CPU check exercises
> - [ ] Re-measure fwd+bwd on real graphs to confirm the speedups reproduce here, since the
>       numbers below were taken against a kernel without anemoi's padding/masking
> - [ ] Check bwd peak memory against `main` on a large graph (`dop_dec_large`), to confirm
>       the `2 * M * H` handoff is as cheap in practice as the arithmetic suggests

## What

The Triton GraphTransformer backward had its src pass recompute everything the dst pass
had already derived. This transplants an optimised backward found by an autoresearch run
(`sophie-xhonneux/gt-gym-auto-research@v1`), adapted onto the padded/masked kernel
structure that `anemoi-core` has since grown.

Forward is untouched. Gradient values and accumulation order are unchanged.

## The change

Before, `_gt_bwd_src_pass` reloaded `k`/`v`/`e`/`m`/`D` per edge, redid the `q·ke` dot
product and rebuilt `alpha` and `dS` — all of which `_gt_bwd_dst_pass` had already
computed — and reached each edge through an `edge_ids` indirection, so its per-edge reads
were scattered.

1. **Per-edge `(alpha, dS)` handoff.** The dst pass writes the `alpha` and `dS` it computes
   anyway into two `[M, H]` fp32 buffers. The src pass then needs no `k`/`v`/`e`/`m`/`D`
   loads and no dot product: per edge it reads only `alpha`/`dS` and the two dst-side rows.
2. **`dE` moves to the dst pass**, where the per-edge stores are contiguous in CSC order
   rather than scattered through `edge_ids`.
3. **The handoff is held in CSR order**, via a new `csr_pos` permutation (CSC edge id → CSR
   slot) plus `edge_dst_csr`. The dst pass scatters its stores, which is fire-and-forget;
   the src pass in exchange reads contiguous streams, and its dependent-load chain shrinks
   from `edge_ids → edge_dst → q` to `edge_dst → q`.
4. **The src pass is tiled** (`BLOCK_E`) so that several of the remaining indirect gathers
   are in flight at once. The tile width follows the mean source degree (`>= 4 → 4`, else
   `1`): at degree ~1 the masked-off lanes are pure overhead. The forward and dst passes are
   already bandwidth-saturated and stay untiled.
5. The per-destination `m_j` load is hoisted out of the dst pass edge loop, and the
   now-unused `D` buffer is gone.

## Measurements

See the screenshots

## Correctness

- Same values, same expressions, same accumulation order, with one exception: the tiled src
  pass sums per tile, which reassociates the adds. On the autoresearch branch, strict fp32
  and scale-8 fp16 checks matched the unoptimised kernel, and tiling diffs were ~2e-6
  against a 4e-5 tolerance.
- `test_triton_gt.py` gains `test_graph_transformer_vs_reference_backward_src_tiling`. It
  pins the tile width by building fixed-degree graphs, so both branches of the heuristic are
  covered, with and without H/C padding.
- Dropping the src pass's `m_j`/`Dj` loads also closes a latent out-of-bounds read: both were
  unmasked, so they read past the padded `H` lanes whenever `num_heads` is not a power of two.

## Memory

The handoff costs `2 * M * H` fp32, transient within the backward — 2/C of the `dE` already
allocated there, so ~1.6% on top of it at C=128.

## API

`edge_index_to_csc(..., reverse=True)` now returns a 4-tuple
`(rowptr, edge_id_per_src, edge_dst_csr, csr_pos)`, and the
`anemoi::graph_transformer_attention` op takes `edge_dst_csr, csr_pos` where it took
`edge_ids, edge_dst` (arity unchanged). `block.py` passes `reverse` through opaquely and
needed no change.

## Evidence

Prior benchmark results
image

Improved
 
image

### DIFF
--- a/models/src/anemoi/models/triton/gt.py
+++ b/models/src/anemoi/models/triton/gt.py
@@ -191,7 +191,10 @@ def _gt_bwd_dst_pass(
     COLPTR_ptr,  # [N_dst + 1]
     D_OUT_ptr,  # [N_dst * H * C]
     D_Q_ptr,  # OUT
-    D_ptr,  # [N_dst * H]
+    D_E_ptr,  # OUT [M * H * C] per-edge dE; stores are contiguous in CSC order here
+    ALPHA_ptr,  # OUT [M * H] fp32 per-edge alpha, written to CSR slots for the src pass
+    DS_ptr,  # OUT [M * H] fp32 per-edge dS (alpha folded in), written to CSR slots
+    CSRPOS_ptr,  # [M] CSC edge id -> CSR slot; makes the src pass's reads contiguous
     N_dst,
     H: tl.constexpr,
     C: tl.constexpr,
@@ -212,9 +215,7 @@ def _gt_bwd_dst_pass(
     num_edges = neigh_end - neigh_start
 
     if num_edges == 0:
-        # store D_j = <d_out, out> = 0 and dQ = 0
-        zeros = tl.zeros((H_pad,), dtype=tl.float32)
-        tl.store(D_ptr + dst_idx * H + tl.arange(0, H_pad), zeros, mask=H_mask)
+        # no incident edges: dQ = 0, and there are no per-edge outputs to write
         zeros = tl.zeros((H_pad * C_pad,), dtype=out_dtype)
         tl.store(D_Q_ptr + dst_off, zeros, mask=H_C_mask)
         return
@@ -227,6 +228,9 @@ def _gt_bwd_dst_pass(
 
     q = tl.load(Q_ptr + dst_off, mask=H_C_mask).to(tl.float32).reshape((H_pad, C_pad))
     dq = tl.zeros((H_pad, C_pad), dtype=tl.float32)
+
+    # m_j is per-destination, so it is loaded once instead of on every edge
+    m_j = tl.load(M_ptr + dst_idx * H + tl.arange(0, H_pad), mask=H_mask).to(tl.float32)
 
     edge_ptr = E_ptr + neigh_start * H * C + H_C_off  # pointer to first edge_attr
     e_idx = neigh_start  # first edge index
@@ -242,7 +246,6 @@ def _gt_bwd_dst_pass(
 
         ke = k + e
         # score and alpha using saved M
-        m_j = tl.load(M_ptr + dst_idx * H + tl.arange(0, H_pad), mask=H_mask).to(tl.float32)
         s_ij = tl.sum(q * ke, axis=-1) * qk_scale
         alpha_ij = tl.exp(s_ij - m_j)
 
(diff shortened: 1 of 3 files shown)


models/src/anemoi/models/triton/gt.py
```diff
@@ -489,13 +509,28 @@ def torch_dtype_to_triton(dtype):
...
+        q, k, v, e, out_saved, m, row, colptr, d_out, dQ, dE, alpha_e, dS_e, csr_pos, N_dst, H, C, grad_dtype
+    )
 
-    # Pass B: source nodes (accumulate dK, dV, dE)
-    _gt_bwd_src_pass[(N_src,)](q, k, v, e, rowptr, edge_ids, edge_dst, D, m, d_out, dK, dV, dE, N_src, H, C, grad_dtype)
+    # Pass B: source nodes (accumulate dK, dV)
+    # The tile width follows the mean source degree: high-degree sources are few
+    # programs with long gather chains and profit from concurrent gathers, whereas at
+    # degree ~1 the masked-off lanes are pure overhead.
+    block_e = 4 if num_edges >= 4 * N_src else 1
```
cathalobrien, line 530: Why 4? you could try use autotune to determine if there are more suitable values?

models/src/anemoi/models/triton/gt.py
```diff
@@ -242,7 +246,6 @@ def _gt_bwd_dst_pass(
 
         ke = k + e
         # score and alpha using saved M
-        m_j = tl.load(M_ptr + dst_idx * H + tl.arange(0, H_pad), mask=H_mask).to(tl.float32)
         s_ij = tl.sum(q * ke, axis=-1) * qk_scale
         alpha_ij = tl.exp(s_ij - m_j)
```
cathalobrien, line 250: you're not doing the exp2 trick here?

cathalobrien (commented):
Thanks for the contribution. it would be great to see a correctness test comparing the loss of two real anemoi training runs, with the same seed with this branch compared to main.

models/src/anemoi/models/triton/gt.py
japols: We might actually need a constant heuristic here since dynamic graphs would lead to re-compilations and re-benchmarking even when using something like `int(avg_src_degree)` as a key.

But I agree that we should further explore the heuristic itself.

japols:
Tested this on a 9 km / lvl7 GT on a single H200. Per training iteration, GT time goes from 336 → 294 ms (1.14× speedup / 12.3% reduction), saving ~41 ms/iteration. With an end-to-end iteration time of ~5.88 s, this translates to roughly a 0.7% overall training speedup.
Looking at the per-kernel breakdown (per iteration) we see that the faster src pass offsets the additional work moved into the dst pass: fwd 139→139 ms, bwd_dst 79→117 ms, bwd_src 118→39 ms.

sophie-xhonneux:
Is there a remaining blocker?

cathalobrien:
still need to see a comparison with a real training run with a fixed seed

plus the accuracy tests are failing, not necessarily a problem since the numerics are changed. but increases the need for a real training run comparison

sophie-xhonneux:
Can you provide a reference run and the relevant configs & commands & datasets to reproduce it so I can do the test you are asking for?

PR #1359, opened 2026-09-02
fix(tests): use importlib import mode for test discovery

bdvllrs:
## Description
As test structure does not include any `__init__.py` files, this allows a better test
discovery as explained here: https://docs.pytest.org/en/stable/explanation/pythonpath.html.

Tests with the same file name were causing problems with the default prepend import mode.

In particular, I had an issue in #1351 because `test_config.py` appears two times in `training/tests/unit`.

##  Additional notes ##

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


bdvllrs:
Well actually it doesn't work because test seem to import one another in the training integration tests which is not possible in importlib mode.

One option is to change the code to avoid importing other tests in tests (e.g. move imported objects out of the test and into the corresponding package in a test.py module), which is the option I personally would prefer.

Another is to keep the prepend mode, but add `__init__.py` files everywhere in the test folders to help module resolution.

HCookie:
I have a feeling that adding the `__init__.py`'s would be easier

VeraChristina:
Yes, we use pytest fixtures in all the integration tests, which are shared between tests imported from the conftest -- that's the imports you see. I would rather not refactor this (as it would change how we handle temp dirs and other aspects) unless we have stronger reasons to reconsider the fixture structure.

I'd be happy with the init files approach.

bdvllrs:
@HCookie  yes, let's change this and add the `__init__.py`. @VeraChristina I don't think using the importlib mode would prevent from using the fixture system and pytest dependency injections. It's just the ability to do `from conftest import ...` in the files that would not be possible.

I'll change this PR to keep the `prepend` mode.

mcocdawc:
can we wait 20mins more. I was doing some testing and there might be a more future proof solution.

mcocdawc:
https://docs.pytest.org/en/stable/how-to/writing_plugins.html

> If you have conftest.py files which do not reside in a python package directory (i.e. one containing an __init__.py) then "import conftest" can be ambiguous because there might be other conftest.py files as well on your PYTHONPATH or sys.path. It is thus good practice for projects to either put conftest.py under a package scope or to never import anything from a conftest.py file.

Under prepend, each conftest is imported under the bare name conftest, so all five compete for one sys.modules["conftest"] slot; the last one loaded wins

This is ambiguos and depends on CLI args order.

-----------------------

If we add the `__init__` files then it becomes unambiguous, since the conftest becomes `tests.a.conftest` — but then `from conftest import ...` no longer resolves at all and must be rewritten anyway

-----------------

With `importlib` its not ambiguos and simply not supported.

see [here](https://docs.pytest.org/en/stable/explanation/pythonpath.html)
> Test modules can't import each other.



--------------------------

Long story short. I would do what @bdvllrs suggested in the beginning and move some stuff to e.g. `training/src/anemoi/training/testing.py`
and have 
```python3
# GetTmpPath, moved from tests/integration/conftest.py
from anemoi.utils.testing import TemporaryDirectoryForTestData

class GetTmpPath:
    def __init__(self, temporary_directory_for_test_data: TemporaryDirectoryForTestData) -> None:
```
  
then in the tests or conftest.py we can do `from anemoi.training.testing import GetTmpPath`

bdvllrs:
Thanks for this comment. I indeed testing using prepend mode but with `__init__.py` and we still have some import issue.

I also agree with moving into a testing.py module in anemoi-training what needs to be imported to be the easiest. Besides there are only two problematic files: one could be rewritten as fixture, the other is just one class needed for typing. I'll make a new commit with this solution as a test but it should not be too much work.

bdvllrs:
I've added 3 commits with fixes for the training, models and graphs tests.

It was straightforward for training and graphs, but the migration tests required some changes on how the migration scripts were loaded. I've changed from loading the files by package path (as the package doesn't exists in the PYTHONPATH in the tests with the importlib import mode anymore) to a using the file path directly.

models/src/anemoi/models/migrations/migrator.py
```diff
@@ -523,11 +559,17 @@ def _resolve_context(self, context: MigrationContext) -> None:
...
-        for full_attribute_path_end, attribute_path_start in context.attribute_paths.items():
+        for (
+            full_attribute_path_end,
+            attribute_path_start,
+        ) in context.attribute_paths.items():
             attribute_path_start, _, mod_name_start = attribute_path_start.rpartition(".")
             attribute_path_end, _, mod_name_end = full_attribute_path_end.rpartition(".")
             LOGGER.debug(
-                "Move attribute %s from %s to %s.", mod_name_start, attribute_path_start, full_attribute_path_end
+                "Move attribute %s from %s to %s.",
```
mcocdawc: there seems to be a lot of reformatting to a shorter column width.

Could you please use the 120 line length configured in the pre-commit.

That should reduce the amount of code changes to be reviewed.

graphs/src/anemoi/graphs/testing.py
```diff
@@ -0,0 +1,76 @@
...
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import numpy as np
+
+
+class DatasetMock:
```
mcocdawc: I would declare the module private as `_testing.py`, it should probably not be part of the Public API.

If the module is public, at least the Mock classes should be private

training/tests/integration/schemas/partial_metadata_schema.py
```diff
@@ -1,63 +0,0 @@
-# (C) Copyright 2026 Anemoi contributors.
-#
-# This software is licensed under the terms of the Apache Licence Version 2.0
-# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
-#
-# In applying this licence, ECMWF does not waive the privileges and immunities
-# granted to it by virtue of its status as an intergovernmental organisation
-# nor does it submit to any jurisdiction.
-
-# This schema defines the expected structure of the metadata dictionary produced by the trainer.
```
mcocdawc, line 10: could you move this comment over to a docstring of the 
`partial_metadata_schema` function in
`training/tests/integration/conftest.py`.

models/src/anemoi/models/migrations/migrator.py
```diff
@@ -149,6 +151,34 @@ def _get_code_digest(content: str) -> str:
...
+        The imported module
+    """
+    spec = importlib.util.spec_from_file_location(location.stem, location)
+    if spec is None or spec.loader is None:
+        raise ValueError(f"{location} does not point to a valid Python file.")
+
+    module = importlib.util.module_from_spec(spec)
+
+    module_name = f"{package}.{location.stem}" if package else location.stem
+    sys.modules[module_name] = module
```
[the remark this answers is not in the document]
mcocdawc: `+`

models/src/anemoi/models/testing.py
```diff
@@ -0,0 +1,144 @@
...
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Utilities for launching distributed pytest checks.
+
+These functions handle the setup and launching of torch multiprocess tests, including
+verification of spawned ranks, backends, and devices. Tests of parallel kernels,
+communication primitives should live in separate test modules that call ``run_distributed_test``,
```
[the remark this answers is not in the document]
mcocdawc: `+`

mcocdawc (commented):
Great improvement to the code! Just a few comments.

models/src/anemoi/models/migrations/migrator.py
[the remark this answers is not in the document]
bdvllrs: Done

models/src/anemoi/models/migrations/migrator.py
bdvllrs: Sorry, I don't configure my editor formatter (ruff) because I expect the setting to be in the pyproject. I do run the pre-commit formatter, but ruff has added trailing commas everywhere which are not converted back. I've fixed (hopefully) all formatting contributions to not pollute this PR.

graphs/src/anemoi/graphs/testing.py
bdvllrs: Done: I've put the classes as private.

training/tests/integration/schemas/partial_metadata_schema.py
bdvllrs: Done

bdvllrs:
Thank you for your review! I've addressed all your comments in independent commits.

graphs/src/anemoi/graphs/testing.py
```diff
@@ -0,0 +1,76 @@
...
+# nor does it submit to any jurisdiction.
+
+import numpy as np
+
+
+class _DatasetMock:
+    """This datasets emulates the most primitive unstructured grid with
+    refinement.
+
+    Enumeration of cells , edges and vertices in netCDF file is 1 based.
```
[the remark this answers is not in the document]
bdvllrs: I think this is out of scope of this PR. The goal is to only allow pytest to use the importlib discovery, not to fix issues with the existing code.

bdvllrs:
> **models/src/anemoi/models/testing.py:35**
> 
>     * Renaming this existing public helper to a private symbol breaks callers importing `anemoi.models.testing.run_distributed_test`; the module documentation also still directs tests to that now-missing name. Importlib mode only requires importing the helper from the package, so preserve `run_distributed_test` and update the changed test imports accordingly.

I have renamee the docstring to match the renamed `run_distributed_test` to `_run_distributed_test`. This change was to answer @mcocdawc's previous comment to make the testing functions private. 

>       **graphs/src/anemoi/graphs/testing.py:22**
> 
>     * The refinement description repeats edge `E4`, but the diagram and `edge_vertices` data show that adding `V4` introduces edges `E4` and `E5`. Correct the second edge identifier so the mock documentation matches its data.
> 
> 
> ```
>     Cell C2 with its additional vertex V4 and edges E4 and E4 were added as
> ```

I would say out of scope of this PR as well.

models/src/anemoi/models/migrations/migrator.py
mcocdawc: Thanks for reverting!

Still, it's strange because we have

```
[tool.ruff]
target-version = "py311"
line-length = 120
```

The different tools and places of definition invite such problems.
https://github.com/ecmwf/anemoi-core/issues/1386

This should of course be no concern for this PR.

mcocdawc (commented):
Thank you for adressing the review!

Ignoring the latest copilots's comments makes sense and fixing the grammar in docstrings is not in this scope.

However, while we are at it and see them would be great to fix factually wrong statements in docstrings. Namely the:
> Cell C2 with its additional vertex V4 and edges E4 and E4

Before merging I'll run the integration tests. 
This might take until Monday, apologies, because we had to fix something there.

Apart from the docstring and pending the integration tests it's ready to merge.

models/src/anemoi/models/migrations/migrator.py
bdvllrs: It's in `training/pyproject.toml` but not in `models/pyproject.toml` (which is the package that caused the formatting issue). Thanks for opening the issue! Switching from black to ruff would be nice :)

mcocdawc:
Hey @bdvllrs,

The integration tests are fixed and @mchantry already added the label to run them.
As soon as you do any push, such as merging main into your branch, they are run in full.

(I had run the test that was known to work and it passed, so there likely won't be any problems. :-) )

bdvllrs:
Ouch, I think we might run into a problem with using multiprocessing in tests according to: https://github.com/pytest-dev/pytest/issues/10965.

I'll try to find a workaround (maybe calling the multiprocessing call from an actual module (`anemoi.training.testing` might do it?)

Edit: it did.

bdvllrs:
@mcocdawc There has been a bug with the integration tests!

mcocdawc:
hmm, we are hitting a problem with `importlib` and the question is if we need it at all.
So far, we conflated 
1. the ambiguity that arises from tests importing each other
2. the necessity to introduce `__init__.py` files
3. the necessity to introduce `importlib`

The refactoring to move helper functions into the anemoi package adressed point one. This is the right move and most important change and enables both `importlib` **and** fixes the ambiguity of the old import system. 

Switching to importlib then makes it more cumbersome to work with multiprocessing spawn (the change you had to make) and made a workaround in the migration scripts necessary.


With the refactoring there might be a much simpler way forward.
How important is it to enable name collisions in the test files? I think it's actually a feature that one cannot have two test files with the same name.
If we can rename `training/tests/unit/migrations/test_config.py -> test_config_migrations.py` (or something else) we can solve it much more easily.

------------------------------------------

Re the failing tests. We just hit too many jobs in the queue (`AssocMaxSubmitJobLimit`) ; no concern for this PR

bdvllrs:
Hi! I agree that it's starting to become a lot of work. This change is not strictly necessary and to me it was a good way forward until the changes required for multiprocessing. I personally don't see it as too much of a bad thing to extract main testing code into a `testing` module rather than using the conftests, or the `tests` folder as a package, but it's more a matter of taste.

> How important is it to enable name collisions in the test files? I think it's actually a feature that one cannot have two test files with the same name.

I agree for the prepend mode where module names are mixed, but in general I'm not sure to follow the full argument.

In any case I can do without this, so we can close this if we feal this is too big a change!

mcocdawc:
Hey @bdvllrs ,

I am very sorry for the back and forth. 
I talked to e.g. @joherics  who work on sharding, multiprocessing and distributed tests.
The way how importlib prevents the use of pickling and multiprocess spawn in the tests makes it, unfortunately, unusable for us.

In addition, being able to have name collisions of test files in different directories, such as `test_util.py` is also a hard requirement by some folks here.

The initial refactoring you did of moving shared test helpers out of sibling-importable modules into real packages is worth keeping. 
For the rest we should simply go with adding `__init__.py` everywhere, fully knowing that this requires rewriting many of the existing imports; and fully acknowledging, that I was initially against it.

I have tried adding the `__init__.py` everywhere and it works; changing the imports is simple mechanical work. I happily push this change to the PR, but I also don't want to hijack your work.

bdvllrs:
Hi @mcocdawc! No problem, just trying to improve the overall code quality! I fully agree with your comment and I don't mind you pushing your changes!

mcocdawc:
The PR is ready to merge, however I cannot approve it anymore as I pushed now myself. @VeraChristina  could you approve please?

Thank you @bdvllrs  and once more sorry for the back and forth.

VeraChristina (approved):
Thanks both for fixing this!

PR #1363, opened 2026-09-03
fix(training): resolve subgrid

### LINKED ISSUE #1342: Error in calling training.losses.combined.CombinedLoss when using spectral loss with subgrid
### What happened?

The following error is obtained when I tried to use CombinedLoss combining MSE and spectral loss with subgrid output_mask
hydra.errors.InstantiationException: Error in call to target 'anemoi.training.losses.combined.CombinedLoss':
InstantiationException("Error in call to target 'anemoi.training.losses.spectral.FourierCorrelationLoss':\nTypeError('slice expected at most 3 arguments, got 11')")

the yaml for the combined loss is the following
```
training_loss:
  datasets:
    data:
      losses:
        - _target_: anemoi.training.losses.MSELoss
        - _target_: anemoi.training.losses.FourierCorrelationLoss
          transform: fft2d
          x_dim: 121
          y_dim: 127
          subgrid: output_mask
      ignore_nans: False
      loss_weights: [0.5, 0.5]
```
Seems the function _resolve_subgrid() in anemoi/training/train/methods/base.py cannot resolve "output_mas"

### Version

anemoi-training v0.15.0

6.8.0-124-generic #124~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 21:05:19 UTC  x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell
hydra.errors.InstantiationException: Error in call to target 'anemoi.training.losses.combined.CombinedLoss':
InstantiationException("Error in call to target 'anemoi.training.losses.spectral.FourierCorrelationLoss':\nTypeError('slice expected at most 3 arguments, got 11')")
```

### Accompanying data

_No response_

### Organisation

CCRS, Singapore

dietervdb-meteo:
## Description
Spectral losses have an optional `subgrid` argument, that is a tuple representation of a slice. To facilitate LAM models, where the subgrid should always equal the output mask (LAM domain) one can simply set `subgrid: output_mask` into the config. But to enable that functionality a `resolve_subgrid` helper function replaces that string with the tuple representation ( [start, finish] in the dataset) under the hood. This helper function was broken, since it assumed a `dict` type of (parts of) the loss part of the config while the latest version of the code had it as `DictConfig`. Here we fix this and also make the code more robust against future such type changes.

closes #1342


dietervdb-meteo:
There is, as far as I'm aware, no integration test for spectral losses (let alone in the lam setup), and it could be overkill to introduce them. I did check that the RMI LAM model with a loss similar to #1342 trains without problems after this fix:
```yaml
training_loss:
    datasets:
      data:
        losses:
          - _target_: anemoi.training.losses.MSELoss
            scalers: ['pressure_level', 'general_variable', 'nan_mask_weights', 'node_weights', 'time_steps']
            ignore_nans: False
          - _target_: anemoi.training.losses.FourierCorrelationLoss
            transform: fft2d
            x_dim: 1069
            y_dim: 1069
            subgrid: output_mask
            scalers: ['pressure_level', 'general_variable', 'nan_mask_weights', 'time_steps']
        ignore_nans: False
        loss_weights: [0.5, 0.5]
```

training/src/anemoi/training/losses/loss.py
```diff
@@ -142,6 +142,12 @@ def _build_wrapped_loss(
     return wrapper
 
 
+def _check_subgrid_resolved(loss_config: dict) -> None:
+    if isinstance(loss_config.get("subgrid"), str):
+        msg = f"'subgrid' must be resolved to a tuple before instantiation, got {loss_config['subgrid']!r}"
```
HCookie, line 147: Would this not be better checking if it is not a tuple?

HCookie (commented):
Looks good to me, thanks for fixing that bug.
I haven't used the subgrid before, so can't exactly check that

training/src/anemoi/training/utils/resolve_config.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    return OmegaConf.is_dict(node) or isinstance(node, Mapping)
+
+
+def is_sequence(node: object) -> bool:
+    """Return True for list-like config nodes, including OmegaConf ``ListConfig`` (excluding ``str``).
+
+    ``ListConfig`` is not a subclass of ``list``/``Sequence``, so both are matched.
+    """
+    return (OmegaConf.is_list(node) or isinstance(node, Sequence)) and not isinstance(node, str)
```
frazane, line 35: Why does `is_sequence` check that the object is not a string? Is this needed?

training/src/anemoi/training/utils/resolve_config.py
```diff
@@ -0,0 +1,61 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+def is_mapping(node: object) -> bool:
+    """Return True for dict-like config nodes, including OmegaConf ``DictConfig``.
+
+    ``DictConfig`` is not a subclass of ``dict``/``Mapping``, so both are matched.
+    """
+    return OmegaConf.is_dict(node) or isinstance(node, Mapping)
```
frazane, line 27: Do we have other mappings in the codebase that justify `isinstance(node, Mapping)`? Why isn't `OmegaConf.is_dict(node)` enough?

training/src/anemoi/training/losses/loss.py
dietervdb-meteo: Although `SpectralLoss` expects a tuple, and `"output_mask"` gets resolved to one, a direct specification of the subgrid in yaml, say `[0,1]` gets translated as a list, which is just fine for `SpectralLoss`. So we can do
```python
if not isinstance(loss_config.get("subgrid"), (list,tuple)):
```
if you prefer.

training/src/anemoi/training/utils/resolve_config.py
```diff
@@ -0,0 +1,61 @@
...
+    elif is_sequence(node):
+        for item in node:
+            resolve_subgrid_node(item, output_mask, dataset_name)
+
+
+def resolve_subgrid(config: Mapping, output_mask: Mapping) -> None:
+    """Resolve ``subgrid: output_mask`` placeholders for every dataset in ``config``."""
+    for dataset_name, dataset_config in config.items():
+        if dataset_config is not None:
+            resolve_subgrid_node(dataset_config, output_mask[dataset_name], dataset_name)
```
frazane: I like the idea of having a module with utilities to help with config resolution, but in my opinion it should only contain generic helpers (like the ones above). Actual config resolutions should stay close the API where the config is given.

frazane (commented):
Thanks for the fix, left a couple of comments / questions!

training/src/anemoi/training/losses/loss.py
dietervdb-meteo: Also I added this guard specifically to have an easy pointer to failing of the resolver. Say for some reason a `datetime` gets passed the code will crash, but the issue is not the resolver and the msg would be misleading.

training/src/anemoi/training/utils/resolve_config.py
dietervdb-meteo: It is kinda crucial, that was the original bug, where the string `"output_mask"` got in the end interpreted as a tuple `("o","u","t",...)` and passed on like that to `SpectralLoss`. Of course that is specific for the resolver of this use case. Maybe we can rename `is_non_list_sequence`?

training/src/anemoi/training/utils/resolve_config.py
dietervdb-meteo: Valid point, let me reconsider.

training/src/anemoi/training/utils/resolve_config.py
frazane: I see that, but for this specific helper it feels a bit out of place. As you say this not-a-string check is very specific for this config resolution, so I would probably move it inside the function doing the resolution.

dietervdb-meteo: ok, makes sense!

dietervdb-meteo: done

training/src/anemoi/training/utils/resolve_config.py
dietervdb-meteo: `OmegaConf.is_dict` gives false on `dict`, rather than adding `dict` could be useful to stay general and just ask for `Mapping`.

training/src/anemoi/training/utils/resolve_config.py
dietervdb-meteo: Ok, did that. The helpers remaining are not specifically related to config resolution but left the filename as it gives a place for future config resolution utilities.

dietervdb-meteo:
arghhh.... forgot to adapt tests after refactor. will do tomorrow....

[edit: Today is yesterdays tomorrow. Fixed. :heavy_check_mark: ]

PR #1364, opened 2026-09-03
ci: fix for pull request triggers in integration test workflow

VeraChristina:
## Description
Update the PR trigger options for the ci workflow for integration tests:
- don't run on opened PRs (two avoid running twice when the run-expensive-tests label is added early)
- filter on added labels and only trigger the workflow when the run-expensive-tests label is added (not when another label is added and the run-expensive-tests is present)

The test workflow was triggered when adding arbitrary labels. For example, the expensive tests ran twice when adding the ATS label and the run-expensive-tests label together.

### DIFF
--- a/.github/workflows/integration-tests-hpc.yml
+++ b/.github/workflows/integration-tests-hpc.yml
@@ -26,7 +26,18 @@ jobs:
     timeout-minutes: 120
     # On a pull_request the expensive tests only run when the author opted in
     # via the 'run-expensive-tests' label; schedule/dispatch/merge_group always run.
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests-hpc.yml
```diff
@@ -146,7 +157,7 @@ jobs:
         with:
           issue-number: 294
           body: |
-            ## 🚨 Nightly Job Failed
+            ## Nightly Job Failed
```
mcocdawc, line 160: :-D

mcocdawc (approved):
lgtm, thank you!

PR #1368, opened 2026-09-05
fix(training): ensemble member sharding across GPUs

ssmmnn11:
Sharding of ensemble members got broken at some point.


mchantry:
Thanks Simon, great contribution. 
For my understanding, the test you've added would allow us to prevent this from reoccuring?

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
```diff
@@ -24,11 +24,20 @@ jobs:
...
               --distributed \
               --distributed-backend=all \
               --distributed-world-size=2
 
+            # Run the combined model and ensemble sharding checks on CPUs.
+            OMP_NUM_THREADS=1 python3 -m pytest -q \
+              models/tests/distributed/test_ensemble.py \
+              --distributed \
+              --distributed-backend=gloo \
+              --distributed-world-size=4
```
joherics: Would it be valuable to increase this to 8? That way we could test data parallelism, model sharding, and ensemble sharding combined.

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
joherics: Regarding the comment above, we would have to increase this from 4 to 8 cpus per task but 8 cores is still a very light job.

models/tests/distributed/test_ensemble.py
```diff
@@ -0,0 +1,170 @@
...
+
+    for name, parameter in model.named_parameters():
+        torch.testing.assert_close(parameter.grad, reference.get_parameter(name).grad, rtol=1e-12, atol=1e-12)
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize("data_parallel", [False, True], ids=["data_groups_1", "data_groups_2"])
+@pytest.mark.parametrize("model_group_size", [1, 2], ids=["model_shards_1", "model_shards_2"])
+@pytest.mark.parametrize("use_checkpoint", [False, True], ids=["no_checkpoint", "checkpoint"])
+def test_gather_ensemble_matches_unsharded_parameter_gradients(
```
joherics, line 150: This test gathers the full spatial grid before calculating the loss. I think it would be good to also add a test that where the loss is calculated locally on each grid shard and then reduced over the model group. E.g. training with sharded ensemble and crps loss.

joherics (changes requested):
I think this is a good solution to the ensemble-gradient scaling problem. It makes sense to add a new autograd wrapper and to place it in  `models/distributed` with the other custom autograd functions. As discussed offline, we can change the design in the future if we would like.

I only have two comments/questions regarding the test coverage but I would be fine with merging this in its current state and addressing them later if we want to get this fix into main quickly.

.github/workflows/distributed-primitive-tests-hpc-gpu.yml
joherics: Increased this from 4 to 8

models/tests/distributed/test_ensemble.py
joherics: I have added a new test that cover the case with sharded losses calculated locally.

joherics (approved):
I have pushed 2 commits addressing my previous comments. I have only added an additional test and increased the number of procs in the distributed workflow.  Feel free to merge this if you are ok with my additions @ssmmnn11

PR #1369, opened 2026-09-06
fix(training): scale prog variable at data index 0 in the tendency scaler

ssmmnn11:
AI summary:

## Background

The tendency scalers (`stdev_tendency` and `var_tendency`) adjust how much each
predicted variable counts towards the training loss. They do this by comparing
how much a variable varies overall with how much it changes from one time step to
the next. A variable that changes only slowly gets a larger weight, so that slow
variables and fast variables both contribute meaningfully to the loss.

## The problem

When the scaler works out the weight for each variable, it first checked whether
the variable exists in the dataset. That check was written as:

    self.data_indices.data.output.name_to_index.get(key)

This returns the position of the variable in the dataset, for example 0, 1, 2 and
so on. In Python the number 0 counts as false. So for the variable that happens to
sit at position 0, the check failed and the variable was skipped entirely.

A skipped variable keeps a weight of 1.0 instead of the weight the scaler was
supposed to give it. Variables in a dataset are ordered by position, and the
variable at position 0 is usually one of the predicted ones, so this affected most
training configurations that use a tendency scaler.

Please note that this changes the loss weights compared with previous runs. Runs
started before this fix used different weights, so results will not match exactly.

## Why the check was removed instead of corrected

The check is not needed at all any more, so it has been deleted rather than
rewritten.

It was added at a time when the model could split one dataset variable into
several model variables. For example, a wind or wave direction was converted into
its cosine and its sine, because an angle jumps from 359 degrees back to 0 degrees
and that is difficult to learn directly. The two new variables existed inside the
model but not in the dataset, so they had no statistics, and the check was there to
skip them.

That feature was removed in #268 and the conversion is now done when the dataset is
built instead. Since then, every variable the model produces also exists in the
dataset, so the check could never fail. Removing it avoids leaving a test in the
code that suggests a situation that can no longer happen.

## Other changes in this pull request

- The code now checks whether tendency statistics are `None` explicitly, rather
  than relying on Python's true or false interpretation of a dictionary. This is
  the same kind of mistake as the one described above.

- The two warning messages in the tendency scaler now go through the normal
  logger and say which scaler and which dataset they refer to. Previously they
  did not, which was confusing because the shipped configuration files create both
  tendency scalers even when the loss does not use them, so the same anonymous
  warning could appear twice.

- The temporal downscaler configuration files now explain why the
  `output_left_boundary` option requires the tendency timestep to be set
  explicitly. When that option is switched on, the first output step is 0 hours
  ahead, and there is no such thing as a change over 0 hours, so no statistics
  exist for it. Without an explicit timestep the scaler picks the first available
  lead time, finds nothing, and training stops with an error at startup.

## Tests

Three unit tests were added in `test_loss_scaling.py`:

1. Every predicted variable, including the one at position 0, receives the weight
   calculated from its own statistics. This test fails if the old check is put
   back.
2. A dataset with no tendency statistics leaves all weights at 1.0 and writes a
   warning naming the scaler and the dataset.
3. When no timestep is configured, the scaler falls back to the first available
   lead time and reports this.


training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -70,6 +70,7 @@ training:
   # Overwrites the tendency timestep used in the scaling to the output timestep so all lead times are normalised
   # by the same per-step statistics rather than their cumulative tendency statistics.
```
mc4117, line 71:
```suggestion
  # by the same per-step statistics rather than their cumulative tendency statistics. This is required if using task scaler with task.output_left_boundary because the 0h output step has no tendency statistics to fall back on.
```

training/src/anemoi/training/config/temporal_downscaler_ensemble.yaml
```diff
@@ -70,6 +70,7 @@ training:
...
   # This does NOT activate the tendency scaler - to do this you must include the scaler in the loss configuration.
+  # Required with task.output_left_boundary: the 0h output step has no tendency statistics to fall back on.
```
mc4117, line 73:
```suggestion
```
moved up

mc4117:
this looks good but I just suggested rewording the comment a bit in the yaml file because it took me a while to understand what the comment meant at first. I've only done it in one of the yaml files as an example but would need to change in both

mc4117 (approved):
LGTM!

PR #1371, opened 2026-09-06
fix(training): scalers and target variable

ssmmnn11:
Fix variable scalers to use model-output width when data.target is set-up. This prevents incorrect weight selection and normalisation. Stacked on #1369.


OpheliaMiralles (dismissed):
Looks OK to me

mc4117 (approved):
LGTM!

PR #1377, opened 2026-09-08
fix(training): sharded ensemble gradient scaling

japols:
## Description
Alternative fix to #1368, correcting grad scaling by registering gradient scaling hooks instead.

credit @ssmmnn11 

Ensemble gradients are currently under-scaled by a factor of `ens_comm_subgroup_size = num_gpus_per_ensemble / num_gpus_per_model`.

##  Additional notes ##


joherics (approved):
Thank you for implementing this approach as well @ssmmnn11! This looks good to me and I can see the benefits of using hooks rather introducing new custom autograd comm. functions.

In the future (non-blocking from my point of view) it would be good to add some further tests with larger problem sizes and also test data+model+ensemble sharding combined. I don't if we hit that scenario in anemoi but it could also be worth adding a tests where the loss is computed locally on shards and then reduced. I can add such tests with the model sharding tests.

PR #1382, opened 2026-09-10
fix(training): benchmark tests run again

cathalobrien:
## Description
benchmark tests pass again.

Screenshot 2026-09-10 at 16 33 05

I have two passing tests [here](https://github.com/ecmwf/anemoi-core/actions/runs/34484988453) and [here](https://github.com/ecmwf/anemoi-core/actions/runs/34485018706)

The problem was related to running distributed pytests, one would clean up the others tmpdir and the other process would fail then when it went to cleanup and there was nothing there. I fixed this by changing the retention policy to 'keep all tests'. this is safe, since the TMPDIRs are cleaned by slurm at the end of the job anyway.

A previous hack fix by me was to add a distributed barrier at the end of a test, to sync the processes right before cleanup. this resulted in a further issue: when the first test (LAM) failed the performance check, process 0 would raise an error and not join the barrier at the end of the LAM test. Process 1 would wait dutifully in the barrier at the end of LAM test, and process 0 would proceed to the next test (graph transformer) and timeout during model initialisation after 30 mins.

lastly, this PR changes the number of layers in the processor for the benchmark tests back to 2, after a recent cleanup PR inadvertently changed there number of layers to 16

----
📚 Documentation preview 📚: https://anemoi-training--1382.org.readthedocs.build/en/1382/

----
📚 Documentation preview 📚: https://anemoi-graphs--1382.org.readthedocs.build/en/1382/

----
📚 Documentation preview 📚: https://anemoi-models--1382.org.readthedocs.build/en/1382/

### DIFF
--- a/training/pytest.ini
+++ b/training/pytest.ini
@@ -11,7 +11,7 @@ markers =
     asyncio: marks tests as async tests
 
 
-tmp_path_retention_policy = all
+tmp_path_retention_policy = none
 log_cli = true
 log_cli_level = INFO
 log_cli_format = %(asctime)s %(levelname)s %(message)s
--- a/training/pytest.ini
+++ b/training/pytest.ini
@@ -1,5 +1,4 @@
 [pytest]
-tmp_path_retention_policy = all
 asyncio_mode = auto
 asyncio_default_fixture_loop_scope = function
 markers =
@@ -12,7 +11,7 @@ markers =
     asyncio: marks tests as async tests
 
 
-tmp_path_retention_policy = none
+tmp_path_retention_policy = all
 log_cli = true
 log_cli_level = INFO
 log_cli_format = %(asctime)s %(levelname)s %(message)s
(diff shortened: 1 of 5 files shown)


training/pytest.ini
```diff
@@ -11,7 +11,7 @@ markers =
     asyncio: marks tests as async tests
 
 
-tmp_path_retention_policy = none
```
VeraChristina, line 14: rather than modifying this for all pytest users (including running unit tests locally), can we pass this as an option in the ci?

VeraChristina (commented):
Thanks so much for fixing this! Just one quick question, otherwise very happy to approve!

training/pytest.ini
cathalobrien: good catch, yeah I pushed a commit now which limits the change just to the benchmark tests on github

VeraChristina (approved):
Thanks again! LGTM

PR #1383, opened 2026-09-10
fix: torch2.6 compatibility

### LINKED ISSUE #1326: anemoi-models fails to import on torch < 2.7 despite declaring torch>=2.3
Since #1282, importing `anemoi.models.distributed.primitives` fails on torch < 2.7, while `anemoi-models` declares `torch>=2.3,<2.11`. Any `anemoi-training` command hits this at startup through the schema import chain.

The `anemoi_distributed::alltoall` custom op is registered with builtin-generic annotations (`list[Tensor]`, `list[int]`). `torch.library.infer_schema` only accepts these spellings from torch 2.7 (pytorch/pytorch#146594); older versions require `typing.List[...]`.

With torch 2.5.1+cu124, Python 3.12:

```
$ python -c "import anemoi.models.distributed.primitives"
ValueError: infer_schema(func): Parameter input_list has unsupported type list[torch.Tensor].
```

Expected: import succeeds on any torch matching the declared requirement.

Possible fixes: switch the op annotations to `typing.List[...]`, or raise the torch floor. Note `torch.library.custom_op` itself only exists since torch 2.4, so with the annotation fix the floor is 2.4, not 2.3.

MeraX:
## Description
Resolves the incompatibility with torch < 2.7 noted in #1326 and restores GPU-compatibility of a graphs test.

bugfix

closes #1326 

### DIFF
--- a/graphs/tests/nodes/attributes/test_weights.py
+++ b/graphs/tests/nodes/attributes/test_weights.py
@@ -144,10 +144,10 @@ def test_masked_planar_area_weights_subset():
 
     weights = MaskedPlanarAreaWeights(mask_node_attr_name="patch").compute(graph, "test_nodes")
     assert torch.all(weights[~mask] == 0)
-    np.testing.assert_allclose(weights[mask].numpy(), weights[mask].max().item(), rtol=1e-6)
+    np.testing.assert_allclose(weights[mask].cpu().numpy(), weights[mask].max().item(), rtol=1e-6)
 
     halved = MaskedPlanarAreaWeights(mask_node_attr_name="patch_half").compute(graph, "test_nodes")
-    np.testing.assert_allclose(halved.numpy(), 0.5 * weights.numpy(), rtol=1e-6)
+    np.testing.assert_allclose(halved.cpu().numpy(), 0.5 * weights.cpu().numpy(), rtol=1e-6)
 
 
 def test_voronoi_region_areas_matches_convexhull():
--- a/models/src/anemoi/models/distributed/primitives.py
+++ b/models/src/anemoi/models/distributed/primitives.py
@@ -8,7 +8,8 @@
 # nor does it submit to any jurisdiction.
 
 
-from typing import List, Optional
+from typing import List
+from typing import Optional
 
 import torch
 import torch.distributed as dist
--- a/models/src/anemoi/models/distributed/primitives.py
+++ b/models/src/anemoi/models/distributed/primitives.py
@@ -8,7 +8,7 @@
 # nor does it submit to any jurisdiction.
 
 
-from typing import Optional
+from typing import List, Optional
 
 import torch
 import torch.distributed as dist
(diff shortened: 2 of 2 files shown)


MeraX:
@VeraChristina @dnerini et al.: Should we avoid regression of this issue? Do you see a more lightweight solution that setting up a runner with torch 2.5 or 2.6 and having (a separate) job executing the graphs tests on GPUs? The latter might be less important in practice, but it is nice if tests just pass, regardless of the system (gpu/no gpu) one is testing on.

frazane (commented):
Thanks a lot @MeraX for addressing this. Perhaps to avoid that somebody breaks this again (likely, since using the new annotations as `list[...]` is quite common) I would add a short comment somewhere explaining why we need to use the old annotations.

MeraX:
@frazane , I think `list[]` in principle is all fine, but it fails when it comes to `torch.compile`. Perhaps we should also inform @cathalobrien to keep this in mind when doing further great performance optimisations.

frazane:
> @frazane , I think `list[]` in principle is all fine, but it fails when it comes to `torch.compile`. Perhaps we should also inform @cathalobrien to keep this in mind when doing further great performance optimisations.

I agree that `list[]` is fine, it is also my personal preference, but that's exactly my point. If I didn't know this breaks things under specific circumstances, I would revert the current changes. A comment would be a cheap way to prevert this without the need to make pytorch 2.6 excplicitly part of the test matrix. But then if we want to pay the cost of those tests, that's the most robust option of course 👍

VeraChristina:
Thanks for fixing this!

I agree with Francesco that adding a comment would be a cheap way to help prevent this from being broken again. Would you mind adding a short comment explaining why we need to use the older annotation syntax here?

If we see this regression again in the future, we can then think about whether there’s a more robust way to catch it. For now, though, I’d prefer not to increase the test matrix just for this.

Also, if I understand the linked issue correctly, this fix would still require raising the minimum supported Torch version from 2.3 to 2.4. Can you clarify if this is the intended minimum version?

MeraX:
I think just commenting this singe occasion does not really prevent regression. I'm assuming that there are other signatures where `list` could  replace `List` in principle. The true save guard are tests. A cheap solution would be using different torch versions with the different python versions. If this suggestions needs more discussion, I'm happy to open a new issue/PR. I would be glad if we get this PR merged rather soon to include it in the next release.

frazane:
My comment was not blocking, just a suggestion, I trust your judgement here @MeraX !

MeraX:
I tested `anemoi-training` with torch 2.3 and 2.4. 2.3 failed, 2.4 passes and it is likely that we will update the minimum required version anyway. see https://github.com/ecmwf/anemoi-core/issues/1396 and https://github.com/ecmwf/anemoi-core/issues/1405

VeraChristina (approved):
Thanks for the fix!

PR #1384, opened 2026-09-10
fix(training): Match validation loss to training rollout.

jakob-schloer:
### Description

`task.validation_rollout` used to control two things at once: how far validation unrolls and how many steps the reported validation loss averages over. With `rollout.maximum: 1` and `validation_rollout: 5`, `train_<loss>_loss` was the loss at one step while `val_<loss>_loss` was the mean over steps 1 to 5. Two different quantities under the same name, which looks like a large train/val gap.

This PR separates the two roles with one rule:

- Losses follow `training. val_<loss>_loss` and the per-dataset `val_<dataset>_<loss>_loss` are averaged over the current training rollout (rollout.step) only, so they are directly comparable to `train_<loss>_loss`.

- Metrics follow the validation horizon. The validation metrics `val_<metric>_metric/<dataset>/<group>/<k>` are logged for every unrolled validation step k. Long-range skill lives here.

- `validation_rollout` is a minimum horizon. Validation unrolls `max(rollout.step, validation_rollout)` steps. The default null follows the training rollout, as before, so early epochs stay cheap.

`SingleTraining._step` and `EnsembleTraining._step` collect per-step losses and metrics and hand them to a new shared helper, `BaseTrainingModule._aggregate_steps`, which applies the rule above. The task's `steps("validation")` is the single source of truth for the validation horizon, including the dataloader offsets.

`RolloutEval` is removed. Since the task refactor it re-ran the identical `_step(validation_mode=True)` as `validation_step` and logged the same numbers under a `val_r<N>_` prefix. It could not choose its own rollout length, and it logged the loss as `val_r<N>_moduledict` because `pl_module.loss` is now a `ModuleDict`. The per-step validation metrics replace it. The callback, its config file, the entry in diagnostics/evaluation_ens.yaml, the docs section and the integration test configs are updated accordingly.

One validation pass now gives both readings: `val_multi_dataset_loss` comparable to the training loss.

image
_Figure: Example training runs with `validation_rollout: 5` and training rollout 1 previous to this PR and now. The training loss is exactly equal but previously the validation loss could not be compared to the training loss._ 

Here you find the test runs on Mlflow:
https://mlflow.ecmwf.int/#/experiments/855/runs?workflowType=machine_learning&searchFilter=&orderByKey=attributes.start_time&orderByAsc=false&startTime=ALL&lifecycleFilter=Active&modelVersionFilter=All+Runs&datasetsFilter=W10%3D

Follow-on to #1136, which fixed the same comparability problem by shortening the validation horizon. This gets the comparable loss while keeping the long-range skill curve.

Breaking changes

- `val_r<N>_*` metrics from `RolloutEval` are no longer logged. Configs that list `anemoi.training.diagnostics.callbacks.evaluation.RolloutEval` under `diagnostics.callbacks` must drop it and set `task.validation_rollout` to the desired horizon instead.

- `val_<loss>_loss` and `val_<dataset>_<loss>_loss` change numerically wherever `validation_rollout` exceeds the training rollout. Integration test configs use `validation_rollout: 2` with a training rollout of 1, so their logged validation loss changes; no assertion depends on its value.

- `validation_rollout: 0` is no longer accepted (schema is now PositiveInt | None). Default of `validation_rollout: null` of course is.

## Additional notes

There is no hard reason for validation metrics to be computed inside `_step` at all. The validation step already returns every per-step prediction for the plotting callbacks, targets are recoverable per step via `task.get_targets`, and `PerTimestepMetrics` already shows the pattern: a callback that delegates to `calculate_val_metrics` on the module. Moving the general validation metrics into a callback would remove the `validation_mode` branch from `compute_loss_metrics` and `reduce _step` to losses only. The metric objects would still live on the module because they share scaler state with the loss.

----
📚 Documentation preview 📚: https://anemoi-training--1384.org.readthedocs.build/en/1384/

----
📚 Documentation preview 📚: https://anemoi-graphs--1384.org.readthedocs.build/en/1384/

----
📚 Documentation preview 📚: https://anemoi-models--1384.org.readthedocs.build/en/1384/


training/src/anemoi/training/train/methods/base.py
```diff
@@ -867,7 +885,9 @@ def compute_loss_metrics(
                 if validation_mode:
                     loss_obj = self.loss[dataset_name]
                     loss_name = getattr(loss_obj, "name", loss_obj.__class__.__name__.lower())
-                    metrics_next[f"{dataset_name}_{loss_name}_loss"] = dataset_loss
+                    step = kwargs.get("rollout_step")
+                    suffix = "" if step is None else f"/{step + 1}"
+                    metrics_next[f"{dataset_name}_{loss_name}_loss{suffix}"] = dataset_loss
```
mc4117, line 890: I'm slightly confused why this is needed because don't we already log per rollout step metrics?

mc4117:
Good spot! Would be great to run the benchmark/accuracy tests on this branch as they might use the validation metrics and hence need to be updated

training/src/anemoi/training/train/methods/base.py
jakob-schloer: You are right. This is a duplication from `calculate_val_metrics`. I'll remove that.

training/src/anemoi/training/config/task/forecaster.yaml
```diff
@@ -11,5 +11,5 @@ rollout:
   # maximum rollout to use
   maximum: 1
 
-# number of rollout steps to use for validation, or null to use the training rollout
+# minimum number of rollout steps unrolled in validation, or null to follow the training rollout.
```
mc4117: I think `minimum` is a bit confusing here? 
Maybe something like

```
set validation rollout length if more rollout is required in validation than training, or set to null to follow the training rollout.
```

training/src/anemoi/training/train/methods/base.py
```diff
@@ -1007,6 +1011,30 @@ def _step(
     ) -> TrainingStepOutput:
         pass
 
+    def _aggregate_steps(
```
mc4117: I think aggregate is a bit confusing here. We already use it in two different contexts in core. How about `_reduce_rollout_steps`?

mc4117: or `average_rollout_steps`?

jakob-schloer: Good point! I went for `_combine_loss_and_metrics`.

mc4117 (dismissed):
LGTM!

training/src/anemoi/training/schemas/tasks.py
```diff
@@ -42,8 +42,8 @@ class ForecasterSchema(BaseModel):
     "Timestep string (e.g. '6H') defining the frequency of the input and output steps."
     rollout: RolloutSchema = Field(...)
     "Rollout configuration for autoregressive training."
-    validation_rollout: NonNegativeInt | None = Field(default=None, example=[None, 6, 12])
-    "Number of rollouts to use for validation. If unset, validation uses the training rollout."
+    validation_rollout: PositiveInt | None = Field(default=None, example=[None, 6, 12])
+    "Number of rollout steps unrolled in validation. The val loss is averaged over the training rollout only."
```
ssmmnn11:
```suggestion
    "Minimum number of rollout steps unrolled in validation; the training rollout is always unrolled. The val loss is averaged over the training rollout only."
```

training/src/anemoi/training/schemas/tasks.py
```diff
@@ -59,8 +59,8 @@ class OffsetForecasterSchema(BaseModel):
     "Time shift applied to the offsets between rollout steps. 'default' infers the largest valid shift."
     rollout: RolloutSchema = Field(...)
     "Rollout configuration for autoregressive training."
-    validation_rollout: NonNegativeInt | None = Field(default=None, example=[None, 6, 12])
-    "Number of rollouts to use for validation. If unset, validation uses the training rollout."
+    validation_rollout: PositiveInt | None = Field(default=None, example=[None, 6, 12])
+    "Number of rollout steps unrolled in validation. The val loss is averaged over the training rollout only."
```
ssmmnn11:
```suggestion
    "Minimum number of rollout steps unrolled in validation; the training rollout is always unrolled. The val loss is averaged over the training rollout only."
```

training/src/anemoi/training/config/task/offset_forecaster.yaml
```diff
@@ -10,5 +10,5 @@ rollout:
   # maximum rollout to use
   maximum: 1
 
-# number of rollout steps to use for validation, or null to use the training rollout
+# minimum number of rollout steps unrolled in validation, or null to follow the training rollout.
```
ssmmnn11:
```suggestion
# set validation rollout length if more rollout is required in validation than
# training, or set to null to follow the training rollout. Values below the
# training rollout have no effect: validation always unrolls the training rollout.
```

ssmmnn11 (changes requested):
Nice Works! Minor comments.

jakob-schloer:
Benchmark tests passed: 
https://github.com/ecmwf/anemoi-core/actions/runs/35600407382

training/src/anemoi/training/config/task/forecaster.yaml
```diff
@@ -11,5 +11,6 @@ rollout:
   # maximum rollout to use
   maximum: 1
 
-# number of rollout steps to use for validation, or null to use the training rollout
+# set validation rollout length if more rollout is required in validation than
+# training, or set to null to follow the training rollout.
```
mc4117:
```suggestion
# training, or set to null to follow the training rollout. Values below the
# training rollout have no effect: validation always unrolls the training rollout.
```

mc4117 (approved):
LGTM! Great work 🚀

ssmmnn11 (approved):
Very good. Thx!

PR #1374, opened 2026-09-07
feat(training): multi domain wiring

dietervdb-meteo:
This is some code on top of #1107 to wire the `MultiDomainDataset` to the config and set up a working example.

WIP: so far can train on a single domain with `MultiDomainDataset` dataloading functionality.

PR #1309, opened 2026-08-05
feat(models): add cross-attention aggregator for dataset latents

ssmmnn11:
revise aggregation: make aggregation via cross attention possible -> does not depend on dataset order

----
📚 Documentation preview 📚: https://anemoi-training--1309.org.readthedocs.build/en/1309/

----
📚 Documentation preview 📚: https://anemoi-graphs--1309.org.readthedocs.build/en/1309/

----
📚 Documentation preview 📚: https://anemoi-models--1309.org.readthedocs.build/en/1309/

PR #1385, opened 2026-09-11
feature(models): first draft of variable dropout for dynamical variable embedding

### LINKED ISSUE #1411: Variable agnostic modelling using metadata representation and embedding
### Motivation
We want a variable-agnostic model that encodes heterogeneous regional datasets into a shared latent representation of the atmosphere and decodes the requested physical fields. The current code requires global and regional datasets to share the same variables when constructing a cutout dataset, and does not support resuming from a checkpoint with a different variable set. These restrictions tie weather services to the variables available in ERA5, IFS, or checkpoints released by other national centres.

The intersection of available variables can shrink as more domains are added. We want to work with the union of those variables, using all available information and, ideally, predicting requested fields through their metadata, including when they are absent from a particular source.

### Proposed representation

For each dataset, store:

```text
values:         [batch, nodes, fields]
field metadata: [batch, fields, metadata]
field IDs:      [batch, fields]
```

Metadata should be stored once per field, ideally not once per grid point. It can be broadcast temporarily while combining values with embedded metadata, processed in bounded node chunks, and pooled over fields before graph encoding.

### Metadata semantics

The learned field description should embed the metadata that defines the physical quantity:

```yaml
param: t
vertical_coordinate:
  type: pressure                 # height_above_ground | height_above_sea | pressure | model
  level: 850
  unit: hPa                      # not mandatory
temporal_operator: instantaneous  # instantaneous | accumulation | mean | maximum | minimum
temporal_window: null             # required for non-instantaneous fields
value_unit: K
```

Pressure and height should be represented as continuous physical coordinates in canonical units to be able to extrapolate to unseen levels. Model levels are not universal, so they require a provenance-specific vertical-coordinate definition or conversion to physical coordinates but could be dealt with in a second space.

### Algorithm

```text
native values + field metadata
    -> per-node field pooling
    -> encoder / processor / decoder
    -> output transform conditioned on requested field metadata
```

Pooling would remove the explicit variable axis before spatial encoding and give the graph components a fixed-width node representation.

### Embedding

Shared across nodes, the field metadata could combine:

- A categorical embedding of `param`.
- A continuous encoding of pressure or height, such as a positional encoding or learned projection.
- A categorical embedding of `vertical_coordinate.type`.
- Encodings of the temporal operator, window and units where relevant.

At each node, the field values, optionally passed through a value encoder, could be combined with these metadata embeddings. Pooling over the available fields could then use attention or another set aggregation method. The specific combination and pooling operations would remain configurable.

On the output side, a corresponding transform could combine the decoded node representation with embeddings of the requested fields. This would allow the output field set to differ from the input field set.

### Pipeline design

Since graph encoding and processing logic is shared with previous iterations of the code, we suggest isolating the embedding logic in an isolated configurable **`input_transform`** module that replaces the preprocessing of **`x_vars`** in **`_assemble_input`** in the current `enc_proc_dec` model:

```python
def _assemble_input(...) -> tuple[torch.Tensor, torch.Tensor, ShardSizes]:
    node_attributes_data = ...
    grid_shard_sizes = ...
    x_skip = ...
    x_vars = self.input_transform(x, node_attributes_data)  # configurable in config file
    return x_vars, x_skip, grid_shard_sizes
```

### Handling heterogeneous field sets

- **Missing fields and values:** omit absent fields from the active values and metadata. Variable dropout could expose the model to different field subsets during training while retaining a fixed-width pooled representation. Per-node validity masks could exclude missing observations from pooling within an otherwise available field. This is particularly relevant in the context multidomain training.

- **Normalisation and checkpoint compatibility:** normalisation statistics could be selected by physical field identity after conversion to canonical units. A persistent vocabulary, metadata definitions and normalisation state could be saved alongside the model, allowing a new dataset’s active fields to be mapped independently of channel order.

- **Residual/skip connections:** construct skip mappings from compatible input and output field identities. Shared fields could retain a residual connection; outputs without a corresponding input could be predicted directly from the latent representation. Field selection and variable dropout would update these mappings consistently.

- **Requested fields absent from a source:** query the output transform with the requested metadata, using relationships learned from training data that provide those targets. Predicting a physical variable never supervised during training would remain an open question.

### Design goals

- Accept different variables, levels and missing fields from each source.
- Store field metadata as `O(fields × metadata)`, while the spatial processor operates without a variable axis.
- Investigate decoding at continuous pressure and height levels, including levels not seen during training.

### Additional context

In the DE-376 project, differences in the variables available across regional datasets limit how much atmospheric information we can use. For example, we cannot currently build a stretched grid from datasets with different variable sets or resume training from a checkpoint with a different set of variables. We propose a metadata-based representation to overcome these restrictions, allowing the model to accommodate missing inputs and predict requested fields even when they are absent from a particular training source.

### Organisation

DE-376

### Additional notes

Would configurable input and output transforms be a suitable integration point? Feedback on the metadata structure and compatibility with existing normalisation and checkpoint handling would also be useful.

### Edge cases identified:
- `mslp` : should be `param` `p` and level `height_above_sea` vs `sp` should be `param` `p` and level `height_above_ground`

einrone:
In this PR we introduce variable dropout for dynamic variable embeddings that will be used in multi-domain training. This functionality supports single or multiple prognostic variable dropout. With this feature, the model learns how to deal with missing variables during training. The implementation is a simple `torch.nn.Module` class.

## Description

This relates to PRs #1076, #1374, and #1245, as well as ongoing implementations in the Destination Earth 376 project.

This is part of the work required to support transfer learning when a given domain does not contain all variables that exist in the pretrained backbone. The goal is to use a pretrained AIFS backbone to fine-tune regional and multi-domain models where the available variables can differ between domains.

Variable dropout introduces additional flexibility by randomly dropping one or multiple prognostic input variables during training. This prevents the model from always relying on the presence of every variable and encourages it to learn representations that are robust to missing variables and differences in variable availability between domains.

For example, AIFS and domains A and B may contain `2t`, while domain C does not. During training, variable dropout can remove `2t` from domains where it is available, exposing the model to situations where this variable is missing. The model therefore learns to operate without always having access to `2t` and can make use of information and correlations contained in the remaining variables.

Combined with dynamic variable embeddings and the multi-domain architecture, this functionality is intended to improve robustness when transferring a pretrained AIFS backbone to domains with heterogeneous sets of prognostic variables.

##  Additional notes ##

----
📚 Documentation preview 📚: https://anemoi-training--1385.org.readthedocs.build/en/1385/

----
📚 Documentation preview 📚: https://anemoi-graphs--1385.org.readthedocs.build/en/1385/

----
📚 Documentation preview 📚: https://anemoi-models--1385.org.readthedocs.build/en/1385/


models/src/anemoi/models/models/base.py
```diff
@@ -71,6 +74,22 @@ def __init__(
 
         self.dataset_names = list(data_indices.keys())
         self._graph_name_hidden = model_config.model.model.hidden_nodes_name
+        self._multiple_hidden_meshes = isinstance(self._graph_name_hidden, (dict, DictConfig))
+        if isinstance(self._graph_name_hidden, (dict, DictConfig)):
+            self.dataset2hidden = dict(self._graph_name_hidden)
+            missing_datasets = set(self.dataset_names) - set(self.dataset2hidden)
+            unknown_datasets = set(self.dataset2hidden) - set(self.dataset_names)
+            assert not missing_datasets, f"Hidden node names are missing for datasets {sorted(missing_datasets)}."
```
OpheliaMiralles, line 82: All of it is part of multidomain we need to rebase on the main branch to get rid of this

einrone: what do you mean?

OpheliaMiralles: to get a clean diff vs main, we need to rebase on main and multidomain changes shouldn't appear in this PR, only variable agnosticity /dropout work

models/src/anemoi/models/variable/variablevocabular.py
```diff
@@ -0,0 +1,608 @@
...
+
+import re
+from dataclasses import dataclass
+from typing import Any
+from typing import Literal
+
+import torch
+
+VerticalCoordinateType = Literal[
+    "surface",
```
OpheliaMiralles, line 18: I would replace surface with heightAboveSea so mslp is indexed with it and tp/sp/etc with heightAboveGround

models/src/anemoi/models/variable/variablevocabular.py
```diff
@@ -0,0 +1,608 @@
...
+    @staticmethod
+    def _fetch_height(
+        mars: dict[str, Any],
+    ) -> tuple[str | None, int | None]:
+        """Extract height from surface parameters such as 2t and 10u.
+
+        Examples:
+            2t  -> ("t", 2)
+            10u -> ("u", 10)
+            sp  -> (None, None)
```
OpheliaMiralles, line 186: sp should be ("p", 0) for vertical coordinate heightAboveGround I think

models/src/anemoi/models/variable/__init__.py
```diff
@@ -0,0 +1,8 @@
+# (C) Copyright 2026 Anemoi contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
```
OpheliaMiralles, line 8: I would revert this change

OpheliaMiralles (commented):
It is much cleaner thanks! A few things about the var vocabulary still, maybe would be nice to have an example config to have an idea where we can insert it in the config. It is not entirely clear yet how it gets fused with the rest of the inputs yet. Maybe we can see what e.g. is done in #1309 and in this branch https://github.com/ecmwf/anemoi-core/compare/main...feat/richer-batch (see model/data/batch.py, the new batch object is allowed to carry metadata) and try to think about how  it would fit with our variable embedding.

PR #1406, opened 2026-09-16
wip: end to end multidomain

OpheliaMiralles:
## Description
Adds end-to-end multidomain model support on top of `feature/multi-domain`: all datasets share one encoder, one hidden latent representation and one decoder, with runtime routing for differently sized regional grids. 
Includes also 

- active-domain training/loss handling and 
- an integration test using two CERRA-derived regions with shared ERA5 context and weighted regional loss.

All based on the logic from the original destination Earth branch https://github.com/destination-earth-digital-twins/anemoi-core/tree/DE376.

Right now the multidomain branch (#1107) supports only the sampling mechanism for multiple domains. This code adds the component allowing to use the same encoder for datasets with different shapes and support for decoding using the same decoder for several domains. 

```
decoders:
    "0":
      target_datasets:
        - sg_1
        - sg_2
```

#1374 #1107 

----
📚 Documentation preview 📚: https://anemoi-training--1406.org.readthedocs.build/en/1406/

----
📚 Documentation preview 📚: https://anemoi-graphs--1406.org.readthedocs.build/en/1406/

----
📚 Documentation preview 📚: https://anemoi-models--1406.org.readthedocs.build/en/1406/


training/tests/integration/conftest.py
```diff
@@ -254,11 +255,77 @@ def multidatasets_config(
...
 def multidomain_config(
-    multidatasets_config: tuple[DictConfig, list[str]],
+    stretched_config: tuple[DictConfig, list[str]],
 ) -> tuple[DictConfig, list[str]]:
-    cfg, urls = multidatasets_config
+    cfg, urls = stretched_config
     use_case_modifications = OmegaConf.load(Path.cwd() / "training/tests/integration/config/test_multidomain.yaml")
+    OmegaConf.set_struct(cfg, False)
     cfg = OmegaConf.merge(cfg, use_case_modifications)
+
```
dietervdb-meteo, line 262: Could be good to move some of the below explictly to a default config in `training/src/training/configs` and then base the test off of that. will provide a more human readable example of a multi-domain config.

OpheliaMiralles: Agreed! Will do

models/src/anemoi/models/models/base.py
```diff
@@ -71,6 +74,22 @@ def __init__(
 
         self.dataset_names = list(data_indices.keys())
         self._graph_name_hidden = model_config.model.model.hidden_nodes_name
+        self._multiple_hidden_meshes = isinstance(self._graph_name_hidden, (dict, DictConfig))
+        if isinstance(self._graph_name_hidden, (dict, DictConfig)):
+            self.dataset2hidden = dict(self._graph_name_hidden)
```
dietervdb-meteo, line 79: I think it would be good to collect all asserts in a helper function, something like `self._validate` or something.

models/src/anemoi/models/models/base.py
```diff
@@ -39,6 +39,9 @@
 class BaseGraphModel(nn.Module):
     """Message passing graph neural network."""
 
+    supports_shared_encoder_decoder = False
```
dietervdb-meteo, line 42: Maybe remove this and only add `supports_shared_encoder_decoder = False` to the classes that don't support it? Then check with `self.getattribute('supports_shared_encoder_decoder', True)`, that way, only classes that don't support it need extra code, and then as we bring support to more and more classes we can remove code, rather than updating an attribute that becomes meaningless.

einrone:
As discussed yesterday, I can enable multiple hidden meshes within this PR.

dietervdb-meteo:
Hi @einrone , [c38efa8](https://github.com/ecmwf/anemoi-core/pull/1406/commits/c38efa82a0331b79a89aa16018ee4cbcc553ccab) already takes care of this. So we should review and finetune what is there now.

dietervdb-meteo:
Integration tests appart from accuracy pass. The accuracy issue lives on main: https://github.com/ecmwf/anemoi-core/actions/runs/35847912233. So we are good to merge.

